PR #1, opened 2015-08-14
Do not allow ProductCovarianceModel of dim > 1

jschueller:
TODO: TensorizedCovarianceModel to allow that


sofianehaddad:
We could also remove useless '#include "IdentityMatrix.hxx"' in ProductCovarianceModel.cxx

jschueller:
Thank you.

PR #2, opened 2015-08-17
Update README with Markdown syntax

felipeam86:
With these modifications the README renders better on Github.


jschueller:
Thank you,
- I would have prefered rst syntax for consistency with what we use in the sphinx doc.
- This change will affect packaging, you need to modify distro/rpm/openturns.spec & distro/debian/docs

Note that you could have moved & modified in one single commit,
also I advise you to create a branch for each PR you create, as this allows to handle multiple PRs at the same time easily.

j

felipeam86:
Ok, my bad. In that case, I propose to ignore this pull-request.

We can keep rst syntax and also have a nice rendering of the README. We just need to add the .rst extension so that github knows how to render.

Take a look at my fork to see the effect. There are some modifications to be done if we want to have a better rendering. For example, the source installation instructions for the [quick reader](https://github.com/felipeam86/openturns#source-installation)

jschueller:
thanks, see #15

PR #4, opened 2015-09-02
Add missing closing parenthesis in Mixture::__str__

dbarbier:


jschueller:
LGTM
By the way travis failed with hmat master last time, is this a wanted change or should we use 1.1.2 ?

dbarbier:
It is a good idea to follow hmat-oss master in order to detect those regressions earlier.  I am investigating this one.

dbarbier:
hmat-oss has been fixed, thanks.

PR #9, opened 2015-09-08
Pretty print big samples

jschueller:
If a sample contains more than 1000 elts, the content is shorten, like numpy arrays.
this affects repr and str, and description print is moved to implementation
```
import openturns as ot
a=ot.NumericalSample(32,32)
a.setDescription(['deded7']*32)

a
class=NumericalSample name=Unnamed implementation=class=NumericalSampleImplementation name=Unnamed size=32 dimension=32 description=[deded7,deded7,deded7,...,deded7,deded7,deded7] data=[[0,0,0,...,0,0,0],[0,0,0,...,0,0,0],[0,0,0,...,0,0,0],...,[0,0,0,...,0,0,0],[0,0,0,...,0,0,0],[0,0,0,...,0,0,0]]

print(a)
     [ deded7 deded7 deded7 ... deded7 deded7 deded7 ]
 0 : [ 0      0      0      ... 0      0      0      ]
 1 : [ 0      0      0      ... 0      0      0      ]
 2 : [ 0      0      0      ... 0      0      0      ]
...
29 : [ 0      0      0      ... 0      0      0      ]
30 : [ 0      0      0      ... 0      0      0      ]
31 : [ 0      0      0      ... 0      0      0      ]
```

it breaks 80 tests
thank you, rodrigue.


dbarbier:
This description is misleading, the main change is not output truncation (which is a nice addition), but the change in `NumericalSampleImplementation::__repr__`.  IMO it would be better to have 2 distinct commits (gathered in a single pull request), 1st one change NumericalSampleImplementation representation, 2nd one implements truncation for large data.  Both could be extended to matrices too, that would be nice.

BTW there are many whitespace change in `NumericalSampleImplementation::__str__`, this is annoying.

PR #12, opened 2015-09-08
Big sample2

jschueller:
see #9


dbarbier:
+1, thanks for updating.
Looks good to me, but please keep it open for several days so that other developers have a chance to give their opinion.

jschueller:
@dbarbier Ok, no problem.

jschueller:
I don't think much can be reused, besides ordering is different for matrices.

regislebrun:
I was thinking about separating the computation of the optimal column width, the decorations (descriptions, row number) and the printing of the raw values with the potential ellipsis. But you are right: the data layount is different between the two objects and it seems to be difficult to reuse this last part.

jschueller:
@regislebrun Now the computation of the optimal width takes the ellipsis into account and the code should be faster, is that what you were thinking ?

jschueller:
@regislebrun so, good so far ? Can I merge this next week ?

PR #15, opened 2015-09-09
Format README to rst

jschueller:
superseeds #2
https://github.com/jschueller/openturns/blob/readme_rst/README.rst


dbarbier:
+1

PR #17, opened 2015-09-15
Major redesign of optimization classes

dbarbier:
OpenTURNS embeds several optimization solvers, but their usage is currently hardcoded when such a solver is needed; for instance, `TrapezoidalFactory` uses Cobyla, and it is not possible to let it use another solver.

This work aims at modifying classes so that user can change optimization solver if they want, exactly as it had been done for FFT.  In the future, more solvers will be wrapped, either directly within OpenTURNS or via modules.

There are two main classes:
  * `OptimizationProblem` defines an optimization problem (objective function, equality constraints, inequality constraints, bounds).
  * `OptimizationSolver` is a wrapper around an optimization solver.  After being initialized, its `run` method is called and result is available via its  `getResult` method

`Cobyla`, `TNC`, `AbdoRackwitz` and `SQP` classes have been modified to derive from `OptimizationSolverImplementation`.  `NearestPointAlgorithm` and `BoundConstrainedAlgorithm` classes are removed.

All classes which were using these solvers now have a `setOptimizationSolver` method to override default optimization solver, which is unchanged.

There are still some unresolved issues:
 * OptimizationSolver contains accessors for a maximum number of iterations; should we keep this number, or instead define the maximum number of evaluations (for the objective function). 
 * There is no smart choice of the best algorithm, default one is still hardcoded
 * Classes which make use of optimization solvers do not expose their `OptimizationProblem`, maybe they should have an `optimizationProblem_` member with a public accessor.


jschueller:
1- NLopt has an API that limits the number of iterations, but I guess it depends on the solver.
Maybe we could add both and the solver could throws an Exception for the one it doesn't support.
We could maybe extend it to the convergence criteria : solvers have their own. Maybe we shouldnt try to enforce our own a posteriori criteria and tremove the : "The Cobyla algorithm could not enforce the convergence criteria".
2- I won't know we could make it smarter than try a list of solvers until it succeeds, sorry.
3- Interesting, what's the use-case you have in mind for this ?

dbarbier:
For 3: I was wondering if client algorithms could expose their `OptimizationProblem` so that users can inspect it and decide of the best available solver for this specific case.  But we discussed about that on Friday with Régis and he convinced me that this is not a good idea, so let us forget about it.

For 1: I was puzzled by `scipy.optimize.minimize`, most solvers provide a `maxiter` option, but sometimes also a `maxfev` option (or named `maxfun`).  I agree that we should provide both, in order to fix the ticket you link to.

Anyway I received only positive feedback about these changes, please merge them, and we will discuss about further changes later.

PR #22, opened 2015-09-23
Move static *_initializer variables from hxx into cxx files and make them const

dbarbier:
Currently one instance of these variables is created by object files including these
header files.  This is harmless because initialization of the real struct is
protected by pthread_once, but it is annoying when debugging initialization issues
of singleton classes.


jschueller:
-1

It would be cleaner to have them in cxx, but the order of initialization is important, the dependencies between the 3 headers forces the right order in the same compilation unit.

http://trac.openturns.org/ticket/400

dbarbier:
Indeed, but it would be much simpler to remove these dependencies, there are threading issues which cannot be solved easily (AFAICT they can only be solved when compiled with C++11).

dbarbier:
I had a closer look at ticket 400, the real problem is that `ResourceMap::GetInstance()` relies on static object being initialized, but it should create it if needed.

dbarbier:
Closing, a better implementation is proposed in #25.

PR #23, opened 2015-09-23
Argument of Factory constructor is unused

dbarbier:
Mark previous constructor as @deprecated, and create a new one
without argument.

Update all classes to call the new constructor.
Make static Factory objects as const, since they are not modified.


jschueller:
+1

jschueller:
cherry-picked https://github.com/openturns/openturns/commit/bbac40295e0d47ce9e7e8464ae1f85d1c7ca34bf

PR #24, opened 2015-09-24
Add timeout option in coupling_tools.execute

jschueller:


dbarbier:
Looks good, but I do not see the point about `check_timeout`; when you set a timeout, you want to be warned if it is hit.

jschueller:
Ok, removed.

dbarbier:
Thanks, +1

regislebrun:
+1

PR #25, opened 2015-09-24
Rewrite singleton classes Catalog, Log, ResourceMap and TBB

dbarbier:
Singletons are challenging in C++, it is very hard to have a thread-safe implementation which takes into account dependencies between singletons.

Current code does not work as expected, we had issues on Windows when loading an Excel plug-in because ResourceMap was not initialized. Current code is difficult to read and debug, mostly because singleton builders (`*_init` structs defined along singletons) are initialized in header files and thus duplicated in each object file.

We propose here a simpler pattern: a `MutexLockSingleton` template class is declared in `MutexLock.hxx`, it is modelled after `ResourceMapInstance`. `Catalog`, `Log`, `ResourceMap` and `TBB` are rewritten to use it. 

A static builder is defined in implementation files, and a second one in `GetInstance()`.  There are no threading issues because static variables are instantiated during library load, before user code.  The second builder ensures that initialization order does not matter.

It is not clear whether problems reported above are fixed by this commit, but at least it should be simpler to follow singleton constructions.

### DIFF
--- a/lib/src/Base/Common/Catalog.cxx
+++ b/lib/src/Base/Common/Catalog.cxx
@@ -32,25 +32,32 @@
 BEGIN_NAMESPACE_OPENTURNS
 
 
-
-
 static pthread_mutex_t Catalog_InstanceMutex_;
 static Catalog * Catalog_P_instance_ = 0;
-static pthread_once_t Catalog_InstanceMutex_once = PTHREAD_ONCE_INIT;
+static const Catalog_init static_initializer_Catalog;
 
 Catalog_init::Catalog_init()
 {
-  int rc = pthread_once( &Catalog_InstanceMutex_once, Catalog::Initialize );
-  if (rc != 0)
+  if (!Catalog_P_instance_)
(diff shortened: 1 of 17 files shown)


dbarbier:
This pull request superseeds #22.

dbarbier:
Rebased, slightly edited 8447fd9 and added e40245a.

jschueller:
Just tested on mingw with the isolated dlls and a single .exe ; it does work even with a missing config.
+1

dbarbier:
Done, last commit amended.

jschueller:
@dbarbier, it breaks the (mingw) build on windows (not wine).
ResourceMap_init::~ResourceMap_init and Catalog_init::~Catalog_init are called twice when the library is unloaded, and the second call to pthread_mutex_destroy results in a crash.
if I comment them the program exits fine:
```
//  pthread_mutex_destroy( &ResourceMap_InstanceMutex_ );
```

PR #26, opened 2015-09-25
Optim

jschueller:
Follows #17


dbarbier:
+1

jschueller:
yes, fixed and squashed

sofianehaddad:
+1

jschueller:
Just tested on mingw with the isolated dlls and a single .exe ; it does work even with a missing config.
+1

PR #19, opened 2015-09-21
Remove library wrapper

jschueller:

PR #27, opened 2015-09-25
Increase NP typecheck priority

jschueller:
Fixes swig warning 509 about NP/NS overloads for various methods after wrappers removal.
```
[0m/home/travis/build/openturns/openturns/lib/src/Base/Geom/Mesh.hxx:84: Warning 509: Overloaded method OT::Mesh::getNearestVertexIndex(OT::NumericalSample const &) const effectively ignored,
/home/travis/build/openturns/openturns/lib/src/Base/Geom/Mesh.hxx:78: Warning 509: as it is shadowed by OT::Mesh::getNearestVertexIndex(OT::NumericalPoint const &) const.
```

See discussion in https://github.com/openturns/openturns/pull/19

I chosed to increase the NP priority instead of NS remembering this: http://trac.openturns.org/changeset/2261
Also Field has this set to 3.

### DIFF
--- a/python/src/NumericalPoint.i
+++ b/python/src/NumericalPoint.i
@@ -28,7 +28,7 @@
 %template(NumericalPointPersistentCollection)  OT::PersistentCollection<OT::NumericalPoint>;
 
 
-#define OT_TYPECHECK_NUMERICALPOINT 5
+#define OT_TYPECHECK_NUMERICALPOINT 4
 
 %typemap(in) const NumericalPoint & ($1_basetype temp) {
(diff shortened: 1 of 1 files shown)


dbarbier:
+1

BTW with current master, NP is checked before NS.  With your change, NS is checked before NP.  Thus it looks like declaration order does not matter, we can set NP and NS typechecks to any value.

jschueller:
I wanted a higher priority for NP than NS, I revert it to 4.

jschueller:
@dbarbier, so what do you think ?

dbarbier:
+1, I believe that NP and NS precedence do not matter, they only have to be different.

PR #28, opened 2015-09-28
Minor update in CovarianceModelImplementation

sofianehaddad:
All methods of the implementation class should be declared as virtual. Indeed some of these methods might be redefined by derived classes


dbarbier:
+1

jschueller:
As a follow-up, these methods are lacking docstrings:
http://openturns.github.io/user_manual/_generated/openturns.CovarianceModel.html

sofianehaddad:
Done

jschueller:
+1

PR #29, opened 2015-09-28
Avoid infinite loops in several distribution factories

dbarbier:
When these factories are called with invalid values (either NaN or Inf),
getSample() loops forever.
Fix http://trac.openturns.org/ticket/796

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,7 @@
  * Enhanced print of samples
 
 === Bug fixes ===
+ * #796 (Beta distribution: if sample contains Inf, freeze on getSample)
  * #798 (Error message misstyping (Gamma distribution))
 
 == 1.6 release (2015-08-14) == #release-1.6
--- a/lib/src/Uncertainty/Distribution/BetaFactory.cxx
+++ b/lib/src/Uncertainty/Distribution/BetaFactory.cxx
@@ -19,6 +19,7 @@
  *
  */
 #include "BetaFactory.hxx"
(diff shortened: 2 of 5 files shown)


dbarbier:
Good idea, commit rebased.  ChangeLog entry also added.

jschueller:
+1

dbarbier:
@jschueller Any idea why `t_VonMises_std.py` failed on travis?  I cannot reproduce this failure, and doubt that it is real.  Could you please force a rebuild?

jschueller:
I already seen it fail on travis twice, restarting.

sofianehaddad:
+1

jschueller:
What's wrong with it ?

dbarbier:
Huh?  Sorry, looks like I closed it by mistake.

jschueller:
As a follow-up from http://trac.openturns.org/ticket/799 on the same topic,
we could do the same patch for gumbel:
-  if (sigma <= 0.0) throw InvalidArgumentException(HERE) << "Error: can build a Gumbel distribution only if sigma > 0, here sigma=" << sigma;
+  if (!(sigma > 0.0) || SpecFunc::IsInf(sigma)) throw InvalidArgumentException(HERE) << "Error: can build a Gumbel distribution only if sigma > 0, here sigma=" << sigma;

dbarbier:
@jschueller Maybe, but there is no infinite loop involved in GumbelFactory.

jschueller:
ok

PR #30, opened 2015-09-30
Cleanup optimization API

jschueller:
1.- The solvers *SpecificParameters classes are deprecated as parameters are transfered to solvers as attributes, docstrings were added so as to solve http://trac.openturns.org/ticket/790
2.- I thought OptimizationSolverImplementationResult could be renamed to OptimizationResult
3.- checkProblem should be private


jschueller:
@dbarbier, ok, fixed and squashed

dbarbier:
I have mixed feelings about the last commit, okay for other ones.

jschueller:
@dbarbier why do you think we'll need to overload OptimProblem ?

sofianehaddad:
@jschueller @dbarbier Maybe in the framework of robust optimization, it will be important to have a specific class which inherits from OptimizationProblem. What do you think?

jschueller:
@sofianehaddad I'm not sure what you're referring to

jschueller:
I removed the last commit, leaving for later.

dbarbier:
Thanks, +1

PR #31, opened 2015-10-02
Native compilation on OS X

sofianehaddad:
We update the library API to let the compilation on Debian & OS X systems using clang/libc++

Indeed, native compilation on OS X system uses the last ones as Mac OS X 10.9+ no longer uses GCC/libstdc++. In that case, dependencies such as muParser & tbb are naturally linked to libc++ (using packager like MacPort/Homebrew or ot-superbuild for example)

Following pull request will focus on a detail "How to install with OS X"

### DIFF
--- a/lib/src/Base/Common/OSS.hxx
+++ b/lib/src/Base/Common/OSS.hxx
@@ -272,13 +272,6 @@ public:
   OSS_iterator&
   operator=(const _Tp& value)
   {
-#ifdef __GNUC__
-#if ( GCC_VERSION >= 30400 )
-    __glibcxx_requires_cond(_M_stream != 0,
-                            _M_message(__gnu_debug::__msg_output_ostream)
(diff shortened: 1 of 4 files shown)


jschueller:
1- do you know if that __glibcxx_requires_cond block is still useful ? I tried removing it and it's ok.
2- usually that test is written sys.platform.startswith('win') which is simpler

dbarbier:
@jschueller Not sure about 2, I believe that this test must also detect cygwin.

dbarbier:
+, I can not check whether these patches work fine on Mac OSX, but AFAICT they do not introduce regressions for current supported systems.

jschueller:
1. seems you can remove that code block ; #32 passed the tests successfully
2. denis is right, you can leave the code as is

jschueller:
+1

dbarbier:
+1

PR #34, opened 2015-10-06
Support infix operator for matrix multiplication (PEP465)

jschueller:
http://legacy.python.org/dev/peps/pep-0465/


dbarbier:
Wow, looks great.  +1, am now willing to upgrade to Python 3.5 to use it ;)

sofianehaddad:
+1

PR #35, opened 2015-10-06
Removed ConfidenceInterval class

jschueller:
As discussed todya, this is just a storage class that provides nothing more than Interval.


adutfoy:
+1

sofianehaddad:
+1

PR #36, opened 2015-10-06
Implement DistributionParameters::inverse

jschueller:
Implemented inverse conversion formulas from native parameters.
Also DistributionParameters::getParametersCollection becomes DistributionParameters::getParameters which returns a NPWD instead of a collection, this compact representation will also happen in distribution
Then merged the DistributionParameters tests into one which allowed to reuse a lot of code


sofianehaddad:
+1

dbarbier:
Is this pull request superseded by #38?  It looks like this branch had been deleted.

jschueller:
no, it's independent

PR #38, opened 2015-10-07
Compact distribution parameters representation

jschueller:
<s>This adds Distribution::[gs]etParameters for the compact representation of parameters, ie no duplicates of parameters that have an effect on marginals as well as copula, in a single sequence.
Currently implemented for univariate distributions.
Distribution::setParametersCollection calls Distribution::setParameters in the 1-d case : all univariate distributions have been simplified to use setParameters instead of setParametersCollection
Distribution::getParametersCollection calls Distribution::getParameters in all cases : all univariate distributions have been simplified too, and multivariate distributions all implement getParametersCollection.
DistributionImplementation::getParameters return an empty NPWD by default.
in classes MinCopula, SubSquareCopula, IndependentCopula and TracyWidom* getParametersCollection have been removed as getParameters also returns empty parameters</s>


sofianehaddad:
+1

jschueller:
I'm not too happy about that code yet :[

sofianehaddad:
I am wondering if getParametesCollection is still usefull?

regislebrun:
Yes it is, for those who want to compute FORM sensitivities!

sofianehaddad:
Yes but we have the getParameters. Could we update for example the FORM/SORM algorithms to take into account getParameters? Why having a the "same" method twice?

regislebrun:
the work has to be done into AnalyticalResult::computeHasoferReliabilityIndexSensitivity(), and it is not totaly trivial. I propose to postpone this change as it could take more time than we can afford now.

jschueller:
UPDATE: new approach: setParametersCollection calls setParameters, getParameters calls getParametersCollection.
In the usual case, a distribution overloads getParametersCollection & setParameters.

- Removed setParametersCollection(NumericalPoint) MCMC, ConditionalDistribution uses this:
```
Distribution pI(conditional_);
-    pI.setParametersCollection(zi);
+    pI.setParameters(zi);
     NumericalScalar logPdf = pI.computeLogPDF(observations_[i]);
```

- Added safeguards in the setParameters (previously setParametersCollection):
```
>>> import openturns as ot
>>> dist=ot.Uniform()
>>> dist.setParametersCollection(ot.NumericalPointCollection(0))
Erreur de segmentation (core dumped)
```

- Implemented TruncatedDistribution::setParameters

jschueller:
- Include marginal index in EllipticalDistribution parameters so getParametersNumber is correct:
before:
```
>>> import openturns as ot
>>> dist=ot.Normal(3)
>>> dist.getParametersCollection()
[class=NumericalPointWithDescription name=marginal 1 dimension=2 description=[mean,standard deviation] values=[0,1],class=NumericalPointWithDescription name=marginal 2 dimension=2 description=[mean,standard deviation] values=[0,1],class=NumericalPointWithDescription name=marginal 3 dimension=2 description=[mean,standard deviation] values=[0,1],class=NumericalPointWithDescription name=dependence dimension=3 description=[R_1_0,R_2_0,R_2_1] values=[0,0,0]]
>>> dist.getParametersNumber()
5L
```

Now:
```
>>> import openturns as ot
>>> dist=ot.Normal(3)
>>> dist.getParametersCollection()
[class=NumericalPointWithDescription name=marginal 1 dimension=2 description=[mean_0,standard_deviation_0] values=[0,1],class=NumericalPointWithDescription name=marginal 2 dimension=2 description=[mean_1,standard_deviation_1] values=[0,1],class=NumericalPointWithDescription name=marginal 3 dimension=2 description=[mean_2,standard_deviation_2] values=[0,1],class=NumericalPointWithDescription name=dependence dimension=3 description=[R_1_0,R_2_0,R_2_1] values=[0,0,0]]
>>> dist.getParametersNumber()
9L
```

jschueller:
@regislebrun what do you think about my new proposal

jschueller:
- Fixes the same problem for Dirichlet.

dbarbier:
Commit log messages are way too short; rationale must be stored in git, not in comments here.  BTW, NMF also provides {set,get}Parameters, we should be consistent.

IMO we can replace {set,get}ParametersCollection by {set,get}Parameters, which is the most common usage, and add  a `Collection<UnsignedInteger> getParametersDimension() const` method to retrieve dimensions.  NMF contains an unused getParameterDimension method, it should be removed or replaced by getParametersDimension.

jschueller:
We need getParametersCollection atm for FORM sensitivities, more on this tomorrow.

PR #41, opened 2015-10-15
Fix  python test environement variable for Darwin systems 

sofianehaddad:
Currently, installcheck fails on Darwin system. The reason is that the environement is uncomplete, in comparaison with other Unixes systems.

To fix it, DYLD_LIBRARY_PATH environment variable should be initialized instead of LD_LIBRARY_PATH  on Apple (https://cmake.org/Wiki/CMake_RPATH_handling)


sofianehaddad:
As mentionned in the previous link, for binaries such tests, the linker locate the libraries using the full recorded paths.

We can see also that in cppcheck, the LD_LIBRARY_PATH is not overloaded ( "LD_LIBRARY_PATH=$ENV{LD_LIBRARY_PATH}")

sofianehaddad:
ok for list, i ll change it

jschueller:
could you also test the change in ottemplate ?

jschueller:
+1

sofianehaddad:
For ottemplate, same changes are required for pycheck. 
For cppcheck, some changes have also to be done as t_* are not directly linked to libOT

dbarbier:
Okay, this patch fixes failures on Mac OSX, thus +1 from me.

But it is wrong: in fact  `LD_LIBRARY_PATH` is not needed on Linux, installation paths are already recorded within rpath.  So the real problem is that rpath settings do not work on Mac OSX.

jschueller:
@dbarbier, I think LD_LIBRARY_PATH is needed when using DESTDIR, for example when building the debian packages : tests should link to the temporary INSTALL_DESTDIR=/tmp/debian... (and for a system install path rpath is not used at all anyway)

dbarbier:
@jschueller Sure, but when you use DESTDIR, you intentionally mess up installation paths, so you have to fix things with LD_LIBRARY_PATH if needed.

dbarbier:
Let me try to better explain my point of view: with this patch, `make installcheck` works, but I am pretty sure that Python bindings won't work unless `DYLD_LIBRARY_PATH` is set.  This is why I told that this patch hides this problem but does not solve it.

@sofianehaddad Can you please tell whether you have to set `DYLD_LIBRARY_PATH` in your environment after installing Python bindings?  Or is this only a DESTDIR issue?

sofianehaddad:
#46 fixes the issue, I close the PR

PR #42, opened 2015-10-15
Allow to override CMAKE_INSTALL_RPATH

jschueller:
Even if non-system install prefix is provided,
one may want to build without rpaths.


dbarbier:
Then use `CMAKE_SKIP_RPATH` or `CMAKE_SKIP_INSTALL_RPATH` options.

jschueller:
Right, thank you.

PR #43, opened 2015-10-15
Implement DistributionParameters::inverse

jschueller:
same as #36, in a new branch


sofianehaddad:
+1

jschueller:
UPDATE: only commits relevant for DistributionParameters::inverse

dbarbier:
+1

PR #46, opened 2015-10-20
Fix RPATH handling on Mac OSX and remove SYSTEM_INSTALL

dbarbier:
The canonical way to prevent rpath from being used is to set
CMAKE_SKIP_RPATH or CMAKE_SKIP_INSTALL_RPATH options.


dbarbier:
It this works fine on OSX, pull request #41 is no more needed.

sofianehaddad:
+1, the patch fixes the issue. 
DYLD_LIBRARY_PATH is no more required for installcheck.
Also, Python bidding works fine without specification of the DYLD_LIBRARY_PATH variable
#41 has been closed.

dbarbier:
Great news, thanks for testing.

jschueller:
I think packages will need to be patched to disable rpaths.

jschueller:
+1

PR #47, opened 2015-10-20
Fix DistributionParameters_std test for Python3

jschueller:


jschueller:
My bad, I should have read the test logs more carefully.

sofianehaddad:
+1

PR #49, opened 2015-10-21
User defined covariance model doc

adutfoy:


jschueller:
+1

jschueller:
Could you squash these commit together ? see my comments for #63.

PR #50, opened 2015-10-21
New parametrization for covariance models

sofianehaddad:
Covariance models are available since OpenTURNS 1.0. Few models were implemented (ExponentialModel, SphericalModel) and their usage was restricted to the stochastic processes.

In OpenTURNS 1.3, the kriging regression was implemented and several models have been added to the library (AbsoluteExponential, SquaredExponential, GeneralizedExponential for example). These models were stricly oriented to the estimation of response surface (in fact they were considered as correlation models).

The objective here is to define a coherant structure for all the models using the same ingredients (A correlation model R,  an amplitude vector and a scaling factor)

For now, we consider that the covariance writes Sigma * factor, where the factor is scalar measuring the spatial correlation. This last one depends both on the space points and the scale factor and it is computed thanks to a new **computeStandardRepresentative** method. The general expression of such function is \rho(x/scale, y/scale) where division is vectorial (x_i/scale_i). 
The scale vector should be now of size spatialDimension to map the various amplitude of inputs.
Sigma is a constant matrix

The computeStandardRepresentative method is then the main method to implement.

In parallel, as the spectral model is the Fourier transform of the stationary covariance model, the new parametrization implies also changes in spectral models classes, which have now similar expression i.e. a matrix/complex product where the complex is the Fourier transform of CovarianceModel::computeStandardRepresentative and the matrix is the same constant matrix Sigma as previously.

Same changes have been done in SpectralModel


jschueller:
about 9179a53: getParameters should return a NumericalPoint with the values only, and a new method getParametersDescription should return a Description, to be consistent to the new distribution parameters accessors.

sofianehaddad:
ok I will take it into account

sofianehaddad:
@jschueller take into account your request in 891b7ea 
I updated also ProcessSample docstring because of new parametrization
I fixed also a missing `ot.` but here I wonder how doctest pass in #56

sofianehaddad:
@jschueller ==> 32b0c2b
Otherwise I wonder how master does not fail doctest. I fixed also a missing ot in docstring

sofianehaddad:
@regislebrun, what do you think?

regislebrun:
It looks very good, minus a small naming trouble: in many places in the code we hesitate between getParameters() and getParameter(). Once and for all we should decide on the correct naming. As in PR#53, I suggest to think about *the* parameter of a given concept, the parameter being most of the time of dimension > 1. Sorry for the burden.

jschueller:
I propose to rename getParameters, setParametersDescription & setParameters for Distribution, NMF & DistributionParameters in the compact-parameter branch while @sofianehaddad renames the covariancemodel ones.

sofianehaddad:
If I understand well, getParameters becomes getParameter, this is it @jschueller @regislebrun ?
What about getParametersDescription?

jschueller:
Yes, the proposal is to rename all of them.

sofianehaddad:
ok getParametersDescription moves to getDescription, get|setParameters to get|set Parameter

jschueller:
no, getParametersDescription becomes getParameterDescription

sofianehaddad:
@regislebrun @jschueller done in 3a4be4f

jschueller:
+1

regislebrun:
+1

PR #53, opened 2015-10-22
Compact parameters

jschueller:


sofianehaddad:
+1

jschueller:
@regislebrun, getParameterDimension or getParametersDimension, I see both in the code.

regislebrun:
getParameterDimension(). In several places we hesitate between singular and plural nouns, but singular is more natural when talking about parameter space. And there is only one parameter, with dimension >=1 ;-)!

jschueller:
@regislebrun, thanks, so getparametersdescription must be renamed too!

regislebrun:
I think so. The same naming problem occurs in PR#50 on the new parameterization of the covariance models.

jschueller:
@regislebrun , maybe I shoud rename DistributionParameters too ?

sofianehaddad:
+1

PR #59, opened 2015-10-23
User defined stationary covariance model doc

adutfoy:


sofianehaddad:
+1

jschueller:
Could you squash these commit together ? see my comments for #63.

dbarbier:
Thanks for resolving conflicts; in order to minimize such conflicts, you could put several commits (squashed like this one, not original ones) into the same pull request.  This is not considered as good practice in general, but for documentation, this is acceptable, in my opinion.

BTW your last commit  a4304d1 adds an extra leading space in `python/doc/user_manual/stochastic_process.rst` at l.38,39,41,43,44, I am pretty sure that this has to be fixed too.

dbarbier:
Oops, in fact you fixed indentation, I read diff output backwards, sorry for the confusion.

+1

PR #62, opened 2015-10-23
Initial NLopt support

jschueller:
For experimentation


regislebrun:
+1
It looks very promising! More work is needed eg to document the current implementation of the NLopt class (no docstring) or to provide a richer interface to the collection of algorithms provided by NLopt (currently the use of the SLSQP algorithm is hard-coded).

jschueller:
@regislebrun added doc & more algos.

sofianehaddad:
+1

lib/src/Base/Optim/OptimizationSolverImplementation.cxx
```diff
@@ -30,6 +30,7 @@ OptimizationSolverImplementation::OptimizationSolverImplementation()
   : PersistentObject()
   , startingPoint_(NumericalPoint(0))
   , maximumIterationsNumber_(ResourceMap::GetAsUnsignedInteger( "OptimizationSolverImplementation-DefaultMaximumIteration" ))
+  , maximumEvaluationNumber_(ResourceMap::GetAsUnsignedInteger("OptimizationSolverImplementation-DefaultMaximumEvaluationNumber"))
```
dbarbier: We should be consistent: either maximumIterationsNumber_/maximumEvaluationsNumber_ or maximumIterationNumber_/maximumEvaluationNumber_, I am not sure which one is right, it seems that we prefer singular form.

jschueller:
@dbarbier, ok, I'll deprecate *IterationsNumber methods.

dbarbier:
It may be nice to have a
```
 static bool NLopt::IsAvailable() const
```
method.

dbarbier:
+1, looks good.

lib/src/Base/Optim/OptimizationSolverImplementation.hxx
```diff
@@ -120,7 +124,13 @@ protected:
 private:
   NumericalPoint startingPoint_;
   OptimizationProblem  problem_;
-  UnsignedInteger    maximumIterationsNumber_; /**< Number of outermost iterations (in case of nested iterations) */
+
+  /** Number of outermost iterations (in case of nested iterations) */
+  UnsignedInteger maximumIterationsNumber_;
+
+  /** Maximum function calls */
+  UnsignedInteger maximumEvaluationNumber_;
```
dbarbier: You forgot to handle this new member in load/save/repr methods.

python/test/t_NLopt_std.py
```diff
@@ -0,0 +1,54 @@
...
+                    problem.setEqualityConstraint(ot.NumericalMathFunction(['x1', 'x2', 'x3', 'x4'], ['eq'], ['x4-2']))
+                try:
+                    algo.setProblem(problem)
+                    algo.setStartingPoint(startingPoint)
+                    print('algo=', algo)
+                    algo.run()
+                    result = algo.getResult()
+                    print('x^=', printNumericalPoint(result.getOptimalPoint(), 4))
+                except:
+                    print('-- Not supported: algo=', algo.__class__.__name__, 'inequality=', inequality, 'equality=', equality)
```
dbarbier: Replace `algo.__class__.__name__` by `algo.getClassName()`?

jschueller: Why ? it work with every Python object, imho getClassName shouldn't even be provided in Python.

jschueller:
@dbarbier, about NLopt::IsAvailable, I agree for the principle, but maybe this could be addressed globally (and separately) to apply to each optional dependency xml, tbb, muparser, ...

python/test/t_NLopt_std.py
dbarbier: For consistency reasons, we already use `getClassName` in few python tests, but no `__class__.__name__`.

jschueller: ok

jschueller:
@dbarbier I fixed the use of the new optimsolver attribute, and deprecated  [sg]etMaximumIterationsNumber

regislebrun:
+1

sofianehaddad:
I think you forgot to update ChangeLog

jschueller:
@sofianehaddad fixed

PR #63, opened 2015-10-25
Bernstein copula factory

regislebrun:
This PR implements a non-parametric copula estimator based on Bernstein polynomial approximation of the empirical copula.


jschueller:
Could you squash the BernsteinCopulaFactory commits together and drop the changes for KernelSmoothing and ClaytonCopula and include them in separate PRS ? 

For this you can use
```
git rebase -i HEAD~10
```
Then delete 3066d63 and 8eacc63 commit lines, and replace pick by squash for all other commits except the first one, save.

At the next screen remove all the commit messages of the remaining 8 entries and add one single message.

dbarbier:
+1, but next time, please give more details in commit message, it does not seem obvious at first glance what "improved" means in your last commit;  is it about performance?  Numerical precision?  Something else?

sofianehaddad:
+1

jschueller:
shouldn't be added to DistributionFactory::GetContinuousMultiVariateFactories ?

regislebrun:
@jschueller Not really as it is a non-parametric factory. It is much more like the KernelSmoothing class.

jschueller:
+1

regislebrun:
@dbarbier I agree with you on the need and the interest for detailed commits, but I understood during our last CT that it was not a good practice with github. BTW, here it was a matter of speed for the same precision but a reduced range of arguments.

dbarbier:
@regislebrun @jschueller I do not remember, maybe you discussed this after I left?

IMO when there is a single commit, as in #70, pull request's detailed description must match commit log message, there is no reason to omit details since you already take time to write them.  When there are several commits, there is no single rule, it depends, but we should try when possible to put all relevant details into commit log messages.

jschueller:
Seems there a problem in t_BernsteinCopulaFactory_std.py indentation for python3:
http://autobuilder.openturns.org/openturns/pull/63/squeeze-x86_64-ot-linux/LastTest.log
```
  File "/home/openturns/distcheck/workdir/python/test/t_BernsteinCopulaFactory_std.py", line 13
    ref_copula = coll[i]
                       ^
TabError: inconsistent use of tabs and spaces in indentation
```

jschueller:
You forgot to install BernsteinCopulaFactory.hxx.

jschueller:
Maybe it's time to close http://trac.openturns.org/ticket/73 ?

PR #67, opened 2015-10-26
Improved the symmetry check in MatrixImplementation

regislebrun:
Due to rounding errors, sometimes a symmetric matrix (obtained ie as L^tCL with C symmetric) is not exactly symmetric due to rounding errors. A more permissive check is needed in order to enforce the symmetry of the result in Python.

### DIFF
--- a/lib/etc/openturns.conf.in
+++ b/lib/etc/openturns.conf.in
@@ -193,6 +193,7 @@
 
   <!-- OT::MatrixImplementation parameters -->
   <MatrixImplementation-DefaultSmallPivot value="1.0e-7"     />
+  <MatrixImplementation-SymmetryThreshold value="1.0e-12"    />
 
   <!-- OT::BurrFactory parameters -->
   <BurrFactory-AbsolutePrecision value="1e-12" />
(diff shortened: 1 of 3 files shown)


jschueller:
+1

jschueller:
+1
Same here, next time try to include these comments inside the commit message:
```
Improved the symmetry check in MatrixImplementation

Due to rounding errors, sometimes a symmetric matrix (obtained ie as L^tCL with C symmetric) is not exactly symmetric due to rounding errors. A more permissive check is needed in order to enforce the symmetry of the result in Python.
```

dbarbier:
+1 in general, but shouldn't we use a relative error instead of an absolute error?  Say for instance

```
if ( std::abs(this->operator[](convertPosition(i, j)) - operator[](convertPosition(j, i))) > epsilon * (std::abs(this->operator[](convertPosition(i, j))) + std::abs(operator[](convertPosition(j, i)))) )
```

regislebrun:
@dbarbier Not clear for me which test is the best one (probably none is the best). As we check the elements of a matrix, we should probably compare the magnitude of the difference to a norm of the matrix, feasible but probably very costly and overkill for most typical situations.
In the case of homogeneous magnitudes and small coefficients, the absolute difference is a bad choice if the threshold is not small enough, but in the case of very heterogeneous magnitudes the test could fail using relative error on very small elements even if the difference is not significant at all. So I am doubtful on my choice and on your proposal.

PR #68, opened 2015-10-26
Fixed a bug preventing the symmetrization of the UserDefinedStationar…

regislebrun:
…yCovarianceModel class.
This covariance model must be built using values associated with positive first component of the parameter space, then it is automatically symmeterized wrt the origin.


jschueller:
+1
Next time try to include the details in the commit message.

dbarbier:
+1

PR #69, opened 2015-10-26
Improved p-value initialization in FittingTest

regislebrun:
In several situations, we would like to get at least one model when testing several parametric probabilistic models for given data, even if its p-value is zero (but the model can be estimated from the data).


dbarbier:
+1

Comments in pull requests are volatile, please write explanations into commit messages instead.

sofianehaddad:
+1

PR #70, opened 2015-10-26
Extend WelchFactory to time grid with non-zero start

regislebrun:
The WelchFactory class made the implicit assumption that the time grid associated to the given process sample was starting from 0, resulting in a wrong estimation if it was not the case. Now, it is fixed.


jschueller:
+1

dbarbier:
+1

PR #71, opened 2015-10-27
Tensorized covariance model

sofianehaddad:
The following PR adds a TensorizedCovarianceModel class, which enables to build a mutlivariate covariance model using a collection of univariate and non univariate models with same spatial dimension


sofianehaddad:
@jschueller fixed missing explicit + python test
We should fix same missing explicit in ProductCovarianceModel

jschueller:
@sofianehaddad  I've got a new warning:
```
/home/schueller/projects/openturns/schueller/lib/src/Base/Stat/TensorizedCovarianceModel.cxx: In member function ‘virtual OT::CovarianceModelImplementation::Implementation OT::TensorizedCovarianceModel::getMarginal(OT::UnsignedInteger) const’:
/home/schueller/projects/openturns/schueller/lib/src/Base/Stat/TensorizedCovarianceModel.cxx:292:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
```

sofianehaddad:
@jschueller I should have read the warning before. I will fix it in a separate PR. Thanks for the report

PR #72, opened 2015-10-27
Add new parametric estimation services

jschueller:


jschueller:
In progress.
@regislebrun On the test,  Student distribution parameter estimation yields a non-SPD covariance matrix using NormalFactory, what to do in such case ?

regislebrun:
At a given point, we will have to think more seriously about degenerated multivariate distributions, ie those taking values in a sub-manifold embedded into a higher dimensional space. In the case of elliptical distributions (eg Student and Normal distributions), it means a semi-definite covariance matrix. I have no clear idea on how to do that in a generic way right now.

For the second point, could you ellaborate a little more on your idea please? Maybe through another media (eg email)?

jschueller:
@regislebrun, I canceled the test changes, added docstrings.

jschueller:
@regislebrun, I have 2 questions:

- the covariance of the (normal) distribution of the parameters estimated by maximumlikelihood is analytical from the inverse of the fisher information matrix and evaluated from the original sample, but what about the mean ? it seems to me only boostrap is available:
```
  for (UnsignedInteger i = 0; i < bootstrapSize; ++ i)
  {
    NumericalSample bootstrapSample(experiment.generate());
    Distribution estimatedDistribution(build(bootstrapSample));
    parameterSample.add(estimatedDistribution.getParameter());
  }
  Distribution parameterDistribution;
  if (isRegular) {
    NumericalPoint mean(parameterSample.computeMean());
    Matrix theta(parameterDimension, parameterDimension);
    const NumericalSample pdf(distribution.computePDF(sample));
    const NumericalSample dpdf(distribution.computePDFGradient(sample));
    for (UnsignedInteger i = 0; i < size; ++ i)
    {
      Matrix dpdfi(parameterDimension, 1, dpdf[i].getCollection());
      dpdfi = dpdfi / pdf[i][0];
      theta = theta + dpdfi * dpdfi.transpose() / size;
    }
    CovarianceMatrix covariance(SymmetricMatrix(theta.getImplementation()).solveLinearSystem(IdentityMatrix(parameterDimension) / size).getImplementation());
    parameterDistribution = Normal(mean, covariance);
  } else {
    KernelSmoothing factory;
```

- How do you transport the parameters of the gaussian given the new parametrization ? Use the image of mean and stddev ?
```
Distribution nativeParameterDistribution(nativeResult.getParameterDistribution());
  if (nativeParameterDistribution.getImplementation()->getClassName() == "Normal")
  {
    NumericalPoint nativeMean(nativeParameterDistribution.getMean());
    NumericalPoint nativeStandardDeviation(nativeParameterDistribution.getStandardDeviation());
    parameterDistribution = Normal(parameters.inverse(nativeMean), parameters.inverse(nativeStandardDeviation), CorrelationMatrix(nativeMean.getDimension()));
  }
  else
  {
    UnsignedInteger bootstrapSize = getBootstrapSize();
```
```

regislebrun:
@jschueller Easy question ;-): the mean is simply the estimated value! As your estimator is consistent (ie it converges toward the exact value), its mean must also converge to the mean of the limit... which is equal to the value of the parameter as the limit is constant!
So you have everything to compute the limit distribution: its mean (the parameter estimate) and its covariance (the inverse of the estimate of Fisher's information matrix).

jschueller:
@regislebrun, ok thanks, that was not obvious to me, and for the 2nd item ?

regislebrun:
@jschueller Sorry I missed the second point ;-)! If \theta_n is the parameter estimate in the first parameterization, \Sigma_n its covariance matrix and g is the reparameterization, then the estimate \delta_n of the new parameterization is \delta_n=g(\theta_n) and its asymptotic distribution is Gaussian with mean \delta_n and covariance \Delta\Sigma\Delta^t where \Delta=grad(g)(\theta_n)

jschueller:
@regislebrun, thanks a bunch

jschueller:
@regislebrun, @sofianehaddad, @aurelieladier, @dbarbier, this one is ready for review
the doc of all factories will be completed in another PR if time allows

regislebrun:
+1

sofianehaddad:
+1

PR #75, opened 2015-10-28
NumericalMathFunction API consistency

jschueller:
For consistency with Distribution, CovarianceModel.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,6 +37,7 @@
  * Deprecated OptimizationSolver::setMaximumIterationsNumber in favor of OptimizationSolver::[sg]etMaximumIterationNumber
  * Moved NumericalMathFunction::[sg]etParameters to NumericalMathFunction::[sg]etParameter
  * Moved NumericalMathFunction::parametersGradient to NumericalMathFunction::parameterGradient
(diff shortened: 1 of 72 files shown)


dbarbier:
+1, please document API changes in ChangeLog too.

sofianehaddad:
+1

jschueller:
@dbarbier, ok

jschueller:
@regislebrun, what do you say about this one ?

regislebrun:
We added the flag on the use of the default implementation for gradients because in many occasions we got questions regarding strange results in QuadraticCumul or FORM/SORM during trainings, the reason of the problem being the bad choice of step in finite difference. We (ie Anne and myself) think that it should be kept, or something equivalent should be implemented.
Otherwise it is a big +1

jschueller:
@regislebrun, ok I think you're just interested in the console warning, maybe these methods should be kept 'protected' to the class ?

regislebrun:
Yes indeed.

jschueller:
@regislebrun: update, these are ignored at the python level, also removed the get/setInitial(grad|hess)Implementation that are useless and also incorrectly implemented

regislebrun:
+1

PR #80, opened 2015-10-30
Encapsulate muParser

jschueller:
Allows to compile without muParser headers
See http://trac.openturns.org/ticket/727

Also abstracts the parser idea so no more muparser-specific code in
AnalyticalNumericalMath(Evaluation|Gradient|Hessian)Implementation.cxx


dbarbier:
+1, but new naming is confusing IMO, AnalyticalParser inherits from mu::Parser, and by reading MuParser.hxx, it is quite hard to see why it is named like this.  IMO it is better to swap AnalyticalParser and MuParser names.

jschueller:
Agreed

jschueller:
@dbarbier, ok removed

dbarbier:
+1, thanks

regislebrun:
+1

PR #82, opened 2015-10-31
Added the computeP1Stiffness() method to the Mesh class. It is mandat…

regislebrun:
…ory to compute eg Karhunen-Loeve expansions of stochastic processes when the modes and the covariance model are approximated using P1 Lagrange finite elements.


sofianehaddad:
+1

jschueller:
+1

dbarbier:
+1

One very minor note: your commit message should  have a short title, a blank line, then long description, like this: 

```
   Added the computeP1Stiffness() method to the Mesh class

   It is mandatory to compute eg Karhunen-Loeve expansions of
   stochastic processes when the modes and the covariance model
   are approximated using P1 Lagrange finite elements.
```

jschueller:
yes, could you adjust the commit message ?

PR #83, opened 2015-11-03
TNC to handle problem bounds correctly

jschueller:
If the problem has no bounds, spawn a new instance with infinite bounds.
Also check for constraints in checkProblem instead of callback.
Removed invalid initialization of starting point in case some bounds are
infinite.


regislebrun:
+1

dbarbier:
+1

PR #85, opened 2015-11-04
Fix missing definition of swig template for OptimizationProblem/OptimizationSolver

sofianehaddad:
Small commit that fixes missing swig templates for the new optimization classes


regislebrun:
+1

dbarbier:
+1

PR #86, opened 2015-11-06
Fix standard space origin evaluation

jschueller:
It's a regression from aec7de2


jschueller:
It's a regression from aec7de2b

sofianehaddad:
ok
+1

regislebrun:
+1

jschueller:
cc @adumasphi

PR #87, opened 2015-11-09
Fix quotes/angle brackets in #include lines

dbarbier:
We use quotes for header files embedded in our repository, and angle brackets for third party files, but few files do not follow this convention.

We plan moving header files into an ```openturns``` subdirectory, this commit is needed so that a script can perform this change automatically.


jschueller:
+1

sofianehaddad:
+1

jschueller:
you missed muparser

dbarbier:
Header files which are not installed will not be moved into `openturns/` by my script.  See my branch `db/move-header-files`.

regislebrun:
+1

PR #88, opened 2015-11-10
Use OPENTURNS_SWIG_INCLUDE_DIRS in python/src/CMakeLists.txt

dbarbier:
This is purely cosmetics since top-level CMakeLists.txt defines

  set (OPENTURNS_SWIG_INCLUDE_DIRS ${INCLUDE_PATH}/swig)


sofianehaddad:
+1

PR #89, opened 2015-11-11
Improved the numerical stability and accuracy of ClaytonCopula

regislebrun:
This copula has a CDF/PDF and a random generator badly conditioned for large values of theta (the limit of this copula is the Min copula, which is singular), leading to overflows when implemented naively, and suffers from cancellation for theta close to zero (the limit being the independent copula). The previous implementation was not robust enough to handle these cases with full accuracy.

### DIFF
--- a/lib/src/Uncertainty/Distribution/ClaytonCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/ClaytonCopula.cxx
@@ -115,17 +115,24 @@ NumericalPoint ClaytonCopula::getRealization() const
   // Laplace transform:    speed= 2719399.57105 rng/s, t= 0.367728233337 s
 #ifdef CLAYTON_USE_LAPLACE
   if (theta_ > 0.0)
-   {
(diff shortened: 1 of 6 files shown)


jschueller:
maybe some test cases in python/test/t_ClaytonCopula_std.py with unusual theta values (<1e-8 and <1e-5) could be exposed so each branch of code is explored at least once.

sofianehaddad:
Same remarks as @jschueller + comments on code to keep in mind why/how we introduce these changes

jschueller:
The following tests FAILED:
	182 - cppcheck_ComposedCopula_std (Failed)
	279 - cppcheck_ClaytonCopula_std (Failed)

jschueller:
+1

jschueller:
@dbarbier, @sofianehaddad thoughts on this one, for me it's ready to be merged

sofianehaddad:
seems ok for me, +1

PR #92, opened 2015-11-16
Generalized linear models

sofianehaddad:
The following PR introduces the GeneralizedLinearModelAlgorithm class, which enables to estimate parameters of a generalized linear model and stores results into a GeneralizedLinearModelResult class.

Btw, TNC is set as default optimizer and bounds parameters might be calibrated using new resource map keys.


regislebrun:
+1

dbarbier:
+1

jschueller:
On the autobuilder some test fail:
- either because nlopt is not there
- or on mingw: TestFailed : Value 0.704 is not close enough to 0.702

sofianehaddad:
@jschueller I've already seen it and it's partially fixed on my local branch.
I used  NLopt because TNC reaches its limits. So I'll split tests into standard/hmat/nlopt

PR #98, opened 2015-11-23
DistributionFactory

jschueller:


dbarbier:
.

sofianehaddad:
I think that the PR should be splitted into two distinct PR because 750dbed fixes docstring whereas e20e46a implies more changes and feedback

BTW, I agree with the two commits so it's +1 for me

regislebrun:
+1

jschueller:
ok i'll do the rename in a separate pr

PR #99, opened 2015-11-24
Remove python tests encoding tags

jschueller:
Python 3.5 throws an exception on windows:
SyntaxError: encoding problem: iso-8859-15


dbarbier:
+1

PR #100, opened 2015-11-24
Rewrite the Fortran language check

jschueller:
This allows to set CMAKE_Fortran_COMPILER through the mingw toolchain file
while not triggering fortran language specifics if not needed.


dbarbier:
+1

PR #101, opened 2015-11-26
Fix empty range segfault in drawDiscreteCDF

jschueller:
When the range of drawDiscreteCDF is disjoint from the support
the size variable is zero and leads to a segfault:

import numpy as np
import openturns as ot
mysample1 = np.arange(0,10,0.1)
mysample = np.split(mysample1,len(mysample1))
myCDF = ot.VisualTest_DrawEmpiricalCDF(mysample,10,100)


regislebrun:
+1

PR #102, opened 2015-11-28
Improved the draw() method of Field

regislebrun:
Now, for fields with spatial dimension 1 and with dimension 2, the default draw() method plots the second output as a function of the first output.


jschueller:
+1

dbarbier:
+1

PR #115, opened 2015-12-08
Use std::abs(double) instead of abs(int)

jschueller:
abs comes from stdlib.h on operates on int


jschueller:
Maybe we could agree to use std::abs everywhere, which is more c++-ish (224 occurences) for consistency instead of fabs (215 occurences), which comes from math.h (c lib).

regislebrun:
Agree. Who is volunteer for the task?

jschueller:
I can do it, it's just a sed command, but I refrained from doing so because of all the active branches, maybe I could do it after the dust settles.

dbarbier:
+1, and there is another occurence in `GeneralizedLinearModelAlgorithm.cxx`.

jschueller:
@dbarbier, yes @sofianehaddad took care of it

dbarbier:
@jschueller There is no PR yet, it is better to fix it in this PR.

jschueller:
@dbarbier, @sofianehaddad, ok, reverted

PR #116, opened 2015-12-08
Multidimensional piecewise linear function

regislebrun:
It is the Karhunen-Loeve companion!


jschueller:
How about the KarhunenLoeveP1Factory::build commented loop/return problem I told you about,
wasn't supposed to be addressed with this PR ?

```
Basis KarhunenLoeveP1Factory::build(const CovarianceModel & covarianceModel,
                                     NumericalPoint & selectedEV) const
{
   const ProcessSample resultAsProcessSample(buildAsProcessSample(covarianceModel, selectedEV));
   Basis result(0);
   for (UnsignedInteger i = 0; i < resultAsProcessSample.getSize(); ++i)
     //    result.add(P1LagrangeEvaluationImplementation(resultAsProcessSample.getField(i)));
   return result;
}
```

regislebrun:
@jschueller I already had a look at python/src/OrthogonalUniVariatePolynomialFamily.i but I am a bit confused by the exact meaning of numinputs, $1 and *$1. In the case of OrthogonalUniVariatePolynomialFamily there is only one argument and it is the input/output argument. In my case I have 3 arguments and only the last one is an input/output argument.

Concerning the build() method, it should have landed to github. I may have miss something working so late at night...

jschueller:
It is only relative to the current set of parameters wrapped, so you can safely ignore that and copy-paste almost everything.

jschueller:
as for the build() method, I dont see it in that PR

jschueller:
+1

jschueller:
@dbarbier, @sofianehaddad, thoughts on this ? afaic it's ready to go to master

sofianehaddad:
it seems that one part of f4fadc9 has been done in 8214235, thus the two commits should be squashed or split should be done correctly
Otherwise it is ok for me

regislebrun:
@sofianehaddad Ok I squash them. No time to learn how to split things.

sofianehaddad:
+1

PR #117, opened 2015-12-09
Fix Interval::intersect for mixed finite/infinite bounds

jschueller:
Method Interval::intersect did not take infinite bounds properly into
account and returned the max/min of the bounds regardless of the finite
flag.
It seems to impact RandomMixture and Distribution arithmetic
multiplication.

In cases both bounds from both interval are infinite, we set the
corresponding bound value to the numerical limit.
In truncated distribution class, we need to override
this behavior by manually setting min/max on distribution support values.


jschueller:
@regislebrun In fact it is still a bit unclear to me what value we should return when both bounds are infinite. +-inf or the actual min/max of the value? We're trying to solve a bug in TruncatedDistribution, and it seems that the range bound value (even if the finite flags are unset) are important for the computation made for distributions.

regislebrun:
@jschueller The interval contains two kind of information: the first is as close as possible to the mathematical concept and the second is the numerical aspect. When a bound is finite, the numerical and theoretical are the same, but when a bound is infinite, its mathematical value is +/-inf while its numerical value is what it is reasonable to consider as a "numerical" infinite value.

To make things more understandable, consider the distribution N(0,1). Its range is (-inf,+inf) but numerically it is [-7.65063,7.65063] because Phi(7.65063) is the largest double precision number less than 1.

A more involved example is the following: consider the N(10, 1) distribution. Its range is also (-inf,+inf) but numerically it is [2.34937, 17.6506]: in this case, -inf is numerically 2.34937, a positive number!

It is of uttermost importance to insure that these numerical values are meaningful from a numerical point of view, because many numerical integration or root finding algorithms rely on these values and fail if we simply take +/-inf = +/-SpecFunc::MaxNumericalScalar.

Send me specific examples if you need more assistance on debugging this part of the code.

jschueller:
@regislebrun here's an example (-inf(-1); +inf(1)) vs (-1; 2):
```
In [3]: ot.Interval([-1],[1],[False], [False]).intersect(ot.Interval([-1],[2],[True],[True]))
Out[3]: class=Interval name=Unnamed dimension=1 lower bound=class=NumericalPoint name=Unnamed dimension=1 values=[-1] upper bound=class=NumericalPoint name=Unnamed dimension=1 values=[1] finite lower bound=[1] finite upper bound=[1]
```
returns (-1; 1)

regislebrun:
@jschueller Yes, it looks weird, but it is the expected behavior if you want the algorithm to run correctly. If you define the interval (-inf,+inf) as being numerically equal to [-1,1] (because it is the range of an unbounded distribution very concentrated around 0), its intersection with [-1,2] can only be [-1,1], and it is bounded because the intersection between an interval and a bounded interval is a bounded interval.
If you want to get [-1,2] you have to define (-inf,+inf) as being numerically equal to [-SpecFun.MaxNumericalScalar, SpecFunc.MaxNumericalScalar].

regislebrun:
-1

PR #123, opened 2015-12-14
Prepare release

jschueller:


jschueller:
@dbarbier, your fix is ok (and a lot cleaner), thanks.

PR #125, opened 2015-12-16
Fix conversion from sphinx to PDF 

dbarbier:
Output file is far from being perfect, but at least, it can be produced now.


regislebrun:
+1

sofianehaddad:
+1

jschueller:
no, but you can smash one more digit

PR #128, opened 2015-12-22
Use of new static constructor Factory<T> for some classes

sofianehaddad:


sofianehaddad:
@jschueller ok
I will update it

jschueller:
+1

PR #130, opened 2015-12-24
Renamed the computeP1Stifness() method into computeP1Gram() in Mesh.

regislebrun:
Sofiane was right when he told me that the P1 Gram matrix should be called the mass matrix instead of the stiffness matrix (if one want to stick with mechanical wording). In fact it is the Gram matrix of P1 functions (ie the matrix of dot products) so the change of name.


sofianehaddad:
+1

PR #131, opened 2015-12-24
Allowed to build MaximumLikelihoodFactory using a distribution

regislebrun:
Most of the interest of the MaximumLikelihoodFactory is to have a generic factory for distributions with no specific factory. In the previous implementation, this class was only an helper class to ease the work of the C++ programmer to implement its factory, now it can be used out of the box by a Python user.


jschueller:
What about removing the constructor from a factory ?

regislebrun:
ok.

jschueller:
+1

sofianehaddad:
+1

dbarbier:
+1

PR #137, opened 2016-01-06
Product distribution cdf

regislebrun:
This PR provides a specific implementation of the computeCDF() and computeProbability() methods for ProductDistribution.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -81,11 +81,13 @@
  * #784 (Troubles with UserDefinedFactory/UserDefined)
  * #790 (AbdoRackwitz parameters)
  * #796 (Beta distribution: if sample contains Inf, freeze on getSample)
+ * #797 (computeProbability might be wrong when distribution arithmetic is done)
  * #798 (Error message misstyping (Gamma distribution))
  * #799 (Error message misstyping (Gumbel distribution factory))
  * #800 (Exponential distribution built on constant sample)
  * #804 (no IntervalMesher documentation content)
  * #805 (Python segfault in computeSilvermanBandwidth)
+ * #806 (DistributionImplementation::computeCDFParallel crash)
  * #808 (Index check of SymmetricTensor fails when embedded within a PythonFunction)
  * #812 (Sphinx documentation build error)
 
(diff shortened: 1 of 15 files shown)


jschueller:
Some tests take too much time and the autobuilder timeout (100s):
```
The following tests FAILED:
        241 - cppcheck_ProductDistribution_std (Timeout)
        733 - pyinstallcheck_Distribution_arithmetic (Timeout)
```

On Travis there's no barrier for now:
```
313/824 Test #241: cppcheck_ProductDistribution_std .......................................   Passed  197.53 sec
824/824 Test #733: pyinstallcheck_Distribution_arithmetic .................................   Passed  528.27 sec
```

dbarbier:
IMO changes in `Normal::computeCDF` should go into another PR.  If I am not mistaken, all other changes are conservative, and thus tests output should be unchanged.

jschueller:
I added a timeout on Travis with #139.

jschueller:
I agree with @dbarbier to split the changes, especially the part that fixes http://trac.openturns.org/ticket/806 in a separate commit to be backported in the 1.7 branch.

jschueller:
If the adaptative quadrature changes are independent from the ProductDistribution changes, they should go into separate commits including changes to the respective tests and changelog entries (you could mention http://trac.openturns.org/ticket/806 and http://trac.openturns.org/ticket/797 in the 1.7 bug fixes section if they should be backported to 1.7 branch).
In my opinion we should have also a separate commit for the Mixture changes.

regislebrun:
@jschueller I agree with you, but the problem is that I don't know how to split commits afterward. Any clue?

jschueller:
you could save patch somewhere: ```git format-patch HEAD~1``` which write a .patch file
then go back to the previous commit ```git reset --hard _commithash_```
re-apply the patch with ```patch -p1 -i 0001-xxxxxx.patch```
finally apply only the relevant changes in differents commits

jschueller:
+1, it's much easier to see what's going on, so d832997 is related to #797, and 3f773f9 is for #806, plus 153b339 as an extra bugfix.

PR #139, opened 2016-01-07
Set a global timeout on all tests

jschueller:


dbarbier:
-1, separate lines are much more readable

jschueller:
@dbarbier, I want to avoid make tests & ctest for a quicker build result if make fails.

dbarbier:
Ok, for reasons I do not understand, travis build does not abort when a command fails :(
Thus +1 from me, sorry for the confusion.

regislebrun:
+1

PR #141, opened 2016-01-11
Improved the computeCDF() method of Normal

regislebrun:
Now, this method uses adaptive quadrature instead of fixed Gauss-Legendre quadrature for moderate dimensions.


jschueller:
+1

dbarbier:
or ```std::log(1.0 * numberOfPoints)```, as in KernelSmoothing.cxx for instance

dbarbier:
+1

jschueller:
cppcheck_Normal_std goes from 0.5s to ~70s on both autobuilder & travis, is it still ok ?

regislebrun:
The previous implementation used a low-order nonadaptive Gauss quadrature for moderate dimensions (4-6), fast but not accurate. The adaptive quadrature gives much more accurate results, but takes ages (probably due to the huge amount of NMF created by binding methods). I reverted the implementation to a fixed Kronrod quadrature and a parallel evaluation of the PDF on the integration nodes. It gives 13 correct digits in dimension 4 in 0.29s on my laptop, 8 in dimension 6 in 3.9s.

dbarbier:
+1

jschueller:
+1

PR #142, opened 2016-01-12
Fix ProductCovarianceModel::getParameterDescription

sofianehaddad:
Use of new convention to set the description which is of size spatialDimension + 1


jschueller:
+1

dbarbier:
+1

regislebrun:
+

regislebrun:
+1

PR #144, opened 2016-01-18
Fix duplicate contour levels

jschueller:
Fixes the default BayesDistribution pdf graph with uniform
distributions with matplotlib.

Matplotlib 1.5.1 now turns this into an error:
https://github.com/matplotlib/matplotlib/pull/5485


jschueller:
See http://trac.openturns.org/ticket/812 (added in commit log)

PR #148, opened 2016-01-26
Publish only from master branch

jschueller:
The previous code did not exit:
```- test "$TRAVIS_PULL_REQUEST" != "true" || exit 0```
As a consequence PRs as well commits to other branches triggered doc
updates


regislebrun:
+1

PR #149, opened 2016-02-03
Fix matrix multiplication operators

jschueller:


sofianehaddad:
+1

regislebrun:
It would be a good thing to be able to multiply real matrices with complex matrices.. Otherwise ok for this PR.

jschueller:
@regislebrun, Matrix/ComplexMatrix product is now possible

jschueller:
@regislebrun, I left aside the matrix/complexmatrix multiplication, see http://trac.openturns.org/ticket/802

PR #150, opened 2016-02-04
Renamed QuadraticCumul to TaylorExpansionsMoments

jschueller:


dbarbier:
+1

jschueller:
@dbarbier, you're right it's only deprecated in Python, see TaylorExpansionsMoments.i

PR #154, opened 2016-02-18
Shorter EllipticalDistribution exception message

jschueller:


jschueller:
@ladier, this fixes the ugly error message we've been having

sofianehaddad:
+1

PR #156, opened 2016-02-23
Renamed __contains__ to contains

jschueller:


regislebrun:
+1

dbarbier:
Is this method really useful?  It is equivalent to `getValues().__contains__(val)`.

jschueller:
@dbarbier, indeed it seems FieldImplementation.contains & TimeSeries are similar ; however there's a difference : TimeSeries iterates until n_ instead of getSize.

dbarbier:
Well, my comment was only about FieldImplementation::contains, I found it strange to have a `contains` method because FieldImplementation is a complex container. In Java, List has a `contains` method, but Map has a `containsValue` method (and `containsKey` too), this seems more logical to me.

When looking at FieldImplementation::contains source code to see what this method is doing exactly, I found that it is strictly equivalent to calling `getValues().contains(val)`, thus I was questioning the utility of this method in this class.

BTW since you mention that, it seems to me that TimeSeries::getSize() returns `n_`, so there is no need to redefine it in TimeSeries.

jschueller:
@dbarbier, I completely agree, removed

dbarbier:
+1

PR #157, opened 2016-02-23
New orthogonal basis

jschueller:


jschueller:
@regislebrun, I wrote a simpler Log2

regislebrun:
No! The previous one was delightful! My previous comment was a compliment!

jschueller:
@regislebrun, ok, I put it back, with comments

regislebrun:
I have some doubts on the implementation of the CanonicalTensor class. It will have to store a collection of DesignProxy that will be shared among the different rank one tensors.

regislebrun:
+1

PR #162, opened 2016-03-01
Fix ComposedDistribution::computeSurvivalFunction

jschueller:
...with a non-independent copula

### DIFF
--- a/lib/src/Uncertainty/Distribution/ComposedDistribution.cxx
+++ b/lib/src/Uncertainty/Distribution/ComposedDistribution.cxx
@@ -376,7 +376,13 @@ NumericalScalar ComposedDistribution::computeCDF(const NumericalPoint & point) c
 
 NumericalScalar ComposedDistribution::computeSurvivalFunction(const NumericalPoint & point) const
 {
-  /* Survival = copula_survival(dist1_survival(p1), ..., distn_survival(pn)) */
+  /* Survival = \hat{F}(x_1, \dots, x_d)
+   *          = \hat{C}(\hat{F}_1(x_1), \dots, \hat{F}_d(x_d))
+   *          = \bar{C}(1-\hat{F}_1(x_1), \dots, 1-\hat{F}_d(x_d))
+   *          = \bar{C}(F_1(x_1), \dots, F_d(x_d))
+   *
+   * With \bar{C} the survival function of the copula, not to be mistaken with the survival copula \hat{C}
+   */
   const UnsignedInteger dimension(getDimension());
   if (point.getDimension() != dimension) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=" << dimension << ", here dimension=" << point.getDimension();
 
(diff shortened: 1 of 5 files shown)


dubourg:
Why so brutal? ;)

Eventually, it seems your fix is correct. Check [this extract from Nelsen's book](https://books.google.fr/books?id=yexFAAAAQBAJ&pg=PA33&dq=nelsen+copula+survival+function&hl=fr&sa=X&ved=0ahUKEwjWvunKx6HLAhXFPhQKHfqbDDsQ6AEIKjAA#v=onepage&q=nelsen%20copula%20survival%20function&f=false).
cc @whereIsRegis?!

jschueller:
I was about to check EllipticalDistribution, maybe this code deserves a comment:
```
NumericalScalar EllipticalDistribution::computeSurvivalFunction(const NumericalPoint & point) const
{
  return computeCDF(2.0 * mean_ - point);
}
```

dubourg:
> Care should be taken not to confuse the survival copula $\hat{C}$ with the joint survival function $\bar{C}$ for two uniform (0, 1) random variables whose joint distribution function is the copula $C$.
Note that $\bar{C}(u, v) = P[U > u, V > v] = 1 - u - v + C(u, v) = \hat{C}(1 - u, 1 -v)$.

-- Nelsen (1999)

dubourg:
My guess is copula::computeSurvivalFunction implements the survival function $\bar{C}$ of the copula $C$ as its name suggests!

Hence your fix is correct, which means the composed distribution's joint survival function reads:
S(x_i, i=1, ..., n) = \bar{C}(F_i(x_i), i=1, ..., n)

Let's wait for Régis review...

regislebrun:
Looks like a bug in CopulaImplementation::computeSurvivalFunction(). I have a look at it ASAP.

regislebrun:
The CopulaImplementation::computeSurvivalFunction() is correct: it computes the survival function \bar{C} of  the copula C, which is NOT the survival copula \hat{C} as noted by Vincent. So the bug is in the ComposedCopula class: the survival function \hat{F} is such that:
\hat{F}(x_1,\dots,x_d)=\hat{C}(\hat{F}_1(x_1),\dots,\hat{F}_d(x_d))
                                  =\bar{C}(1-\hat{F}_1(x_1),\dots,1-\hat{F}_d(x_d))
                                  =\bar{C}(F_1(x_1),\dots,F_d(x_d))
so the correction proposed by Julien is perfectly right. Thanks for the report, for the fix and shame on me for the bug!

regislebrun:
+1

jschueller:
@regislebrun, @dubourg, thank you for your feedback on this

dubourg:
Good job @jschueller! :+1: Thank you all.

dbarbier:
As shown by this discussion, comments in this portion of code should be added to avoid confusion.

jschueller:
@dbarbier, replaced the erroneous comments

PR #169, opened 2016-03-17
Add subset-sampling

jschueller:
This integrates code from the otsubset module, see
Au, S. K. "Estimation of small failure probabilities in high
dimensions by subset simulation"


lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
```diff
@@ -0,0 +1,572 @@
...
+      // conditional sampling
+      TruncatedDistribution truncatedChiSquare(ChiSquare(dimension_), betaMin_ * betaMin_, TruncatedDistribution::LOWER);
+      Normal normal(dimension_);
+      inputSample = NumericalSample(0, dimension_);
+      for (UnsignedInteger j = 0; j < blockSize; ++ j)
+      {
+        NumericalPoint direction = normal.getRealization();
+        NumericalScalar norm = direction.norm();
+        NumericalScalar radius = sqrt(truncatedChiSquare.getRealization()[0]);
+        if (fabs(norm) > 1e-12)
```
regislebrun: norm is always positive. No fabs. BTW, std::abs is prefered.

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
```diff
@@ -0,0 +1,572 @@
...
+  for (UnsignedInteger i = 0; i < maximumOuterSampling; ++ i)
+  {
+    NumericalSample inputSample;
+    if (!iSubset_) {
+      // crude MC
+      inputSample = standardEvent_.getAntecedent()->getDistribution().getSample(blockSize);
+    }
+    else {
+      // conditional sampling
+      TruncatedDistribution truncatedChiSquare(ChiSquare(dimension_), betaMin_ * betaMin_, TruncatedDistribution::LOWER);
```
regislebrun: Here you need a truncated Chi distribution, not a truncated ChiSquare distribution. The lower bound is betaMin_.

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
```diff
@@ -0,0 +1,572 @@
...
+      Normal normal(dimension_);
+      inputSample = NumericalSample(0, dimension_);
+      for (UnsignedInteger j = 0; j < blockSize; ++ j)
+      {
+        NumericalPoint direction = normal.getRealization();
```
regislebrun: NumericalPoint direction(normal.getRealization());

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
```diff
@@ -0,0 +1,572 @@
...
+    currentThreshold = getEvent().getThreshold();
+  }  
+  thresholdPerStep_.add(currentThreshold);
+
+  // compute monte carlo probability estimate
+  NumericalScalar probabilityEstimate = computeProbability(1.0, currentThreshold);
+
+  if (iSubset_)
+  {
+    NumericalScalar correction = 1.0 - ChiSquare(standardEvent_.getImplementation()->getAntecedent()->getDistribution().getDimension()).computeCDF(betaMin_ * betaMin_);
```
regislebrun: Use computeComplementaryCDF here.

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
```diff
@@ -0,0 +1,572 @@
...
+      // generate a new point
+      NumericalPoint oldPoint(currentPointSample_[i * blockSize + j]);
+      NumericalPoint newPoint(oldPoint + randomWalk.getRealization());
+
+      // 1. accept / reject new components
+      NumericalPoint uniform(RandomGenerator::Generate(dimension_));
+      for (UnsignedInteger k = 0; k < dimension_; ++ k)
+      {
+        // compute ratio
+        NumericalScalar ratio = std::min(1.0, exp(0.5 * (oldPoint[k] * oldPoint[k] - newPoint[k] * newPoint[k])));
```
regislebrun: No need to take the minimum as the next test will reject the proposal as soon as the ratio is not less than 1

jschueller:
@regislebrun, the formula was correct with the sqrt, but I could simplify
I hesitated to rename to SubsetSimulation, what do you think ? Is DirectionalSampling correct ?

regislebrun:
@jschueller my point was: using ChiSquare you need sqrt, using Chi you don't. Concerning the names, most of the algorithms have the 'Sampling' suffix. It may be the time to change everything in favor of 'Simulation' but it will break many scripts. I prefer to keep the 'Sampling' suffix for all the algorithms.

dbarbier:
+1

PR #171, opened 2016-03-18
Fix analytical function test on Py3.5

jschueller:
We're not testing python3.5 yet, just 2.7/3.1 on the autobuilder.


adutfoy:
+1

regislebrun:
+1

PR #172, opened 2016-03-18
Distribution getprobabilities

jschueller:


lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -2458,6 +2458,15 @@ NumericalSample DistributionImplementation::getSupport() const
   return getSupport(getRange());
 }
 
+/* Get the discrete probability levels */
+NumericalPoint DistributionImplementation::getProbabilities() const
+{
+  if (!isDiscrete())
+    throw InternalException(HERE) << "Error: cannot return probability levels of a non discrete distribution.";
+
+  return computeCDF(getSupport()).getImplementation()->getData();
```
adutfoy: -1 It is the PDF which gives the probability to take a given value for a discrete random vector, not the CDF

regislebrun:
No chance to get the C++ tests updated? There are still situations in which python tests fail due to a bug in the swig layer, and a C++ test is very useful to detect these situations.

regislebrun:
Otherwise the PR is ok.

jschueller:
@regislebrun, tested in c++ too.

regislebrun:
+1

sofianehaddad:
+1

jschueller:
Turns out I dont need it for otrobopt.

regislebrun:
But this service is useful even if it is not needed by your development! It is the natural counterpart of getSupport(), why not add it to the platform?

PR #174, opened 2016-03-21
Parametric function

jschueller:
- 2056158: The parameters description did not take into account the parameters
position indices.

- d0e3ee7: The referencePoint argument is now only the values of the parameters,
previously, to define a parametric function F(x1,x2=4) from f(x)
referencePoint had to be of size the dimension of x:
   f=NMF(['x1', 'x2'], ['y'], ['x1+x2])
   set = [1]
   ref = [0.0, 4.0]
   F=NMF(f, set, ref)
now ref is only of size the size of set, the parameter indices in x:
   ref = [4.0]
   F=NMF(f, set, ref)

- e842457: The parametric function constructor without parameters values
is removed, as the user may not know they have been set to zero.


jschueller:
@regislebrun, also, maybe the constructor without referencePoint is a bit dangerous ?

regislebrun:
@jschueller ok, remove this constructor. It will break the script on sequential Monte Carlo for robust optimization, but it is not critical.

jschueller:
@regislebrun it is removed.

regislebrun:
+1

jschueller:
@dbarbier, @sofianehaddad, any thoughts here ?

regislebrun:
@dbarbier, @sofianehaddad, any thoughts here ? We need this change for the robust optimization module...

sofianehaddad:
Seems ok for me, +1

PR #177, opened 2016-03-23
Use adaptive Stieltjes by default in StandardDistributionPolynomialFactory

regislebrun:
The adaptive Stieltjes algorithm is the most robust algorithm to generate orthonormal polynomials wrt an arbitrary 1D continuous distribution, so it is a better choice than the modified Gram Schmidt algorithm as a default algorithm for StandardDistributionPolynomialFactory


jschueller:
First commit message is too long:
"StandardDistributionPolynomialFactory to default to Stieltjes algo" would avoid line wrapping.

dbarbier:
+1
BTW this change should appear in ChangeLog IMHO, as output is modified.

PR #179, opened 2016-03-24
Pythondistribution parallel

jschueller:
Python memory management is not thread-safe and parallelization
cannot be used to wrap the callbacks.


regislebrun:
+1

dbarbier:
+1

PR #180, opened 2016-03-30
Deprecate distribution pairs

jschueller:
Using array accessors is easier, from both c++ and Python apis,
in order to define an Histogram or UserDefined distribution.


lib/src/Uncertainty/Distribution/openturns/Histogram.hxx
```diff
@@ -111,6 +111,12 @@ public:
   void setFirst(const NumericalScalar first);
   NumericalScalar getFirst() const;
 
+  /** Data accessor */
+  void setData(const NumericalPoint & l,
+               const NumericalPoint & h);
+  NumericalPoint getL() const;
```
dbarbier: Why l and h member names, and not width/height?

jschueller: @dbarbier, ok, i'll rename the attributes

regislebrun:
+1

jschueller:
@dbarbier, @sofianehaddad, any thoughts here ?

jschueller:
@dbarbier, @sofianehaddad, any thoughts here ?

jschueller:
@dbarbier, @sofianehaddad, any thoughts here ?

sofianehaddad:
seems ok, +1

PR #183, opened 2016-03-31
Made more methods virtual in NumericalMathFunctionImplementation

jschueller:
It allows to define locally specializations of NumericalMathFunction without sticking on the pattern EvaluationImplementation/GradientImplementation/HessianImplementation, typically to avoid the use of the bindMethod trick which is SLOW.


jschueller:
cherry-picked from lebrun/avoid-bindmethod, needed for otrobopt

regislebrun:
+1

PR #184, opened 2016-03-31
Add NumericalMathFunction::setParameter(NP)

jschueller:
The method is available in NumericalMathFunctionImplementation but not
in NumericalMathFunction.


sofianehaddad:
+1

regislebrun:
+1

PR #187, opened 2016-04-05
SquareMatrix from SymmetricMatrix conversion

jschueller:
Adds a contructor to SquareMatrix from SymmetricMatrix,
which symmetrizes the underlying implementation.
Fixes http://trac.openturns.org/ticket/822.
Extended to complex matrix.


jschueller:
As a follow-up, I saw that the Hermitian matrix accessor doesn't allow to modify the upper diagonal.
Either we allow it for HermitianMatrix, or we deny it for SymmetricMatrix too:
```
sm = ot.HermitianMatrix(2)
sm[0,1] = 0.3j
TypeError: InvalidArgumentException : Error: only the lower triangle of an Hermitian matrix can be filled directly.
```

sofianehaddad:
+1

regislebrun:
-1 For an Hermitian matrix, the entries on the upper triangle are the conjugates of the entries in the lower triangle. You cannot simply store M_{ij} at M_{ji} when i<j.

jschueller:
yes, that was stupid

regislebrun:
+1

PR #188, opened 2016-04-06
Fix hasIndependentCopula() and hasEllipticalCopula()

regislebrun:
By default all the distributions were declared as having an independent and elliptical copula, resulting in a wrong behaviour for user defined multivariate distributions (eg using PythonDistribution).


jschueller:
+1

dbarbier:
+1

PR #190, opened 2016-04-11
Add two samples Kolmogorov test

jschueller:


regislebrun:
@jschueller There must be a more efficient algorithm than the O((n1+n2)^2) you implemented. Otherwise it is a very good contribution.

regislebrun:
@jschueller You have useful information here:
https://commons.apache.org/proper/commons-math/apidocs/src-html/org/apache/commons/math3/stat/inference/KolmogorovSmirnovTest.html

lib/src/Uncertainty/StatTests/FittingTest.cxx
```diff
@@ -287,6 +287,47 @@ TestResult FittingTest::Kolmogorov(const NumericalSample & sample,
...
+  NumericalScalar value = 0.0;
+  for (UnsignedInteger i = 0; i < sizeAll; ++i)
+  {
+    NumericalScalar cdf1 = 0.;
+    for (UnsignedInteger j = 0; j < size1; ++j)
+    {
+      if (sampleAllSorted[j][0] <= sampleAllSorted[i][0])
+      {
+        cdf1 = (1.0 * j + 1.0) / (1.0 * size1);
+      }
```
dbarbier: Since sampleAllSorted is sorted, I guess inner loop can be discarded when this test fails by adding
```
    else break;
```
Ditto below for cdf2.

dbarbier:
+1

lib/src/Uncertainty/StatTests/FittingTest.cxx
dbarbier: PS1: or better:
```
   for (UnsignedInteger j = 0; j < size1 && sampleAllSorted[j][0] <= sampleAllSorted[i][0]; ++j)
   {
     cdf1 = (1.0 * j + 1.0) / (1.0 * size1);
   }
```

PS2: it may be a good idea to put `sampleAllSorted[i][0]` into a local variable

jschueller:
@regislebrun There seems to be what I need in the function at line 326 in the file you pointed,
but it seems I'm missing something to make it work.
That function is supposed to return supD=n*m*D_{n,m}, but if I use supD/(n*m)  I can't seem to get the right value to pass to pKomolgorov.

dbarbier:
My previous comment still stands, you should exit on inner loops as soon as test on sampleAllSorted being small fails.

jschueller:
@dbarbier, sampleAllSorted is not always sorted ; it is only the concatenation of two sorted samples.

dbarbier:
Agreed, but each loop iterates over a single sorted block.  Or in other words, once test is false, all subsequent tests within the same inner loop are also false.

jschueller:
@dbarbier, ok

regislebrun:
+1 I think a more efficient algorithm is possible but this new version is reasonably good as a first try. No time to study in details the reference to propose something better...

PR #193, opened 2016-04-20
Use assignment operator for simple datatypes

dbarbier:
For performance reasons, constructors (and not assignments) are called to instantiate any object in local variables.  Current coding policy also enforces using constructors for plain old data structures.  There is no performance gain in this case, the only reason is consistency.

But several of us found assignment operators more readable.  Here are two commits to replace all occurences of
```
  const UnsignedInteger var(...);
```
by
```
  const UnsignedInteger var = ...;
```
Since lots of files are modified, this PR will conflict with any other PR.  So it should not be merged as is, but regenerated by running commands mentioned in these commits.


jschueller:
+1, will apply this before branching 1.8

regislebrun:
+1

adutfoy:
+1

jschueller:
see #271

PR #194, opened 2016-04-20
Split function parameters value and description

jschueller:
This adds [gs]etParameterDescription
and modifies [gs]etParameter to return an NP instead of a
NPWithDescription for functions/evaluation/gradient/hessian classes.

This change make the API consistent with Distribution and
CovarianceModel.

It also fixes a problem with the overload of setParameter in several evaluation
implementation (discovered it with ComposedNMEI fot robopt module).
The evaluation implementation would have to overload both
setParameter(NP) and setParameter(NPWD) for the accessor to work.
I prefered to have a single accessor, and it should be a little faster
than copying NPWD.

Also some implementations of setParameter were missing in evaluation.


jschueller:
Test case for the bug (expected result is 801 instead of 3001):
```
    Description inVar(Description::BuildDefault(4, "x"));
    Description outVar(Description::BuildDefault(3, "y"));
    Description formulas(3, "5");
    formulas[0] = "x0 + 10*x1 +100 *x2 + 1000*x3";
    NumericalMathFunction f(inVar, outVar, formulas);
    Indices set(1);
    set[0] = 3;
    Bool parametersSet(true);
    NumericalPoint x(inVar.getSize()-set.getSize());
    x[0] = 1.0;
    NumericalPoint referencePoint(1, 3.0);
    // with reference point
    NumericalMathFunction g2(f, set, referencePoint, parametersSet);
    NumericalMathFunction g2_0(g2.getMarginal(0));
    NumericalPoint theta(1, 0.8);
    fullprint << "g2_0(x)=" << g2_0(x, theta) << std::endl;
```

sofianehaddad:
+1

dbarbier:
+1

regislebrun:
+1

PR #195, opened 2016-04-21
Implement NMF::parameterGradient

jschueller:
needed for otrobopt


lib/src/Base/Func/NumericalMathEvaluationImplementation.cxx
```diff
@@ -285,7 +285,35 @@ HistoryStrategy NumericalMathEvaluationImplementation::getHistoryOutput() const
...
+  Matrix grad(parameterDimension, outputDimension);
+  for (UnsignedInteger i = 0; i < parameterDimension; ++ i)
+  {
+    for (UnsignedInteger j = 0; j < outputDimension; ++ j)
+    {
+      grad(i, j) = (outS[1 + i][j] - outS[0][j]) / epsilon;
+    }
+  }
+  return grad;
 }
```
regislebrun: To rewrite when operator(const NP &, const NS &) will be available ;-)

jschueller:
@regislebrun, yes, that was the plan

regislebrun:
+1

sofianehaddad:
+1

PR #196, opened 2016-04-21
Change behavior of NumericalSampleImplementation::sort

dbarbier:
After a container is being sorted, one can assume that
   c[i] <= c[j] for all i <= j
Previous implementation was breaking this assumption.  This fixes trac#824.

Also add a NS::unique() method to sort a NS and remove duplicated points.


jschueller:
+1

regislebrun:
+1

jschueller:
you can mention in the changelog that it closes http://trac.openturns.org/ticket/824

dbarbier:
I had to rename unique() into sortUnique() because this definition shadows Pointer::unique().

jschueller:
+1

regislebrun:
+1

PR #197, opened 2016-04-22
Add NMF::operator(NP x, NS parameters)

jschueller:
Overloaded in ParametricEvaluation to efficiently
evaluate a function on a single point and multiple parameters.


lib/src/Base/Func/ParametricEvaluationImplementation.cxx
```diff
@@ -187,6 +187,31 @@ NumericalSample ParametricEvaluationImplementation::operator() (const NumericalS
...
+  NumericalSample input(size, inputDimension);
+  for (UnsignedInteger i = 0; i < size; ++i)
+  {
+    for (UnsignedInteger j = 0; j < parametersDimension; ++j) input[i][parametersPositions_[j]] = parameters[i][j];
+    for (UnsignedInteger j = 0; j < pointDimension; ++j) input[i][inputPositions_[j]] = point[j];
+  }
+  const NumericalSample output(function_(input));
+  if (isHistoryEnabled_)
+  {
+    inputStrategy_.store(NumericalSample(size, point));
```
regislebrun: It's an interesting point here: do we have to include the value of the parameter in the history mechanism? Up to now we did'nt bother with the parameter, which was an empty NumericalPoint almost everytime. Maybe it would be better to join the input point and the parameter and to store the result into the input history, else the user will have the feeling that a given deterministic function was able to give different values with the exact same input point. New accessors should then be added, such as getInputPointHistory() and getInputParameterHistory() in addition to the getInputHistory() method with the obvious optimization getInputPointHistory()==getInputHistory() if the parameter dimension is zero.

jschueller:
@regislebrun, Implemented this, using NS as return values.

regislebrun:
+1

sofianehaddad:
+1

PR #200, opened 2016-05-01
Improved the NLopt interface

regislebrun:
It is now possible to map nlopt algorithms names with nlopt algorithms IDs. New shortcuts are also provided for the following algorithms:
MMA/CCSAQ/COBYLA/BOBYQA


jschueller:
+1

jschueller:
None of the new classes/methods are documented.

python/test/t_NLopt_std.py
```diff
@@ -49,6 +54,6 @@ def printNumericalPoint(point, digits):
                     print('algo=', algo)
                     algo.run()
                     result = algo.getResult()
-                    print('x^=', printNumericalPoint(result.getOptimalPoint(), 4))
+                    #print('x^=', printNumericalPoint(result.getOptimalPoint(), 3))
```
jschueller: why did you remove the test output ?

regislebrun: Because for some NLopt algorithms (the random ones) the output change slightly from run to run. I don't know how to manage NLopt algorithms: we provide access to all the algorithms proposed by the nlopt library, but we don't test them, and if we do we find that some algorithms generate segmentation violation, others provide varying results that can be very far away from the correct result in some cases... Any idea is welcomed!

regislebrun: By the way, the current situation is worse: nothing is tested at all using the generic constuctor!

jschueller:
I'd like to restore all the tests for algorithms that give reproducible results.
For non-reproducible results, maybe nlopt uses the C interface to rand, maybe we could use srand ?
For algos that crash, maybe a bug report could be submited upstream ? And in the mean time a warning could be emitted from OT side ?

adutfoy:
+1

dbarbier:
nlopt defines its own RNG, and it seems that results are reproducible if `nlopt::srand` is called.  Now the hard question is: where to put this call?
Maybe in NLopt::run, use RandomGenerator::GetState to compute some seed to pass to nlopt::srand?

jschueller:
@dbarbier I'd rather set it only once (using a key in the ResourceMap), when the first instance of NLopt is created, with the possibility to also re-initialize it manually with NLopt::SetSeed(n). What do you think ?

regislebrun:
Many thanks for the analysis. I prefer to define a specific seed for NLopt as its prng has a very specific usage. The solution proposed by @jschueller seems very reasonable, I will give it a try and see how many issues it solves.

dbarbier:
@jschueller ok for your solution

jschueller:
Or it can even be just manually cally NLopt::SetSeed at the begining of the test, without any initiliation in the lib by default.
cc @dbarbier

dbarbier:
@jschueller Indeed

jschueller:
+1

PR #201, opened 2016-05-02
Fix NMEvaluation::::getParameterDimension

jschueller:
to use the virtual getParameter method if redefined in inherited classes


dbarbier:
+1

sofianehaddad:
+1

PR #202, opened 2016-05-02
Fix compilation of Python bindings

dbarbier:
Swig adds several -I flags to c++ command line option, in particular
-I/usr/include on Debian.  If OpenTURNS had already been installed
into /usr/include, headers are then taken from /usr/include/openturns
and not from source directory.
Pass BEFORE option to CMake include_directories() command to ensure
that correct header files are being used.


sofianehaddad:
+1

regislebrun:
+1

PR #206, opened 2016-05-06
Move Path::GetWinDirectory>GetExecutableDirectory

jschueller:
..and make it protected instead of public.

This is used to guess the configuration file location from
the Python executable path when the compile-time install path
differs from the runtime install path on Windows, and now Conda.


regislebrun:
+1

dbarbier:
+1

adutfoy:
+1

sofianehaddad:
+1

PR #207, opened 2016-05-09
Use svg math format

jschueller:
requires sphinx 1.4 and the dvisvgm utility.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,22 @@ addons:
     - texlive-fonts-recommended
     - texlive-latex-extra
     - dvipng
+    - libkpathsea-dev
+    - libpotrace-dev
+    - python-lxml
+    - xmlto
+    - asciidoc
 
 install:
+# dvisvgm
+  - git clone https://github.com/mgieseki/dvisvgm.git
+  - pushd dvisvgm
+  - ./autogen.sh
+  - ./configure --prefix=$HOME/.local
+  - make -j2
+  - make install
+  - export PATH=$PATH:$HOME/.local/bin
+  - popd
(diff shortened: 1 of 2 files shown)


regislebrun:
dvisvgm is not yet a very polished tool to say the least, which can be guessed by its version number (0.7.1). Its configuration procedure is almost inexistent: no dependency checking, a configure script that configures almost nothing, all the relevant environment variables have to be setted by hand in several places. If the official documentation is generated using this script we have to check that all the contributors are able to install and use it locally.

jschueller:
@regislebrun, having just compiled it I it seems very well polished to me: it's got autotools to check dependencies https://github.com/mgieseki/dvisvgm/releases, I did not have to set any env variables.
Seems part of recent texlive distributions, at least it in ubuntu 14.04.

regislebrun:
@jschueller Ouuppss, I mispelled the name of the tool and found dvisvg at http://dvisvg.sourceforge.net/ instead of dvisvgm. I agree with you: dvisvgm looks very mature. More feedback soon.

jschueller:
@regislebrun, I applied this in otrobopt if you want to see the result in the published doc

regislebrun:
@jschueller The equations look much better, in particular when I make a zoom on it. I find it disturbing to have sans-serif fonts for the main text and serif fonts for the equations. Could it be possible to use the exact same fonts in both environments?

regislebrun:
+1

dbarbier:
+1

sofianehaddad:
+1

adutfoy:
+1

PR #218, opened 2016-05-17
Fix iterated quadrature

regislebrun:
The case of multivalued functions was not properly handled by IteratedQuadrature


sofianehaddad:
+1

jschueller:
+1

PR #219, opened 2016-05-18
Fix math operators

adutfoy:
This PR fixes math operators not in Roman shape. It also fixes missing decima parts in numbers, and uses the specific command to write integration measure.


dbarbier:
+1

regislebrun:
+1

PR #220, opened 2016-05-19
Fix gauss product experiment

regislebrun:
Fixed a bug in GaussProductExperiment:
The weighted experiment was always built for the standard representative of the distribution.

Fixed a bug in selection mechanism in StandardDistributionPolynomialFactory:
In the case of a distribution for which a dedicated factory exists for its standard representative, this dedicated factory was selected regardless of the parameters of the distribution.

Rewrote AdaptiveStieltjesAlgorithm algorithm:
Now it uses monic orthogonal polynomials for the internal computations instead of orthonormal polynomials to avoid some possible cancellation. The number of sub-intervals is also adapted to the degree of the polynomial.


lib/src/Uncertainty/Algorithm/OrthogonalBasis/AdaptiveStieltjesAlgorithm.cxx
```diff
@@ -65,81 +78,97 @@ AdaptiveStieltjesAlgorithm * AdaptiveStieltjesAlgorithm::clone() const
...
+  // the size of the cache is increased by each call to getRecurrenceCoefficients()
+  while (n >= monicRecurrenceCoefficients_.getSize()) getRecurrenceCoefficients(n - 1);
+  // Here we know that n == cacheSize - 1. In order to compute the recurrence coefficients of the orthonormal polynomial pN
+  // we need \alpha_{n-1}, \beta_{n-1} and \beta_n. The first two values are in the cache, we need \beta_n. We compute both
+  // \alpha_n and \beta_n by numerical integration.
+  Coefficients monicCoefficients(3);
+  monicCoefficients[0] = 1.0;
+  // Build the monic orthogonal polynomial of degree cacheSize
+  const OrthogonalUniVariatePolynomial qN(monicRecurrenceCoefficients_);
+  const GaussKronrod algo(GaussKronrod(ResourceMap::GetAsUnsignedInteger("AdaptiveStieltjesAlgorithm-MaximumSubIntervalsBetweenRoots") * (n + 1), ResourceMap::GetAsNumericalScalar("AdaptiveStieltjesAlgorithm-MaximumError"), GaussKronrodRule(GaussKronrodRule::G7K15)));
```
jschueller: the inner GaussKronrod is not needed

lib/test/t_StandardDistributionPolynomialFactory_std.cxx
```diff
@@ -43,18 +43,19 @@ int main(int argc, char *argv[])
...
     distributionCollection.add(Triangular(-1.0, 0.3, 1.0));
     distributionCollection.add(Uniform(-1.0, 1.0));
+    distributionCollection.add(Uniform(-1.0, 3.0));
     distributionCollection.add(Weibull(1.0, 3.0));
     for (UnsignedInteger n = 0; n < distributionCollection.getSize(); ++n)
     {
       const Distribution distribution(distributionCollection[n]);
       const String name(distribution.getImplementation()->getClassName());
-      StandardDistributionPolynomialFactory polynomialFactory(distribution);
+      StandardDistributionPolynomialFactory polynomialFactory = StandardDistributionPolynomialFactory(AdaptiveStieltjesAlgorithm(distribution));
```
jschueller: use copy constructor instead

lib/src/Uncertainty/Algorithm/OrthogonalBasis/StandardDistributionPolynomialFactory.cxx
```diff
@@ -94,60 +94,89 @@ void StandardDistributionPolynomialFactory::checkSpecificFamily()
...
-  }
+    {
+      referenceFamily = LegendreFactory();
+      if (referenceFamily.getMeasure() == measure_)
+	{
+	  specificFamily_ = referenceFamily;
+	  hasSpecificFamily_ = true;
+	  return;
+	}
+    }
```
jschueller: you added always the same code blocks, I bet they can be merged into one

jschueller: why did you change the indentation ?

regislebrun: The point is to test first the class name (cheap) then the parameters (maybe costly), and since the number of parameters needed to build a specific factory depends on the distribution there is no straightforward way to simplify the code. I will propose something slightly shorter but no revolution here.
For the indentation, it is just a confusion between astyle indent (the whole buffer) and emacs indent (the current line).

jschueller:
+9000

dbarbier:
+1

PR #221, opened 2016-05-24
Fix Path::GetExecutableDirectory on osx

jschueller:
/proc/ is not available to read the exe path
use _NSGetExecutablePath system call

### DIFF
--- a/lib/src/Base/Common/Path.cxx
+++ b/lib/src/Base/Common/Path.cxx
@@ -66,6 +66,10 @@
 #error "OPENTURNS_HOME_ENV_VAR is NOT defined. Check configuration."
 #endif
 
+#ifdef __APPLE__
+#include <mach-o/dyld.h>
+#endif
+
 BEGIN_NAMESPACE_OPENTURNS
 
 /* The environment variable name */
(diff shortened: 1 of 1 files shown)


dbarbier:
This method does not make sense, couldn't we instead just drop it?

jschueller:
@dbarbier I use it to guess the config dir location when the install tree is moved around, like in conda.
I agree this is a bit hackish, but it's the only thing I could come up with.

jschueller:
@dbarbier any other idea ?

dbarbier:
@jschueller I do not know conda, it seems that http://conda.pydata.org/docs/building/meta-yaml.html#relocatable discusses this issue

jschueller:
@dbarbier
Our problem is that ot searches for the config file in env vars or compile-time prefix.
In the context of conda we cannot set env vars easily, and the installation prefix is different from the compile-time prefix and conda doesn't help us to replace the compile-time prefix as it resides in the library and not text files.

dbarbier:
Prefix can be replaced in binary file too, see the 3rd item.

jschueller:
@dbarbier, indeed

sofianehaddad:
hmmm the problem with conda is indeed solved. But the current master still generates error due to bad readlink (thus tests fail and library loads with an error...)

jschueller:
@sofianehaddad, I propose to only use GetExecutableDirectory on windows.

sofianehaddad:
@jschueller ok

jschueller:
@sofianehaddad, see #225

PR #222, opened 2016-05-25
Fixed a bug in GaussProductExperiment

regislebrun:
The constructor based only on marginal indices was wrong for dimension>1 as the dimension of the indices was not compatible with the dimension of the default distribution (=1).


jschueller:
+1

dbarbier:
+1

jschueller:
@regislebrun this is weird, this broke otrobopt non-uniform results

regislebrun:
@jschueller Ongoing compilation, more news this evening!

PR #224, opened 2016-05-25
Fixed a bug in GaussProductExperiment introduced in PR#222

regislebrun:
The bug is due to the incremental way an attribute is built during the construction of the object, which is reused when the distribution underlying this weighted experiment is changed. The reset of the attribute was missing.


dbarbier:
+1

jschueller:
great!
otrobopt is happy again

PR #228, opened 2016-05-30
Parametrize NLopt by algo names

jschueller:
and remove subclasses as NLopt does not expose any algorithm-specific interface.
Also for consistency, why add only some shortcuts (to SLSQP, BFGS, ...) and not all of them.


dbarbier:
+1

lib/src/Base/Optim/NLopt.cxx
```diff
@@ -137,18 +119,19 @@ void NLopt::SetSeed(const UnsignedInteger seed)
 
 
 /* Default constructor */
-NLopt::NLopt(const UnsignedInteger algoType)
+NLopt::NLopt(const String algoName)
```
regislebrun: const String & algoName

lib/src/Base/Optim/NLopt.cxx
```diff
@@ -137,18 +119,19 @@ void NLopt::SetSeed(const UnsignedInteger seed)
...
-  , algoType_(algoType)
+  , algoName_(algoName)
 {
-  // Nothing to do
+  GetAlgorithmCode(algoName);
 }
 
 NLopt::NLopt(const OptimizationProblem & problem,
-             const UnsignedInteger algoType)
+             const String algoName)
```
regislebrun: const String & algoName

lib/src/Base/Optim/openturns/NLopt.hxx
```diff
@@ -38,11 +38,11 @@ class OT_API NLopt :
 public:
 
   /** Default constructor */
-  explicit NLopt(const UnsignedInteger algoType = 0);
+  explicit NLopt(const String algoName = "GN_DIRECT");
```
regislebrun: const String & algoName = "LD_SLSQP"
+ a complex type such as String must be passed by reference
+ LD_SLSQP is the most versatile algorithm provided by NLopt

lib/src/Base/Optim/openturns/NLopt.hxx
```diff
@@ -38,11 +38,11 @@ class OT_API NLopt :
...
 
   /** Constructor with parameters */
   explicit NLopt(const OptimizationProblem & problem,
-                 const UnsignedInteger algoType = 0);
+                 const String algoName = "GN_DIRECT");
```
regislebrun: const String & algoName = "LD_SLSQP"
+ a complex type such as String must be passed by reference
+ LD_SLSQP is the most versatile algorithm provided by NLopt

jschueller:
@regislebrun, ok, fixed

sofianehaddad:
+1

regislebrun:
I started to rework the NLopt interface (but stopped everything due to a lack of time) in order to make the nlopt::opt instance an attribute of the NLopt class. It would allow to call the set_default_initial_step() method before to start the optimization, mandatory for the COBYLA algorithm provided by NLopt.

Even if the nlopt library does not provide a lot of control on the underlying algorithms, this control should be given to the user as much as possible. In my mind, it is the main difference between creating eg SLSQP() or NLopt("LD_SLSQP"): the first constructor should give the maximum control on the algorithm.

regislebrun:
otherwise +1, the point being to not suppress the SLSQP() & Co classes too fast.

jschueller:
@regislebrun, Again this can be done independently from the algorithm, see my last commit.

jschueller:
@regislebrun 
btw there's set_default_initial_step & set_initial_step, I called set_default_initial_step:
https://github.com/stevengj/nlopt/blob/master/api/options.c#L704

regislebrun:
@jschueller A simple test-case is to enable the version of Cobyla provided by NLopt in t_NLopt_std. With a meaningful value for the initial step (the default is zero in nlopt, a value that is certain to fail) it should solve the problem without trouble!

regislebrun:
Ok I got it regarding the initial step. There remains a problem: how can we provide the auxiliary solver to the AUGLAG* family of algorithms with the current interface? With this additional feature it would give access to the full potential of NLopt!

jschueller:
@regislebrun AUGLAG seems to run now, couldn't say the same for Cobyla.

jschueller:
@regislebrun I tried several values to pass to setInitialStep for Cobyla, I get either a segfault or an error message.

PR #229, opened 2016-05-30
Removed MeshFactory

jschueller:
This interface is not unused: as the IntervalMesher method defined shadows
the generic method and the resulting code is invalid:
Mesh MeshFactory::build(const Domain & domain) const;
Mesh IntervalMesher::build(const Interval & interval,const Bool diamond = false) const;


regislebrun:
-1 @jschueller When I implemented this class I had in mind to provide a LevelSetMesher class but had no time to implement it. Now, with the new getLowerBound() and getUpperBound methods it is fairly easy to provide a rough version of such a class: mesh the bounding box using the IntervalMesher class and retains all the simplices that have at least half of the vertices into the LevelSet.

I agree on the fact that IntervalMesher::build() should not shadow the generic build() method. BTW the diamond flag is unused in the current implementation of the build() method in IntervalMesher. Another development that has to be postpone due to a lack of time...

regislebrun:
+1 @jschueller could you reopen this PR?

PR #233, opened 2016-06-01
Preliminary unity build work

jschueller:
These changes could appear as random cleanups but in fact it prepares the work of a unity build patch, that is to be able to build the library in a single compilation unit.
This is useful to speed up build when you do not need incremental compilation (just one full build), and when you have only 1 cpu available (Travis/OSX).
To be continued after the tensor works.


dbarbier:
+1

regislebrun:
+1

PR #235, opened 2016-06-01
Symmetrize matrices on array conversion

jschueller:


dbarbier:
+1
Wow, I don't understand how it works, thanks for fixing this.

jschueller:
@dbarbier the trick is to overload the array interface query '__array_interface__' already intercepted by Matrix.getattribute in SymmetricMatrix.getattribute to symmetrize the object.

regislebrun:
+1

sofianehaddad:
+1

PR #238, opened 2016-06-02
Fix a segfault in CovarianceModelImplementation::discretizeHMatrix

dbarbier:
CovarianceModelImplementation::discretizeHMatrix contains this snippet:
```
  if (dimension_ == 1)
  {
    CovarianceAssemblyFunction simple(*this, vertices, nuggetFactor);
    covarianceHMatrix.assemble(simple, 'L');
  }
```

In fact, `*this` is converted into a local `CovarianceModel` variable,
and `CovarianceAssemblyFunction::covarianceModel_` is a reference to
this local variable.  When exiting from `CovarianceAssemblyFunction`
constructor, this local variable is destroyed, but `covarianceModel_`
is still referencing it and `covarianceHMatrix.assemble()` crashes.

Also remove reference for vertices too, in order to be safe.

### DIFF
--- a/lib/src/Base/Stat/openturns/HMatrixImplementation.hxx
+++ b/lib/src/Base/Stat/openturns/HMatrixImplementation.hxx
@@ -197,8 +197,8 @@ private:
 class CovarianceAssemblyFunction : public HMatrixRealAssemblyFunction
 {
 private:
-  const CovarianceModel & covarianceModel_;
-  const NumericalSample & vertices_;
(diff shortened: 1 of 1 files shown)


dbarbier:
I used references in order to avoid copies, but this proved to be a very bad idea.

regislebrun:
+1

jschueller:
+1
NumericalSample & CovarianceModel are pointer classes anyway

PR #240, opened 2016-06-02
Centered Sobol estimation 

jschueller:


sofianehaddad:
@jschueller not sure about this. Even if the estimators look better (particularly Mauntz-Kucherenko/Saltelli, which are known to lack efficiency in case of decentered output), the algorithm could not be referenced as MK/Saltelli. ( as you mentioned in http://trac.openturns.org/ticket/786, it becomes another estimator...)

regislebrun:
+1

regislebrun:
@jschueller @sofianehaddad In fact NOBODY excepted us implement this method in its initial form. Either we remove it from the library (there is room for it in a module), or we adopt Julien's patch. I am quite surprised by this patch as I thank it was already done!

jschueller:
@dbarbier some thoughts ?

sofianehaddad:
@jschueller still disagree with the changes. But if you plan to integrate it you must update docstrings as it is not the original methods that are implemented

dbarbier:
+1

PR #242, opened 2016-06-02
Fix Cobyla optimal value dimension

jschueller:
A 2nd component is used to return the constraint error through the
callback.


dbarbier:
+1

PR #247, opened 2016-06-03
Fix Pearson doctest with R 3.3.0

jschueller:


regislebrun:
+1

sofianehaddad:
+1
Thanks, bug is oldest (R 3.2)

PR #251, opened 2016-06-10
Fix tests with NLopt 2.4.1

jschueller:
We also use Ubuntu Trusty at Phimeca, which provides NLopt 2.4.1 by default.


regislebrun:
+1 I don't know how to deal with problematic nlopt algorithms. Either we drop them, or we inform our users about the bad behaviour of these algorithms. We could also send a feedback to the nlopt team with specific examples showing the bugs, but it needs time and it is a precious resource.

jschueller:
@regislebrun I don't think any of them are in bad shape, there's just a small gap between available algos and those we can test reliably across different compilers and platforms, nlopt version etc.

dbarbier:
+1

PR #252, opened 2016-06-15
Unity builds

jschueller:
Use cotire cmake module: https://github.com/sakra/cotire

This allows to speed-up builds of libOT by using fewer compilation units.
This option speeds up one time builds but incremental builds will be slower.
This can be used for continuous integration or producing binaries.

Using USE_COTIRE=ON disables the sphinx doc and the bison/flex parser as it does not play well with custom rules.


dbarbier:
Description?

jschueller:
@dbarbier see 3987ff0:

This allows to speed-up builds of libOT by using fewer compilation
units.
This option speeds up one time builds but incremental builds will be
slower.
This can be used for continuous integration or producing binaries.

dbarbier:
+1

regislebrun:
+1

PR #257, opened 2016-06-16
Better use of CovarianceModelCollection

sofianehaddad:
The PR fixes http://trac.openturns.org/ticket/841

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -57,6 +57,7 @@
  * #830 (more optim algos examples)
  * #837 (TruncatedDistribution::setParameter segfaults)
  * #838 (Symmetry of SymmetricMatrix not always enforced)
+ * #841 (Better CovarianceModelCollection in Python)
 
 == 1.7 release (2016-01-27) == #release-1.7
 
--- a/python/src/CovarianceModel.i
+++ b/python/src/CovarianceModel.i
@@ -6,7 +6,10 @@
 
 %include CovarianceModel_doc.i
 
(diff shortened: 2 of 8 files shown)


regislebrun:
+1

jschueller:
nice, I did a similar patch while in the train for this, but using the macros:
https://github.com/jschueller/openturns/commit/7b9395820fa329e3fbe77ebb4d466535a32f6ff9

sofianehaddad:
@jschueller looks great. Did not know about the macros

jschueller:
I opened #260

sofianehaddad:
@jschueller ouch I was fixing the CovarianceModel.i by using your tip...

jschueller:
@sofianehaddad, oh sorry, I'll close mine

sofianehaddad:
@jschueller you are right for the PyObject* constructor but extra constructor with other arguments is not supported ( TensorizedCovModel(PyObject*, NP &)  for example)

jschueller:
If TensorizedCovarianceModel(const CovarianceModelCollection & collection) works with a list of cov model, then TensorizedCovarianceModel(const CovarianceModelCollection & collection, const NumericalPoint & scale) should work too.

sofianehaddad:
strange because one moment I had TCM([cov1,cov2]) ok but TCM([cov1;cov2], [1,2]) nok
It seems now to be fixed, thanks

jschueller:
+1

PR #261, opened 2016-06-17
Homogenity in covariance models

sofianehaddad:
Covariance models should have the same constructors (scale, amplitude) or (scale, amplitude, *args)
This PR fixes inconsistency in some model classes (fixes a part of http://trac.openturns.org/ticket/833)

Also constructors that uses both spatialDimension/scale are deprecated

Finally some default parameter values (theta) are updated


jschueller:
+1

regislebrun:
+1

sofianehaddad:
@jschueller yes

PR #262, opened 2016-06-17
Fix Kriging random failures

jschueller:


regislebrun:
@jschueller I am very surprised by the fact that this simple change works. In the new implementation of KrigingAlgorithm, the parameters of the covariance model are always optimized, so even if you think you have fixed the values of the scale parameter, it is recomputed. A numerical miracle?

regislebrun:
+1

jschueller:
@regislebrun, the tests do not fail anymore, I ran them a few 100s time on 2 different boxes.
tip: ctest has a decidated ```--repeat-until-fail N``` option.

regislebrun:
@jschueller  Ok good news, I was not able to catch it! To merge ASAP ;-)!

sofianehaddad:
+1

PR #265, opened 2016-06-23
Level set bounds

regislebrun:
Improved LevelSet:
Now the user can specify a bounding box by using the setLowerBound()/setUpperBound() methods. The level set is cliped by this bounding box, ie the contains() method will return false as soon as the given point is outside of the user-defined bounding box.

Improved KarhunenLoeveQuadratureFactory:
Now it fully support multidimensional covariance models. It is also more verbose.

### DIFF
--- a/lib/src/Base/Geom/LevelSet.cxx
+++ b/lib/src/Base/Geom/LevelSet.cxx
@@ -39,6 +39,8 @@ LevelSet::LevelSet(const UnsignedInteger dimension)
   : DomainImplementation(dimension)
   , function_(NumericalMathFunction(Description::BuildDefault(dimension, "x"), Description(1, "1.0")))
   , level_(0.0)
+  , lowerBound_(0)
+  , upperBound_(0)
 {
   // Nothing to do
 }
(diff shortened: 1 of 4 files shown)


jschueller:
@regislebrun, Have you considered using an Interval object instead of 2 points ? 
Maybe it would help simplify some redundant code such as:
```
if ((lowerBound_.getDimension() == dimension_) &&
      (upperBound_.getDimension() == dimension_) &&
      (other.lowerBound_.getDimension() == dimension_) &&
      (other.upperBound_.getDimension() == dimension_))
```

regislebrun:
@jschueller The test is here to check if bounding boxes are available for both level sets without computing them on the fly. I use two points to be consistent with DomainImplementation, where I cannot use an Interval without introducing circular dependencies. BTW, the test will not benefit from the introduction of an Interval as we check the dimensions of the bounds as flags to tell if they have already been set/computed. Did I missed something?

jschueller:
+1

dbarbier:
+1

PR #266, opened 2016-06-24
Make ot.PythonFunction (optionally) distributed with multiprocessing

felipeam86:
This pull request adds the optional argument `n_cpus` to PythonFunction. If it is set, it distributes `func` with multiprocessing on the number of cpus specified by `n_cpus`. If `n_cpus=-1`, it uses all the cores of the machine. if `n_cpus=1`, it does nothing. If `n_cpus` and `func_sample` are both given as arguments, `n_cpus` will be ignored and samples will be handled by `func_sample`.

This function is limited to distributing `func` calls on the machine you are working on. More advanced distribution techniques will be develloped for [otwrapy.Parallelizer](http://felipeam86.github.io/otwrapy/_generated/otwrapy.Parallelizer.html#otwrapy.Parallelizer)

### DIFF
--- a/python/test/t_NumericalMathFunction_python.py
+++ b/python/test/t_NumericalMathFunction_python.py
@@ -124,6 +124,15 @@ myFunc = PythonFunction(2, 1, func_sample=a_exec_sample)
 outSample = myFunc(a_sample)
 print(outSample)
 
+print('distributed exec only on a point')
+myFunc = PythonFunction(2, 1, a_exec, n_cpus=-1)
+outSample = myFunc([100., 100.])
+print(outSample)
+
+print('distributed exec only on a sample')
+myFunc = PythonFunction(2, 1, a_exec, n_cpus=-1)
+outSample = myFunc(a_sample)
+print(outSample)
 
 def a_grad(X):
     # wrong but allows to verify
--- a/python/test/t_NumericalMathFunction_python.py
+++ b/python/test/t_NumericalMathFunction_python.py
@@ -2,6 +2,7 @@
 
 from __future__ import print_function
 from openturns import *
+import sys
 
 TESTPREAMBLE()
 
@@ -125,12 +126,12 @@ outSample = myFunc(a_sample)
 print(outSample)
 
 print('distributed exec only on a point')
-myFunc = PythonFunction(2, 1, a_exec, n_cpus=-1)
+myFunc = PythonFunction(2, 1, a_exec, n_cpus=None if sys.platform.startswith('win') else -1)
 outSample = myFunc([100., 100.])
 print(outSample)
 
 print('distributed exec only on a sample')
-myFunc = PythonFunction(2, 1, a_exec, n_cpus=-1)
+myFunc = PythonFunction(2, 1, a_exec, n_cpus=None if sys.platform.startswith('win') else -1)
 outSample = myFunc(a_sample)
 print(outSample)
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,6 +11,7 @@ individual contributors:
     Antoine Dumas <dumas@phimeca.com>
     Aurelie Ladier <ladier@phimeca.com>
     Denis Barbier <barbier@imacs.polytechnique.fr>
+    Felipe Aguirre Martinez <aguirre@phimeca.com>
     Ivan Dutka-Malen <ivan.dutka-malen@edf.fr>
     Julien Schueller <schueller@phimeca.com>
     Mathieu Lapointe <lapointe@phimeca.com>
--- a/ChangeLog
+++ b/ChangeLog
@@ -37,6 +37,7 @@
  * Deprecated LAR in favor of LARS
 
 === Python module ===
+ * Add the possibility to distribute PythonFunction calls with multiprocessing
 
 === Miscellaneous ===
  * Improved the computeCDF() method of Normal
--- a/python/src/NumericalMathFunction.i
+++ b/python/src/NumericalMathFunction.i
@@ -186,6 +186,32 @@ class OpenTURNSPythonFunction(object):
     def getParameterDescription(self):
         return self.__paramDesc
 
+def _exec_sample_multiprocessing(func, n_cpus):
+    """Return a distributed function using multiprocessing
+
+    Parameters
+    ----------
+    func : Function or calable
+        A callable python object, usually a function. The function should take
+        an input vector as argument and return an output vector.
+
+    n_cpus : int
+        Number of CPUs on which to distribute the function calls.
+
+    Returns
+    -------
+    _exec_sample : Function or callable
+        The parallelized funtion.
+    """
+
+    def _exec_sample(X):
+        from multiprocessing import Pool
+        p = Pool(processes=n_cpus)
+        rs = p.map_async(func, X)
+        p.close()
+        return rs.get()
+    return _exec_sample
+
 class PythonFunction(NumericalMathFunction):
     """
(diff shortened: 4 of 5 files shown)


jschueller:
You can squash all these commits.

felipeam86:
@jschueller commits squashed, tests passed !

jschueller:
You can mention it the ChangeLog, in the "Python module" entry.

regislebrun:
+1

felipeam86:
Initially I thought about adding this functionality with multiprocessing to avoid imposing a dependency. But the thing is that I really think that `joblib` is worth the trouble ! My two main reasons are :

- If an exception is raised, it prints the original traceback of what raised the exception (mulriprocessing does not do this)
- It has a progress "bar" by default

What do you think if we add joblib as the default backend, but optional. That is, if joblib is installed, it is used by default, otherwise func is distributed with multiprocessing. Take a look at this [commit](https://github.com/openturns/openturns/commit/556f7e9dbf699308387ff25147539530babb55a1) and let me know what you think

jschueller:
I fear users getting only the multiprocessing backend could be deceived by the lack of traceback support. Depending on joblib is not an option, but we could put it in a separate import (not an external module) like we do for the matplotlib backend: 'import openturns.viewer'.

felipeam86:
After some discussion with @jschueller, it is true that the idea of including joblib may not be that good for the moment. On one hand, this might lead to two different behaviors on different machines without having the control to choose. On the other hand, forcing the dependency on joblib is not a nice idea either.

This discussion raised an additional concern : when the function is distributed with multiprocessing, we loose the traceback of any given raised exception during the execution of the distributed function. This might be considered as a degraded used of the PythonFunction --> we loose visibility of exceptions raised which makes it harder to debug. However, for this kind of situations, users can use  [otwrapy.Parallelizer](http://felipeam86.github.io/otwrapy/_generated/otwrapy.Parallelizer.html#otwrapy.Parallelizer) with joblib backend instead.

I propose the following : 
- Keep this pull request as is (with only multiprocessing). It provides a very cheap and easy way to have a distributed wrapper.
- Warn the users that they loose the traceback of raised exceptions by the function.
- Advice them to use otwrapy with joblib for more complicated cases where the external code is unstable and tends to raise exceptions for some input parameters.
- Follow up on joblib devellopment and on how it can be integrated to openturns. Scikit learn for example integrates it in [sklearn.external](https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/externals/joblib). By the way, joblib is developed by the developers of scikit learn

jschueller:
Ok for me, as long as this limitation appears clearly in the doc.

Maybe we could also add comments related to the parallelization aspect in wrapper_development.rst as it also deals with PythonFunction: http://openturns.github.io/developer_guide/wrapper_development.html.

jschueller:
+1

dbarbier:
+1

jschueller:
It fails on our autobuilder:
```
459/844 Testing: pyinstallcheck_NumericalMathFunction_python
459/844 Test: pyinstallcheck_NumericalMathFunction_python
Command: "/bin/sh" "-c" "(/usr/bin/python2.6 /home/openturns/distcheck/workdir/python/test/t_NumericalMathFunction_python.py   | cat  > /home/openturns/distcheck/workdir/build/python/test/t_NumericalMathFunction_python.out) && diff -u /home/openturns/distcheck/workdir/python/test/t_NumericalMathFunction_python.expout /home/openturns/distcheck/workdir/build/python/test/t_NumericalMathFunction_python.out"
Directory: /home/openturns/distcheck/workdir/build/python/test
"pyinstallcheck_NumericalMathFunction_python" start time: Jun 27 16:14 CEST
Output:
----------------------------------------------------------
--- /home/openturns/distcheck/workdir/python/test/t_NumericalMathFunction_python.expout	2016-06-27 15:44:17.682753520 +0200
+++ /home/openturns/distcheck/workdir/build/python/test/t_NumericalMathFunction_python.out	2016-06-27 16:14:08.530992476 +0200
@@ -40,20 +40,3 @@
 distributed exec only on a point
 [200]
 distributed exec only on a sample
-    [ y0  ]
-0 : [ 200 ]
-1 : [ 202 ]
-2 : [ 204 ]
-gradient
-[[  1 ]
- [ -1 ]]
-hessian
-sheet #0
-[[ 0.1 0.3 ]
- [ 0.3 0.1 ]]
-no func
-no function detected : ok.
-exception handling: ok
-exception handling: ok
-x= [4.0, 5.0] theta= [1.2] f(x, theta)= [10]
-[1.2] [theta]
  File "/home/openturns/distcheck/last/squeeze-x86_64-ot-linux/lib/python2.6/site-packages/openturns/func.py", line 10338, in __call__
    Y = self._exec_sample(ns)
  File "/home/openturns/distcheck/last/squeeze-x86_64-ot-linux/lib/python2.6/site-packages/openturns/func.py", line 10388, in _exec_sample
    p = Pool(processes=n_cpus)
  File "/usr/lib/python2.6/multiprocessing/__init__.py", line 227, in Pool
    return Pool(processes, initializer, initargs)
  File "/usr/lib/python2.6/multiprocessing/pool.py", line 84, in __init__
    self._setup_queues()
  File "/usr/lib/python2.6/multiprocessing/pool.py", line 131, in _setup_queues
    self._inqueue = SimpleQueue()
  File "/usr/lib/python2.6/multiprocessing/queues.py", line 328, in __init__
    self._rlock = Lock()
  File "/usr/lib/python2.6/multiprocessing/synchronize.py", line 117, in __init__
    SemLock.__init__(self, SEMAPHORE, 1, 1)
  File "/usr/lib/python2.6/multiprocessing/synchronize.py", line 49, in __init__
    sl = self._semlock = _multiprocessing.SemLock(kind, value, maxvalue)
OSError: [Errno 13] Permission denied
  File "/home/openturns/distcheck/workdir/python/test/t_NumericalMathFunction_python.py", line 134, in <module>
    outSample = myFunc(a_sample)
  File "/home/openturns/distcheck/last/squeeze-x86_64-ot-linux/lib/python2.6/site-packages/openturns/func.py", line 9753, in __call__
    def __call__(self, *args): return _func.NumericalMathFunction___call__(self, *args)
RuntimeError: InternalException : Python exception: OSError
<end of output>
Test time =   0.30 sec
----------------------------------------------------------
Test Failed.
"pyinstallcheck_NumericalMathFunction_python" end time: Jun 27 16:14 CEST
"pyinstallcheck_NumericalMathFunction_python" time elapsed: 00:00:00
----------------------------------------------------------
```

jschueller:
```
D400: First line should end with a period (not 'g')

 11376: def _exec_sample_multiprocessing(func, n_cpus):
 11377:     """Return a distributed function using multiprocessing
```

jschueller:
Seems multiprocessing does not play well in wine we use for testing: http://bugs.python.org/issue9099
Could you disable it ?
```PythonFunction(2, 1, a_exec, n_cpus=None if sys.platform.startswith('win') else -1)```

File: python/test/t_NumericalMathFunction_python.py [hunk 1]
```diff
@@ -125,12 +125,12 @@ def a_exec_sample(Xs):
 print(outSample)
 
 print('distributed exec only on a point')
-myFunc = PythonFunction(2, 1, a_exec, n_cpus=-1)
+myFunc = PythonFunction(2, 1, a_exec, n_cpus=None if sys.platform.startswith('win') else -1)
```
jschueller: import sys ?

PR #267, opened 2016-06-27
Improved IntegrationStrategy.

regislebrun:
Now, it uses the DesignProxy class efficiently, 50x speedup on typical applications.


jschueller:
+1

dbarbier:
`ProjectionStrategyImplementation::setMeasure` and `ProjectionStrategyImplementation::setExperiment` must reinitialize inputSample_, otherwise wrong cached values may be used.

regislebrun:
@dbarbier Thanks, fixed!

dbarbier:
There is another problem, `proxy_` takes `basis` argument in its constructor, thus you must ensure in `IntegrationStrategy::computeCoefficients` that `basis` is the same object that has been used to construct `proxy_`.

regislebrun:
@dbarbier I think you found a true bug here! I have to refresh my memory on how the classes interact through their natural use in the context of functional chaos decomposition AND when they are used independently. Initially, they were not supposed to be used independently but a customer asked me to add the ability to use IntegrationStrategy::computeCoefficients() as a general service, so I added the basis in the arguments and it changes a lot of things. I also have to insure that the cached quantities are not invalidated if one change the measure or the experiment to the exact same values as the one stored in the members, taking into account that the initial value of the members can be pure garbage... So it will take some quiet work to do things correctly...

dbarbier:
> I also have to insure that the cached quantities are not invalidated if one change the measure or the experiment to the exact same values as the one stored in the members

I understand why you want to do that, but this is overkill IMO, unless those setters are called internally by OT.

About Basis, calls to `copyOnWrite()` within non-const `Basis::operator[]` and `Basis::add` methods should be added. You could then store a `BasisImplementation*` member into `IntegrationStrategy`; when it changes, `proxy_` is invalidated.  These changes are less invasive than rewriting `IntegrationStrategy::computeCoefficients()`.

regislebrun:
@dbarbier As far as I remind, these setters are called internally, which is my main concern. About the other points you mention, a copyOnWrite() is mandatory for all the non-const methods of all the interface objects so you found a bug (in fact 2) in Basis. I am not fond of C++ pointers as members, can a smart pointer be an alternative?

dbarbier:
Right, there is no good reason to use a raw pointer here, sorry.  In fact, there is a simple solution, this should work (not tested):
```
  if (proxy_.getInputSample().getSize() == 0 || proxy_.getBasis().getImplementation() != basis.getImplementation())
```

jschueller:
@regislebrun do you plan on including it in the 1.8rc1 ?

regislebrun:
@jschueller Yes I am working hard on it. Denis found a bug that is deeply rooted in the way FunctionalChaosAlgorithm and ProjectionStrategyImplementation interact.

jschueller:
+1

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/IntegrationStrategy.cxx
```diff
@@ -92,6 +92,47 @@ String IntegrationStrategy::__repr__() const
...
+    // Nothing to do
+  }
+
+  inline void operator()( const TBB::BlockedRange<UnsignedInteger> & r ) const
+  {
+    for (UnsignedInteger j = r.begin(); j != r.end(); ++j)
+      {
+	for (UnsignedInteger i = 0; i < outputSample_.getSize(); ++i)
+	  {
+	    const UnsignedInteger indexAdded(addedRanks_[j]);
```
dbarbier: Should be moved 2 lines above.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/ProjectionStrategyImplementation.cxx
```diff
@@ -142,9 +138,13 @@ String ProjectionStrategyImplementation::__repr__() const
 /* Measure accessor */
 void ProjectionStrategyImplementation::setMeasure(const Distribution & measure)
 {
-  measure_ = measure;
-  // Set the measure as the distribution of the weighted experiment
-  weightedExperiment_.setDistribution(measure);
+  if (measure != measure_)
```
dbarbier: `if (!(measure == measure_))`

dbarbier:
+1 with or without my comments being taken into account, they are not critical.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/IntegrationStrategy.cxx
```diff
@@ -92,6 +92,47 @@ String IntegrationStrategy::__repr__() const
...
+	    // The integral is approximated by a weighted sum of f * Phi
+	    alpha_[j] += weights_[i] * outputSample_[i][marginalIndex_] * designMatrix_(i, indexAdded);
```
regislebrun: @dbarbier I first wrote a wrong code where the i and j loops were interchanged, then I fixed the mess but leaving things in the innermost loop by mistake. Concerning the linear algebra aspects, the situation is more involved than it looks at first sight. For some of our applications, the database can be VERY large, say 1e6 points in dimension 50, so it is not a good idea to store things twice. For the algorithm, only the weighted version of the output sample is needed, but the user may want to access to the output sample, so both the weighted and unweighted versions of the output sample are needed. I chose to keep only the unweighted sample and to recompute the weighted version on the fly, but other options are possible, such as a threshold below which things are stored and above which they are recomputed on the fly. Do you have a better idea to cope with these constraints?

dbarbier: @regislebrun 
> For the algorithm, only the weighted version of the output sample is needed, but the user may want to access to the output sample, so both the weighted and unweighted versions of the output sample are needed.

You indeed have to keep the unweighted sample, but you only need a weighted marginal, not the whole weighted sample, so this is affordable.
In `IntegrationStrategy::computeCoefficients`, create a NP weightedMarginal of dimension `outputSample_.getSize()`, fill it with
```
weightedMarginal[i] = weights_[i] * outputSample_[i][marginalIndex_];
```
(no idea yet how to do that efficiently) and replace the first 3 arguments of IntegrationStrategyCoefficientsPolicy constructor by this NP.

regislebrun: @dbarbier I implemented your idea, using iterators in the innermost loop in order to avoid expansive index arithmetic to access the matrix elements. It gives me a 5% improvement when used inside the FunctionalChaosAlgorithm class, and the computation of the coefficients is no more the bottleneck of the whole algorithm. Thanks.

dbarbier: Excellent!

regislebrun:
This PR has been replaced by a series of PR.

PR #268, opened 2016-06-27
Fixed BernsteinCopulaFactory

regislebrun:
The BernsteinCopulaFactory class produces a Mixture which is a copula but it cannot be deduced from the individual properties of the atoms in the mixture.

### DIFF
--- a/lib/src/Uncertainty/Distribution/BernsteinCopulaFactory.cxx
+++ b/lib/src/Uncertainty/Distribution/BernsteinCopulaFactory.cxx
@@ -104,7 +104,10 @@ Distribution BernsteinCopulaFactory::build(const NumericalSample & sample,
   BernsteinCopulaFactoryPolicy policy(empiricalCopulaSample, binNumber, atomsMixture);
   LOGINFO("BernsteinCopulaFactory - Create the resulting Bernstein copula");
   TBB::ParallelFor( 0, size, policy );
-  return Mixture(atomsMixture);
+  Mixture result(atomsMixture);
+  // Here we know that the mixture is a copula even if none of its atoms is.
+  result.isCopula_ = true;
+  return result;
 }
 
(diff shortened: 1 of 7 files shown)


jschueller:
so if you save/load the result of BernsteinCopulaFactory the isCopula flag goes from true to false as it's not saved in distributionimplementation and takes its default value

dbarbier:
+1

jschueller:
+1

jschueller:
Maybe CopulaImplementation:::isCopula can be removed too while only its flag would be set

regislebrun:
@jschueller Yes, isCopula() should be removed from all the distributions/copulas, but it would take more time than I can afford right now, and the key point was to fix a bug with this PR. Of course it has to be done at a point, so I open a ticket.

jschueller:
No, it's just that one:
```
$ grep -nr '::isCopula' schueller/lib/src/
schueller/lib/src/Uncertainty/Model/Distribution.cxx:611:Bool Distribution::isCopula() const
schueller/lib/src/Uncertainty/Model/DistributionImplementation.cxx:2391:Bool DistributionImplementation::isCopula() const
schueller/lib/src/Uncertainty/Model/CopulaImplementation.cxx:287:Bool CopulaImplementation::isCopula() const
schueller/lib/src/Uncertainty/Distribution/Mixture.cxx:528:Bool Mixture::isCopula() const
```

regislebrun:
@jschueller I made a confusion between isCopula (only a few instances) and isIndependent() or isElliptical()!

jschueller:
oh, theres also PythonDistributionImplementation

regislebrun:
@jschueller This one has to stay like this, because the user may want to implement a copula using the PythonDistributionImplementation class and there is no way to set the isCopula_ member directly from python (or tell me how to do it ;-)).

PR #269, opened 2016-06-28
Enhance matplotlib backend

jschueller:
- Fixes a problem with PolygonArray when colors are defined with names instead of hex codes
- For contours, take account colors, label flag and legend.
- Control image size in pixels

### DIFF
--- a/python/src/viewer.py
+++ b/python/src/viewer.py
@@ -323,7 +323,8 @@ class View(object):
                 colors = drawable.getPalette()
                 colorsRGBA = []
                 for i in range(polygonsNumber):
-                    rgba = drawable.ConvertToRGBA(colors[i])
+                    hex_code = ot.Drawable.ConvertFromName(colors[i])
+                    rgba = ot.Drawable.ConvertToRGBA(hex_code)
                     colorsRGBA.append(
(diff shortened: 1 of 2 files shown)


jschueller:
@regislebrun, to set figure size figure has a default size of 10x10in and the default dpi is 80,(=> 800x800) so you can play with the dpi to get higher resolution with figure_kwargs={'dpi':xx}

regislebrun:
@jschueller It allows to scale the figure, but how to change the aspect ratio? A square figure means a rectangular drawing region, and sometimes one need a square drawing region (and hence a rectangular figure).

jschueller:
You can use the figsize keyword:
```view = View(graph, figure_kwargs={'figsize':(20,10)})```

regislebrun:
@jschueller So natural, compared to view.save('fig.png', 800, 600) ;-)!

regislebrun:
@jschueller So I have to divide the dimensions I would have by 80 and specify the result as the value of figsize? So long since I did my last mental calculation exam...

regislebrun:
+1

jschueller:
@regislebrun I added witdth, height keywords to control image size

dbarbier:
+1

PR #270, opened 2016-06-30
Tensor approximation

jschueller:
- It supports greedy rank-one
- The estimation of the coefficients in the alternating least-squares can be optionaly L-1 regularized
by default it's a plain estimation
- The optimal rank can be optionally estimated be k-fold cross-validation, by default it goes to the provided maximum rank
- I'm not quite sure about the convergence criteria for the rank-1 approximation: I added a relative residual convergence criterion to the radius convergence criteria to improve further the quality of the metamodel
- KFold could'nt be reused in TAA::greedyRankOneSelection, I hope to be able to change that
- I hesitated to use the less-specific term low-rank instead of tensor, what do you think ?
- Didn't have time to implement Sobol' indices : chapter 5.2 of rai 2015


lib/src/Base/Func/UniVariateFunctionEvaluationImplementation.cxx
```diff
@@ -0,0 +1,115 @@
...
+ */
+#include "openturns/UniVariateFunctionEvaluationImplementation.hxx"
+#include "openturns/OSS.hxx"
+#include "openturns/PersistentObjectFactory.hxx"
+
+BEGIN_NAMESPACE_OPENTURNS
+
+CLASSNAMEINIT(UniVariateFunctionEvaluationImplementation);
+
+static const Factory<UniVariateFunctionEvaluationImplementation> RegisteredFactory;
```
dbarbier: This static variable name  must be changed according to #233

dbarbier:
I see no reason to have RankOneTensor defined in CanonicalTensor.{cxx,hxx}.

lib/src/Base/Func/UniVariateFunctionEvaluationImplementation.cxx
```diff
@@ -0,0 +1,115 @@
...
+  oss << offset << __repr__();
+  return oss;
+}
+
+
+/* Operator (): Evaluate a product of 1D polynomials for one sample */
+NumericalPoint UniVariateFunctionEvaluationImplementation::operator() (const NumericalPoint & inP) const
+{
+  const UnsignedInteger inDimension(inP.getDimension());
+  if (inDimension != getInputDimension()) throw InvalidArgumentException(HERE) << "Error: trying to evaluate a ProductPolynomialFunction with an argument of invalid dimension";
```
dbarbier: Wrong class name in message

lib/src/Uncertainty/Algorithm/MetaModel/TensorApproximation/TensorApproximationAlgorithm.cxx
```diff
@@ -0,0 +1,647 @@
...
+
+    // compute residual
+    NumericalMathFunction combination(prodColl, rk);
+    yRes = y - combination(x);
+
+  }
+}
+
+void TensorApproximationAlgorithm::rankOne (const UnsignedInteger marginalIndex,
+                                            RankOneTensor & rank1Tensor,
```
dbarbier: rank1Tensor should be passed as return value instead of non-const reference

lib/src/Uncertainty/Algorithm/MetaModel/TensorApproximation/TensorApproximationAlgorithm.cxx
```diff
@@ -0,0 +1,647 @@
...
+                                            const NumericalSample & y,
+                                            NumericalScalar & marginalResidual,
+                                            NumericalScalar & marginalRelativeError)
+{
+  Bool convergence = false;
+  const UnsignedInteger dimension = x.getDimension();
+  const UnsignedInteger size = x.getSize();
+  NumericalPoint weight(size, 1.0 / size);
+  UnsignedInteger r = tensor_[marginalIndex].getRank();
+  Collection<Collection<Basis> > subBasis(r);
```
dbarbier: A 'Collection<Collection>' may hurt performance, but fortunately this variable is unused :-D

lib/src/Uncertainty/Algorithm/MetaModel/TensorApproximation/openturns/TensorApproximationResult.hxx
```diff
@@ -0,0 +1,155 @@
...
+  typedef PersistentCollection<NumericalMathFunction> NumericalMathFunctionPersistentCollection;
+
+  // friend class Factory<TensorApproximationResult>;
+
+  /** Default constructor */
+  TensorApproximationResult();
+#ifndef SWIG
+  /** Parameter constructor */
+  TensorApproximationResult(
+//     const NumericalMathFunction & model,
```
dbarbier: These comments are ugly, please remove dead code (ditto in cxx)

dbarbier:
Several header files are missing from OTTensorApproximation.hxx

lib/src/Uncertainty/Algorithm/MetaModel/TensorApproximation/TensorApproximationAlgorithm.cxx
jschueller: Actually I need some of it's parameters as input of the algorithm (basis and nk), so that would be weird to pass it as input, make a copy, modify the coefficients and copy the tensor back.

dbarbier:
To be honest, I found this PR very hard to review.  There are friend classes, arguments are non-const references so that their members are modified, several long methods have no comment, etc.

jschueller:
@dbarbier I agree, maybe I can try to improve the tensor structure and the related methods.

jschueller:
@dbarbier, The CanonicalTensor/RankOneTensor classes are clarified, and no more references or friend classes except for TAA.

lib/src/Uncertainty/Algorithm/MetaModel/TensorApproximation/CanonicalTensor.cxx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+
+CanonicalTensor::FunctionFamilyCollection CanonicalTensor::getFunctionFamilies() const
+{
+  return functionFamilies_;
+}
+
+
+Indices CanonicalTensor::getNk() const
```
dbarbier: Replace nk_ by degrees_ and getNk by getDegrees?

dbarbier:
I still don't understand which mathematical objects are represented by CanonicalTensor/RankOneTensor, and how they are used by TensorApproximationAlgorithm.  Its constructor contains
```
  tensor_ = Collection<CanonicalTensor>(outputSample.getDimension(), CanonicalTensor(functionFamilies, nk, maxRank));
```
thus all CanonicalTensor share the same functionFamilies and nk.  In TensorApproximationAlgorithm::runMarginal, unless I am mistaken, at l. 205
```
      proxy_[i] = DesignProxy(transformedInputSample_.getMarginal(i), tensor_[marginalIndex].rank1tensors_[0].getBasis(i));
```
getBasis() will return the same object for any marginalIndex.

I have the feeling that
 * coefficients_ and radius_ should be moved away from RankOneTensor, this class will then only define basis functions
 * CanonicalTensor will contain a collection of RankOneTensor like now, but also a collection of coefficients and radius; coefficients_ would then be a NumericalSample, and radius_ a NumericalPoint.  In TensorApproximationAlgorithm, tensor_ will then no more be a Collection<CanonicalTensor>, but a CanonicalTensor.

jschueller:
@dbarbier ok for renaming nk
RankOneTensor gathers the information of the product (a product term for each components) of a sum (each term in the sum is an element of the basis).
CanonicalTensor is a linear combination of rank-1 tensors (the rank of the tensor is the number of rank-1 combinations).

A CanonicalTensor represents a scalar function so we need one per marginal hence the collection.
I don't think grouping the data for several marginals makes sense in regard to the algorithms : we know how to learn rank-1 tensors ; and we know learning canonical format by aggregation of rank-1 (or the canonical form directly, but it doesn't work well here), but certainly not a multivariate representation in one go.

PR #272, opened 2016-07-06
Make C cobyla function private

dbarbier:
Another cobyla function is provided by NLopt.
On Windows with MSVC, linking fails with this error:
```
  nlopt.lib(cobyla.obj) : error LNK2005: cobyla already defined in algocobyla.obj
```
Since this function is needed only by Cobyla.cxx, we can embed it into Cobyla.cxx with few adjustements so that algocobyla.c compiles with a C++ compiler.

An alternative would be to rename cobyla into ot_cobyla, I can do that if requested.

### DIFF
--- a/lib/src/Base/Optim/Cobyla/Cobyla.cxx
+++ b/lib/src/Base/Optim/Cobyla/Cobyla.cxx
@@ -126,7 +126,7 @@ void Cobyla::run()
    * extern int cobyla(int n, int m, double *x, double rhobeg, double rhoend,
    *  int message, int *maxfun, cobyla_function *calcfc, void *state);
    */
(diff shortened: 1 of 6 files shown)


regislebrun:
+1

sofianehaddad:
+1, you should rebase with master

jschueller:
+1 Though renaming was way simpler.

dbarbier:
Please cherry-pick commit e3e4e24 into js/release for native Windows builds, thanks.

jschueller:
thanks!

jschueller:
ok, rebased

jschueller:
thanks for renaming

PR #274, opened 2016-07-21
[travis] Use CMake instead of autotools

sofianehaddad:
nlopt does not support autotools anymore since the integration of https://github.com/stevengj/nlopt/pull/49


sofianehaddad:
@jschueller @dbarbier I merge the PR in order to make possible travis compilation (PR affects only travis script and not the library)
Future enhancements are welcome

PR #275, opened 2016-07-25
Improved projection

regislebrun:
1) Improved IntegrationStrategy and LeastSquaresStrategy.
+ It uses the DesignProxy class efficiently,
+ The computation of the new coefficients is done in parallel
+ The computation of the residual is done using BLAS
There was also a bug in ProjectionStrategyImplementation: the inputSample_ member was not reinitialized when either the measure or the experiment were changed.

2) Added missing copyOnWrite() in Basis.

3) Fixed a bug in LowDiscrepancyExperiment.
+ The marginals were not correctly initialized when using constructors with no distribution.

4) Fixed a bug in FunctionalChaosAlgorithm
+ When the input sample is provided, no need to change the measure in the projection strategy.


sofianehaddad:
+1

PR #276, opened 2016-07-25
Nontensorized chaos

regislebrun:
1) Improved FunctionalChaosAlgorithm
+ Now it can work with bases with non-product measures (ie non-tensorized bases).

2) Fixed comparison operator for distributions.
+ It is a partial solution for ticket #845.

3) Added the identity function
+ It allows to improve the chaos result when the input distribution is equel to the orthogonality measure


sofianehaddad:
+1

jschueller:
@regislebrun I'm puzzled, I had to write new classes to use non-tensorized basis for low-rank tensors, how would you use non-tensorized basis since it uses OrthogonalProductPolynomialFactory. In this change I only see the transformation simplification.

regislebrun:
@jschueller The title is probably not the best one. The key point is in the mapping between the initial distribution (arbitrary margins and copula) and the orthogonality measure. Before this PR, the fact that the orthogonality measure had an independent copula was hard-coded, now it can support arbitrary copulas too. BTW, you are not supposed to use nontensorized bases for low-rank tensors as you should use only 1D orthogonal bases in the canonical tensor format.

PR #277, opened 2016-07-26
Performance enhancement

regislebrun:
1) Improved MatrixImplementation
+ Now, the constructor based on a collection of scalars uses std::copy for better performance

2) Improved LinearNumericalMathEvaluationImplementation
+ Now, the evaluation over a sample uses a matrix/matrix product instead of a loop over matrix/vector products

3) Added regularization in KarhunenLoeve[P1|Quadrature]Factory

4) Parallelized SoizeGhanemFactorEvaluation

5) Improved the getKendalTau() method in Copula
+ Now, it uses analytical formulas instead of numerical integration when the copula is elliptical

6) Made some methods return const references instead of copies
+ It greatly improves the speed of some algorithms when the returned object is large and is not modified, in which case a const reference is enough to have.

7) Improved the speed of the P1LagrangeEvaluationImplementation class.
+ Now the operator() method benefits from the const references returned by the Field and Mesh objects instead of costly copies. It greatly improves the speed of the evaluation over a point. The most demanding part of the evaluation over a sample is done in parallel, resulting in an additional speed-up. The cache mechanism is not activated in the evaluation over a sample.


sofianehaddad:
+1

PR #278, opened 2016-07-26
Fixed ticket #828 (CSV import)

regislebrun:
This PR fixes the CSV import problem reported in ticket #828

### DIFF
--- a/lib/src/Base/Stat/csv_lexer.ll
+++ b/lib/src/Base/Stat/csv_lexer.ll
@@ -43,7 +43,7 @@ integer   {digit}+
 exponent  [eE]{sign}?{integer}
 real      {sign}?({integer}("."({integer})?)?|"."{integer}){exponent}?
 
-alpha     [a-zA-Z_]
+alpha     [a-zA-Z_()]
 string    {alpha}({alpha}|{integer})*
 
 space     [ \t]
--- a/ChangeLog
+++ b/ChangeLog
@@ -57,6 +57,7 @@
  * #820 (Python distribution fails randomly when computing the PDF over a sample)
  * #822 (Incorect Matrix / point operations with cast)
  * #824 (Confusing behavior of NumericalSample::sort)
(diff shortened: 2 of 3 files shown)


sofianehaddad:
Could you add a test?
Also missing update in changeLog

regislebrun:
@sofianehaddad There was already a test in the library testsuite, and I just updated the ChangeLog.

lib/src/Base/Stat/csv_lexer.ll
```diff
@@ -43,7 +43,7 @@ integer   {digit}+
 exponent  [eE]{sign}?{integer}
 real      {sign}?({integer}("."({integer})?)?|"."{integer}){exponent}?
 
-alpha     [a-zA-Z_]
+alpha     [a-zA-Z_()]
 string    {alpha}({alpha}|{integer})*
```
jschueller: This seems restrictive still a bit restrictive compared to the autorized chars in the norm:
https://tools.ietf.org/html/rfc4180#page-2

it mentions TEXTDATA =  %x20-21 / %x23-2B / %x2D-7E
that is [space]! / #$%&'()*+ / -./[0-9]:;<=>?@[A-Z][\]^_`[a-z]{|}
according to https://fr.wikipedia.org/wiki/American_Standard_Code_for_Information_Interchange
so that would mean these characters should also be allowed.

regislebrun:
@jschueller The first idea was to fix *exactly* what was reported in ticket #828, then I tried to add all the printable characters but with no success. Either escaping or protecting the reserved characters failed. If you know how to do it please tell it to me.

jschueller:
I've come up with this patch: alpha-numerics plus explicitely all the symbols escaped as needed:
```
-alpha     [a-zA-Z_]
-string    {alpha}({alpha}|{integer})*
+string    [[:alnum:]!#\$%&'()\*\+\-./:<=>\?@\[\\\]\^_`\{\|\}~]*
```

jschueller:
see #285

PR #281, opened 2016-07-27
Trim description quotes on text import

jschueller:
If the description contains quotes they're not trimmed unliked the flex/bison
implementation.


sofianehaddad:
+1

regislebrun:
+1

PR #282, opened 2016-07-28
Improve composed distribution comparison

regislebrun:
1) Improved copula comparison.
+ It allows to detect independent copulas either at the CopulaImplementation level or at the ComposedDistribution level.

2) Fixed comments in FunctionalChaos/FunctionalChaosAlgorithm

3) Improved the simplified interface to FunctionalChaosAlgorithm
+ Now, it favors the use of sparse regression as the performace of this method has been greatly improved.


jschueller:
+1

PR #286, opened 2016-08-05
Set Sobol positve default boostrap size

jschueller:
otherwise the confidence interval length is null


sofianehaddad:
Maybe should also be documented

dbarbier:
+1

PR #289, opened 2016-08-09
Fix Collection objects Python comparison

jschueller:


sofianehaddad:
+1

dbarbier:
+1, but why do we have to define `__eq__` in the first place?  This fix is too specific IMHO, the problem is broader.  BTW python operator '!=' is wrong, we must also define `__ne__` methods.

jschueller:
@dbarbier I added a new macro to fix all these

sofianehaddad:
+1

PR #294, opened 2016-08-16
Include block size criteria in Subset condition

jschueller:
A condition used is p*N>1, unfortunately N was defined as
N=maximumoutersampling instead of maximumoutersampling * blocksize.


.travis.yml
```diff
@@ -19,6 +19,7 @@ addons:
     - python-scipy
     - python-matplotlib
     - python-sphinx
+    - python-six
     - texlive-latex-recommended
```
sofianehaddad: It seems that the used version for 'python-six' is 1.1.0-2 wheras required one is at least 1.4
Otherwise I think this should be in a separate PR

dbarbier:
+1

PR #298, opened 2016-08-25
Conditional include of headers

sofianehaddad:
In Os/Path, unistd.h, sys/types.h and sys/stat.h are systematically included (for rmdir, unlink, stat)

We should first check that OPENTURNS_HAVE_XXX variables are defined to include these last ones.

### DIFF
--- a/lib/include/OTconfig.h.in
+++ b/lib/include/OTconfig.h.in
@@ -63,10 +63,6 @@
 /* Support for regular expression library */
 #cmakedefine OPENTURNS_HAVE_LIBXML2
 
-/* Define to 1 if your system has a GNU libc compatible `malloc' function, and
-   to 0 otherwise. */
-#cmakedefine HAVE_MALLOC
-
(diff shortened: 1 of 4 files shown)


jschueller:
+1

sofianehaddad:
@jschueller rmdir/unlink are called in OS inconditionnaly to OPENTURNS_HAVE_XXX. Same for stat in Path.
What should happen if these headers are missing?

jschueller:
we would have noticed way sooner as it should result in a compilation error, but as linux, osx, and even msvc provide them we're fine.

sofianehaddad:
@jschueller agreed. Thus what is the interest of OPENTURNS_HAVE_XXX ? 
I think unistd.h, sys/types.h and sys/stat.h should be mandatory

jschueller:
Yes, for these it's useless.

jschueller:
:)

PR #301, opened 2016-08-29
Add Os::IsDirectory, IsFile

jschueller:
In particular use IsFile in Path::FindFileByNameInDirectoryList
instead of just 'stat' to check for a regular file.
This allows to work around what seems a bug compiling on conda/osx
with libcpp where this function returns a directory.

### DIFF
--- a/lib/src/Base/Common/Os.cxx
+++ b/lib/src/Base/Common/Os.cxx
@@ -274,45 +274,32 @@ int Os::DeleteDirectory(const String & path)
   }
 #endif
 
-  struct stat file_stat;
-  int rc = 0;
-  rc = stat( directory, &file_stat );
-  if (rc == 0)
-  {
-    if (! S_ISDIR(file_stat.st_mode))
-    {
-      // Not a directory
-      return 1;
-    }
-  }
-
 #ifndef WIN32
 
-  rc = nftw( directory, deleteRegularFileOrDirectory, 20, FTW_DEPTH );
-  if ( rc != 0 ) return 1;
+  int rc = nftw(directory, deleteRegularFileOrDirectory, 20, FTW_DEPTH);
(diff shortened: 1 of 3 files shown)


regislebrun:
+1

lib/src/Base/Common/Os.cxx
```diff
@@ -266,45 +274,32 @@ int Os::DeleteDirectory(const String & path)
...
+  int rc = nftw(directory, deleteRegularFileOrDirectory, 20, FTW_DEPTH);
+  if (rc != 0) return 1;
 
 #else /* WIN32 */
 
-  size_t timeout = ResourceMap::GetAsUnsignedInteger("output-files-timeout");
-  size_t countdown = timeout;
-  OT::String rmdirCmd = OT::String("rmdir /Q /S \"") + directory + "\"";
-  int directoryExists;
+  SignedInteger timeout = ResourceMap::GetAsUnsignedInteger("output-files-timeout");
```
dbarbier: Why signed?  Make them UnsignedInteger, ie. size_t

lib/src/Base/Common/Os.cxx
```diff
@@ -266,45 +274,32 @@ int Os::DeleteDirectory(const String & path)
...
     // check if directory still there (rmdir dos command always return 0)
-    struct stat dir_stat;
-    directoryExists = stat( directory, &dir_stat );
-    if( directoryExists == 0 )
+    directoryExists = IsDirectory(path);
+    if (directoryExists)
     {
-      if  ( countdown <= 0 ) return 1;
-      --countdown;
+      if (countdown <= 0) return 1;
```
dbarbier: if (countdown == 0)

dbarbier:
+1

PR #302, opened 2016-08-29
Remove LINK_PYTHON_LIBRARY option

jschueller:
Doesn't seem to be needed anywhere.


sofianehaddad:
-1, the option is mandatory for OSX systems

jschueller:
indeed

PR #309, opened 2016-09-08
Bugfix in CorrelationAnalysis::SRC

sofianehaddad:
Normalization factor should be the variance estimated thanks to the linear
regression and not the initial data.


regislebrun:
+1

dbarbier:
+1

PR #310, opened 2016-09-09
Generalize WeightedExperiment::generateWithWeights

jschueller:
All weighted experiment to implement generateWithWeights
generate is calling generateWithWeights
getWeight is deprecated
Fixes http://trac.openturns.org/ticket/840

PR #311, opened 2016-09-12
Fix LHSExperiment(size) constructor

jschueller:


TODO
```diff
@@ -1 +1,2 @@
 Remove deprecated LAR class in 1.10 version
+Remove deprecated WeightedExperiment::getWeight
```
sofianehaddad: Where did you remove it?

jschueller:
in #310 but I forgot to mention it

TODO
jschueller: in #310 but I forgot to mention it

sofianehaddad:
@jschueller no in #310 you deprecated it but it is not removed

jschueller:
yes, deprecated for 1.8 and to be removed in the next version

sofianehaddad:
ah sorry it is the TODO file

sofianehaddad:
+1

regislebrun:
+1

PR #312, opened 2016-09-14
Add SpecFunc::BinomialCoefficient

jschueller:


jschueller:
I mostly needed the integer number of combinations ; do you think it's ok to add UInt SpecFunc::Combinations(n,k) that calls BinomialCoefficient(n,k) and throws if overflow ?

regislebrun:
Only the integer version seems to be needed, so BinomialCoefficient should return an integer. In order to get an accurate value, one should think about performing only integer arithmetic, I am not 10% sure that the rounded floating point computation is accurate enough for large arguments. Have a look here:
http://stackoverflow.com/questions/11032781/fastest-way-to-generate-binomial-coefficients

jschueller:
Depends if large is greater than the integer range.
Anyway integer computation is much cheaper for our needs.

regislebrun:
+1

PR #315, opened 2016-09-15
Avoid 'total order' 

jschueller:
For consistency with FunctionalChaosRandomVector, dont use 'total order' indice, just total indice when mentioning sensitivity indices, as remarked by @aurelieladier.
- in SobolIndicesAlgorithm
- and FAST


dbarbier:
I am not convinced that this change is useful, FunctionalChaosRandomVector has a getSobolTotalIndex method, not getTotalIndex.

sofianehaddad:
I think that the same work should be done on the FunctionalChaosRandomVector to get the same method names in all classes

jschueller:
ok, not convinced either

PR #316, opened 2016-09-16
Fix problems with SWIG binding of ComplexMatrix::solveLinearSystem

dbarbier:
Fixes trac#802.

Not sure why this change is needed though. 
Other *Matrix::solveLinearSystem methods were not affected.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -73,6 +73,7 @@
  * #604 (Bug concerning the NonCentralStudent distribution)
  * #698 (KernelSmoothing() as a factory)
  * #786 (Bug in sensitivity analysis)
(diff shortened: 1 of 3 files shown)


regislebrun:
+1

sofianehaddad:
Please add notes on the ChangeLog

sofianehaddad:
Could you please add a test that illustrates the problem is solved?

jschueller:
yes please a small test-case, and squash them except 295e73c

dbarbier:
I can't add a test, AFAICT this problem only occurs from an interactive session.

jschueller:
ok

PR #317, opened 2016-09-19
Deprecate alternative parameter set flag

jschueller:


lib/src/Uncertainty/Distribution/NormalFactory.cxx
```diff
@@ -87,4 +89,26 @@ Normal NormalFactory::buildAsNormal() const
...
+DistributionFactoryResult NormalFactory::buildEstimator(const NumericalSample & sample) const
+{
+  Normal distribution(buildAsNormal(sample));
+  NumericalPoint mu(distribution.getMean());
+  NumericalPoint sigma(distribution.getSigma());
+  const UnsignedInteger size = sample.getSize();
+  const UnsignedInteger dimension = distribution.getDimension();
+  NumericalPoint parameter(distribution.getParameter());
+  ComposedDistribution::DistributionCollection coll(parameter.getSize());
+  for (UnsignedInteger j = 0; j < dimension; ++ j)
```
sofianehaddad: No the asymptotic estimate is for dimension 1 only
For upper dimension it is more complex (Wishart product...)

jschueller: ok, fixed

lib/src/Uncertainty/Distribution/NormalFactory.cxx
```diff
@@ -87,4 +89,22 @@ Normal NormalFactory::buildAsNormal() const
   return Normal();
 }
 
+
+DistributionFactoryResult NormalFactory::buildEstimator(const NumericalSample & sample) const
+{
+  if (sample.getDimension() != 1) return DistributionFactoryImplementation::buildEstimator(sample);
+  Normal distribution(buildAsNormal(sample));
```
sofianehaddad: No bootstrap for dimension > 1?

jschueller: ok

lib/src/Uncertainty/Distribution/NormalFactory.cxx
```diff
@@ -87,4 +89,22 @@ Normal NormalFactory::buildAsNormal() const
   return Normal();
 }
 
+
+DistributionFactoryResult NormalFactory::buildEstimator(const NumericalSample & sample) const
+{
+  if (sample.getDimension() > 1) return buildBootStrapEstimator(sample);
```
jschueller: What about size=1 ? buildAsNormal will throw as usual

jschueller: @sofianehaddad I see what you mean, added 46c8e9e

lib/src/Uncertainty/Distribution/Arcsine.cxx
```diff
@@ -43,13 +43,28 @@ Arcsine::Arcsine()
...
 
 /* Parameters constructor */
+Arcsine::Arcsine(const NumericalScalar a,
+                 const NumericalScalar b)
+  : ContinuousDistribution()
+  , a_(a)
+  , b_(b)
+{
+  if (a >= b)
+    throw InvalidArgumentException(HERE) << "in Arcsine : a must be smaller than b";
```
dbarbier: As already discussed, it would be better to negate these checks: if you want to ensure that a is smaller than b, then it should be written as
```
if (!(a < b))
    throw InvalidArgumentException(HERE) << "in Arcsine : a must be smaller than b";
```
This way, it catches all gotchas (a or b being NaN, etc).

lib/test/t_Arcsine_std.cxx
```diff
@@ -135,14 +135,6 @@ int main(int argc, char *argv[])
...
     for (UnsignedInteger i = 0; i < 6; ++i) fullprint << "standard moment n=" << i << ", value=" << distribution.getStandardMoment(i) << std::endl;
     fullprint << "Standard representative=" << distribution.getStandardRepresentative()->__str__() << std::endl;
-    // Specific to this distribution
-    NumericalScalar mu = distribution.getMu();
-    fullprint << "mu=" << mu << std::endl;
-    NumericalScalar sigma = distribution.getSigma();
-    fullprint << "sigma=" << sigma << std::endl;
-    Arcsine newDistribution(mu, sigma, Arcsine::MUSIGMA);
-    fullprint << "a from (mu, sigma)=" << newDistribution.getA() << std::endl;
-    fullprint << "b from (mu, sigma)=" << newDistribution.getB() << std::endl;
```
dbarbier: Shouldn't we keep tests unchanged to make sure that the compatibility layer is working as expected, and remove them only when methods are dropped?

jschueller: I guess you're right

lib/src/Uncertainty/Distribution/Arcsine.cxx
jschueller: ok

lib/src/Uncertainty/Distribution/NormalFactory.cxx
sofianehaddad: +1

regislebrun:
+1

dbarbier:
+1

PR #319, opened 2016-09-19
Fix Python bindings for HMatrix::compressionRatio

dbarbier:
Without this change, swig is confused:
```
 >>> cov.compressionRatio()
 <Swig Object of type 'std::pair< size_t,size_t > *' at 0x11bb13180>
```
With this change:
```
 >>> cov.compressionRatio()
 (36L, 36L)
```


regislebrun:
+1

PR #321, opened 2016-09-20
Fix and enhance InverseWishart.computeLogPDF

mathieu-couplet:
See http://trac.openturns.org/ticket/854
A problem was the inialization of the logNormalizationFactor and its use inside the method computeLogPDF.
Moreover the computation of the term Trace(...) is now more efficient.


lib/src/Uncertainty/Distribution/InverseWishart.cxx
```diff
@@ -48,11 +48,12 @@ InverseWishart::InverseWishart(const CovarianceMatrix & v,
                                const NumericalScalar nu)
   : ContinuousDistribution()
   , cholesky_()
-  , nu_(nu)
+  , nu_(-1.) // implies nu_ != nu: see setNu
```
regislebrun: Please use -1.0 instead of -1.

lib/src/Uncertainty/Distribution/InverseWishart.cxx
```diff
@@ -189,15 +190,16 @@ NumericalScalar InverseWishart::computeLogPDF(const CovarianceMatrix & m) const
     // If the Cholesky factor is not defined, it means that M is not symmetric positive definite (an exception is thrown) and the PDF is zero
     TriangularMatrix X(CovarianceMatrix(m).computeCholesky());
     // Compute the determinant of the Cholesky factor, ie the square-root of the determinant of M
-    NumericalScalar logPDF = logNormalizationFactor_;
+    NumericalScalar logPDF = 0.;
```
regislebrun: Same remark here: 0.0 instead of 0.

lib/src/Uncertainty/Distribution/InverseWishart.cxx
```diff
@@ -189,15 +190,16 @@ NumericalScalar InverseWishart::computeLogPDF(const CovarianceMatrix & m) const
...
-    // V^{-1}M = (CC')^{-1}(XX')
-    // = C'^{-1}(C^{-1}X)X'
+    // Add the term which does not depend on M
+    logPDF += logNormalizationFactor_;
+    // Trace(V M^{-1}) = Trace(C C' X'^{-1} X^{-1}) = Trace(C'X'^{-1} X^{-1}C)
+    //                 = Trace(A'A) with A = X^{-1}C
     TriangularMatrix A(X.solveLinearSystem(cholesky_).getImplementation());
-    SquareMatrix B((A * cholesky_.transpose()).getImplementation());
-    SquareMatrix C(X.transpose().solveLinearSystem(B).getImplementation());
+    SquareMatrix C((A.transpose() * A).getImplementation());
```
regislebrun: Here use logPDF -= 0.5 * A.computeGram().computeTrace(), no need to cast the result into a SquareMatrix to have access to computeTrace() as the result of A.computeGram() is a CovarianceMatrix, hence a SquareMatrix. The computeGram() method uses a specific BLAS implementation, so is twice the speed of A.transpose() * A and avoid the data copy done by A.transpose()

regislebrun:
Great! Thanks

jschueller:
This also needs a testcase to be added.

regislebrun:
Not clear yet if it solves the ticket. I only checked the code, not the maths. Are you confident in your patch on this point?

mathieu-couplet:
My confidence is based on three different tests which are implemented by a Python script provided as an attachment of the trac ticket (http://trac.openturns.org/ticket/854) :
1) A random checking that InverseWishart.computeLogPDF = InverseGamma.computeLogPDF in the 1D  case
2) A random "test of coherence" between the n-dimensional case (only available for *diagonal* input and parameter diagonal matrices) and the 1D case (checking that the log ratio of the n-dimensional PDF by the 1D PDF is equal to the theoretical value).
3) A comparison in a random n-dimensional case (n>1) with the PDF computed with the R package MCMCpack.

You can check the tests 1) and  2) with the R package MCMCpack using the R script I've just added to the trac ticket.

mathieu-couplet:
Concerning 1), a quick and dirty checking that the integral of the PDF over the real line is equal to 1 was also performed.
MC.

jschueller:
Nevertheless a test case should be added into openturns testsuite.
Currently there isn't even a python/test/t_Whishart_std.py file, I suggest you add one (you can copy-paste from another distribution).

mathieu-couplet:
I agree. There is also the ticket http://trac.openturns.org/ticket/848...
I will try to have some time to propose a python test before october 14.

jschueller:
What the status of this ?

mathieu-couplet:
I sent a email about the status of this at developers@openturns.org.
What is the good way to move forward in the github workflow to propose a new commit ? Thanks.

jschueller:
You can just commit to the same branch and push.
I did not receive any email, the best is to comment here.

python/test/t_InverseWishart_std.py
```diff
@@ -0,0 +1,171 @@
...
+            print("Run " + member_name + "... ", end="")
+            sys.stdout.flush()
+            try:
+                test.debug()
+                print("SUCCESS")
+            except  Exception, exception:
+                print("FAILURE")
+                print(exception)
+            test.tearDown()
+    cls.tearDownClass()
```
jschueller: Unfortunately tearDownClass is new for 2.7, and we want to support python 2.6 a little longer,
would you mind not using it if version is < 2.7 ?
https://docs.python.org/2/library/unittest.html#unittest.TestCase.tearDownClass

jschueller:
I did not see that unittest was a default module, so sorry for my previous erroneous remark, apart from the tearDownClass usage this looks good.

mathieu-couplet:
Thank you. I've just fixed the problem (I hope : I did not test on python 2.6 because I cannot do it right now on my machine).

python/test/t_InverseWishart_std.py
```diff
@@ -0,0 +1,180 @@
...
+            is_tearDownClass_there = True
+    for test_name in test_names:
+        test = cls(test_name)
+        test.setUp()
+        print("Run " + test_name + "... ", end="")
+        sys.stdout.flush()
+        try:
+            test.debug()
+            print("SUCCESS")
+        except  Exception, exception:
```
jschueller:
```
  File "/home/openturns/distcheck/workdir/python/test/t_InverseWishart_std.py", line 37
    except  Exception, exception:
                     ^
SyntaxError: invalid syntax
```

mathieu-couplet: It is an obsolete syntax forbidden in python 3 indeed. It is fixed.

regislebrun:
@mathieu-couplet Should it be possible to enhance the C++ test too? Python tests are not checked in the  production of the native Windows C++ library.

jschueller:
@regislebrun no msvc c++ test is even ci-tested yet

regislebrun:
@jschueller 
Yes BUT:
+ it will probably change sooner or later and I don't want to inspect all the tests this day
+ it is a good practice anyway, agreeded in the coding rules

jschueller:
@regislebrun I just meant native windows tests are currently not run in the ci.

mathieu-couplet:
Unfortunately I cannot commit some c++ tests before the 7th of november.
From my point of view, accepting this pull request (after taking account the last remarks) could allow to close the tickets and would be an improvement.
Moreover, the InverseWishart class does not really exists since it is not documented, and the Wishart class is neither tested nor documented yet, as well.
A real problem would be no-tested but documented classes.
The work could be completed for release 1.9.
What do you think?

jschueller:
Ok, be sure to update the Changelog file to mention the tickets it solves, plus you can add yourself to the AUTHORS file!

regislebrun:
Yes BUT:
+ it will probably change sooner or later and I don't want to inspect all the tests this day
+ it is a good practice anyway, agreeded in the coding rules

regislebrun:
+1

mathieu-couplet:
Could you tell me how to get details about failed tests please? Thanks.

jschueller:
Hello, the tests are not even run because of merge conflicts in the ChangeLog file ; you edited an obsolete version of the file. Please rebase to master and resolve conflicts.

mathieu-couplet:
I am sorry but, unfortunately, I cannot fetch and merge anymore today to resolve the conflict of ChangeLog.

mathieu-couplet:
@jschueller Could you please add these two lines in the ChangeLog, please ?

=== Bug fixes ===
 * #854 (InverseWishart.computeLogPDF)
 * #848 (Inverse Wishart sampling)

Thanks.

PR #323, opened 2016-09-22
Missing abs in LinearModelTest::LinearModelDurbinWatson

jschueller:


jschueller:
cc @adumasphi

regislebrun:
Is there a reference to check if the maths are correct? Looks good but I have no time to dig into my math books to check the formula...

jschueller:
@adumasphi is there a reference for the formula somewhere ?

adumasphi:
Yes, there are references : Testing for Serial Correlation in Least Squares Regression. I, II et III.
DOI : 10.2307/2332391
DOI : 10.2307/2332325
DOI : 10.2307/2334313
I have the papers if you want to read it.

The formulation I have coded is coming from the third paper in which it corresponds with a normal approximation of the computed statistic. In this third paper, Durbin and Watson compare several methods to estimate the p-value. They conclude that this method is accurate enough for common practice. More accurate methods exist such as one using the "pan" algorithm or the Henshaw method.

dbarbier:
I did not read those papers, but approved these changes after checking that the same formula is used in R lmtest package.

PR #324, opened 2016-09-22
Fixed ticket 853

regislebrun:
Changed the default value of the 'normalize' flag from true to false in GeneralizedLinearModelFactory.


python/test/t_BoxCoxFactory_glm.expout
```diff
@@ -1,2 +1,2 @@
 myBoxCox (GLM) = BoxCox(lambda=[1.80518], shift=[0.1])
```
sofianehaddad: I am wondering if we should not define a new ResourceMap key for BoxCoxFactory and GLM?

jschueller:
Unsetting this flag makes GLMF test deterministic again??

PR #325, opened 2016-09-27
Allow to change TBB threads number at runtime

jschueller:
Added TBB::SetNumberOfThreads which is called in ResourceMap::SetAsUnsignedInteger.

This can be used to get a deterministic behavior of KrigingAlgorithm with normalization enabled.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -71,6 +71,7 @@
  * Shortened ResourceMap keys to not contain 'Implementation'
 
 === Bug fixes ===
+ * #535 (parallel-threads option cannot be changed at runtime with TBB)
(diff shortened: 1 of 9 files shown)


dbarbier:
IMO it would be better to add static methods `TBB::Disable()` and `TBB::Enable()` to temporarily disable TBB, I cannot think of any usage where you call `TBB::SetNumberOfThreads` with values different from 1 and total number of processors.

jschueller:
@dbarbier, ok, it's not more changed through resourcemap

regislebrun:
+1 Usefull when there is a mix between TBB threads and OpenBLAS threads!

PR #326, opened 2016-09-28
Fix MinGW compilation with newer GCC

jschueller:
A macro from pyconfigh.h conflicts with a definition in cmath:
error: '::hypot' has not been declared

http://stackoverflow.com/questions/10660524/error-building-boost-1-49-0-with-gcc-4-7-0/12124708#12124708


sofianehaddad:
+1

PR #329, opened 2016-10-06
Homogenization in covariance models

sofianehaddad:
1) Do not define new covariance attribut in DiracCovarianceModel but use this already defined one
2) We should add nugget factor if values are very small and not exactly zero (as done in Matern)


lib/src/Base/Stat/DiracCovarianceModel.cxx
```diff
@@ -384,15 +380,14 @@ String DiracCovarianceModel::__str__(const String & offset) const
 void DiracCovarianceModel::save(Advocate & adv) const
 {
   StationaryCovarianceModel::save(adv);
-  adv.saveAttribute("covariance_", covariance_);
+  adv.saveAttribute("spatialCovariance_", spatialCovariance_);
```
jschueller: saved but not loaded, as it's mutable I guess it should not be

jschueller:
+1

regislebrun:
+1

PR #331, opened 2016-10-16
Confidence region computations

regislebrun:
This pull request adds new algorithms to compute various kind of confidence regions:
- minimum volume level set
- product of marginal minimum volume intervals
- product of bilateral confidence intervals
- unilateral confidence intervals

It also contains some code cleanup, some performance improvement (mainly the switch from binded methods to specific NumericalMathFunction) and a few bug fixes.


jschueller:
- The build fails because hmat-oss master is in a bad shape, but I'm sure @jeromerobert or @gsylvand will resolve this soon : )
- That'be great to have the definition of each confidence region, but I couldn't find the documentation.
- The binding of new methods like computeMinimumVolumeInterval could be enhanced so as to return a tuple, see https://github.com/openturns/openturns/blob/master/python/src/WeightedExperimentImplementation.i

regislebrun:
The documentation was lost during the merge process, it is back! For the SWIG sorcellery, I will have a look at your pointer. Hallowen is coming, maybe I will find the trick!

regislebrun:
For the SWIG trick, the trouble is that each new method comes in two flavors:
  * one which returns only an Interval or a LevelSet
  * one which gives  also the marginal probability or the PDF threshold

I would have to rename these methods to disambiguate the call when one want the auxiliary value or not, and it is too much work to be done in a short term (all the tests have to be changed!)

jschueller:
I don't understand why there's a NP signature, each time it's of dim 1.

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -761,14 +862,45 @@ NumericalScalar DistributionImplementation::computeProbabilityContinuous(const I
   if (reducedInterval == getRange()) return 1.0;
   // Use adaptive multidimensional integration of the PDF on the reduced interval
   const PDFWrapper pdfWrapper(this);
-  const NumericalMathFunction fPDF(bindMethod<PDFWrapper, NumericalPoint, NumericalPoint>(pdfWrapper, &PDFWrapper::computePDF, dimension_, 1));
   NumericalScalar probability;
```
jschueller: = 0

regislebrun:
The Np signature is here to allow for side-effect in Python, ie arguments that are modified. The C++ version uses a NumericalScalar as expected.

jschueller:
Oh I see, so it's useless for C++ only, this definitively needs to change, I'll provide a patch.

jschueller:
What's the name of the return argument, I see 'marginalprob' and 'threshold' in the cxx, but in the docsting I see level, what should I use to name the new methods:
```
  virtual Interval computeMinimumVolumeIntervalWithProbability(const NumericalScalar prob, NumericalScalar & marginalProb) const;

virtual Interval computeBilateralConfidenceIntervalWithProbability(const NumericalScalar prob, NumericalScalar & marginalProb) const;
  virtual Interval computeUnilateralConfidenceIntervalWithProbability(const NumericalScalar prob, const Bool tail, NumericalScalar & marginalProb) const;
  virtual LevelSet computeMinimumVolumeLevelSetWithThreshold(const NumericalScalar prob, NumericalScalar & threshold) const;


```

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -761,14 +862,45 @@ NumericalScalar DistributionImplementation::computeProbabilityContinuous(const I
...
+    if (singularitiesNumber == 0 || singularities[0] >= upper || singularities[singularitiesNumber - 1] <= lower) probability = GaussKronrod().integrate(pdfWrapper, reducedInterval, error)[0];
+    else
+    {
+      NumericalScalar a = lower;
+      for (UnsignedInteger i = 0; i < singularitiesNumber; ++i)
+      {
+        const NumericalScalar b = singularities[i];
+        if (b > lower && b < upper)
+        {
+          probability += GaussKronrod().integrate(pdfWrapper, Interval(a, b), error)[0];
```
jschueller: maybe that's causing the meixner troubles ?

jschueller:
+1

jschueller:
@regislebrun, could you rebase to master ?

regislebrun:
@jschueller No problem, but you merge new PRs quicker than I rebase/test/push!

lib/test/t_Normal_std.cxx
```diff
@@ -41,6 +41,8 @@ inline NumericalPoint clean(NumericalPoint in)
 int main(int argc, char *argv[])
 {
   TESTPREAMBLE;
+  Log::Show(Log::INFO);
```
sofianehaddad: I think we should remove such debugs in tests

lib/src/Uncertainty/Model/ContinuousDistribution.cxx
```diff
@@ -183,8 +182,8 @@ Collection<PiecewiseHermiteEvaluationImplementation> ContinuousDistribution::int
     NumericalSample fi;
     NumericalPoint ei;
     NumericalScalar error;
-    valuesCDF[i] = valuesCDF[i - 1] + algo.integrate(fPDF, xCDFOld, xCDF, error, ai, bi, fi, ei)[0];
-    valuesCCDF[n - i - 1] = valuesCCDF[n - i] + algo.integrate(fPDF, xCCDF, xCCDFOld, error, ai, bi, fi, ei)[0];
+    valuesCDF[i] = valuesCDF[i - 1] + algo.integrate(pdfWrapper, xCDFOld, xCDF, error, ai, bi, fi, ei)[0];
+    valuesCCDF[n - i - 1] = valuesCCDF[n - i] + algo.integrate(pdfWrapper, xCCDF, xCCDFOld, error, ai, bi, fi, ei)[0];
     derivativesCDF[i] = computePDF(xCDF);
     derivativesCCDF[n - i - 1] = -computePDF(xCCDF);
     xCDFOld = xCDF;
```
sofianehaddad: The main part of the switch is done in e0e8bf2

jschueller:
As I said earlier the tests do not pass without boost:
- disable the Unilateral confidence interval test in t_NonCentralStudent
- the following must be applied in distfunc:

```
diff --git a/lib/src/Uncertainty/Distribution/DistFunc.cxx b/lib/src/Uncertainty/Distribution/DistFunc.cxx
index 08d5847..234d196 100644
--- a/lib/src/Uncertainty/Distribution/DistFunc.cxx
+++ b/lib/src/Uncertainty/Distribution/DistFunc.cxx
@@ -451,8 +451,11 @@ NumericalScalar DistFunc::qGamma(const NumericalScalar k,
                                  const NumericalScalar p,
                                  const Bool tail)
 {
-  if (!tail && (p >= 1.0 - SpecFunc::NumericalScalarEpsilon)) return SpecFunc::RegularizedIncompleteGammaInverse(k, p, tail);
-  return SpecFunc::RegularizedIncompleteGammaInverse(k, p, tail);
+  NumericalScalar r = p;
+  if (r <= SpecFunc::MinNumericalScalar) r = SpecFunc::MinNumericalScalar;
+  //if (r >= 1.0 - SpecFunc::NumericalScalarEpsilon) r = 1.0 - SpecFunc::NumericalScalarEpsilon;
+  //if (!tail && (p >= 1.0 - SpecFunc::NumericalScalarEpsilon)) return SpecFunc::RegularizedIncompleteGammaInverse(k, 1.0 - SpecFunc::NumericalScalarEpsilon, !tail);
+  return SpecFunc::RegularizedIncompleteGammaInverse(k, r, tail);
 }
```

regislebrun:
@jschueller There is definitely something strange in the current DistFunc::qGamma() function, but I am not convinced by your patch. If it fixes the problem let's go for it, but something deeper has to be understood here.

regislebrun:
@jschueller Both Maple and Mathematica have trouble to compute non-central Student CDF in the far tails, so it is hard to say which of the two implementations (boost or native) is correct. Let's adopt your patch waiting for a more satisfactory answer.

jschueller:
Could you rebase to master ?

regislebrun:
@jshueller ongoing

regislebrun:
@jschueller YES! YES! YES! Please merge it ASAP! No more bug without boost! No more bug on i386! No more fancy behaviour from fancy distributions! No more timeout! No more rebase due to other PRs being merged!

PR #334, opened 2016-10-18
Covariance model parameter management

jschueller:
Add an api to control the active parameters with indices (from adutofy/covariance_model_parameter)
the covariance models implement:
```
  /** Parameter accessor */
  virtual void setFullParameter(const NumericalPoint & parameter);
  virtual NumericalPoint getFullParameter() const;
  virtual Description getFullParameterDescription() const;
```
Then get/setParameter filters according to activeParameter_ indices


regislebrun:
+1

PR #335, opened 2016-10-19
LeastSquaresMethod/DesignProxy swig

jschueller:
- Expose these to SWIG
- Add tests, doc
- Fix bugs


regislebrun:
+1

lib/src/Base/Algo/DesignProxy.cxx
```diff
@@ -126,23 +126,35 @@ MatrixImplementation DesignProxy::computeDesign(const Indices & indices) const
...
+    design = filteredDesign;
+  } // hasRowFilter()
+  if (hasWeight())
+  {
+    UnsignedInteger linearIndex = 0;
+    for (UnsignedInteger j = 0; j < design.getNbColumns(); ++ j)
+    {
+      for (UnsignedInteger i = 0; i < design.getNbRows(); ++ i)
+      {
+        design[linearIndex] *= weight_[i];
```
regislebrun: But it is only SLIGHTLY wrong... Who cares? Ok I will fix it ;-)

jschueller: Ok, fixed.

PR #341, opened 2016-10-20
New Text class, derived from DrawableImplementation

dbarbier:
Add four new methods to Drawable:
  * getTextAnnotations()/setTextAnnotations()
  * getTextPositions()/setTextPositions()

This class allows adding text to plots.
Text position indicates where text is placed with respect to coordinates:
 * 1 (below)
 * 2 (left)
 * 3 (above)
 * 4 (right)


regislebrun:
I am not very fond of numerical IDs for the position, a string is more readable in my opinion. Why not adopt something similar to the legend position? Otherwise it is a good job!

python/src/viewer.py
```diff
@@ -427,6 +429,37 @@ def __init__(self,
...
+                    horizontal = horizontal_default
+                    vertical = vertical_default
+                    if positions[i] == 2:
+                        horizontal='right'
+                    elif positions[i] == 4:
+                        horizontal='left'
+                    if positions[i] == 3:
+                        vertical='bottom'
+                    elif positions[i] == 1:
+                        vertical='top'
```
jschueller: I'd use a map here:
```
position_map = {1: 'top', 2: 'right', ...}
horizontal = position_map.get(i, 'center')
```

jschueller:
+1

regislebrun:
+1

python/src/viewer.py
```diff
@@ -427,6 +429,37 @@ def __init__(self,
...
+                    horizontal_default = text_kwargs['horizontalalignment']
+                    del text_kwargs['horizontalalignment']
+                else:
+                    horizontal_default = 'center'
+                if 'verticalalignment' in text_kwargs_default:
+                    vertical_default = text_kwargs['verticalalignment']
+                    del text_kwargs['verticalalignment']
+                else:
+                    vertical_default = 'center'
+                positions = drawable.getTextPositions()
```
jschueller: positions is now a description, seems you forgot to change that part of the code

dbarbier: Good catch, fixed now, thanks.

jschueller:
+1

PR #346, opened 2016-10-21
Added drawLogPDF() to Distribution

regislebrun:


jschueller:
+1

sofianehaddad:
small question : how to handle - SpecFunc::MaxNumericalScalar values ?

regislebrun:
It all depends on what we mean by being close to zero. Is it to be equal to SpecFunc::NumericalScalarEpsilon or to be equal to SpecFunc::LogMinNumericalScalar? If all cases, it should not be -SpecFunc::MaxNumericalScalar.

Otherwise, as soon as there are only normal numbers into the NS defining a Graph, everything should by allright.

sofianehaddad:
@regislebrun accoring to the convention in OT, if pdf=0 then logPdf = -SpecFunc::NumericalScalarEpsilon 
Default parameters of the drawLogPDF method make the evaluation correct (ie no point for which pdf=0). However we might face to trouble. 
For example, if we consider this distribution, ot.Mixture([ot.Normal(-1,0.3), ot.Normal(15,1)])
Lots of "zeros" are to be considered by the drawPDF method. 
The question is how to handle this case ?

regislebrun:
@sofianehaddad I checked in the code: the convention is more to take logPdf = -SpecFunc::MaxNumericalScalar (otherwise you get -2.e-16, not really -\inf ;-)). It should definitely be SpecFunc::LogMinNumericalScalar  for log-likelihood/log-PDF computations, as it insures a kind of "numerical continuity".
Concerning the case where you have a lot of zeros in the PDF, you can play with the Graph bounding box to adapt the axes to something meaningful, waiting for a clever solution.

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -2808,6 +2835,219 @@ Graph DistributionImplementation::drawPDF() const
...
+  marginalGraph.setTitle(OSS() << getDescription() << "->" << description_[marginalIndex] << " component log PDF");
+  return marginalGraph;
+}
+
+/* Draw the log-PDF of the distribution when its dimension is 2 */
+Graph DistributionImplementation::drawLogPDF(const NumericalPoint & xMin,
+    const NumericalPoint & xMax,
+    const Indices & pointNumber) const
+{
+  if ((pointNumber[0] <= 2) || (pointNumber[1] <= 2)) throw InvalidArgumentException(HERE) << "Error: the discretization must have at least 2 points per component";
```
dbarbier: Error message says that pointNumber must be at least 2, but test is wrong, it checks whether it is strictly superior to 2.
BTW please write rewrite tests to handle the case where value is not a number:
```
  if (!(pointNumber[0] >= 2 && pointNumber[1] >= 2)) throw...
```

regislebrun:
@jschueller  Impossible to see what is going wrong on travis: no access to the log file...

jschueller:
(py)t_FittingTest_std.expout seems out of date and TrapezoidalFactory timeouts.

PR #347, opened 2016-10-21
Improved Classifier/MixtureClassifier/ExpertMixture

regislebrun:
Now, both grade() and classify() methods are parallelized and/or vectorized.

### DIFF
--- a/lib/src/Base/Algo/openturns/ClassifierImplementation.hxx
+++ b/lib/src/Base/Algo/openturns/ClassifierImplementation.hxx
@@ -51,11 +51,23 @@ public:
   virtual UnsignedInteger classify(const NumericalPoint & inP) const;
   virtual Indices classify(const NumericalSample & inS) const;
 
+protected:
+  virtual Indices classifyParallel(const NumericalSample & inS) const;
+  virtual Indices classifySequential(const NumericalSample & inS) const;
+public:
+
   /** Grade a point as if it were associated to a class */
   virtual NumericalScalar grade(const NumericalPoint & inP,
                                 const UnsignedInteger outC) const;
   virtual NumericalPoint grade(const NumericalSample & inS,
                                const Indices & outC) const;
+
+protected:
+  virtual NumericalPoint gradeTBB(const NumericalSample & inS,
+                           const Indices & outC) const;
+  virtual NumericalPoint gradeSequential(const NumericalSample & inS,
+                                  const Indices & outC) const;
+public:
 
   /** String converter */
   virtual String __repr__() const;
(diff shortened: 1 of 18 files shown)


jschueller:
It's not consistent with the rest of the code to use a flag to use the parallel implementation ; everywhere else you have to use the global TBB::Disable.

dbarbier:
@jschueller There is no use of TBB::Enable/Disable currently, so for consistency reasons I suggest to not use them here ;-)
We have MaximumLikelihoodFactory-Parallel and Distribution-Parallel resource keys, I would thus prefer to rename Classifier-UseTBB into Classifier-Parallel, and also mimic DistributionImplementation: add an isParallel_ member, and isParallel/setParallel methods.

jschueller:
I agree

lib/src/Uncertainty/Algorithm/Classification/MixtureClassifier.cxx
```diff
@@ -77,6 +78,48 @@ UnsignedInteger MixtureClassifier::classify(const NumericalPoint& inP) const
...
 
+Indices MixtureClassifier::classify(const NumericalSample & inS) const
+{
+  if (inS.getDimension() != getDimension()) throw InvalidArgumentException(HERE) << "Error: the sample to classify has dimension=" << inS.getDimension() << " but the classifier expects dimension=" << mixture_.getDimension();
+  if (ResourceMap::GetAsBool("Classifier-Parallel"))
+    return classifyParallel(inS);
+  return classifySample(inS);
+}
+
+Indices MixtureClassifier::classifySample(const NumericalSample & inS) const
```
dbarbier: There is no need to redefine classify, just rename classifySample into classifySequential.

jschueller:
+1

lib/src/Base/Algo/openturns/ClassifierImplementation.hxx
```diff
@@ -51,12 +51,24 @@ public:
...
+public:
+
   /** Grade a point as if it were associated to a class */
   virtual NumericalScalar grade(const NumericalPoint & inP,
                                 const UnsignedInteger outC) const;
   virtual NumericalPoint grade(const NumericalSample & inS,
                                const Indices & outC) const;
 
+protected:
+  virtual NumericalPoint gradeTBB(const NumericalSample & inS,
```
dbarbier: I would prefer gradeParallel over gradeTBB, but this is not a big deal, +1 from me

sofianehaddad:
+1, thanks @regislebrun

PR #353, opened 2016-11-08
Integrate otlhs module

jschueller:


regislebrun:
+1

lib/src/Uncertainty/Algorithm/WeightedExperiments/OptimalLHSImplementation.cxx
```diff
@@ -0,0 +1,152 @@
...
+
+using namespace OT;
+
+namespace OT
+{
+
+CLASSNAMEINIT(OptimalLHSImplementation);
+
+static const Factory<OptimalLHSImplementation> Factory_OptimalLHSImplementation;
+
```
sofianehaddad: I thought it will inherits from WeightedExperiment with your changes?

jschueller: That was lhsdesign, which was removed.

sofianehaddad: yes or generateResult..., that is the idea

jschueller:
@regislebrun @sofianehaddad I changed the API so as to inherit WExperiment.

sofianehaddad:
hmmm in https://github.com/openturns/otlhs/pull/21, yes
It seems that here API did not change. Or am I wrong?

jschueller:
@sofianehaddad I pushed the wrong local copy when rebasing, everything is back in order

regislebrun:
+1

PR #354, opened 2016-11-10
Fixed a bug in KarhunenLoeveResultImplementation::liftAsField().

regislebrun:


jschueller:
Could it be tested somewhere?

PR #356, opened 2016-11-11
Fixed GeneralizedLinearModelAlgorithm::computeF()

regislebrun:
When basis_ was of size 0, the matrix F was not correctly initialized.


sofianehaddad:
+1

PR #357, opened 2016-11-11
Fixed StationaryCovarianceModel::discretize()

regislebrun:
When called using a RegularGrid, the diagonal blocks were not properly computed.


sofianehaddad:
+1

PR #349, opened 2016-10-27
Distribution::computeLogPDF gradient method and application in MaximumLikelihoodFactory

sofianehaddad:
This PR implements a generic Distribution::computeLogPDFGradient (defined as PDFGradient/PDF)
The objective is to use the method as an exact gradient when maximizing the log-likelihood function for estimating distribution's parameters in MaximumLikelihoodFactoryLogLikelihood. 

Indeed this class provides a buildParameter method which purpose is to maximize the log-likelihood function previously defined as bind method function (+ difference finite gradient). We rework the internal structure "MaximumLikelihoodFactoryLogLikelihood" as follows:
- It inherits now from NumericalMathEvaluationImplementation;
- It implements the operator() (reuse the computeLogLikelihood method)
- A gradient wrapper is added (maps Distribution::computeLogPDFGradient)
  Thus we get the function and is exact gradient for the evaluation.

Note that here the separation between NumericalMathEvaluationImplementation and NumericalMathGradientImplementation is required as some optimization algorithms require explicitely the gradientImplementation and not only the gradient operator.

Finally to fix/speed up some tests, we implement specific computeLogPDFGradient to TruncatedNormal and Fisher-Snedecor

PR #358, opened 2016-11-14
MaximumLikelihoodFactory enhancements

sofianehaddad:
This PR implements a generic Distribution::computeLogPDFGradient (defined as PDFGradient/PDF), which purpose is to help for estimating distributions (see #349)

Indeed, when estimating distribution's parameters thanks to the MaximumLikelihoodFactory class, the function to maximize (the log-likelihood) was
previously defined as bind method function (+ difference finite gradient). We rework the internal structure "MaximumLikelihoodFactoryLogLikelihood" as follows:

 - It inherits now from NumericalMathEvaluationImplementation;
  - It implements the operator() (reuse the previous computeLogLikelihood method)
  - A gradient wrapper is added (maps Distribution::computeLogPDFGradient)
    Thus we get the function and is exact gradient for the evaluation.

Note that here the separation between NumericalMathEvaluationImplementation and NumericalMathGradientImplementation is mandatory as some optimization algorithms require explicitely the gradientImplementation and not only the gradient operator.

Finally to fix/speed up some tests, we implement specific computeLogPDFGradient to Fisher-Snedecor, TruncatedNormal and Trapezoidal distributions

### DIFF
--- a/lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
+++ b/lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
@@ -142,6 +142,30 @@ NumericalScalar FisherSnedecor::computeLogPDF(const NumericalPoint & point) cons
   return normalizationFactor_ + (0.5 * d1_ - 1.0) * std::log(x) - 0.5 * (d1_ + d2_) * log1p(d1_ * x / d2_);
 }
 
+NumericalPoint FisherSnedecor::computeLogPDFGradient(const NumericalPoint & point) const
+{
+  const NumericalScalar x = point[0];
+  NumericalPoint logPdfGradient(2, 0.0);
+  if (!(x > 0.0)) return logPdfGradient;
+  const NumericalScalar d1xd2 = d1_ * x + d2_;
+  // First derivate the normlizationFactor as function of d1_, d2_ (see expression above in LogPDF)
+  // As the term is a combinations of LnBeta(d1/2, d2/2) := log(Beta(d1/2, d2/2), dLnBeta = dBeta/Beta
+  // As dBeta(x,y) = B(x,y) * (DiGamma(X) - DiGamma(x+y)) (see  https://en.wikipedia.org/wiki/Beta_function#Derivatives)
+  // it follows that d(LnBeta(x,y)) = dBeta(x,y) / Beta(x,y) = DiGamma(X) - DiGamma(x+y)
+  // Rest is very easy to derivate
+  logPdfGradient[0] = 0.5 * ( std::log(d1_ * x / d1xd2) + 1.0 - SpecFunc::DiGamma(0.5 * d1_) + SpecFunc::DiGamma(0.5 * d1_ + 0.5 * d2_) - (d1_ + d2_) * x / d1xd2);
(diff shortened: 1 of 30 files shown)


regislebrun:
+1

jschueller:
+1
"as some optimization algorithms require explicitely the gradientImplementation" : where exactly ?

regislebrun:
@jschueller Not exactly needed by any optimization solver but mandatory for the composition of function.

jschueller:
ah... yes

lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
```diff
@@ -142,6 +142,32 @@ NumericalScalar FisherSnedecor::computeLogPDF(const NumericalPoint & point) cons
...
+  const NumericalScalar d1xd2 = d1_ * x + d2_;
+  // First derivate the normlizationFactor as function of d1_, d2_ (see expression above in LogPDF)
+  // As the term is a combinations of LnBeta(d1/2, d2/2) := log(Beta(d1/2, d2/2), dLnBeta = dBeta/Beta
+  // As dBeta(x,y) = B(x,y) * (DiGamma(X) - DiGamma(x+y)) (see  https://en.wikipedia.org/wiki/Beta_function#Derivatives)
+  // it follows that d(LnBeta(x,y)) = dBeta(x,y) / Beta(x,y) = DiGamma(X) - DiGamma(x+y)
+  logPdfGradient[0] = 0.5 * ( std::log(d1_ / d2_)  + 1.0 - SpecFunc::DiGamma(0.5 * d1_) + SpecFunc::DiGamma(0.5 * d1_ + 0.5 * d2_) );
+  logPdfGradient[1] = 0.5 * (-d1_  / d2_ - SpecFunc::DiGamma(0.5 * d2_) + SpecFunc::DiGamma(0.5 * d1_ + 0.5 * d2_) );
+  // Second part is less complex
+  logPdfGradient[0] += 0.5 * (std::log(x) - log1p(d1_ * x / d2_) - (d1_ + d2_) * x / d1xd2);
+  logPdfGradient[1] += 0.5 * (-log1p(d1_ * x / d2_) + (d1_  + d2_) * (d1_ * x  / d2_) / d1xd2);
```
dbarbier: I do not understand why both [0] and [1] lines are splitted.  If you factor, you have for the first term
```
std::log(d1_ / d2_)  + std::log(x) - log1p(d1_ * x / d2_) = std::log(x * d1_ /(d2_ + x * d1_))
```

sofianehaddad: ok

lib/test/t_TrapezoidalFactory_std.cxx
```diff
@@ -31,6 +31,7 @@ int main(int argc, char *argv[])
   setRandomGenerator();
 
   ResourceMap::SetAsBool("MaximumLikelihoodFactory-Parallel", false);
+  TBB::Disable();
```
sofianehaddad: It seems that disabling TBB fixes numerical precision issue

dbarbier:
IMO the real problem is that TBB should never had been introduced in NSI where results depend on the order of operations, ie. compute{Mean,Covariance,Variance,KendallTau,Skewness,Kurtosis,CenteredMoment,RawMoment}.
Cc'ing @regislebrun

jschueller:
apriori I'd say results in other classes also depend on ordering

PR #359, opened 2016-11-14
Allow to serialize DistributionFactory

jschueller:


jschueller:
cc @aurelieladier, due in 1.9

lib/src/Uncertainty/Model/DistributionFactoryImplementation.cxx
```diff
@@ -18,29 +18,22 @@
...
-#include <cstdlib>
-#include <iomanip>
-#include <fstream>
+
 #include "openturns/DistributionFactoryImplementation.hxx"
-#include "openturns/Description.hxx"
-#include "openturns/Path.hxx"
-#include "openturns/Exception.hxx"
-#include "openturns/OTconfig.hxx"
-#include "openturns/Log.hxx"
```
dbarbier: Log.hxx and Exception.hxx are used in this file, they should not be removed

jschueller: ok for Exceptions but Log ?

sofianehaddad:
hmm why serializing factories?

jschueller:
We need to serialize an object that has factories as attributes.

sofianehaddad:
ok

PR #361, opened 2016-11-16
Really allow to serialize factories

jschueller:


regislebrun:
+1 @jschueller And the next one will be *really really* allow to serialize factories ;-)?

jschueller:
sorry

sofianehaddad:
+1

PR #370, opened 2016-11-19
Add CovarianceModel::parameterGradient

jschueller:


regislebrun:
-1 Why did you defined this method only for stationary models? I expected a signature taking two NPs as arguments.

jschueller:
sorry

regislebrun:
+1 Good job!

sofianehaddad:
+1

PR #372, opened 2016-11-21
Rewrite some NSI methods to not use TBB

dbarbier:
Rewrite computeMean, computeVariance and computeCovariance to not use TBB.  These methods are called by many internal code in OT, and the only solution to have fully reproducible results is to discard TBB.

With this commit, parallelism at a higher level is still possible, and will provide good performance with reproducible results.

There was also a subtle glitch in `LogLikelihoodEvaluation::operator()`, sequential code computes `logPdfSample(k, 0) / size` whereas parallel code computes `logPdfSample(k, 0) * (1.0 / size)`, and results do differ noticeably when an optimizer tries to find the best parameters.

### DIFF
--- a/lib/etc/openturns.conf.in
+++ b/lib/etc/openturns.conf.in
@@ -345,7 +345,6 @@
   <MaximumLikelihoodFactory-MaximumRelativeError    value="1.0e-10" />
   <MaximumLikelihoodFactory-MaximumObjectiveError   value="1.0e-10" />
   <MaximumLikelihoodFactory-MaximumConstraintError  value="1.0e-10" />
-  <MaximumLikelihoodFactory-Parallel                value="true" />
 
   <!-- OT::Student parameters -->
   <Student-MaximumNumberOfPoints          value="10000000" />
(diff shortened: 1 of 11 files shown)


jschueller:
I agree these should be memory-bound anyway.

regislebrun:
@dbarbier @jschueller 
I fully agree on the fact that NSI accessors should be more efficient, and the propose change is probably the most efficient way to access data, but I cannot agree on the fact that parallelism should be removed because it leads to non-reproducible results:
  * if no cancellation occur, any order should give essentially the same result (say up to the 14th digit) so parallelism is harmless
  * if cancellation do occur, then the problem at hand is badly conditioned and it is not a problem of parallelism: the canonical order of summation has no better properties than a random order obtained through parallelism. It could even be a good thing that the result differ, as it would point out this bad conditioning! Think about CADNA and the random rounding used to detect the actual significant results in a computation.
Anyway, it should be a user choice to activate parallelism or not, as soon as it leads to a significant speed improvement. Some benchmark is needed here!

dbarbier:
The hours we spend to debug precision issues is more important to me than those tenth of seconds, thus I won't lose my time anymore.

jschueller:
this looked fine, regis was just afraid we loose to much perf, could you show some benchmarks to reassure everyone ?

regislebrun:
@dbarbier @jschueller Ok guys, we all had hard time these last days (at least on my side), and we may be a little tense. So keep cool, being sent to Toulouse is not such a big deal after all ;-)!
If removing the TBBs is just a matter of tenth of second of course it is not an issue, and the PR is ok. But it was not mentioned in the description of the PR... So *please* reopen the PR and merge it. And *thanks* Denis for the good job!

jschueller:
@dbarbier, @regislebrun I propose to reopen this and accept it. In my opinion the time we loose debugging tests is not worth the performance of basic statistical results. If we really need parallel esimation (kendall) this should come in a dedicated class.

PR #374, opened 2016-11-23
Covariance model enhancements

jschueller:


regislebrun:
+1

jschueller:
@sofianehaddad the matrix returned is symmetric, so it does not affect other computations doesn't it ?

jschueller:
@sofianehaddad unfortunately I wont have time to go further for the gradients, would it be okay to include it as is, or should I thrash this ?

sofianehaddad:
oki for me

PR #375, opened 2016-11-24
Maximumlikelihoodfactory enhancements

jschueller:


jschueller:
@sofianehaddad, @regislebrun I'm not sure about d223794
also there are several places where we compute loglikelihoods and sometimes we return logminscalar, sometimes -maxscalar

python/test/t_MaximumLikelihoodFactory_std.py
```diff
@@ -15,3 +15,45 @@
 inf_distribution = factory.build(sample)
 print(inf_distribution)
 
+class TruncatedFactory(object):
+    def __init__(self, distribution, lb=None, ub=None):
```
regislebrun: @jchueller any hope to have it in the C++ library? eg as the TruncatedDistributionFactory class?

jschueller: @regislebrun after more testing it seems optimizer have trouble getting the bounds rights, so I forget TruncatedDistributionFactory for now.

sofianehaddad: @jschueller why did you remove it? The problem is that we should set an inequality constraint lb < ub + constrains if parameters should be in [lb, ub]

regislebrun:
+1 @jschueller @sofianehaddad we should return SpecFunc::LogMinNumericalScalar each time we compute the log of a zero PDF/likelihood, but I got invalid results in TrapezoidalFactory when I did it in MaximumLikelihoodFactory. Under time pressure, I prefered to leave the code unchanged just before the release.

python/test/t_MaximumLikelihoodFactory_std.py
jschueller: @sofianehaddad yes, but there's no generic way of knowing which parameters should be in bounds

PR #380, opened 2016-11-30
Rewrite some NSI methods to not use TBB

jschueller:


regislebrun:
+1 Nice to see this PR back ;-)!

dbarbier:
Maybe I misunderstood, I thought that benchmarks were required.

jschueller:
I think we're all convinced now.

PR #382, opened 2016-12-01
Indices::check(n) <=> seq[j] < n

jschueller:
It's less error-prone to change the definition of Indices::check
to seq[j] < n for all j instead of seq[j] <= n for all j


jschueller:
We cannot check for [0] anymore, closing :[

regislebrun:
+1

PR #385, opened 2016-12-07
Fix MaximumLikelihoodFactory use of check

jschueller:


jschueller:
#375 should have been rebased on #382

PR #388, opened 2016-12-13
Fix several bugs in LeastSquaresMethod

dbarbier:
These methods are bogus: 
   * CholeskyMethod::getGramInverseTrace
   * QRMethod::getGramInverse
   * SVDMethod::getGramInverse
   * SVDMethod::getGramInverseTrace

Add LeastSquaresMethod::getGramInverseDiag generic method and specific implementations.


lib/src/Base/Algo/openturns/LeastSquaresMethodImplementation.hxx
```diff
@@ -94,6 +94,9 @@ public:
   /** Get the diagonal of H */
   virtual NumericalPoint getHDiag() const;
 
+  /** Get the diagonal of the Gram inverse */
+  virtual NumericalPoint getGramInverseDiag() const;
```
jschueller: getGramInverseDiag is not documented (neither are getGramInverse(), getGramInverseTrace() getH() 	getHDiag() but that's my bad)

dbarbier: Arghhh, I always forgot.  I will document all these methods.

dbarbier:
BTW `update()` method must be called before calling these accessors, this is annoying.  Here are some options:
 1. Keep as is
 2. Call `update()` from these accessors and make them non-const 
 3. Call `update()` from these accessors and make fields `mutable`

jschueller:
I'd say option 3

dbarbier:
Docstrings added; I wanted to give more details in `{QR,SVD,Cholesky}Methods` but gave up, usage of DesignProxy makes explanations cumbersome.  These classes should deal with NumericalSample only, Basis are useless here.

regislebrun:
+1 @dbarbier @jschueller The actual needs on DesignProxy is to be able to add a new function and to keep the values which have already been computed. It is needed by the GreedyRankOne algorithm: you have learnt m rank-one tensors one by one and after the learning on each rank one tensor you have to solve a least-squares problem involving all the previous rank-one tensors and the newly created one. I had no other idea than to use a DesignProxy to keep the values of the previous rank-one tensors on the learning basis and to add the column corresponding to the new tensor. I will think about that and propose a solution to get rid of the basis and stay on a NumericalSample or a Matrix in a separate thread (by email).
In the mean time, I propose to merge this PR as it solves many issues! Great work!

dbarbier:
My main grief against DesignProxy is that it caches function evaluations, whereas it would be much better to avoid them.  For instance, I benchmarked CorrectedLeaveOneOut; on a large sample, CorrectedLeaveOneOut::run from start to method.solve call takes 1.3s, and evaluation of yHat=metamodel(x) afterwards takes 13s.  So if we compute yHat directly from design matrix,  CorrectedLeaveOneOut::run() method runs 10x faster (at least on this example).

jschueller:
@dbarbier see #392

PR #390, opened 2016-12-14
Fixed ticket #871 (generalized exponential parameter)

regislebrun:


lib/src/Base/Stat/GeneralizedExponential.cxx
```diff
@@ -164,6 +165,7 @@ NumericalScalar GeneralizedExponential::getP() const
 void GeneralizedExponential::setP(const NumericalScalar p)
 {
   if (p <= 0.0) throw InvalidArgumentException(HERE) << "Error: p must be positive.";
+  if (p > 2.0) throw InvalidArgumentException(HERE) << "Error: p must be less or equal to 2.";
```
dbarbier: please write instead if (!(p <= 2.0))

PR #391, opened 2016-12-15
Move NMF::GetValidFunctions to AnalyticalFunction

jschueller:


sofianehaddad:
Ok for mooving but I think we should first deprecate methods on the NMF side ?

jschueller:
@sofianehaddad ok

PR #392, opened 2016-12-15
Re-use design evaluations in CorrectedLeaveOneOut

jschueller:


dbarbier:
+1 for changes in CorrectedLeaveOneOut, -1 for ProxyDesign::add.  As mentioned in #388, I believe that ProxyDesign should not manage NMF.

dbarbier:
A similar change can be applied in KFold; I pushed a branch db/improve-LeastSquares to my repository with the commit I worked on yesterday to show how to do it.

jschueller:
ok. maybe it's more convenient if you push it then.

PR #397, opened 2017-01-02
Add missing rst entry for kFactor

jschueller:


regislebrun:
+1 Happy new year folks!

PR #398, opened 2017-01-03
Actually test DistFunc.kFactorPooled

jschueller:
kFactorPooled was not used, fixed it to match cxx results
Also fix a formatting issue with Python 3.5


regislebrun:
+1 Sorry ;-)

PR #401, opened 2017-01-06
Function api changes

jschueller:
Mainly enhances DatabaseFunction with a cache flag (as in the eval impl) and removes the specific Python parametric accessor (follow-up to http://trac.openturns.org/ticket/873)


regislebrun:
+1 but the ability to use parameters in a pure Python context was appreciated by some of our users from EDF. Not sure we can drop this capability silently...

PR #403, opened 2017-01-09
Fix #875 in trac

sofianehaddad:
Fix http://trac.openturns.org/ticket/875

TruncatedNormalFactory randomly crashes on osx systems. This is due to
the new implementation in the MaximumLikelihoodFactory class (reduced
objective function but not accurate bounds). Thus to fix the problem:

 - We update OptimizationProblem's bounds :  they should be zero size or
 should have the same size as objective's function input dimension
 Thus we redefine new bounds in case
 - We update also Equality/Inequality constraints :  we raise an exception
 if their input dimension size is not compatible with objective function's
 input dimension

Same work has been done on the MethodOfMoment class as we should get the
same problems if we face reduced objective functions

### DIFF
--- a/lib/src/Base/Optim/OptimizationProblem.cxx
+++ b/lib/src/Base/Optim/OptimizationProblem.cxx
@@ -180,6 +180,11 @@ Bool OptimizationProblem::isMinimization() const
   return getImplementation()->isMinimization();
 }
 
+Bool OptimizationProblem::isValid() const
+{
+  return getImplementation()->isValid();
+}
+
 /* String converter */
 String OptimizationProblem::__repr__() const
 {
--- a/lib/src/Base/Optim/OptimizationProblemImplementation.cxx
+++ b/lib/src/Base/Optim/OptimizationProblemImplementation.cxx
@@ -232,6 +232,45 @@ Bool OptimizationProblemImplementation::isMinimization() const
   return minimization_;
 }
 
+Bool OptimizationProblemImplementation::isValid() const
+{
+  // isValid checks if the optimization problem is 'valid' (all functions have accurate dimension) :
+  // 1) Accurate bounds dimension if defined
+  // 2) Accuracy between equality constraint and objective function input dimensions
+  // 3) Accuracy between inequality constraint and objective function input dimensions
+  // 4) Accuracy between level function and objective function input dimensions
+  if (hasBounds() && (getDimension() != getBounds().getDimension()))
+  {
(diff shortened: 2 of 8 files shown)


jschueller:
The dimension checks should rather go in OptimizationSolverImplementation::setProblem.

sofianehaddad:
@jschueller for inequality constraints ok
In the case of objective function, there is a check + definition of new bounds as we suppose that initial problem have been reduced

So I could add all checks in OptimizationSolverImplementation::setProblem and keep the definition of new bounds if required. What do ou think ?

jschueller:
ok

sofianehaddad:
@jschueller finally I did differently. The reason is that we should not check the validity before setting all the various elements (objective, bounds, constraints...)

jschueller:
yes, when setProblem is reached isn't the problem complete ?

sofianehaddad:
@jschueller no. Have a look a kriging. Objective function is defined in the run method but setProblem is reached at the init step

jschueller:
then we'd rather check the validity in each solver's run

sofianehaddad:
yes

regislebrun:
@sofianehaddad @jschueller Here is my understanding:
+ As we decided to allow for the creation of a "default" OptimizationProblem to which we add an objective function and eventually some constraints, it is complex to check the validity of the problem on the fly. One way to do that in a robust way is:
  * The input dimension of the objective function defines the problem dimension
  * If the objective function is changed, all the constraints (equality, inequality, bounds) with incompatible dimensions are invalidated with a warning. It must be eiher none of the constraint or all the constraints.
  * If a constraint is added or modified, its input dimension is checked. If it is not compatible it throws an exception
  This way, the optimization problem is always valid as it has been built valid and any change leads to a valid problem. No need to check its validity in the OptimizationProblem class, only its compatibility with the solver capabilities needs to be checked.
+ Another way is to allow for the creation of invalid problems, eg using the default constructor of OptimizationProblem then the manipulation with no check of its elements (objective function, constraints and so on). Then, the user can check if the problem is valid using a specific method of OptimizationProblem, eg an isValid() method
+ In the setProblem() method of the OptimizationAlgorithm class, the first thing to do is to check the problem validity using its isValid() method, then to check the problem against the solver capabilities.

My preference is to adopt the first approach, as one gets only valid optimisation problems.

jschueller:
As sofiane said sometimes the objective is not known in advance but we'd like to set constraints/bounds, so go for option2, but isvalid has to be checked in run.

jschueller:
superceded by https://github.com/openturns/openturns/pull/404

PR #404, opened 2017-01-09
Keep OptimizationProblem valid

jschueller:
Bounds and constraints have always the dimension of the objective.
They are cleared when the objective is set with a different input dimension.
So, we cannot use an OptimizationProblem to pass a partial problem to (eg GLM), so add accessors to set bounds etc.
Fixes http://trac.openturns.org/ticket/875

### DIFF
--- a/lib/src/Base/Optim/OptimizationProblemImplementation.cxx
+++ b/lib/src/Base/Optim/OptimizationProblemImplementation.cxx
@@ -47,6 +47,15 @@ OptimizationProblemImplementation::OptimizationProblemImplementation()
   // Nothing to do
 }
 
+OptimizationProblemImplementation::OptimizationProblemImplementation(const NumericalMathFunction & objective)
+  : PersistentObject()
+  , objective_(objective)
+  , minimization_(true)
+  , dimension_(objective.getInputDimension())
+{
+  // nothing to do
+}
+
 /*
  * @brief General multi-objective equality, inequality and bound constraints
  */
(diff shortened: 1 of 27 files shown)


lib/src/Base/Optim/OptimizationProblemImplementation.cxx
```diff
@@ -97,6 +93,12 @@ NumericalMathFunction OptimizationProblemImplementation::getObjective() const
 
 void OptimizationProblemImplementation::setObjective(const NumericalMathFunction & objective)
 {
+  // clear constraint, bounds
+  LOGWARN(OSS() << "Clearing constraints and bounds");
+  equalityConstraint_ = NumericalMathFunction();
+  inequalityConstraint_ = NumericalMathFunction();
+  bounds_ = Interval(0);
+
   clearLevelFunction();
```
regislebrun: The constraints should be removed only if the new objective function has an incompatible dimension. It allows eg to switch to a finer discretized robust objective function while keeping the constraints as is.

lib/src/Uncertainty/Distribution/MeixnerDistribution.cxx
```diff
@@ -422,19 +422,20 @@ void MeixnerDistribution::update()
...
+  problem.setBounds(getRange());
+  solver_.setStartingPoint(getMean());
   solver_.setProblem(problem);
   solver_.run();
   b_ = std::sqrt(solver_.getResult().getOptimalValue()[0]);
 
   // Define Optimization problem2 : minimization fCD
   problem.setMinimization(true);
   problem.setObjective(fCD);
+  problem.setBounds(getRange());
```
regislebrun: No need to set the bounds another time if the setObjective() method is modified according to my proposal

jschueller:
@regislebrun I implemented your remark

sofianehaddad:
This new behaviour (first objective function then other elements) has to be documented

jschueller:
@sofianehaddad, ok, see OptimizationProblemImplementation_doc.i.in

PR #405, opened 2017-01-10
Check DistFactory::buildEstimator parametrization

jschueller:


lib/src/Uncertainty/Model/DistributionFactoryImplementation.cxx
```diff
@@ -93,6 +93,10 @@ DistributionFactoryResult DistributionFactoryImplementation::buildEstimator(cons
 {
   DistributionFactoryResult nativeResult(buildEstimator(sample));
   Distribution nativeDistribution(nativeResult.getDistribution());
+  Distribution parametersDistribution(parameters.getDistribution());
```
regislebrun: It may be more efficient to test the compatibility between the factory and the parameter set before to build the estimator, as it can be time consuming (bootstrap). A way to do this is to rely on the build() method of the factory, which call the default constructor of the parametric model and which is supposed to be fast. Something like:
String referenceDistributionName(build()->getClassName());
String parametersDistributionName(parameters.getDistribution().getImplementation()->getClassName());
then the test and the actual estimation if the test is ok.

jschueller: ok

regislebrun:
+1

PR #406, opened 2017-01-11
Compile Python bindings with appveyor

jschueller:
Various fixes to allow:
- Compiling the bindings
- Support Visual Studio 14 2015

Side note: when compiling with vs14 yields lots of these:
```Exception.hxx(92): warning C4275: non dll-interface class 'std::exception' used as base for dll-interface class 'OT::Exception'```
apparently the solution would be inlining exceptions instead of exporting them, what do you think ?

### CODE REVIEW

**Comment by regislebrun** [comment]
+1 @jschueller No idea on what should be done regarding exceptions...

File: lib/src/Base/Stat/openturns/CovarianceMatrix.hxx [hunk 1]
```diff
@@ -67,8 +67,13 @@ public:
   using SymmetricMatrix::operator +;
 
   /** CovarianceMatrix multiplication (must have consistent dimensions) */
+#ifdef _MSC_VER // VS2010 does not like 'using' being called after overloads
+  using SymmetricMatrix::operator *;
+#endif
   CovarianceMatrix operator * (const IdentityMatrix & m) const;
+#ifndef _MSC_VER // VS2010 does not like 'using' being called after overloads
   using SymmetricMatrix::operator *;
+#endif
```
dbarbier: You can switch the 2 lines unconditionnally:
```
  using SymmetricMatrix::operator *; 
  CovarianceMatrix operator * (const IdentityMatrix & m) const; 
```

BTW it is strange that there is no problem with `operator+` just above.

jschueller: ok, only in CovarianceMatrix.hxx ?

dbarbier:
IMO we can silence this warning for now

lib/src/Base/Stat/openturns/CovarianceMatrix.hxx
jschueller: No, that fails in travis/linux.

PR #407, opened 2017-01-16
Inline exceptions

jschueller:
- Fixes the exception symbols visibility with vs14+:
```Exception.hxx(92): warning C4275: non dll-interface class 'std::exception' used as base for dll-interface class 'OT::Exception'```
- I suspect usage of CMAKE_lang_IMPLICIT_LINK_LIBRARIES libs to be useless


jschueller:
this does not seem to work on squeeze:
http://autobuilder.openturns.org/openturns/pull/407/squeeze-x86_64-ot-linux/distcheck-launcher.log

PR #410, opened 2017-01-25
Viewer fixes

jschueller:
- A fix for svg file bytes conversion to unicode string for html when rendering a graph inside notebook
- View.show to show only current graph, not all queued plots (not available in matplotlib < 1.3.1)
side effect: the block keyword is removed -> use pyplot api for blocking show all in non-interactive mode

### DIFF
--- a/python/doc/sphinxext/execforresourcemap_directive.py
+++ b/python/doc/sphinxext/execforresourcemap_directive.py
@@ -3,11 +3,7 @@ from os.path import basename
 import openturns as ot
 
 
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
-
+from io import StringIO
 from sphinx.util.compat import Directive
 from docutils import nodes, statemachine
 
--- a/python/src/__init__.py
+++ b/python/src/__init__.py
@@ -89,7 +89,7 @@ from .uncertainty import *
 from .coupling_tools import *
 
 
-def Show(graph, block=False):
+def Show(graph):
     """
     Display a graph pop-up.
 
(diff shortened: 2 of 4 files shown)


sofianehaddad:
@jschueller seems ok for both py2 & py3

sofianehaddad:
@jschueller I think you should also fix View.show

diff --git a/python/src/viewer.py b/python/src/viewer.py
index e8d3fb35..2000a71e 100644
--- a/python/src/viewer.py
+++ b/python/src/viewer.py
@@ -513,7 +513,7 @@ class View(object):
 
             These parameters are passed to matplotlib.pyplot.show()
         """
-       plt.show(**kwargs)
+      self._fig.show(**kwargs)

jschueller:
do you have a testcase ?

sofianehaddad:
Try this : 

> import openturns as ot
> import matplotlib.pylab as plt
> x = ot.Uniform(0,1).getSample(10000)
> plt.plot(x)
> ot.Show(ot.Uniform(0, 1).drawPDF())

Instead of getting PDF plot, you will have **all** figures already defined.

sofianehaddad:
@jschueller have a quick look at https://gist.github.com/sofianehaddad/cb54d2373871cd333f1805cbf005a485

jschueller:
got it

sofianehaddad:
@jschueller close seems ok

Travis fails. Matplotlib is too old (1.1.1 in gcc config) and fig.show has been introduced after 1.1.1 (see https://github.com/ipython/ipython/pull/1615)

python/test/t_Viewer should be also updated (`block` argument is not defined for `fig.show` : `block` is disabled)

regislebrun:
+1

PR #413, opened 2017-01-31
MetaModel enhancements

jschueller:
- Add MetaModelAlgorithm::buildTransformation (factorize code for tensor)
- Check eigen values in SVD (possible numerical issue with specific doe)

### DIFF
--- a/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosAlgorithm.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosAlgorithm.cxx
@@ -27,13 +27,7 @@
 #include "openturns/NumericalPoint.hxx"
 #include "openturns/NumericalMathFunctionImplementation.hxx"
 #include "openturns/ComposedFunction.hxx"
-#include "openturns/IdentityFunction.hxx"
 #include "openturns/DatabaseFunction.hxx"
-#include "openturns/RosenblattEvaluation.hxx"
-#include "openturns/InverseRosenblattEvaluation.hxx"
-#include "openturns/MarginalTransformationEvaluation.hxx"
-#include "openturns/MarginalTransformationGradient.hxx"
-#include "openturns/MarginalTransformationHessian.hxx"
 #include "openturns/FixedStrategy.hxx"
 #include "openturns/CleaningStrategy.hxx"
 #include "openturns/FixedExperiment.hxx"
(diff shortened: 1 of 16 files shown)


jschueller:
in Uncertainty/Algorithm/Transformation/MetaModelTransformation.cxx ?:
```
MetaModelTransformation::MetaModelTransformation(const Distribution & measure)
MetaModelTransformation::build(NMF & transformation, NMF & inverseTransformation)
```

regislebrun:
@jschueller I would remove the reference to the meta-models. Ideally, this class should be the only entry point for all the distribution transformations, hiding the other distribution-transformation related classes (Nataf, Rosenblatt, marginal transformation etc), in the spirit of the StandardDistributionPolynomialFactory for the orthogonal polynomials.
So DistributionTransformation should be a reasonable name, and it should be a NMF. It should have an inverse() method, which should return a DistributionTransformation with swapped arguments. Based on this class, the getIsoProbabilisticTransformation() and getInverseIsoProbabilisticTransformation() could be implemented in a generic way.

jschueller:
@regislebrun, I created the DistributionTransformation class, but I dont really know where to use it except in FChaosAlgo.

regislebrun:
@jschueller It could be used in DistributionImplementation::getIsoProbabilisticTransformation() & co for example!

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosAlgorithm.cxx
```diff
@@ -287,105 +289,13 @@ void FunctionalChaosAlgorithm::run()
...
-          LOGINFO("Non-normal standard space for basis");
-          TZ = NumericalMathFunction(NumericalMathFunctionImplementation(RosenblattEvaluation(measure.getImplementation()).clone()));
-          invTZ = NumericalMathFunction(NumericalMathFunctionImplementation(InverseRosenblattEvaluation(measure.getImplementation()).clone()));
-        }
-        transformation_ = ComposedFunction(invTZ, TX);
-        inverseTransformation_ = ComposedFunction(invTX, TZ);
-      }
-    } // Non-independent input copula
-  }
+  DistributionTransformation transformation(distribution_, measure);
```
regislebrun: const DistributionTransformation?

lib/src/Uncertainty/Algorithm/Transformation/DistributionTransformation.cxx
```diff
@@ -0,0 +1,167 @@
...
+
+typedef Collection<Distribution> DistributionCollection;
+
+CLASSNAMEINIT(DistributionTransformation);
+
+/* Default constructor */
+DistributionTransformation::DistributionTransformation ()
+  : NumericalMathFunction()
+{
+}
```
regislebrun: // Nothing to do

jschueller:
@regislebrun the thing is DistributionTransformation already calls DistributionImplementation::getIsoProbabilisticTransformation : this can result in segfaults.

regislebrun:
@jschueller Yes, but it is because some code is still in some distributions instead of being in this new class. I think the work has to be done in two steps:
+ first you merge the new class
+ second we (ie not necessarily you) move the relevant code into this class

jschueller:
ok

PR #414, opened 2017-01-31
Embed notebook in sphinx doc

jschueller:
This allows to embed notebooks in the sphinx doc using nbsphinx (http://nbsphinx.rtfd.io/).
I chose to not render notebooks when building the doc, they are tested with t_notebook.py.
For discussion, more on this on february 9.


jschueller:
@dbarbier I restored the travis script

sofianehaddad:
@jschueller maybe fa084e5 should be in a separate PR?

jschueller:
@sofianehaddad if you want

sofianehaddad:
@jschueller could we do the same thing for t_notebook as for t_docstring? I mean split the tests
Examples will be added gradually so in order to keep time < timeout and for easy debugging it could be better. What do you think?

jschueller:
Yes, absolutely, maybe later when we have more notebooks and then we could decide on how to split them.

sofianehaddad:
ok

regislebrun:
+1

PR #419, opened 2017-02-03
Add about section

jschueller:


regislebrun:
@jschueller What kind of help is wanted?

jschueller:
What do you think could be added in that section ? Is the history section ok to you ? Do you want to add more informations on authors ? Maybe some of your publications ?

sofianehaddad:
@jschueller : 
 - openturns user days
 - How to cite OT

jschueller:
@sofianehaddad ok, added

sofianehaddad:
I think we could merge it and update it progressively

PR #420, opened 2017-02-10
Minor cosmetics

sofianehaddad:
The PR provides minor changes : 
 - Fix potential circular/recursive import in viewer
 - Provide a SVG export to drawables

### DIFF
--- a/python/src/viewer.py
+++ b/python/src/viewer.py
@@ -514,6 +514,13 @@ class View(object):
             # mpl < 1.3.1, see https://github.com/ipython/ipython/pull/1615
             plt.show(**kwargs)
 
+    @staticmethod
+    def ShowAll(**kwargs):
+        """
+        Display all graphs.
+        """
+        plt.show(**kwargs)
+
     def save(self, fname, **kwargs):
         """
         Save the graph as file.
--- a/python/src/viewer.py
+++ b/python/src/viewer.py
@@ -23,6 +23,7 @@ import warnings
 import io
 import sys
 
+__all__ = ["View", "PlotDesign"]
 
 class View(object):
     """
--- a/python/src/DrawableImplementation.i
+++ b/python/src/DrawableImplementation.i
@@ -7,4 +7,21 @@
 %include DrawableImplementation_doc.i
 
 %include openturns/DrawableImplementation.hxx
+
+%pythoncode %{
+def DrawableImplementation__repr_svg_(self):
+    """Get the SVG representation."""
+    from .viewer import ToSVGString
+    return ToSVGString(self)
+
+DrawableImplementation._repr_svg_ = DrawableImplementation__repr_svg_
+
+
+def DrawableImplementation__repr_html_(self):
+    """Get the HTML representation."""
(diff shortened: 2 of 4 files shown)


sofianehaddad:
Ah exact! 
No idea, except a static View.Show
Any idea?

jschueller:
Yeah, I was thinking the same thing. Maybe 'ShowAll' to differenciate it more ?

sofianehaddad:
ok I will add it

python/src/viewer.py
```diff
@@ -514,6 +514,14 @@ def show(self, **kwargs):
             # mpl < 1.3.1, see https://github.com/ipython/ipython/pull/1615
             plt.show(**kwargs)
 
+    @staticmethod
+    def ShowAll(**kwargs):
+        """
+        Display all graphs.
+
```
jschueller: no new line here (pep257)

sofianehaddad: fixed

sofianehaddad: Done

jschueller:
There's a Show function in __init__.py that also assumes a blocking behavior.
But I suggest we remove it ; its undocumented anyways.

sofianehaddad:
@jschueller no the Show function does not assumes a blocking behaviour
It depends on matplotlib version + used backend

jschueller:
if you use the default backend on an oridnary debian it's blocking.

sofianehaddad:
@jschueller I use default backend with an ordinary debian ==> not blocking

Btw the Show method is very usefull; I think we should not remove it

sofianehaddad:
@jschueller Show function in **__init__** is documented

    Display a graph pop-up.

    Parameters
    ----------
    graph : Graph
        the graph to display

jschueller:
I don't think there's an rst entry

sofianehaddad:
yep

regislebrun:
@sofianehaddad I agree with you: the Show() command is REALLY usefull. It is a one-line version of the 3 to 4 lines viewer + the specific import. I use it mainly for debuging purpose, for production graphics Viewer is much more adapted. I confirm that the Show() command was blocking in the 1.8 version. It is no more the case with the current trunk, and it is a regression from my point of view (even if it solved other issues).

otherwise +1

jschueller:
@sofianehaddad it's blocking for me too, matplotlib.get_backend() return TkAgg, what's yours ? and @regislebrun ?

sofianehaddad:
@jschueller same backend but not blocking

jschueller:
Anyways, it uses view.show which used plt.show(), to get the same behavior it should use View.ShowAll now.

sofianehaddad:
@jschueller did not understand
you want Show using View.ShowAll instead of View.show?

jschueller:
yes, to result in a call to plt.show as before

sofianehaddad:
Done

jschueller:
Thanks
Could you rebase to master ? I tried to fix the circle-ci script from a forked repo without passing secret env vars.

PR #422, opened 2017-02-13
Use CircleCI as primary provider for Linux builds

jschueller:
We keep Travis but disable documentation upload.


regislebrun:
+1

PR #424, opened 2017-02-15
Keep output sample out of LeastSquaresMethod

jschueller:
Output sample is contained in LeastSquaresMethod for convenience.
This has an impact over the BasisSequenceFactory & FittingAlgorithm APIs.
The result is we are able to use these prior to instanciating the output sample.


regislebrun:
+1

PR #428, opened 2017-02-22
Remove the centered WhiteNoise constraint

jschueller:
Only require it in the context of an ARMA process.


regislebrun:
+1 Good!

PR #430, opened 2017-02-23
Add EGO stochastic optimization algorithm

jschueller:
New class EfficientGlobalOptimization


lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+  : OptimizationAlgorithmImplementation()
+  , solver_(new Cobyla)
+  , useDefaultSolver_(true)
+  , multiStartExperimentSize_(ResourceMap::GetAsUnsignedInteger("EfficientGlobalOptimization-DefaultMultiStartExperimentSize"))
+  , multiStartNumber_(ResourceMap::GetAsUnsignedInteger("EfficientGlobalOptimization-DefaultMultiStartNumber"))
+  , parameterEstimationPeriod_(ResourceMap::GetAsUnsignedInteger("EfficientGlobalOptimization-DefaultParameterEstimationPeriod"))
+  , improvementFactor_(1.0)
+  , correlationLengthFactor_(1.0)
+  , aeiTradeoff_(1.0)
+{
```
sofianehaddad: These parameters should be set in ResourceMap

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+  {
+    const NumericalScalar mx = metaModelResult_.getMetaModel()(x)[0];
+    const NumericalScalar fmMk = optimalValue_ - mx;
+    const NumericalScalar sk2 = metaModelResult_.getConditionalCovariance(x)(0, 0);
+    const NumericalScalar sk = sqrt(sk2);
+    if (!SpecFunc::IsNormal(sk)) return NumericalPoint(1, -SpecFunc::MaxNumericalScalar);
+    const NumericalScalar ratio = fmMk / sk;
+    NumericalScalar ei = fmMk * normal_.computeCDF(ratio) + sk * normal_.computePDF(ratio);
+    if (noiseModel_.getOutputDimension() == 1) // if provided
+    {
```
sofianehaddad: you are encapsulating both EI and AEI, either for run method :-/
I think it is not a good idea. Maybe two different classes are required as you are implementing two different methods!

jschueller: Only the handling of noise would differ between the two classes, the core of the algorithm is the same. I chose to expose only one class.

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
jschueller: ok

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
sofianehaddad: ok thus you should have a global hasNoise_ argument (that rely on given model for the noise), two implementations of EfficientGlobalOptimization::run, two NumericalMathFunction classes (ExpectedImprovementEvaluation and AugmentedExpectedImprovement) to make the code easy readable and easy to debug and avoid internal checking at each iteration of the algo

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      for (UnsignedInteger i2 = 0; i2 < i1; ++ i2)
+      {
+        for (UnsignedInteger j = 0; j < dimension; ++ j)
+        {
+          NumericalScalar distance = std::abs(inputSample[i1][j] - inputSample[i2][j]);
+          if (distance < minimumDistance[j])
+          {
+            minimumDistance[j] = distance;
+          }
```
regislebrun: The cost is dn(n-1)/2. If n is large, it can be time-consuming. If you sort each marginal sample (cost: d*n*log(n)) then you can find the minimum distance per component in an additional d*n step, for a total cost dominated by the first step.

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+        }
+      }
+    }
+
+    NumericalMathFunction improvementObjective(new ExpectedImprovementEvaluation(optimalValueSubstitute, metaModelResult, noiseModel_));
+
+    if (useDefaultSolver_ && problem.hasBounds())
+    {
+      // Sample uniformly into the bounds
+      Interval bounds(problem.getBounds());
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      NumericalPoint lowerBound(bounds.getLowerBound());
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      NumericalPoint upperBound(bounds.getUpperBound());
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      Interval::BoolCollection finiteLowerBound(bounds.getFiniteLowerBound());
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      Interval::BoolCollection finiteUpperBound(bounds.getFiniteUpperBound());
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      ComposedDistribution::DistributionCollection coll;
+      for (UnsignedInteger i = 0; i < dimension; ++ i)
+      {
+        if (!finiteLowerBound[i] || !finiteUpperBound[i])
+          throw InvalidArgumentException(HERE) << "Bounds must be finite";
+        coll.add(Uniform(lowerBound[i], upperBound[i]));
+      }
+      ComposedDistribution distribution(coll);
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      NumericalSample improvementExperiment(distribution.getSample(multiStartExperimentSize_));
+      // retain best P/N points as starting points
+      improvementExperiment.stack(improvementObjective(improvementExperiment));
+      Indices inputs(dimension);
+      inputs.fill();
+      NumericalSample sortedImprovement(improvementExperiment.sortAccordingToAComponent(dimension).getMarginal(inputs));
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      // handle multiStartExperimentSize_ < multiStartNumber_
+      UnsignedInteger pointNumber = std::min(multiStartNumber_, multiStartExperimentSize_);
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+      NumericalSample startingPoints(sortedImprovement, multiStartExperimentSize_ - pointNumber, multiStartExperimentSize_);
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -0,0 +1,542 @@
...
+    // build problem
+    OptimizationProblem maximizeImprovement;
+    maximizeImprovement.setObjective(improvementObjective);
+    maximizeImprovement.setMinimization(false);
+    if (problem.hasBounds())
+      maximizeImprovement.setBounds(problem.getBounds());
+    solver_.setProblem(maximizeImprovement);
+    solver_.setStartingPoint(optimizer);
+    solver_.run();
+    OptimizationResult improvementResult(solver_.getResult());
```
regislebrun: const and in many other places in the remaining of the code

regislebrun:
+1 excepted for minor remarks

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
jschueller: the cost should be in the kriging step

regislebrun: ok. I will probably introduce the findClosestPair() and findClosestPairsPerComponent() methods in NumericalSample to encapsulate these computations as they could be reused eg to maximize the minimum distance over n points in a given box.

PR #431, opened 2017-02-23
Rename some classes

jschueller:


regislebrun:
+1

sofianehaddad:
@jschueller what about [gs]etOptimizationSolver ? I think it is mandatory to keep these methods deprecated until 1.10

dbarbier:
I agree with @sofianehaddad, [gs]etOptimizationSolver should be deprecated and replaced by [gs]etOptimizationAlgorithm.

PR #438, opened 2017-03-15
Initiatize p_evaluation in FDGradient to fix segfault

jschueller:
cc @aurelieladier


dbarbier:
Which segfault?

jschueller:
this one:
```print(ot.CenteredFiniteDifferenceGradient())```

**Comment by sofianehaddad** [comment]
@jschueller not exactly! repr(ot.CenteredFiniteDifferenceGradient()) segfaults.
It is (due to the internal pointer which is not instancied in that case)

BTW, the empty constructor should not be (is never) used. 

+1

**Comment by dbarbier** [comment]
I disagree with this fix, the problem is in FiniteDifferenceGradient.

Add
```
    , p_evaluation_(new NoNumericalMathEvaluationImplementation())
```
in default constructor without argument.

BTW CenteredFiniteDifferenceGradient should define save/load methods to serialize epsilon.

jschueller:
@dbarbier ok

jschueller:
@dbarbier the epsilon attribute is already in FiniteDifferenceGradient through FiniteDifferenceStep

regislebrun:
+1

PR #439, opened 2017-03-15
Try to bisect swig/osx failure

jschueller:
maybe swig is the source of the osx failures in the last days
https://github.com/swig/swig/commits/master


jschueller:
@sofianehaddad  maybe brew ?

PR #442, opened 2017-03-17
Improve p1 lagrange

regislebrun:
This PR provides a more efficient implementation of the evaluation operator of P1LagrangeEvaluationImplementation. It also reverted the use of const references as return values of some methods in Field and Mesh, fixing random segfaults.


jschueller:
could you rebase to master to take advantage of sofiane's osx fix ?

PR #444, opened 2017-03-17
Add optimization callbacks

jschueller:
A possible use-case is to control FORM optimization by the number of calls (http://trac.openturns.org/ticket/351)

This is not implemented by all solvers, the behavior of each solver is documented if it doesnt implement one of the callbacks.


regislebrun:
+1

PR #445, opened 2017-03-17
Fix osx failure on travis

sofianehaddad:
The PR fixes the recent `OS X` failures on the travis-ci platform

Problem is due to the used `tbb` imported thanks to Homebrew. For now, `tbb` sources are compiled

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_install:
   - if test "$TRAVIS_OS_NAME" = "osx"; then
         brew update &&
         brew tap homebrew/science &&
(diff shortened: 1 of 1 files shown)


sofianehaddad:
@jschueller Compilation is done for `osx` only
 I think `tbb` compilation is quite fast ? Total time allowed for osx is 48 mn. For the compilation we spent here about 36 mn

jschueller:
maybe it's the update to 207u5:
https://github.com/Homebrew/homebrew-core/pull/10569

have you tried to pin tbb to the previous version ?

PR #449, opened 2017-03-23
Fix ticket 883  (VisualTest.DrawHistogram should rely on Histogram.drawPDF)

regislebrun:
* Improved HistogramFactory: now, the user can specify either the bin width or the bin number, and there is a specific method to compute Silverman's bin width.

* Based VisualTest::DrawHistogram on Histogram::drawPDF() as built by HistogramFactory.

The documentation of HistogramFactory is missing in master. It will be given in a separate PR.


jschueller:
This needs a rebase.

PR #453, opened 2017-03-27
More kl algorithms

jschueller:
Various corrections over lebrun's #452 as he called in sick : (


jschueller:
looks like he's back!

PR #455, opened 2017-03-28
Added draw() to CovarianceModel

regislebrun:


lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -704,6 +706,89 @@ CovarianceModelImplementation::Implementation CovarianceModelImplementation::get
...
+  if (asStationary && isStationary())
+    {
+      // Here we compute the normalization for the correlation instead of using
+      // the amplitude attribute for models for which it is not given
+      NumericalScalar ratio = 1.0;
+      if (correlationFlag)
+	{
+	  ratio = (*this)(0.0)(rowIndex, columnIndex);
+	  // If ratio == 0, the covariance iz zero everywhere
+	  if (ratio == 0.0) ratio == 1.0;
```
jschueller: ratio = 1.0 !!!

jschueller: iz -> is

regislebrun: ratio is the enominator of the correlation. It is zero if and only if the numerator is also zero for the off-diagonal terms, so setting it to 1.0 is a useful trick to avoid a test in the normalization loop.

jschueller: yes, but you used the comparison operator

regislebrun: :-0!

jschueller: :-D

jschueller:
Also the test fails on the autobuilder when comparing graph values.

PR #457, opened 2017-03-28
Added a specialization of Field::draw() for 2D fields over 2D domains.

regislebrun:
Also fixed a bug in Field::drawMarginal() when interpolate == true


jschueller:
test still fail when printing graph values

regislebrun:
@jschueller hopefully this time it is fixed...

jschueller:
The python test still fails, maybe avoid printing graphs ?

regislebrun:
Done

regislebrun:
@jschueller found another bug, commit in a few minutes

regislebrun:
pyinstallcheck_Viewer failed, is it solved by PR#458?

jschueller:
maybe unrelated, #458 only fixes a segfault on osx

regislebrun:
ok. I removed all the output of the tests as muparser gives slightly different results on the different machines, resulting in different rounding and a different number of segments in the output.

PR #458, opened 2017-03-29
 	Use viewer in Mesh docstrings

jschueller:
This works around a segfault in docstring_geom test on osx, cc @sofianehaddad


jschueller:
Is this ok to go in @sofianehaddad @regislebrun @dbarbier

regislebrun:
+1

sofianehaddad:
This fix the random segfault on osx?

jschueller:
@sofianehaddad It avoids it, drawing R graphs in docstrings resulted in segfault.

sofianehaddad:
ok for the fix, thanks. +1

I should investigate later the real problem.

PR #459, opened 2017-03-31
Fixed CovarianceModel::draw()

regislebrun:
The normalization was wrong when (!asStationary && correlationFlag)


sofianehaddad:
:-) exact

PR #461, opened 2017-04-02
Added the stop callback to Cobyla.

regislebrun:


regislebrun:
@jschueller I don't understand your comment. Cobyla already implements the ability to stop the computation by checking the return code of the function evaluation, and any non-zero value stops the algorithm. The interesting lines are L307-314 in algocobyla.c:
  if (calcfc(*n, *m, &x[1], &f, &con[1], state))
    {
      if (*iprint >= 1) {
        fprintf(stderr, "cobyla: user requested end of minimization.\n");
      }
      rc = 3;
      goto L600;
    }

Do I miss something here?

PR #464, opened 2017-04-06
Add WhittleFactory::buildWithCriteria

jschueller:
- Allows to disambiguate whether to return information criteria
- Return it in a tuple instead of by reference
- Removed useless ARMAFactoryImplementation::build with reference

### DIFF
--- a/lib/src/Uncertainty/Process/ARMAFactoryImplementation.cxx
+++ b/lib/src/Uncertainty/Process/ARMAFactoryImplementation.cxx
@@ -155,22 +155,10 @@ ARMA ARMAFactoryImplementation::build(const TimeSeries & timeSeries) const
   throw NotYetImplementedException(HERE) << "In ARMAFactoryImplementation::build(const TimeSeries & timeSeries) const";
 }
 
-ARMA ARMAFactoryImplementation::build(const TimeSeries & timeSeries,
-                                      Point & informationCriteria) const
-{
-  throw NotYetImplementedException(HERE) << "In ARMAFactoryImplementation::build(const TimeSeries & timeSeries, Point & informationCriteria) const";
-}
-
 /* Build method */
 ARMA ARMAFactoryImplementation::build(const ProcessSample & sample) const
 {
   throw NotYetImplementedException(HERE) << "In ARMAFactoryImplementation::build(const ProcessSample & sample) const";
-}
-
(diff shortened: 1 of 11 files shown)


sofianehaddad:
@jschueller agreed that the changes allow "to disambiguate"
But IMHO in order to have an "homogeneous code" if changes are done here, we should have the same kind of **buildWithXXX* methods for all ones that had a Point ref. What do you think ?

jschueller:
That's the idea, are there remaining others ?

lib/src/Uncertainty/Process/ARMAFactoryImplementation.cxx
```diff
@@ -155,24 +155,12 @@ ARMA ARMAFactoryImplementation::build(const TimeSeries & timeSeries) const
...
 
-ARMA ARMAFactoryImplementation::build(const TimeSeries & timeSeries,
-                                      Point & informationCriteria) const
-{
-  throw NotYetImplementedException(HERE) << "In ARMAFactoryImplementation::build(const TimeSeries & timeSeries, Point & informationCriteria) const";
-}
-
 /* Build method */
 ARMA ARMAFactoryImplementation::build(const ProcessSample & sample) const
 {
```
sofianehaddad: I think we should keep only ARMAFactoryImplementation::build methods : 

  *ARMAFactoryImplementation::build(const TimeSeries & timeSeries,  Point & informationCriteria) const*

  *ARMAFactoryImplementation::build(const ProcessSample & sample,  Point & informationCriteria) const*

Both **ARMAFactoryImplementation::build(const TimeSeries & timeSeries) const** and 
  **ARMAFactoryImplementation::build(const ProcessSample & sample) const** should be removed

What do you think ?

jschueller: I thought returning a Point for the criteria is specific to WhittleFactory criteria, so I ditched them. Anyways it's not available from ARMAFactory interface. But as there are no other ARMAFactoryImplementation other than WhittleFactory it's hard to tell.

sofianehaddad: ok

regislebrun:
+1

jschueller:
tests fails, related to hmat: https://github.com/jeromerobert/hmat-oss/pull/51

PR #465, opened 2017-04-06
Allow notebook testing without nbsphinx

jschueller:
nbsphinx is not necessary to enable notebook tests, only nbconvert, mpl...


regislebrun:
+1 A good idea would be to allow to generate the sphinx documentation even if the tools to convert the notebooks are missing.

PR #466, opened 2017-04-06
Fix (Linear|Quadratic)NumericalMathFunction aliases

jschueller:
These are deprecated, and that also fixes the documentation as it shadows the dostrings of LinearFunction, QuadraticFunction


regislebrun:
+1

PR #467, opened 2017-04-10
Disable hmat-oss progress bar during factorization

dbarbier:
This progress bar had been added in latest hmat-oss (> 1.5.0).
We can replace calls to factorize by factorize_generic (added in hmat-oss 1.2.0)
to disable this progress bar.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,7 @@ if (MUPARSER_FOUND)
 endif ()
 
 if (USE_HMAT)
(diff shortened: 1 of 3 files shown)


dbarbier:
Should be merged into 1.9 branch too.

jschueller:
in CMakeLists.txt we have ```find_package (HMAT 1.1 NO_MODULE)```

dbarbier:
@jschueller done, thanks

PR #468, opened 2017-04-11
Fix RandomMixture::isContinuous

jschueller:
(doesn't fix last ticket)


dbarbier:
Description is useless, I am unable to understand the rationale of this PR.

jschueller:
I thought RandomMixture was only continuous if all variables are continuous, but I went too quick.

sofianehaddad:
@jschueller exact but finally the PR highlights other problems (see http://trac.openturns.org/ticket/887)

jschueller:
Yes, also, it seems it doesn't play well for discrete+continuous.

PR #469, opened 2017-04-12
Fix RandomMixture::computePDF with 2 atoms

sofianehaddad:
This fix http://trac.openturns.org/ticket/887 : the kernel convolution should be performed only if all atoms are continuous.

Fix also a small bug in RandomMixture::isIntegral

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -115,6 +115,7 @@
  * #879 (Incomplete arguments in FunctionalChaosRandomVector docstrings)
  * #882 (RandomMixture segfaults with Dirac)
  * #883 (VisualTest.DrawHistogram should rely on Histogram.drawPDF)
+ * #887 (Bogus PDF evaluation in RandomMixture with mix of continuous/discrete variables)
 
 == 1.8 release (2016-11-18) == #release-1.8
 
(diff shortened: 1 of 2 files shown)


jschueller:
the ticket must be listed in the changelog

sofianehaddad:
@jschueller you are right, thanks
In the 1.9 section or current Bug fixes ?

jschueller:
in the 1.9 section as I'll backport it

regislebrun:
-1 a RandomMixture is continuous as soon as one of the atoms is continuous. The bug is elsewhere (the use of a numerical convolution when there are 2 atoms without checking that both atoms are continuous). I have a fix on my laptop with no way to commit it before next week.

sofianehaddad:
@regislebrun agreed that RM is continuous as soon as one of the atoms is continuous.
That is why we check that **both the two atoms should be continuous** in order to use the convolution

regislebrun:
+1 I have also implemented a simplification mechanism to deal with the sum of discrete atoms (resulting into a unique discrete atom) and a discrete and a continuous atom (leading to a Mixture atom). This PR is related to the PR concerning getSupport() as we have to deal with the potential huge number of points in the support, thus in the resulting unique discrete distribution or the number of atoms in the Mixture.

PR #470, opened 2017-04-12
Implement RandomMixture::getSupport with more than one atom

dbarbier:
Also fix Multinomial::getSupport() which was bogus.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -112,6 +112,7 @@
  * #879 (Incomplete arguments in FunctionalChaosRandomVector docstrings)
  * #882 (RandomMixture segfaults with Dirac)
  * #883 (VisualTest.DrawHistogram should rely on Histogram.drawPDF)
+ * #886 (problem with getSupport)
  * #887 (Bogus PDF evaluation in RandomMixture with mix of continuous/discrete variables)
  * #888 (Bogus RandomMixture::getSample)
 
--- a/lib/src/Uncertainty/Distribution/RandomMixture.cxx
+++ b/lib/src/Uncertainty/Distribution/RandomMixture.cxx
@@ -2688,20 +2688,47 @@ Bool RandomMixture::isIntegral() const
   return true;
 }
 
+namespace {
+// Compute the support by looping over atoms.
+// Since the number of atoms is only known at runtime, it is not easy to implement
+// nested for-loops.  We use a recursive algorithm, by aggregating intermediate
+// results into 'accumulator' and final sample into 'supports'.
+void computeSupport(const UnsignedInteger start, const UnsignedInteger end,
+                    const Matrix & weights, const Point & constant, Collection<Sample> & localSupports,
+                    const Point & accumulator, Sample & supports)
+{
+  if (start == end)
+  {
+    supports.add(accumulator + constant);
+    return;
+  }
+  const Point currentWeight(*weights.getColumn(start).getImplementation());
+  Pointer<SampleImplementation> current = localSupports[start].getImplementation();
+  for (SampleImplementation::const_iterator cit = current->begin(); cit != current->end(); ++ cit)
+  {
+    const Scalar value = (*cit)[0];
+    const Point newAccumulator(accumulator + value * currentWeight);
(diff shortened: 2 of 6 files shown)


ChangeLog
```diff
@@ -27,6 +27,7 @@
 
 === Bug fixes ===
 
+ * #886 (problem with getSupport)
```
jschueller: could you squeeze this in the 1.9 part of the changelog so it merges nicely in the 1.9 branch ?

dbarbier: @jschueller Done

jschueller:
Could you rebase after Sofiane's changes ?

dbarbier:
I will rebase it after #472 is merged, it is uncontroversial IMO.

dbarbier:
I have rebased branches db/RandomMixture-getSupport and db/RandomMixture-getSupport-Tuples; IMO any branch can be merged, but we can wait for @regislebrun's input.

lib/src/Uncertainty/Distribution/Multinomial.cxx
```diff
@@ -477,9 +477,10 @@ Sample Multinomial::getSupport() const
 {
   const UnsignedInteger dimension = getDimension();
   LinearEnumerateFunction enumerate(dimension);
-  const UnsignedInteger size = enumerate.getStrataCumulatedCardinal(n_);
+  const UnsignedInteger firstIndex = n_ == 0 ? 0 : enumerate.getStrataCumulatedCardinal(n_ - 1);
+  const UnsignedInteger lastIndex = enumerate.getStrataCumulatedCardinal(n_);
   Sample support(0, dimension);
-  for (UnsignedInteger i = 0; i < size; ++i)
+  for (UnsignedInteger i = firstIndex; i < lastIndex; ++i)
```
regislebrun: @dbarbier There is a getStrataCardinal() to get the number of points in a given strata, no need to get it as a difference between two cumulated cardinals. BTW, there was no bug here, re-read the definition of the Multinomial class here:
http://openturns.github.io/openturns/master/user_manual/_generated/openturns.Multinomial.html?highlight=multinomial
In OpenTURNS we use a definition which reduces to the definition of the Binomial class when dimension=1. The classical definition leads to a singular discrete distribution, which is not a good thing. The classical definition is recovered by taking the last value of p equal to 0.
I agree that in this case, the support should be reduced (only the 'generic' support is implemented), but this change has also to be made into the Binomial and Bernoulli classes.

regislebrun:
-1 for the fix of Multinomial::getSupport()
+1 for the fix of RandomMixture::getSupport() in the recursive version, waiting for something more efficient in the (many) special cases. It is not reasonable to spend up to 2GB and 15s to produce the sequence of integers between 0 and 24 in the case of a sum of 24 Bernoulli atoms!

regislebrun:
@dbarbier In order to reduce the size of the intermediate results, the support can be computed by aggregating the atoms one by one and by compacting the support as soon as its size is too big. The rational behind this approach is that the support of a_1X_1+a_2X_2+a_3X_3 is the support of (a_1X_1+a_2X_2) + a_3X_3. If you do it for the sum of n Bernoulli(p), at the ith step you ahave to compute the support of the sum of 2 discrete distributions, one with a support of size i+1 and the other of size 2, so an intermediate support of size 2(i+1) and a compacted support of size i+2.
This kind of optimization is likely to occur only in the case of integral atoms, but it can also occur if there is a mix of integral/nonintegral atoms, so it should be activated in any case (or a clever use of isIntegral() is required for each of the atoms)

dbarbier:
@regislebrun Thanks for your review; I dropped the first commit (for Multinomial) and kept the 2nd one unchanged.

regislebrun:
@dbarbier your welcome ;-)! I have an alternative implementation of getSupport() which avoids to build a large sample before to reduce it, but it works currently only when dimension=1. After the merge of your PR I will see if I am able to catch the bug on time for the release or if I introduce the alternative as a special case when dimension=1.

dbarbier:
Closing, we are working on an alternative implementation

jschueller:
see #474

PR #471, opened 2017-04-13
Fix Mixture::isContinuous

sofianehaddad:
Distribution is continuous as soon as one of the atoms is continuous

### DIFF
--- a/lib/src/Uncertainty/Distribution/Mixture.cxx
+++ b/lib/src/Uncertainty/Distribution/Mixture.cxx
@@ -544,8 +544,8 @@ Bool Mixture::isElliptical() const
 Bool Mixture::isContinuous() const
 {
   const UnsignedInteger size = distributionCollection_.getSize();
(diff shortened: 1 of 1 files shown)


jschueller:
@dbarbier ```ot.Mixture([ot.Normal(), ot.Bernoulli()]).drawPDF()``` is continuous

dbarbier:
It *looks* continuous, like `ot.Uniform().drawPDF()`

jschueller:
what do you mean ?

sofianehaddad:
hmmm I think @dbarbier is right
@jschueller with the same example, have a look at the CDF

jschueller:
so it's almost continuous, except on the support of the bernouilli.

sofianehaddad:
@jschueller but CDF(0+) - CDF(0-) is not zero
If we refer to the mathematical definition, as CDF is a linear combination of CDF functions, it is continuous **if and only if** all CDF functions are **continuous**

jschueller:
yes, I agree.

PR #474, opened 2017-04-18
Improve random mixture

regislebrun:
This PR implements the following improvements in RandomMixture:
  * Made the positivity test of probabilities stricter in UserDefined
  * Removed useless include in Sample.
  * Made minor optimizations in RandomMixture
  * Fixed computePDF() in the case of an analytical discrete RandomMixture
  * Fixed constructors of RandomMixture: the constructors based on a Sample to define the weights were wrong: the size and dimension of the sample were tested in a wrong order and the associated matrix was transposed.
  * Improved the simplification mechanism of RandomMixture
  * Improved getSupport() in RandomMixture
    
The following simplifications are made:
For any dimension:
  * The Dirac atoms are merged into the constant
  * The RandomMixture atoms are flatten

For 1D RandomMixture:
  * The Uniform atoms can be merged two by two into a Trapezoidal or Triangular
  * The Exponential, ChiSquare and Gamma atoms can be merged into a Gamma atom if there scale parameters are all equal after standardization (so we must group these atoms by scale*weight parameter). The possible translation has to be accumulated into the constant.
  * The Normal atoms can be merged into a unique Normal
  * The Bernoulli and Binomial atoms can be merged into a unique Binomial as soon as they share the same value for p and the same weight
  * The Poisson atoms can be merged into a unique Poisson as soon as they share the same weight
  * The Discrete atoms are grouped into Discrete atoms of larger support, if these merged atoms have a support of reasonable size.
  * A continuous atom can be merged with a discrete atom to form a Mixture. This simplification can be done for each pair (continuous,discrete). It is not clear if some pairings are to prefer to others.


dbarbier:
Looks good to me, I have a cosmetics commit in my repository to improve exception messages.

@regislebrun You can include it by running these commands:
```
  git remote add dbarbier https://github.com/dbarbier/openturns.git
  git fetch dbarbier
  git reset --hard dbarbier/improve_random_mixture
```
if you want.

regislebrun:
@dbarbier Done. I just found a small bug when the coefficient of an Exponential, Gamma or ChiSquare is negative.

regislebrun:
Everything looks ok on my side. Crossing fingers...

PR #475, opened 2017-04-18
Prepare marking getGaussNodesAndWeights private

jschueller:


lib/src/Uncertainty/Algorithm/Analytical/AnalyticalResult.cxx
```diff
@@ -233,15 +234,16 @@ void AnalyticalResult::computeMeanPointInStandardEventDomain() const
   Scalar scaling = hasoferReliabilityIndex_ * p_standardDistribution->computeRadialDistributionCDF(hasoferReliabilityIndex_, true);
   Scalar a = hasoferReliabilityIndex_;
   // Quadrature rule
-  Point weights;
-  const Point nodes(p_standardDistribution->getGaussNodesAndWeights(weights));
-  const UnsignedInteger nodesSize = nodes.getSize();
+  const UnsignedInteger integrationNodesNumber = ResourceMap::GetAsUnsignedInteger("Distribution-DefaultIntegrationNodesNumber");
```
sofianehaddad: Maybe we should also rename the RM key? Distribution-DefaultIntegrationNodesNumber should be removed also

regislebrun: @sofianehaddad The key Distribution-DefaultIntegrationNodesNumber is used to tune the internal integrations done in many distributions. Don't touch it!

jschueller: The thing is getGaussNodesAndWeights (and integrationNodesNumber_...) will continue to be used internally as the number of nodes has to be changed depending on the distribution through setIntegrationNodesNumber (see Student, MaxEntropy, RandomMixture, Conditional, Normal...)

sofianehaddad: @jschueller for these distributions, we use other keys (Student-MarginalIntegrationNodesNumber",  "MaximumEntropyOrderStatisticsDistribution-CDFIntegrationNodesNumber", ...)

@regislebrun ok for internal use but the specific AnalyticalResult should have it is own key

regislebrun:
+1

lib/src/Uncertainty/Algorithm/Analytical/AnalyticalResult.cxx
jschueller: ok, done

PR #479, opened 2017-05-04
Fix http://trac.openturns.org/ticket/890

sofianehaddad:
The bounds of the new estimated Triangular distribution correspond now to the min/max of the data +/- the same delta which is the scaled range

### DIFF
--- a/lib/src/Uncertainty/Distribution/TriangularFactory.cxx
+++ b/lib/src/Uncertainty/Distribution/TriangularFactory.cxx
@@ -66,13 +66,14 @@ Triangular TriangularFactory::buildAsTriangular(const Sample & sample) const
   if (sample.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: can build a Triangular distribution only from a sample of dimension 1, here dimension=" << sample.getDimension();
 
   const Scalar xMin = sample.getMin()[0];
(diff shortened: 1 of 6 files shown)


lib/src/Uncertainty/Distribution/TriangularFactory.cxx
```diff
@@ -66,13 +66,14 @@ Triangular TriangularFactory::buildAsTriangular(const Sample & sample) const
   if (sample.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: can build a Triangular distribution only from a sample of dimension 1, here dimension=" << sample.getDimension();
 
   const Scalar xMin = sample.getMin()[0];
-  const Scalar a = xMin - std::abs(xMin) / (size + 2);
   const Scalar xMax = sample.getMax()[0];
-  const Scalar b = xMax + std::abs(xMax) / (size + 2);
+  Scalar delta = xMax - xMin;
```
regislebrun: const Scalar

sofianehaddad: @regislebrun the same delta is used and redifined if xMin == xMax

regislebrun: ok! +1

PR #485, opened 2017-05-17
Add callback examples

jschueller:
- FORM termination example using iteration limit
- MonteCarlo termination using time limit


regislebrun:
+1

PR #492, opened 2017-06-08
Improve performance of CovarianceModelImplementation::discretizeAndFactorize

dbarbier:
The covariance matrix is obtained as the Kronecker product of two matrices:
  C=Kron(A,B),
where B is the spatial covariance, which is precomputed.
If we compute the Cholesky decomposition of A and B:
  A = La La^T
  B = Lb Lb^T
Then Lc = Kron(La,Lb).  As complexity of Cholesky decomposition is O(N^3), this method is much faster in high dimension.

Remove CovarianceModel::compute{AsScalar,StandardRepresentative} overloads with a single argument, these methods are only defined for stationary covariance models.

Avoid creation of CovarianceMatrix instance in computeAsScalar.

Remove useless definitions of isStationary().


dbarbier:
UserDefinedStationaryCovarianceModel has to be updated before merging, but I am unsure how to change it.

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -267,44 +271,64 @@ struct CovarianceModelDiscretizePolicy
...
                                   const CovarianceModelImplementation & model)
     : input_(input)
     , output_(output)
     , model_(model)
-    , dimension_(model.getDimension())
   {}
 
   inline void operator()( const TBB::BlockedRange<UnsignedInteger> & r ) const
   {
+    const UnsignedInteger size = input_.getSize();
```
jschueller: unused ?

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -327,11 +351,21 @@ TriangularMatrix CovarianceModelImplementation::discretizeAndFactorize(const Mes
...
-  // We do not catch InternalException
-  // Incremeant nugget factor to make matrix positive definite
-  CovarianceMatrix covariance = discretize(vertices);
-  TriangularMatrix choleskyFactor = covariance.computeCholesky();
-  return choleskyFactor;
+  const UnsignedInteger size = vertices.getSize();
+  CovarianceMatrix rhoMatrix(size);
+  const CovarianceModelDiscretizePolicy policy( vertices, rhoMatrix, *this );
+  // The loop is over the lower block-triangular part
+  TBB::ParallelFor( 0, size * (size + 1) / 2, policy );
```
jschueller: why did you put spaces inside parenthesis ?

lib/src/Base/Stat/openturns/CovarianceModel.hxx
```diff
@@ -72,19 +72,13 @@ public:
   virtual Scalar computeStandardRepresentative(const Scalar & s,
       const Scalar & t) const;
 
-  virtual Scalar computeStandardRepresentative(const Point & tau) const;
-
-  virtual Scalar computeStandardRepresentative(const Scalar & tau) const;
```
jschueller: This should be reflected in the doc:
http://openturns.github.io/openturns/master/user_manual/_generated/openturns.CovarianceModel.html?highlight=covariancemodel#openturns.CovarianceModel.computeStandardRepresentative

lib/src/Base/Stat/CovarianceModelImplementation.cxx
dbarbier: copy-pasted from discretize()

lib/src/Base/Stat/CovarianceModelImplementation.cxx
dbarbier: indeed, I will drop it.

lib/src/Base/Stat/openturns/CovarianceModel.hxx
dbarbier: I was unsure about this one; there is no StationaryCovarianceModel_doc, and this page tells that the method with a single argument is designed for stationary covariance models, so this information looks still valid to me.

jschueller: ok, but now it's never a Scalar, only Point isn't it ?

dbarbier: Indeed, operator(), computeStandardRepresentative and computeAsScalar do not operate on Scalars, I'll fix that.

dbarbier: I was wrong about operator(), I'll only fix the other ones.

dbarbier:
Branch looks good to me now.
I still do not understand why UserDefinedStationaryCovarianceModel::operator() checks sign of its argument, but this is an unrelated issue.

jschueller:
It's good for me too.

dbarbier:
Please do not merge, there are still corner cases, I am working on a cleaner implementation.

regislebrun:
@dbarbier @jschueller Hi guys, it is quite complex for me to be focused on the code at the moment. A point I would like to emphasize is that the Kronecker product form of covariance models is by no way linked to its stationary property, but rather due to its particular form when it can be written in the form given at the end of the Notes section here:
[http://openturns.github.io/openturns/master/user_manual/_generated/openturns.CovarianceModel.html?highlight=covariancemodel](url)

The fact that all the usual stationary models can be written in this form is a coincidence, and in my view a new property should be added.

Thinking about this optimization, it raises new questions to me:
- the amplitude parameter of these models should not be constrained to be 1D vectors
- the dimension of the spatial correlation should be checked against the dimension of the amplitude
- another kind of optimization should be possible when the covariance function is indeed diagonal (which is the case by default in many models)

In my view, this PR should be merged as soon as Denis is ok, but additional work in the same spirit has yet to be done. So +1 but more work is definitely needed here (and thanks for the good job already done)

dbarbier:
@regislebrun Your first point has already been addressed, I published this PR after implementation has been moved away from StationaryCovarianceModel.

This branch has been rebased and current implementation is much cleaner.  Up to now, discretize calls operator() to build local covarianced matrices.  In the initial version of this PR, I switched to computeStandardRepresentative().  But the problem is that some covariance models do not override this method, and thus discretize would crash.  In this new implementation, a flag is set in covariance models to tell whether they define computeStandardRepresentative or not, and discretize can then either call operator() or computeStandardRepresentative().

Branch db/speed-up-discretize2 will provide some speed-up too, but I am not sure if we want to make a distinct PR or not.

dbarbier:
For the record, there is no hurry to merge this PR, AFAICT there are no conflicting developments just now.  I am working on some benchmarks to show current results and remaining issues, we will then discuss whether we are happy with this PR or are willing to fix remaining issues.

dbarbier:
I pushed a branch db/benchmark-discretize which contains a benchmark file and timings.
Here are current results:

[timings.xlsx](https://github.com/openturns/openturns/files/1068395/timings.xlsx)

Please test and discuss; commit 1660873 is too slow; maybe we can try to use TBB there, but I am not sure it would work.

dbarbier:
@jschueller Please merge this pull request.  I reviewed it yesterday with @regislebrun, and will address his remarks in another PR.

PR #493, opened 2017-06-08
Use Lapack dger for rank-1 updates

dbarbier:


dbarbier:
This is mostly cosmetics, the aim is to use the right Lapack function when available.
It should be slightly faster since there is no allocation of temporary matrices, but hardly measurable on real use cases.

PR #496, opened 2017-06-16
New method SampleImplementation::operator()(i,j)

dbarbier:
Call it from Sample::operator()(i,j) to avoid creation of
NSI_point or NSI_const_point objects.

PR #498, opened 2017-06-19
Add ProbabilitySimulation

jschueller:
This class allows to estimate an event probability by simulation using sequential experiments.
It only handles composite events, to evaluate non-composite events the class Simulation can now be used on it's own, for example to simulate process events by the equivalent of Monte Carlo.

The sequential experiments that can be used are:
    - MonteCarloExperiment
    - ImportanceSamplingExperiment
    - LowDiscrepancyExperiment
    - LHSExperiment
    
The following classes are deprecated in the Python interface:
    - MonteCarlo
    - ImportanceSampling
    - QuasiMonteCarlo
    - RandomizedQuasiMonteCarlo
    - RandomizedLHS

The other simulation algos (DS, LHS, ..) now also know how to handle domain events.

Example1 : MonteCarlo
```
experiment = ot.MonteCarloExperiment()
myAlgo = ot.ProbabilitySimulation(myEvent, experiment)
myAlgo.setMaximumOuterSampling(150)
myAlgo.setBlockSize(4)
myAlgo.setMaximumCoefficientOfVariation(0.1)
myAlgo.run()
```
Example2 : ImportanceSampling
```
myImportance = ot.Normal([49.969, 1.84194, 10.4454, 4.66776], [1.0]*4, ot.IdentityMatrix(4))
experiment = ot.ImportanceSamplingExperiment(myImportance)
myAlgo = ot.ProbabilitySimulation(myEvent, experiment)
myAlgo.setMaximumOuterSampling(150)
myAlgo.setBlockSize(4)
myAlgo.setMaximumCoefficientOfVariation(0.1)
myAlgo.run()
```
Example3 : Quasi Monte Carlo
```
experiment = ot.LowDiscrepancyExperiment()
myAlgo = ot.ProbabilitySimulation(myEvent, experiment)
myAlgo.setMaximumOuterSampling(150)
myAlgo.setBlockSize(4)
myAlgo.setMaximumCoefficientOfVariation(0.1)
myAlgo.run()
print('Probability estimate=%.6f' % myAlgo.getResult().getProbabilityEstimate())
Probability estimate=0.141667
```

A new method Bool WeightedExperiment::hasUniformWeights() const is added.

Also added a randomization flag to LHSExperiment in order to replace RandomizedLHS

Note that the LHS class is kept, as it's not currently used as a sequential algorithm as the total outersampling must be known to compute the shuffle matrix. This is potentially ambiguous as one may also use ProbabilitySimulation with LHSExperiment to get the behavior of the old RandomizedLHS class.

I think PostAnalyticalImportanceSampling can be transformed into a weighted experiment, but I did not do it because we cannot do the same for PostAnalyticalControlledImportanceSampling, which inherits PostAnalyticalSimulation too.


kstuyen:
Thanks

lib/src/Uncertainty/Algorithm/WeightedExperiments/LowDiscrepancyExperiment.cxx
```diff
@@ -108,7 +114,8 @@ String LowDiscrepancyExperiment::__str__(const String & offset) const
       << "(sequence=" << sequence_
       << ", distribution=" << distribution_
       << ", size" << size_
-      << ", restart=" << (restart_ ? "true" : "false");
+      << ", restart=" << (restart_ ? "true" : "false")
```
regislebrun: Use std::boolalpha here and in the next line

jschueller: Turns out boolalpha is useless as OSS has a bool overload, just use ```<< ", restart=" << restart_```

jschueller:
hi @kstuyen, who are you ???

kstuyen:
Hi Jschueller, I am a beginner who wants to use OpenTURNS in Salome-Meca for my calculations. I can not install OpenTURNS for my UBUNTU 12.04 (I use Xubuntu - CAElinux2013).
 I don't know which files are suitable in a number of files that instead in the link: http://ubuntu.openturns.org/. 
I would like to thank you in advance.
TRAN Tuyen,

jschueller:
@kstuyen, please consider using the mailling list "users@openturns.org" instead to join us.

We do provide packages for ubuntu 12.04, see instructions here:
http://openturns.github.io/openturns/master/install.html#debian-ubuntu

lib/src/Uncertainty/Algorithm/Simulation/ProbabilitySimulation.cxx
```diff
@@ -0,0 +1,123 @@
...
+ProbabilitySimulation * ProbabilitySimulation::clone() const
+{
+  return new ProbabilitySimulation(*this);
+}
+
+
+void ProbabilitySimulation::setExperiment(const WeightedExperiment & experiment)
+{
+  experiment_ = experiment;
+  experiment_.setDistribution(getEvent().getImplementation()->getAntecedent()->getDistribution());
```
regislebrun: Boom! This line imposes to work with an event of the composite type (eg f(X)>s), and it is no more possible to work with more general events (eg based on processes)

jschueller: I don't see simulation tests with such events. In fact I don't see tests involving ProcessEvents. Do you have a script with that type of simulation ?

regislebrun: Yes, I was writing it in parallel of the review. Here I create 4 events (one by Event constructor) and run MonteCarlo on it. With OT 1.9 I get:

> ##################################################
composite vector/comparison event
MonteCarlo result= probabilityEstimate=2.473684e-01 varianceEstimate=6.053604e-04 standard deviation=2.46e-02 coefficient of variation=9.95e-02 confidenceLength(0.95)=9.64e-02 outerSampling=95 blockSize=4
##################################################
composite vector/domain event
MonteCarlo result= probabilityEstimate=3.333333e-01 varianceEstimate=1.102293e-03 standard deviation=3.32e-02 coefficient of variation=9.96e-02 confidenceLength(0.95)=1.30e-01 outerSampling=63 blockSize=4
##################################################
composite vector/interval event
MonteCarlo result= probabilityEstimate=1.927481e-01 varianceEstimate=3.688685e-04 standard deviation=1.92e-02 coefficient of variation=9.96e-02 confidenceLength(0.95)=7.53e-02 outerSampling=131 blockSize=4
##################################################
process/domain event
MonteCarlo result= probabilityEstimate=9.166667e-01 varianceEstimate=8.101852e-03 standard deviation=9.00e-02 coefficient of variation=9.82e-02 confidenceLength(0.95)=3.53e-01 outerSampling=3 blockSize=4

and with your PR:

> ##################################################
composite vector/comparison event
WRN - class MonteCarlo is deprecated in favor of ProbabilitySimulation
MonteCarlo result= probabilityEstimate=2.473684e-01 varianceEstimate=6.053604e-04 standard deviation=2.46e-02 coefficient of variation=9.95e-02 confidenceLength(0.95)=9.64e-02 outerSampling=95 blockSize=4
##################################################
composite vector/domain event
Not supported by MonteCarlo
##################################################
composite vector/interval event
MonteCarlo result= probabilityEstimate=2.079832e-01 varianceEstimate=4.288171e-04 standard deviation=2.07e-02 coefficient of variation=9.96e-02 confidenceLength(0.95)=8.12e-02 outerSampling=119 blockSize=4
##################################################
process/domain event
Not supported by MonteCarlo

The code is:

```python
import openturns as ot
description = ot.Description()
description.add("composite vector/comparison event")
dim = 2
distribution = ot.Normal(dim)
Xvector = ot.RandomVector(distribution)
f = ot.SymbolicFunction(["x0", "x1"], ["x0+x1"])
Yvector = ot.RandomVector(f, Xvector)
s = 1.0
event1 = ot.Event(Yvector, ot.Greater(), s)
description.add("composite vector/domain event")
domain1D = ot.LevelSet(ot.SymbolicFunction(["x0"], ["sin(x0)"]), -0.5)
event2 = ot.Event(Yvector, domain1D)
vector/comparison event
description.add("composite vector/interval event")
interval = ot.Interval(0.5, 1.5)
event3 = ot.Event(Yvector, interval)
description.add("process/domain event")
Xprocess = ot.WhiteNoise(distribution, ot.RegularGrid(0.0, 0.1, 10))
domain2D = ot.LevelSet(ot.SymbolicFunction(["x0", "x1"], ["(x0-1)^2+x1^2"]), 1.0)
event4 = ot.Event(Xprocess, domain2D)
all_events = [event1, event2, event3, event4]
for i,event in enumerate(all_events):
    print "#"*50
    print description[i]
    try:
        myAlgo = ot.MonteCarlo(event)
        myAlgo.setMaximumOuterSampling(250)
        myAlgo.setBlockSize(4)
        myAlgo.setMaximumCoefficientOfVariation(0.1)
        myAlgo.run()
        print "MonteCarlo result=", myAlgo.getResult()
    except:
        print "Not supported by MonteCarlo"
```

jschueller:
@regislebrun now ProbabilitySimulation only handles composite events and also handles domain events.
To handle process/domain events Simulation can now be used.
Simulation/ProbabilitySimulation will probably have to be renamed when we introduce quantile simulations.
Other simulation algorithms (LHS, DS, ) also check if event is composite and handle domain events when possible.
I included your test in t_ProbabilitySimulation_std.py (your original script works as expected).

regislebrun:
+1 @jschueller I will improve the Event constructor based on a CompositeRandomVector and a Domain to produce a composite Event when the domain is a LevelSet. It will enlarge the applicability of the generic algorithm.

regislebrun:
+1

lib/src/Uncertainty/Algorithm/WeightedExperiments/LowDiscrepancyExperiment.cxx
```diff
@@ -143,15 +150,44 @@ void LowDiscrepancyExperiment::setRestart(const Bool restart)
...
   sample.setDescription(distribution_.getDescription());
   const UnsignedInteger dimension = marginals_.getSize();
+  Scalar tmp = -1.0;
   for (UnsignedInteger i = 0; i < size_; ++ i)
-    for (UnsignedInteger j = 0; j < dimension; ++ j) sample[i][j] = marginals_[j].computeQuantile(sample[i][j])[0];
+  {
+    for (UnsignedInteger j = 0; j < dimension; ++ j)
+    {
+      Scalar p = sample[i][j];
+      if (randomize_)
```
sofianehaddad: small detail, we could use accessors introduced in #496

ChangeLog
```diff
@@ -18,6 +19,7 @@
  * Removed deprecated GramSchmidtAlgorithm, ChebychevAlgorithm classes
  * Removed [gs]etOptimizationSolver methods
  * Removed CovarianceModel::compute{AsScalar,StandardRepresentative} overloads
+ * Deprecated MonteCarlo, ImportanceSampling, QuasiMonteCarlo, RandomizedQuasiMonteCarlo, RandomizedLHS classes
```
sofianehaddad: what about LHS?

jschueller: "Note that the LHS class is kept, as it's not currently used as a sequential algorithm as the total outersampling must be known to compute the shuffle matrix. This is potentially ambiguous as one may also use ProbabilitySimulation with LHSExperiment to get the behavior of the old RandomizedLHS class."

sofianehaddad:
@jschueller @regislebrun I suggest an *Algorithm* suffix (class contains a **run** method and returns a *XResult* class) to be coherent with OT architecture. What do you think?

ChangeLog
sofianehaddad: Indeed it is ambiguous

jschueller:
What about removing the others ones ? : )

regislebrun:
@jschueller @sofianehaddad I agree with Sofiane. The name 'Simulation' comes from the early days of OpenTURNS and should be rationalized with the name of other classes of algorithm type.

jschueller:
My point is removing the rigidity of these name suffix goes well with the simplification fashion we've done recently: kriging instead of krigingalgorithm, functionalchaos instead of functionalchaosalgorithm, tensorapproximation instead of tensorapproximation algorith... It doesn't look so bad to me.

regislebrun:
@jschueller @sofianehaddad I am not an adept of long names for the pleasure of it. A long name is here only if a shorter one would be ambiguous. Some names are really over-used in statistics and probability, such as exponential or normal, and if the names are not explicit enough many confusions could occur. For example, in my view AbsoluteExponential, GeneralizedExponential, SquaredExponential should have a Model suffix. The Laplace distribution is also called the absolute exponential distribution for example. So I agree on the fact that the Algorithm suffix could be removed if it does not lead to confusion, in particular wrt the associated result class. But it should be done in another PR dedicated to name simplification.

jschueller:
So what do you want me to do exactly ?

regislebrun:
@jschueller Call it ProbabilitySimulationAlgorithm, merge the PR and start a discussion on the renaming of classes/methods/argument order ;-)!

jschueller:
I do not really see the point, but ok it's 2vs1 : )
I cannot merge it right away as @sofianehaddad did not approve (or denis).

regislebrun:
+1

sofianehaddad:
@jschueller @regislebrun @dbarbier IMHO we should also handle the LHS case.
As mentionned previously by @jschueller the used algorithm is not *sequential*. So the question is why LHS::computeBlockSample does not handle LHS experiments of size **blockSize** and keep the *sequential* algorithm ?

jschueller:
I guess ComputeShuffle doesnt shuffle anything when blocksize=1. Maybe it's wrong to use the ProbabilitySimulationAlgorithm sequential version with LHSExperiment.

sofianehaddad:
@jschueller the current LHS class does not also handle the case blockSize=outerSampling=1
Btw, if it is only size restirctions, thus we could add more checks in experiments classes (How to handle blockSize=1 with LowDiscrepancyExperiment ?). 
If it is a more complex algorithmic limitation, thus I will be interested in a return

jschueller:
The case blockSize=1 is indeed problematic when the underlying experiment has to be generated all at once.
Maybe an Experiment::isSequential method ?

dbarbier:
You replaced many occurrences of
```
  event.getOperator().operator()(x, event.getThreshold()))
```
by
```
  event.getDomain().contains(x)
```
Maybe getOperator() and getThreshold() could be deprecated?  Moreover I do not understand why these methods belong to RandomVector and not Event.

IMHO there is no reason to add a getDomain() method, it looks like Event has to provide a service to tell whether a given point lies in event's domain, thus you should add an Event::contains(Point) method (with a better name?).

jschueller:
I cannot just add a isRealizedAt method only to Event as the implementation classes (EventRV, DomainEvent) all inherit from RandomVectorImplementation.
Also, some algorithms explicitely require the threshold.

sofianehaddad:
@regislebrun is there any reason to keep LHS 'non sequential' ?

regislebrun:
Here is a short explanation of the spirit of simulation methods:
+ they are all based on the strong law of large numbers (SLLN, convergence of an empirical mean toward an expectation) controlled by a central limit theorem (TCL, the convergence is monitored essentially by the variance, the empirical variance in practice).
+ as a consequence, all the block samples have to be independent in the probabilistic sense
+ there is no convergence check inside of the computation of a block, and the estimator associated to a block does not have to be controlled by either the SSLN (allowing for low discrepancy sequences) and the TCL (allowing for low discrepancy sequences or directional sampling or LHS)

With that in mind, here are the consequences on the existing algorithms:
+ the QuasiMonteCarlo and LHS algorithms are NOT simulation algorithms. The only mathematically sound way to use a LowDiscrepancyExperiment or a LHSExperiment in this framework is to generate only one block: the algorithm stops when the full simulation budget has been exhausted with no control on the convergence. In the case of the QuasiMonteCarlo algorithm, it is nevertheless possible to use an outer sampling greater than one but with no control of the convergence by the TCL. So the use of the simulation algorithm framework is essentially a trick.
+ the RandomizedQuasiMonteCarlo and RandomizedLHSExperiment are simulation algorithms. The LowDiscrepancyExperiment class is not currently able to reimplement the RandomizedQuasiMonteCarlo algorithm in this new framework but it can be easily fixed by providing a flag to activate a random shift in the spirit of the LHSExperiment. The RandomizedLHSExperiment is what one gets by taking a number of outer sampling greater than one in the new framework.
+ the DirectionalSampling algorithm could be inserted into this framework. For that, the SamplingStrategy class has to be converted into a WeightedExperiment class.

sofianehaddad:
@regislebrun ok thus we can merge this PR and fix the problems mentionned in new PR

jschueller:
- There's an example usage of QMC in the documentation of ProbabilitySimulation, so I guess 
- LowDiscrepancyExperiment has a random shift flag for this purpose, see 4a9eb1d
- I'll look into DS

PR #499, opened 2017-06-19
Use Trusty container in Travis

jschueller:
This should speed up the linux build that timeouts nowadays by not uselessly compiling dependencies.


jschueller:
This should speed up the linux build that timeouts nowadays by not uselessly compiling dependencies.

PR #500, opened 2017-06-19
Fix loading configuration file on OSX when installing with conda

jschueller:
Conda performs prefix substitution within libraries so that packages
are relocatable.  On OSX, line
  directory = String(SYSCONFIG_PATH) + PrefixConfigSubdirectory_;
causes trouble.  When compiling, conda makes SYSCONFIG_PATH a very long
string, and it is replaced during installation.
On Linux, it works as expected, string contains the new path, and /openturns
is appended to it afterwards.
On OSX, string also contains the new path, but it keeps its original length,
and thus /openturns is appended after lots of \0.

With
  directory = String() + String(SYSCONFIG_PATH).c_str() + PrefixConfigSubdirectory_;
SYSCONFIG_PATH is adjusted to have the new length.

### DIFF
--- a/lib/src/Base/Common/Path.cxx
+++ b/lib/src/Base/Common/Path.cxx
@@ -111,7 +111,11 @@ FileName Path::GetInstallationDirectory()
   }
   if (!dirExists)
   {
-    directory = String(INSTALL_PATH);
+    // NOTE: When installing with conda, INSTALL_PATH is substituted during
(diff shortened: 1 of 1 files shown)


jschueller:
+1: https://github.com/conda-forge/openturns-feedstock/pull/38
https://travis-ci.org/conda-forge/openturns-feedstock/jobs/244559893

sofianehaddad:
+1

PR #505, opened 2017-06-23
PythonDistribution enhancements

jschueller:
- Add PythonDistribution::getMarginal(int)
- Fix DistributionImplementation::getMarginal

### DIFF
--- a/lib/src/Uncertainty/Model/DistributionImplementation.cxx
+++ b/lib/src/Uncertainty/Model/DistributionImplementation.cxx
@@ -3044,17 +3044,20 @@ Scalar DistributionImplementation::computeDensityGeneratorSecondDerivative(const
 /* Get the i-th marginal distribution */
 DistributionImplementation::Implementation DistributionImplementation::getMarginal(const UnsignedInteger i) const
 {
-  if ((dimension_ == 1) && (i == 0)) return clone();
-  if (isCopula() && (i < dimension_)) return new Uniform(0.0, 1.0);
+  if (!(i < dimension_)) throw InvalidArgumentException(HERE) << "Marginal index cannot exceed dimension";
+  if (dimension_ == 1) return clone();
+  if (isCopula()) return new Uniform(0.0, 1.0);
   return MarginalDistribution(*this, i).clone();
 }
 
 /* Get the distribution of the marginal distribution corresponding to indices dimensions */
 DistributionImplementation::Implementation DistributionImplementation::getMarginal(const Indices & indices) const
 {
-  if ((dimension_ == 1) && (indices[0] == 0)) return clone();
-  if ((dimension_ == 2) && (indices.getSize() == 1) && (indices[0] < dimension_)) return new Uniform(0.0, 1.0);
-  if ((dimension_ == 2) && (indices.getSize() == 2) && (indices[0] == 0) && (indices[1] == 1)) return clone();
+  if (!indices.check(dimension_)) throw InvalidArgumentException(HERE) << "Marginal indices cannot exceed dimension";
+  Indices full(dimension_);
+  full.fill();
+  if (indices == full) return clone();
+  if (isCopula() && (indices.getSize() == 1)) return new Uniform(0.0, 1.0);
   return MarginalDistribution(*this, indices).clone();
 }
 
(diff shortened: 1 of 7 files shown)


lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -3044,17 +3044,24 @@ Scalar DistributionImplementation::computeDensityGeneratorSecondDerivative(const
...
-  if ((dimension_ == 2) && (indices.getSize() == 1) && (indices[0] < dimension_)) return new Uniform(0.0, 1.0);
-  if ((dimension_ == 2) && (indices.getSize() == 2) && (indices[0] == 0) && (indices[1] == 1)) return clone();
+  if (!indices.check(dimension_)) throw InvalidArgumentException(HERE) << "Marginal indices cannot exceed dimension";
+  Indices full(dimension_);
+  full.fill();
+  if (indices == full) return clone();
+  if (isCopula())
+  {
+    if (indices.getSize() == 1) return new Uniform(0.0, 1.0);
+    else return ComposedDistribution(DistributionCollection(indices.getSize(), Uniform(0.0, 1.0))).clone();
```
regislebrun: -1
This is WRONG! You are saying that all the lower dimensional marginal distributions of a copula have INDEPENDENT uniform distributions over [0,1], which is cearly false. BTW, you could have used the more compact IndependentCopula(indices.getSize()) construction to get a distribution with the axact same mathematical behaviour and a much more efficient implementation

jschueller: Yes, my bad.
I found a new simplification: if iscopula && indices.size=1 -> return Uniform(0,1)

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -3044,17 +3044,20 @@ Scalar DistributionImplementation::computeDensityGeneratorSecondDerivative(const
 /* Get the i-th marginal distribution */
 DistributionImplementation::Implementation DistributionImplementation::getMarginal(const UnsignedInteger i) const
 {
-  if ((dimension_ == 1) && (i == 0)) return clone();
-  if (isCopula() && (i < dimension_)) return new Uniform(0.0, 1.0);
+  if (!(i < dimension_)) throw InvalidArgumentException(HERE) << "Marginal indice cannot exceed dimension";
```
regislebrun: Marginal index

regislebrun:
+1

sofianehaddad:
How could we test bb9106d?

jschueller:
@sofianehaddad some of it is tested with PythonDistribution as it previously went into the copula path and returned an Uniform.

sofianehaddad:
@jschueller in your test, getMarginal(int) refers to getMarginal(Indices&)
I am wondering if we could have a more "robust" test which only invoke "DistributionImplementation::getMarginal"

jschueller:
It's difficult to test it because it's often overloaded.

jschueller:
@regislebrun CopulaImplementation::getMarginal(i) returns IndependentCopula(1)
I guess the only difference is the copula flag with DistributionImplementation::getMarginal returning Uniform(0, 1).
So I propose to remove CopulaImplementation::getMarginal.

regislebrun:
@jschueller You can use CumulativeDistributionNetwork for this test. Here is an example:
```python
import openturns as ot

# Build a distribution with no specific implementation of getMarginal()
atom = ot.ComposedDistribution([ot.Uniform(0.0, 1.0)]*3)
graph = ot.BipartiteGraph([[0, 1, 2], [0, 1, 2]])
distribution = ot.CumulativeDistributionNetwork([atom]*2, graph)

# Extract its marginal using the generic implementation
marginal = distribution.getMarginal([0,1])

# Build by hands the exact marginal
atom_ref = ot.ComposedDistribution([ot.Uniform(0.0, 1.0)]*2)
graph_ref = ot.BipartiteGraph([[0, 1], [0, 1]])
ref = ot.CumulativeDistributionNetwork([atom_ref]*2, graph_ref)

graph = marginal.drawPDF()
graph.add(ref.drawPDF())
ot.Show(graph)
```

jschueller:
Thanks!

PR #506, opened 2017-06-28
Various Python layer enhancements

jschueller:
- Use one entry for all contour iso-values using legend attribute not labels, like the R backend
- Do not set Pairs text color to point color (keep it black by default)
- Add operator*(Scalar, Sample)


regislebrun:
+1

python/test/t_Sample_operators.expout
```diff
@@ -12,6 +12,10 @@ sample*2:     [ a0 a1 ]
 0 : [ 60 80 ]
 1 : [ 62 82 ]
 2 : [ 64 84 ]
+2*sample:     [ a0 a1 ]
+0 : [ 60 80 ]
+1 : [ 62 82 ]
+2 : [ 64 84 ]
 sample/2:     [ a0   a1   ]
```
sofianehaddad: @jschueller  f9796be should be in a new PR ?
Btw seems a bug here no ? 2*Sample == Sample ???

jschueller: no:
```sample5 =     [ a0 a1 ]
 0 : [ 30 40 ]
 1 : [ 31 41 ]
2 : [ 32 42 ]
```

sofianehaddad: ok but I think the commit should be outside this PR

jschueller: I renamed the PR: Various Python layer enhancements

jschueller:
it think it's reasonable : we have already scalar*point, scalar*matrix, also np.array supports this

jschueller:
So, what do you think @dbarbier ?

dbarbier:
We define Sample+int, but neither Point+int nor int+Sample are defined.  IMO we should be consistent.  But that is not a big deal for me, merge it if you want.

jschueller:
I agree sample+int is weird, I'll try to address that in another PR

PR #507, opened 2017-06-29
Improved Sample::ImportFromTextFile performance

regislebrun:
Now, the speed is 10% better than pandas using blank separators and on par using a one-character separator.
In the new implementation I restricted the separator to be of size 1, but it could easily be removed at the expense of a significant performance penalty


dbarbier:
@regislebrun Let t_NumericalSample_csv.py also test ImportFromTextFile to reproduce this error on Linux.

dbarbier:
It is caused by the trailing space on last line.

regislebrun:
@dbarbier What I don't understand is why the CSV import is broken with clang/osx with the exact same code as in master for the lexer and the exact same test. It seems that there have been some discrepancies between gcc/clang when streaming floats some times ago(https://stackoverflow.com/questions/22233868/another-istream-discrepancy-between-libstdc-and-libc, not very recent but looks close to what appends here), maybe something similar is in action here.

jschueller:
BuildFromCSVFile calls BuildFromTextFile if bison/flex is not present, that's what causing the issue on mac.

sofianehaddad:
@jschueller we could add flex/bison on travis  (brew install bison flex) if required

sofianehaddad:
@jschueller currently conda packages do not rely on bison/flex even if both these last one exist!
Should we put these dependencies mandatory ?

dbarbier:
Fix pushed in my repository, branch db/improve_import_from_text_file.  I only checked C++ test file, Python test may have to be edited.

jschueller:
@sofianehaddad why not

regislebrun:
@dbarbier I saw your fix, nice! How do we proceed? I copy-paste the fix in my branch? A PR?

jschueller:
@regislebrun you can cherry-pick his commits to your branch: 
- git remote add dbarbier https:/...dbarbier/openturns.git
- git fetch dbarbier
- git cherry-pick xxxxxxxx

dbarbier:
Once you fetched my repository, you can run
  git reset --hard dbarbier/db/improve_import_from_text_file
to move your branch tip to mine.

jschueller:
superseded by #508

sofianehaddad:
@jschueller for bison/flex, the problem is that it is forced to be "off" when using Cotire
How could we fix it ? Is there any restriction that makes incompatible using both cotire and flex/bison?

jschueller:
I think cotire cannot be used together with our add_custom_command (we cannot build sphinx either) because cotire uses a distinct set of targets that doesn't get along with our rules. It would have been fine if cmake allowed cotire to replace the existing targets by unity targets unfortunately that doesnt seem possible cf https://github.com/sakra/cotire/issues/55.

jschueller:
@sofianehaddad I think I have a solution for this.

PR #508, opened 2017-06-30
Improve import from text file

regislebrun:
Now, the speed is 10% better than pandas using blank separators and on par using a one-character separator.
In the new implementation I restricted the separator to be of size 1, but it could easily be removed at the expense of a significant performance penalty


jschueller:
a rebase is needed

dbarbier:
setlocale is thread-unsafe; on Linux and Mac we can use uselocale, and on Windows _configthreadlocale.
I pushed a commit to my branch, it should work fine but is not fully tested.

jschueller:
there should be no merge commit

dbarbier:
@regislebrun Please stop adding controversial changes to your PR, unless it fixes regressions recently introduced.
If this is not a regression, I would prefer if you move b5c85d7 into a separate PR, and at the same time support escaped quotes.

regislebrun:
@jschueller How to fix this merge commit? I am not used to share development inside a branch, and I followed the same instruction to update my locale copy of the branch as I do when synchronizing master.
@dbarbier I introduced this bug when implementing the ParseAsDescription() method. I reused a bogus code and saw its flaws after the PR, it is why it appeared on a second time. This PR can wait for a better moment for its review, but IMHO it is the right place for this code.

dbarbier:
@regislebrun AFAICT quoted semicolons also do not work in master, it is not a regression.  Quoting in CSV is tricky; if we claim to support that, we should do it right, but your commit does not handle all situations, this is why I suggest to move it into another PR.

regislebrun:
@jschueller @dbarbier  Here I face another problem: my lack of expertise on git! In particular:

- how to fix the merge commit?
- how to reuse a commit done in db/improve_import_from_text_file into a new branch of mine?

jschueller:
to fix the merge commit you could use reset --hard to rollback to a previous commit
then use cherry-pick to include commits, possibly from another branch provided you added and fetched it's remote repo

regislebrun:
I close this PR as it is by no means critical and it takes me ways too much time to fix/learn git.

PR #509, opened 2017-07-03
Perform sample html repr

sofianehaddad:
Currently print(ot.Sample) generates an HTML table of the full sample in a notebook environment, which is incoherant regarding the str method.
Furthermore, the full representation makes sometimes the kernel crash if sample is 'huge' 

This PR fixes the issue by setting an html repr similar to the standard str one

### DIFF
--- a/python/src/Sample.i
+++ b/python/src/Sample.i
@@ -60,18 +60,43 @@ def Sample__repr_html_(self):
     """Get HTML representation."""
     html = '<TABLE>'
     desc = self.getDescription()
+    printEllipsisThreshold = openturns.common.ResourceMap.GetAsUnsignedInteger("Sample-PrintEllipsisThreshold")
+    printEllipsisSize = openturns.common.ResourceMap.GetAsUnsignedInteger("Sample-PrintEllipsisSize")
+    sampleSize = self.getSize()
+    sampleDimension = self.getDimension()
+    ellipsis = sampleSize * sampleDimension > printEllipsisThreshold
     if not desc.isBlank():
-        html += '<TR>'
-        html += '<TD></TD>'
-        for j in range(self.getDimension()):
-            html += '<TH>' + desc[j] + '</TH>'
-        html += '</TR>'
-    for i in range(self.getSize()):
-        html += '<TR>'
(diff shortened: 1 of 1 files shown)


dbarbier:
-1, your repr_html unnecessarily differs from repr:
 * Sample-PrintEllipsisSize controls the number of elements displayed when ellipsis are printed.  But if Sample-PrintEllipsisSize < sample size or dimension < 2*Sample-PrintEllipsisSize, ellipsis should not be used 
 * Use Sample-PrintEllipsisThreshold resource key
 * colspan attribute must be added to <td> element when writing ellipsis

dbarbier:
@sofianehaddad  Please check db/perform-sample_html_repr in my repository

sofianehaddad:
@dbarbier ok I will have a look

python/src/Sample.i
```diff
@@ -58,20 +58,46 @@ Sample.__getattribute__ = Sample___getattribute__
 
 def Sample__repr_html_(self):
     """Get HTML representation."""
+    from openturns.common import ResourceMap
```
jschueller: common is already imported

python/src/Sample.i
```diff
@@ -58,20 +58,46 @@ Sample.__getattribute__ = Sample___getattribute__
...
     html = '<TABLE>'
     desc = self.getDescription()
+    printEllipsisThreshold = ResourceMap.GetAsUnsignedInteger("Sample-PrintEllipsisThreshold")
+    printEllipsisSize = ResourceMap.GetAsUnsignedInteger("Sample-PrintEllipsisSize")
```
jschueller: you can probably do printEllipsisSize = openturns.common.ResourceMap.GetAsUnsignedInteger("Sample-PrintEllipsisSize")

jschueller:
thanks, I needed this for my documentation examples

python/src/Sample.i
sofianehaddad: Fixed

jschueller:
@regislebrun good to go ?

regislebrun:
+1

PR #510, opened 2017-07-03
Remove Sample Scalar translation operators

jschueller:
Removes Sample::operator+/-[=](Scalar) for consistency with other objects.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -18,6 +18,7 @@
  * Removed deprecated GramSchmidtAlgorithm, ChebychevAlgorithm classes
  * Removed [gs]etOptimizationSolver methods
  * Removed CovarianceModel::compute{AsScalar,StandardRepresentative} overloads
+ * Removed Sample Scalar translation operators
 
 === Python module ===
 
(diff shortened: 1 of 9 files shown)


sofianehaddad:
@jschueller why ? For consistency, we can overload other objects ?

jschueller:
I'm not sure if we should allow it, I mean what the meaning of scalar+matrix, that's not part of the algebra.

sofianehaddad:
@jschueller scalar + Matrix --> Matrix, scalar + CovarianceMatrix --> SymmetricMatrix
IMHO, for the easy use, we should keep the translation operators

dbarbier:
There are other issues:
 * Sample.add and Sample.operator+ have different meanings, which is very confusing
 * Sample.operator+(Point) does not make more sense than with a Scalar
 * Matrix has no in-place operator

On the other hand, we could state that operations with Scalar are performed element-wise for any scalar container, like numpy.

jschueller:
Solution doesn't seems obvious, leaving this for later, at next CT ?

PR #511, opened 2017-07-04
Enhance unity build

jschueller:
Instead of adding XXX_unity targets, the new macro ot_add_library adds an unity target with the same name.
This integrates cleanly with other rules while removing the need to change the build workflow (make python_unity && make install/fast).


jschueller:
seems I cannot enable it in appveyor yet ; I get an error in dsfmt.cxx even if I exclude it from cotire:
https://ci.appveyor.com/project/openturns/openturns/build/1.0.832/job/lggubimolnpgpnj4

regislebrun:
+1

PR #512, opened 2017-07-05
Fix BayesDistribution::getMarginal

jschueller:
The indice value was not taken into account.

I copied the test from conditional distribution, the covariance cannot be computed because the generic service requires getMarginal with indices in both conditioned and conditioning distributions, so there may be still room for improvement.


regislebrun:
+1

PR #514, opened 2017-07-06
Remove Sampler class

jschueller:
Rationale: Sampler has the same interface as RandomVector (getRealization/getSample),
so MCMC can inherit RandomVector.

This makes PosteriorRandomVector useless.

Also MCMC now sets the description of the prior when getSample is called (that was my original goal).


regislebrun:
+1

PR #516, opened 2017-07-07
EGO: do not reevaluate model without noise

jschueller:


dbarbier:
outputSample must indeed be added to KrigingResult.
In EGO, I do not understand why minimumDistance is a Point, it should be a Scalar; create a KDTree and call getNearestNeighbour() to find the minimum distance.
Loop at l.252-264  looks bogus to me, you reevaluate mk, sk2 and u at the same points over and over, only the new inserted point is relevant. And you should initially evaluate the metamodel on inputSample, not on each point.
Likewise for ExpectedImprovementEvaluation(Sample), it should evaluate the model on the whole sample, not on each point.
By the way, I cannot comment more, EfficientGlobalOptimization::run() is too difficult to read for me.

jschueller:
- The criteria on the minimum distance is per component as we compare to the correlation length of the kernel.

dbarbier:
That does not make sense to me, iterations stop as soon as a point is found near from other points along one dimension, but the minimum distance along other dimensions may still be huge.
If you have different correlation lengths, use normalized Euclidian distance: d(x,y)=sum_i((x_i-y_i)^2/s_i^2).  But I am pretty sure that you can normalize sample and deal with usual Euclidian distance.

jschueller:
We want to consider the metamodel unfit to add treat new points as soon as the the correlation is smaller than the minimum distance over at least one component because that means the model is noisy, and the EGO formulation is not adapted (but AEI is).

jschueller:
- I evaluated the model on the whole design at once.
Also we did not evaluate the updated model that's why it looked like it could be evaluated on the design only once.

jschueller:
Thanks for the review.

dbarbier:
Performances are getting worse, now you compute a potentially large covariance matrix but use only its diagonal.  Currently KrigingResult::getConditionalCovariance(Point) creates a Sample of size 1 and performs complex computations; if I am not mistaken, matrix dimension is 1 in this case, and I am pretty sure than these computations can be greatly simplified when working on a single point.

If you want better performances, you have to ensure that computations are performed only once, and thus store intermediate results (at least mx and sk2 diagonal) and invalidate them when kriging is rebuilt.

BTW please reconsider @sofianehaddad suggestion in #430 to split EGO into two classes, at least the EI-case should be much simpler.

PS: there are typos in [gs]etAEITradeoff methods (AIE -> AEI).

jschueller:
Yes, but there's no method to compute only the diagonal part isn't ?

regislebrun:
@jschueller @dbarbier Is it worth the effort to implement a multi-point marginal variance method, like the computeVariance() method of the Sample class? Should it be in the KrigingResult class or in the EGO class? In the very specific case of a 1D model, it can be implemented much more efficiently than in the general case as done in the KrigingResult class.

jschueller:
@regislebrun otpod would need that variance too, for 1d models too.

dbarbier:
@regislebrun No idea, I am afraid that I cannot help here.

sofianehaddad:
@regislebrun no idea about the efforts but as it is used both by EGO/OTPOD and other algo classes, it should be implemented in KrigingResult

PR #518, opened 2017-07-17
Debug osx build

jschueller:


sofianehaddad:
@jschueller the problem is not tbb but R
Have a look at #519

PR #520, opened 2017-07-19
[viewer] Let matplotlib define the bounding box

jschueller:
When plotting Cloud drawables, the bounding box is set to match exactly the
data and the points are not completely visible.
We let matplotlib draw its nice 5% margin (default) for all graphs except for
polygons where the bounding box stays [0,1]^2

import openturns as ot
from openturns.viewer import View

levels = [1.0, 1.5, 3.0]
experiment = ot.Axial(2, levels)
sample = experiment.generate()
View(ot.Cloud(sample)).show()

### DIFF
--- a/python/src/viewer.py
+++ b/python/src/viewer.py
@@ -197,12 +197,6 @@ class View(object):
         if (graph.getLogScale() == ot.GraphImplementation.LOGY) or (graph.getLogScale() == ot.GraphImplementation.LOGXY):
             axes_kwargs.setdefault('yscale', 'log')
 
-        # set bounding box
-        axes_kwargs.setdefault(
-            'xlim', [graph.getBoundingBox()[0], graph.getBoundingBox()[1]])
-        axes_kwargs.setdefault(
-            'ylim', [graph.getBoundingBox()[2], graph.getBoundingBox()[3]])
-
         # set figure
         if figure is None:
             if len(axes) == 0:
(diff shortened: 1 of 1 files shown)


dbarbier:
Not sure whether this is a good idea, you can no more set bounding box via axes_kwargs argument.

jschueller:
axes_kwargs is still used so you can still define it by setting xlim & ylim.

dbarbier:
You call set_xlim for Polygon and PolygonArray after axes_kwargs is being used, thus user settings are discarded.

regislebrun:
From my point of view, the Graph class is responsible for the bounding box. We could add a setMargin(const Scalar percent) method to ease the modification of the bounding box, but:
+ the bounding box management should not differ depending on the kind of drawables present into a graph
+ the ability to vizualize a Drawable directly is and should remain a shortcut for putting it into a Graph then vusualize the graph.

jschueller:
@denis, oh you're right, sorry, fixed

@regislebrun I think it's a better idea to let the graphing lib manage the bounding box according to the data present, as it is clever to draw the right ticks etc, besides that's consistent with what the R backend does.

jschueller:
On the other hand, methods that return graphs (SobolIndicesAlgorithmImplementation, DistributionImplementation, VisualTest, ...) also try to override the bounding box.

PR #521, opened 2017-07-20
Bounds evaluation in DsitributionFactory classes

sofianehaddad:
Some factory classes (UniformFactory/BetaFactory) require evaluation of both lower/upper bounds.

The implementation is improved to avoid potential problems like in http://trac.openturns.org/ticket/890. Only Uniform & Beta classes are impacted


regislebrun:
+1

PR #522, opened 2017-07-20
Use of flex/bison for parsing CSV in travis

sofianehaddad:
We enforce bison/flex in travis. This prepares work on the conda packages


regislebrun:
+1

PR #523, opened 2017-07-24
Fix http://trac.openturns.org/ticket/897

sofianehaddad:
Bug was introduced in https://github.com/regislebrun/openturns/commit/9ea891af251523e0ce599c59ed354901acc42f74


regislebrun:
+1

PR #525, opened 2017-07-25
Make MatrixImplementation::isPositiveDefinite const

dbarbier:
Drop its argument.

No code actually calls this method with keepIntact=false, and this is
normal since it does not make sense to overwrite the matrix.


regislebrun:
+1

dbarbier:
Branch rebased due to conflict in ChangeLog

PR #526, opened 2017-07-25
Let ProcessSample::getField() method throw OutOfBoundException

dbarbier:
This method is called from `ProcessSample::__getitem__`.
OTexception.i transforms OutOfBoundException into IndexError,
which is needed by Python to end its for loops.
Fix http://trac.openturns.org/ticket/896

For consistency reason, the same change is applied to setField and operator[].

In `Field::__getitem__`, call `at()` method instead of `getValueAtIndex()`;
the former checks its argument and will thus generate an IndexError.
Copy `FieldImplementation::__setitem__` into Field class, there is no reason
to have different implementations.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -28,6 +28,7 @@
 === Bug fixes ===
  * #890 (Cannot build triangular distribution)
  * #891 (Viewer issue with Pairs drawables)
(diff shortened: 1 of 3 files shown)


regislebrun:
+1

dbarbier:
Branch rebased due to conflicts in ChangeLog.

PR #527, opened 2017-07-25
Use of cotire in travis (linux)

sofianehaddad:


regislebrun:
+1

dbarbier:
Maybe we could reenable C++ checks now on travis?

sofianehaddad:
@dbarbier I am trying but we already reached the limit time

PR #528, opened 2017-07-26
Fixed ticket #899

regislebrun:
The SklarCopula and MarginalDistribution classes were not taking into account the isParalel flag of their underlying distribution

### DIFF
--- a/lib/src/Uncertainty/Model/SklarCopula.cxx
+++ b/lib/src/Uncertainty/Model/SklarCopula.cxx
@@ -61,6 +61,8 @@ SklarCopula::SklarCopula(const Distribution & distribution)
   , marginalCollection_(distribution.getDimension())
 {
   setName( "SklarCopula" );
+  // Manage parallelism
+  setParallel(distribution.getImplementation()->isParallel());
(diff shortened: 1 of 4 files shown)


lib/src/Uncertainty/Model/SklarCopula.cxx
```diff
@@ -155,12 +157,12 @@ Scalar SklarCopula::computePDF(const Point & point) const
 {
   const UnsignedInteger dimension = getDimension();
   // Early exit for the independent case
-  if (distribution_.hasIndependentCopula()) return IndependentCopula(dimension).computePDF(point);
   for (UnsignedInteger i = 0; i < dimension; ++i)
   {
     const Scalar ui = point[i];
     if ((ui <= 0.0) || ui > 1.0) return 0.0;
```
jschueller: the code of IndependentCopula::computePDF threats the case were x==1.0:
```if ((x <= 0.0) || (x >= 1.0)) return 0.0;```

jschueller:
perhaps rebasing to include #527 will fix the travis build ?

jschueller:
please update the changelog

PR #530, opened 2017-07-31
Reenable cxx tests in travis

sofianehaddad:
The objective here is to reenable c ++ tests on travis since sphinx documentation is disabled.

We speed up also some builds & disable useless install (sphinx, numpydoc, matplotlib, texlive)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
     - cmake
     - bison
     - flex
+    - swig
     - libmuparser-dev
     - liblapack-dev
     - libxml2-dev
@@ -17,14 +18,11 @@ addons:
     - python-dev
     - python-scipy
     - python-matplotlib
-    - texlive-latex-recommended
-    - texlive-fonts-recommended
-    - texlive-latex-extra
 
 matrix:
   include:
     - os: linux
-      compiler: gcc
+      compiler: clang
     - os: osx
       compiler: clang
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -95,7 +95,13 @@ script:
...
               -DCOTIRE_MAXIMUM_NUMBER_OF_UNITY_INCLUDES="-j2"
               . &&
-        make install -j2 &&
+        pushd lib &&
+        make -j4 &&
+        popd &&
+        pushd python &&
+        make -j2 &&
+        popd &&
+        make install -j4 &&
```
jschueller: a bit complicated ; ```make OT -j4 && make install -j2``` ?

jschueller:
also we could avoid recompiling swig as it's already done in circleci

.travis.yml
sofianehaddad: Unfortunatly,
> make -j4 

segfaults.  A better solution seems keeping
> make -j2

.travis.yml
```diff
@@ -17,14 +17,11 @@ addons:
...
     - python-matplotlib
-    - texlive-latex-recommended
-    - texlive-fonts-recommended
-    - texlive-latex-extra
 
 matrix:
   include:
     - os: linux
-      compiler: gcc
+      compiler: clang
```
dbarbier: Please keep gcc here, it is a good idea to test different compilers

sofianehaddad: Done

jschueller:
Please just delete commit 7e42 instead of reverting.

sofianehaddad:
@jschueller at the end I will squash commits

.travis.yml
sofianehaddad: It seems gcc is much slower and thus limit time is reached. As circle-ci relies on gcc, I moved to clang

jschueller: I guess this is fine as it was the same version as in circleci anyways

PR #531, opened 2017-07-31
Minor swig improvement

sofianehaddad:
This PR performs some {set,get}item methods : index in python should be in [-size, size-1].

The current implementation throws an OutOfBoundException in case index not in the previous interval. However there is a confusion between SignedInteger/UnsignedInteger and thus leads to a *strange* message error, which is fixed here.

### DIFF
--- a/python/src/Field.i
+++ b/python/src/Field.i
@@ -20,8 +20,12 @@ Field(const Field & other)
 
 Point __getitem__ (SignedInteger index) const
 {
+  OT::UnsignedInteger size = self->getSize();
   if (index < 0) {
     index += self->getSize();
+  }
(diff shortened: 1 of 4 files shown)


sofianehaddad:
@jschueller I think msg lenght's is <= 50

jschueller:
I count 85

sofianehaddad:
@jschueller you talk about the cxx or msg output? By the way, it is updated

jschueller:
oh, I was talking about the git commit message:
```Perfom __{get,set}item__ index checking :ignedInteger should be in [...]```

sofianehaddad:
oh, ok

PR #534, opened 2017-08-10
Add content to example/theory documentation

jschueller:
- Mostly import content from first section (step B: quantification of the uncertainty sources) of latex Reference guide into the new theoric sphinx doc section, mostly under "data analysis" subsection (typos in screen are fixed):
theory
- Added/enriched corresponding notebook examples from UC guide
- Reorganized the new theoric section into the 6 site themes
- Documented estimators of parametric models factories in their docstrings
- Fixes http://trac.openturns.org/ticket/454, http://trac.openturns.org/ticket/605


jschueller:
@adutfoy could you have a look and approve if it's ok, or suggest changes if necessary?

adutfoy:
+1 good job!

PR #536, opened 2017-08-17
Add Graph margin

jschueller:
Another solution for https://github.com/openturns/openturns/pull/520:
we add a margin at the cloud drawable level.

edit: at the graph level

### DIFF
--- a/lib/src/Base/Common/ResourceMap.cxx
+++ b/lib/src/Base/Common/ResourceMap.cxx
@@ -359,6 +359,8 @@ void ResourceMap::loadDefaultConfiguration()
   setAsUnsignedInteger( "Graph-DefaultWidth", 640 );
   setAsUnsignedInteger( "Graph-DefaultHeight", 480 );
   setAsScalar( "Graph-DefaultLegendFontSize", 1.0 );
+  setAsScalar( "Graph-DefaultHorizontalMargin", 0.05 );
+  setAsScalar( "Graph-DefaultVerticalMargin", 0.05 );
 
   // Contour parameters //
   setAsUnsignedInteger( "Contour-DefaultLevelsNumber", 10 );
--- a/lib/src/Base/Graph/GraphImplementation.cxx
+++ b/lib/src/Base/Graph/GraphImplementation.cxx
@@ -574,14 +574,14 @@ void GraphImplementation::clean()
 }
 
 /* Margin accessor */
-void GraphImplementation::setHorizontalMargin(const Scalar xMargin)
+void GraphImplementation::setXMargin(const Scalar xMargin)
 {
   if (!(xMargin >= 0.0) || !(xMargin <= 1.0))
     throw InvalidArgumentException(HERE) << "Horizontal margin must be in [0, 1].";
   xMargin_ = xMargin;
 }
 
-void GraphImplementation::setVerticalMargin(const Scalar yMargin)
+void GraphImplementation::setYMargin(const Scalar yMargin)
 {
   if (!(yMargin >= 0.0) || !(yMargin <= 1.0))
     throw InvalidArgumentException(HERE) << "Vertical margin must be in [0, 1].";
--- a/lib/src/Base/Graph/GraphImplementation.cxx
+++ b/lib/src/Base/Graph/GraphImplementation.cxx
@@ -74,6 +74,8 @@ GraphImplementation::GraphImplementation(const String & title)
   , logScale_(NONE)
   , showGrid_(false)
   , gridColor_("gray")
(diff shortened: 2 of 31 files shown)


lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -337,6 +337,10 @@ void ResourceMap::loadDefaultConfiguration()
   // UniVariatePolynomialImplementation parameters //
   setAsUnsignedInteger( "UniVariatePolynomial-SmallDegree", 400 );
```
sofianehaddad: Good idea but what are the units ? pourcent ? inch ? Maybe we could add a "unit" suffix to the RM keys

Btw, it should be referenced in docstrings

lib/src/Base/Graph/Cloud.cxx
```diff
@@ -112,6 +112,22 @@ String Cloud::__repr__() const
...
 
+/* Accessor for boundingbox */
+Cloud::BoundingBox Cloud::getBoundingBox() const
+{
+  BoundingBox boundingBox(BoundingBoxSize);
+  const Point max(data_.getMax());
+  const Point min(data_.getMin());
+  const Scalar x_margin = ResourceMap::GetAsScalar("Cloud-HorizontalMargin");
+  const Scalar y_margin = ResourceMap::GetAsScalar("Cloud-VerticalMargin");
+  const Point delta(max - min);
```
sofianehaddad: xMargin/yMargin (see OT coding rules)
Btw we should check that values are not negative

lib/src/Base/Graph/Cloud.cxx
```diff
@@ -112,6 +112,22 @@ String Cloud::__repr__() const
...
+  boundingBox[0] = min[0] - x_margin * delta[0];
+  boundingBox[1] = max[0] + x_margin * delta[0];
+  boundingBox[2] = min[1] - y_margin * delta[1];
+  boundingBox[3] = max[1] + y_margin * delta[1];
+  return boundingBox;
```
sofianehaddad: I dont know if it is the best way; what is the issue if delta is biggest and there is only a few points ?

jschueller: That should be fine.

lib/src/Base/Graph/Cloud.cxx
jschueller: ok

lib/src/Base/Common/ResourceMap.cxx
jschueller: ok

jschueller:
@sofianehaddad about rescale, the graph bounding box already takes care of all the drawables bounding boxes; it is updated each time a drawable is added.

regislebrun:
@jschueller @sofianehaddad To make things short, I totally disagree with the way you address the problem. A Drawable is responsible for the presentation of its own data and no more. A Graph is responsible for the axes and the bounding box. So any new service regarding the bounding box management should be done at the Graph level. It can of course be complemented by a bounding box management at the back-end level. Concerning the ticks, it should be delegated to the axes, but we don't have such a concept in OT for now. Matlab (and so matplotlib) has done a great job and we should definitely look at their solution.

jschueller:
I added x/y margins attributes at the Graph level.
Graph::getBoundingBox returns the bounding box plus the margins.
It simplifies several draw methods that wanted to add margins in the same fashion.
A few other DistributionImplementation draw methods wanted to set a custom y margin depending on a quantile computation, for these we have to set y margin to 0 prior to calling getBoundingBox in order not to add some margins twice.

lib/src/Base/Graph/GraphImplementation.cxx
```diff
@@ -569,6 +573,21 @@ void GraphImplementation::clean()
   }
 }
 
+/* Margin accessor */
+void GraphImplementation::setHorizontalMargin(const Scalar xMargin)
+{
+  if (!(xMargin >= 0.0))
```
sofianehaddad: As you consider {x,y}Margin in pourcent, we should also check that it is < 1

jschueller: yes

jschueller:
@sofianehaddad we dont have 3d drawables and this is not the object of the PR

jschueller:
@regislebrun 
- now the BoundingBox attribute is defined by an Interval object
- we dont even need to manually override the bounding box in distributions ; using the margin defined by a 5% ratio on each side seems fine, in fact this is only used in BipartiteGraph::draw.

regislebrun:
+1

PR #537, opened 2017-08-23
Fix http://trac.openturns.org/ticket/905

sofianehaddad:
MaternModel::setFullParameter should update **all internal parameters** and
check the size/accuracy of the input argument  (input parameter should be of size spatialDimension_ + dimension_ +1 for the nu_ parameter)

It fixes some issues with the Kriging calibration


regislebrun:
+1

PR #538, opened 2017-08-24
Remove unused variables from openturns-config

dbarbier:
They prevent reproducible builds.


jschueller:
Maybe we could get rid of openturns-config ? It was useful when the modules used autotools, but now ?

regislebrun:
+1

dbarbier:
@jschueller I agree, openturns-config looks useless to me.

PR #539, opened 2017-08-25
Various build system changes

jschueller:


jschueller:
@jeromerobert seems https://github.com/jeromerobert/hmat-oss/commit/699b22f6baa086e4b76b7896a6824faf9ace24e8 broken compilation with older gcc

dbarbier:
@jeromerobert Fix committed, @jschueller please restart your builds.

PR #541, opened 2017-08-30
Fix Function._exec_sample for exotic platforms

dbarbier:
On some platforms, Python multiprocessing module is not working.  In this case, gracefully fallback to sequential execution.


dbarbier:
See https://buildd.debian.org/status/fetch.php?pkg=openturns&arch=hurd-i386&ver=1.9-3&stamp=1503925275&raw=0

regislebrun:
+1

PR #544, opened 2017-09-07
ProbabilitySimulation to handle general events

jschueller:
ProbabilitySimulation was able to handle only composite events (Y=f(X)) now added a ctor without experiment that handles general events. Also reverts the Simulation change to make it abstract.

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Simulation/ProbabilitySimulationAlgorithm.cxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/ProbabilitySimulationAlgorithm.cxx
@@ -36,6 +36,23 @@ CLASSNAMEINIT(ProbabilitySimulationAlgorithm);
 
 static const Factory<ProbabilitySimulationAlgorithm> Factory_ProbabilitySimulationAlgorithm;
 
+ProbabilitySimulationAlgorithm::ProbabilitySimulationAlgorithm(const Bool verbose,
+                                             const HistoryStrategy & convergenceStrategy)
+  : Simulation(verbose, convergenceStrategy)
+  , isExperimentProvided_(false)
+{
+  // Nothing to do
+}
+
+/* Constructor with parameters */
+ProbabilitySimulationAlgorithm::ProbabilitySimulationAlgorithm(const Event & event,
+                                             const Bool verbose,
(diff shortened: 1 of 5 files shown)


lib/src/Uncertainty/Algorithm/Simulation/ProbabilitySimulationAlgorithm.cxx
```diff
@@ -36,23 +36,33 @@ CLASSNAMEINIT(ProbabilitySimulationAlgorithm);
...
 ProbabilitySimulationAlgorithm::ProbabilitySimulationAlgorithm(const Event & event,
-                                             const WeightedExperiment & experiment,
                                              const Bool verbose,
                                              const HistoryStrategy & convergenceStrategy)
   : Simulation(event, verbose, convergenceStrategy)
+  , isExperimentProvided_(false)
 {
-  if (!getEvent().isComposite()) throw InvalidArgumentException(HERE) << "ProbabilitySimulationAlgorithm requires a composite event";
-  setExperiment(experiment);
 }
```
regislebrun: // Nothing to do

lib/src/Uncertainty/Algorithm/Simulation/ProbabilitySimulationAlgorithm.cxx
```diff
@@ -89,6 +99,13 @@ String ProbabilitySimulationAlgorithm::__repr__() const
...
 {
+  if (isExperimentProvided_)
+    return computeBlockSampleComposite();
+  return event_.getSample(blockSize_);
+}
+
+Sample ProbabilitySimulationAlgorithm::computeBlockSampleComposite()
+{
   Point weights;
   const Sample inputSample(experiment_.generateWithWeights(weights));
```
regislebrun: Maybe a (parallel) contains() method acting on Sample could be added to the Domain class? It can speed-up a lot the case where the domain is a complex mesh.

regislebrun:
-1 Either this class is renamed ProbabilitySimulationAlgorithm or we drop the 'Algorithm' suffix to all the other algorithm classes (ie those with a run() method and a dedicated Result class)

jschueller:
@regislebrun it's still named ProbabilitySimulationAlgorithm

regislebrun:
+1

PR #546, opened 2017-09-12
Fix bug 902 (NormalGamma constructor builds wrong link function)

regislebrun:


lib/src/Uncertainty/Distribution/NormalGamma.cxx
```diff
@@ -69,11 +66,11 @@ NormalGamma::NormalGamma(const Scalar mu,
...
-  OSS oss;
-  oss.setPrecision(17);
-  formulas[0] = String(oss << mu_);
-  oss.clear();
-  formulas[1] = String(oss << "1.0 / sqrt(" << kappa_ << "*lambda)");
+  formulas[0] = String(OSS(true) << mu_);
+  if (kappa_ == 1.0)
+    formulas[1] = String(OSS(true) << "1.0 / lambda");
+  else
+    formulas[1] = String(OSS(true) << "1.0 / (" << kappa_ << "*lambda)");
```
jschueller: Maybe the sqrt got lost here ?

jschueller:
it needs a rebase

dbarbier:
I do not like your 2nd commit (f7d91fbe); using a different OSS instance is only a workaround, you really fixed this bug in your 3rd commit (236e285).  Moreover,  ```oss.setPrecision(17)``` has been lost, is this a problem?

IMO f7d91fbe should be amended to only simplify formula when kappa is 1.  And in this case, there is no need for OSS, simply write ```formulas[1] = String("1.0 / sqrt(lambda)");```.

regislebrun:
@jschueller I am currently writing a test... And I found additional problems, one of them being the default implementation of computeCDF() not able to compute the CDF in a reliable way for this peaky distribution. So I developed a specific algorithm for this task and I am currently testing it.
@dbarbier I agree with you regarding the fact that OSS() is useless when kappa_==1. For the other points:
  * The oss.setprecision(17) is no more needed. With OSS(true), we get the full numerical precision for doubles.
  * Many discussions about the reuse of a string stream vs the instanciation of a new one are, in my understanding, lead to the conclusion that a new instanciation is safer and no more expansive than a reuse, so I switched to a new instanciation. It was the correct way to implement what I had in mind, so I identified this commit as the bug fix for NormalGamma
  * Concerning the clear() method of OSS, in a STL perspective there was no bug (OSS and ostringstream had the exact same behavior) bug in an OT perspective it was not the expected behavior. So I changed this behavior (and could have wrote that it was a bug fix for OSS).

I have some significant changes to commit regarding the NormalGamma distribution. I will make the requested changes after these commits.

dbarbier:
> The oss.setprecision(17) is no more needed. With OSS(true), we get the full numerical precision for doubles.

Hmmm, does not look so to me:
```
#include <openturns/OSS.hxx>
#include <iostream>

using namespace OT;

int main()
{
  const Scalar d=1.234567890123456789;
  std::cout << "OSS(true): " << String(OT::OSS(true) << d) << std::endl;
  std::cout << "OSS(false): " << String(OT::OSS(false) << d) << std::endl;
  OSS oss;
  oss.setPrecision(17);
  std::cout << "setPrecision: " << String(oss << d) << std::endl;
  return 0;
}
```

Output:
```
OSS(true): 1.23457
OSS(false): 1.23457
setPrecision: 1.2345678901234567
```

sofianehaddad:
As mentioned by @dbarbier, with OSS(true), we get the numerical precision set in PlatformInfo::GetNumericalPrecision. Maybe your are losing some precision when removing the setPrecision

regislebrun:
@dbarbier @jschueller You are right, the default precision is hard-coded into PlatformInfo and I had in mind that, by default, we were using the full precision. The 'full' flag of OSS is here only to switch between __repr__ and __str__ for the streaming of complex types. I am still struggling with nasty numerical behaviors for this class...

jschueller:
As Denis and Sofiane mentionned, the precision issue is still not resolved here.

The same kind of issue may happen in at least Uniform, if I look for ```OSS(true)```:
```
/* Get the minimum volume level set containing a given probability of the distribution */
LevelSet Uniform::computeMinimumVolumeLevelSetWithThreshold(const Scalar prob, Scalar & threshold) const
{
  const SymbolicFunction function("x", String(OSS(true) << 2.0 / (b_ - a_) << " * abs(x - (" << 0.5 * (a_ + b_) << "))"));
  threshold = prob;
  return LevelSet(function, threshold);
}
```

jschueller:
On mingw, in DistributionImplementation::computeInverseSurvivalFunction, the code path takes the branch "if (leftSurvival > prob)" (leftsurvival and prob are very very close) maybe tail=1 is not taken into account here ?

regislebrun:
@jschueller Thanks for your help. I think I found the bug, the test x > -xBound should have been x < -xBound. I added some traces to check what is happening on appveyor

jschueller:
MinGW test are ok now, waiting for the autobuilder to finish.

PR #547, opened 2017-09-13
Add SobolIndicesExperiment

jschueller:
Moves SobolIndicesAlgorithm::Generate to SobolIndicesExperiment

Available constructors:
    SobolIndicesExperiment(*distribution, size, computeSecondOrder=True*)

    SobolIndicesExperiment(*experiment, computeSecondOrder=True*)

Checks that the undelying experiment has uniform bounds and a distribution with independent copula.


dbarbier:
SobolIndicesExperiment derives from PersistentObject, save/load methods must be defined.

jschueller:
ok, added

sofianehaddad:
@jschueller why do we need a sepcific class ? Are the experiments to be used outside the SobolAlgorithm classes?

jschueller:
Yes

sofianehaddad:
ok, thus +1

regislebrun:
+1

PR #552, opened 2017-09-26
Tighten error history graph Y axis bbox

jschueller:
We dont want the axis to go into negative values, the minimum is very close to zero.

before:
figure_4

after:
figure_4-1


regislebrun:
+1

PR #555, opened 2017-10-02
Add OptimizationAlgorithm::Build(solverName)

jschueller:


regislebrun:
+1 but should be part of an OptimizationAlgorithmFactory class

PR #557, opened 2017-10-04
Fix long long being only available since c++11

jschueller:


dbarbier:
In general I am reluctant to add #ifdef in header files, there may be trouble if you build OT with a compiler and link against it with a different compiler.  Why not simply drop these definitions?  Long long is not used anywhere in OpenTURNS.

jschueller:
Ok, i'll try to remove them.

jschueller:
Seems it wants (u)int64_t to stream Unsigned64BitsInteger types.

dbarbier:
Indeed; if it causes too much trouble, reintroduce your initial changes.

jschueller:
It's the same issue we had in StorageManager.hxx on osx : unsigned long is the same as uint64, so I have to use the same macro to avoid defining twice the same functions on this platform.

regislebrun:
IMO we should use only types with known size to avoid portability troubles as much as possible. Would it be possible to have only uint64_t, int64_t, uint32_t and int32_t, with SignedInteger and UnsignedInteger aliasing int64_t and uint64_t? Long, int and so on are just painful!

PR #558, opened 2017-10-05
Add a few more examples

jschueller:


regislebrun:
@jschueller You mean 'strong maximum test' and not 'string maximum test' in the description of your commit? Also, 'gaussian' should be 'Gaussian' in the description of the field metamodel example.
Otherwise, +1

jschueller:
ok, fixed

PR #567, opened 2017-10-12
Add missing parameter accessors

jschueller:


lib/src/Uncertainty/Distribution/KernelMixture.cxx
```diff
@@ -661,6 +661,55 @@ KernelMixture::PointWithDescriptionCollection KernelMixture::getParametersCollec
...
+Point KernelMixture::getParameter() const
+{
+  const UnsignedInteger dimension = getDimension();
+  const UnsignedInteger size = sample_.getSize();
+  Point parameter;
+  for (UnsignedInteger i = 0; i < dimension; ++ i)
+  {
+    for (UnsignedInteger j = 0; j < size; ++ j)
+    {
+      parameter.add(sample_[j][i]);
```
dbarbier: For performance reasons, always write `sample_(j, i)` instead of `sample_[j][i]`.  Same thing in setParameter.

lib/src/Uncertainty/Distribution/KernelMixture.cxx
```diff
@@ -661,6 +661,55 @@ KernelMixture::PointWithDescriptionCollection KernelMixture::getParametersCollec
...
+Description KernelMixture::getParameterDescription() const
+{
+  const UnsignedInteger dimension = getDimension();
+  PointWithDescriptionCollection parametersCollection(getParametersCollection());
+  Description parameterDescription;
+  for (UnsignedInteger i = 0; i < dimension; ++i)
+  {
+    parameterDescription.add(parametersCollection[i].getDescription());
+  }
+  return parameterDescription;
```
dbarbier: It would be much more efficient to do it the other way, move lines from getParametersCollection to set description, and call getParameterDescription from getParametersCollection.

lib/src/Uncertainty/Distribution/KernelMixture.cxx
jschueller: Ok, I guess it doesnt matter if the order is not consistent with getParametersCollection

lib/src/Uncertainty/Distribution/KernelMixture.cxx
jschueller: As I change the order I dont want to mix, so I rewrote getParameterDescription from scratch.

dbarbier: I do not get it, getParametersCollection must call getParameterDescription, not the other way.

jschueller: I rewrote getParameterDescription to not call getParametersCollection.
getParametersCollection cannot call getParameterDescription as the order is different now because I changed the order of the i/j loops.

dbarbier:
I give up, after looking at source code (OT, not this PR) for 1 hour I am still unable to understand the logic behind getParameter/getParametersCollection.

jschueller:
getParameter gives you a flattened list of the unique parameters.
getParametersCollection gives the parameters of each marginal plus the parameters of the copula if any.
As for the relationship between the two, in 1-d they are equivalent so only getParameter has to be overloaded, but in n-d you have to reimplement both. And in that case I think there's no rule as which one should call the other.

dbarbier:
Thanks, but then why does DistributionImplementation::getParametersCollection return a collection of size 1 and not dim+1?  I expect that `setParametersCollection(getParametersCollection())` does not modify objects, and never throws.
BTW Wishart is bogus, it returns an empty list.

jschueller:
I think that DistributionImplementation::getParametersCollection is only valid in 1-d and should throw otherwise.

regislebrun:
@jschueller @dbarbier getParametersCollection() has been introduced a long time ago in order to compute FORM sensitivities. We had to separate the sensitivity wrt the marginal parameters and the dependence structure, so for a multidimensional model it is a collection of size dim+1. For dimension 1, there is no dependence parameter and it was convenient at one point to return a collection of size 1. It was a mistake and it should have been a collection of size 2 with an empty Point in the last position.
Note that some parameters of a multidimensional distribution can act on both the marginals and the copula, eg the nu parameter of the multivariate Student distribution.
The parameter management as we know it in the recent work (reparameterization, generic likelihood-based or moment-based estimation) has another purpose: it is a way to flatten all the parameters into a unique collection of Scalar instead of eg a Point and a CovarianceMatrix.

I agree with Denis: setParameter(getParameter()) and setParameterscollection(getParametersCollection()) should leave the distribution unchanged. One way to achieve this is to compare the input parameter/parameters collection to the current one before to make any modification. A special care must be paid to the inherited attributes when doing things like *this=MyClass(...), for distributions it is limited to its weight (otherwise it brokes the RandomMixture) but it is fragile.

dbarbier:
Okay, I get it now, thanks.

dbarbier:
@regislebrun To clarify, I did not request to add checks, I was only saying that it should be clear for anyone reading source code that `setParametersCollection(getParametersCollection())` is a no-op, which is not clear at all when only one of these methods is defined, the other one being inherited.

jschueller:
My issue was that now getParameter is throwing if not implemented (https://github.com/openturns/openturns/pull/546) getParameterDimension throws too now for the specific distributions missing getParameter. I propose to square things up in another PR.

lib/src/Uncertainty/Distribution/KernelMixture.cxx
```diff
@@ -661,6 +661,64 @@ KernelMixture::PointWithDescriptionCollection KernelMixture::getParametersCollec
...
+        description.add(OSS() << "x_" << i << "^" << j);
+    else
+      description.add(OSS() << "x_" << i);
+  }
+
+  if (dimension > 1)
+    for (UnsignedInteger j = 0; j < dimension; ++ j)
+      description.add(OSS() << "h_" << j);
+  else
+    description.add(OSS() << "h");
```
dbarbier: Could be simplified to `description.add("h")`, but not a big deal.

dbarbier:
@jschueller I agree, we can polish things in another PR and fix breakage for now.

lib/src/Uncertainty/Distribution/KernelMixture.cxx
jschueller: ok, done

lib/src/Uncertainty/Distribution/Wishart.cxx
```diff
@@ -266,7 +266,7 @@ Wishart::PointWithDescriptionCollection Wishart::getParametersCollection() const
   point[index] = nu_;
   description[index] = "nu";
   point.setDescription(description);
-  PointWithDescriptionCollection parameters(point.getDimension());
+  PointWithDescriptionCollection parameters(1, point);
   return parameters;
 }
```
regislebrun: This method is COMPLETELY WRONG, and I am the only responsible for that! The Wishart distribution is a multidimensional distribution, and as such, the getParametersCollection() method should return a collection of size dim+1. The difficult part is to describe the marginal parameters for such a distribution, as its natural range is over SPD matrices and not vectors. IMO a NotYetImplementedException should be thrown here, or at least a collection of a correct size should be returned. Or you are ready to wait a little bit for a contribution from my side (check your email! ;-))

jschueller: I removed my wishart fix waiting for a better alternative.

PR #568, opened 2017-10-12
Fix build without boost

jschueller:


jschueller:
This function ctor has just been removed, but that code isnt used if boost is present.
We dont have boost on the old linux autobuilder, so it fixes that build.

PR #569, opened 2017-10-13
Fix several Python bindings for methods returning values by reference

dbarbier:
Fix trac 921.
There were several problems:
 * signature of generateWithWeights was wrong in .i files
 * at the end of WeightedExperimentImplementation.i
  %clear OT::Sample & weights;
instead of
  %clear OT::Point & weights;
 * a FixedExperiment constructor has a 'weights' argument

The last one is problematic because constructor with weights was corrupted.
Argument of generateWithWeights has been renamed to weightsOut to avoid
this problem.

All files using typemaps(argout) have been checked to ensure that typemaps are
properly cleaned up, and thus typemap must be added to derived class which
redefine generateWithWeights().

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -107,6 +107,7 @@
  * #914 (KarhunenLoeveQuadratureAlgorithm crashes for covariance models of dimension>1)
  * #917 (Bug in RandomMixture::computeCDF())
  * #919 (Wrong simplification mechanism in MarginalTransformationEvaluation for Exponential distribution)
+ * #921 (Cannot print a FixedExperiment when built from sample and weight)
 
 == 1.9 release (2017-04-18) == #release-1.9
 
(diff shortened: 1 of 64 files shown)


regislebrun:
+1 The problem was much more annoying than no pretty-print: it was impossible to use the experiment to build eg an IntegrationStrategy, needed by Pamphil. Thanks!

dbarbier:
The solution implemented by these commits is fragile, it would be better to automatically handle arguments passed by reference, but this is impossible for now because such arguments are sometimes passed as Python arguments (eg. GaussKronrod::integrate), sometimes returned via a tuple.

jschueller:
please mention http://trac.openturns.org/ticket/921 in changelog (in the 1.10 section as it will be backported)

PR #571, opened 2017-10-16
[TravisCI] Split CXX/Python tests on Linux

jschueller:
Latests PRs seem to timeout because of them

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ install:
 # use latest nlopt
   - git clone https://github.com/stevengj/nlopt.git
   - pushd nlopt
-  - cmake -DCMAKE_INSTALL_PREFIX=$HOME/.local -DBUILD_SHARED_LIBS:BOOL=ON -DBUILD_PYTHON=OFF -DBUILD_OCTAVE=OFF -DBUILD_GUILE=OFF -DCMAKE_MACOSX_RPATH=ON .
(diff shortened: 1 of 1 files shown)


dbarbier:
I disagree, it is always good to keep an eye on C++ tests.

jschueller:
I want to check them too, but the fact is that we often exceed the 50 minutes limit.

regislebrun:
How can we get rid of this 50 minutes limit? Skipping tests because they cost too much does not make sense to me! Is it possible to start several compilation targets, one focusing on the C++ library (with tests) and the other on the Python binding (with tests+docstring tests+notebooks+...)?

jschueller:
Ci tests should always succeed. But instead of skipping tests I can add another job that runs only the cxx tests.

dbarbier:
Maybe we could now reenable gcc on Linux? ;-)

jschueller:
@dbarbier Ok

regislebrun:
+1

.travis.yml
```diff
@@ -22,7 +22,11 @@ addons:
 matrix:
   include:
     - os: linux
-      compiler: clang
+      compiler: gcc
+      env: BUILD_PYTHON=ON
```
sofianehaddad: We could use both 'gcc' and 'clang' with two specific env ?

jschueller: Ok, whatever

jschueller:
Seems the gcc jobs are less reliable: I got failures yesterday.

PR #572, opened 2017-10-16
Performance improvements for ConditionalDistribution

dbarbier:
Call computePDF(Sample) instead of looping over computePDF(Point),
this is much faster.

### DIFF
--- a/lib/src/Uncertainty/Distribution/ConditionalDistribution.cxx
+++ b/lib/src/Uncertainty/Distribution/ConditionalDistribution.cxx
@@ -334,17 +334,20 @@ void ConditionalDistribution::setConditionedAndConditioningDistributionsAndLinkF
       y[diracMarginalsIndices_[i]] = diracValues_[i];
     // Second, the discrete components
     // For each combination of the discrete components
+    Sample currentY(discreteAtomsNumber, y);
     for (UnsignedInteger i = 0; i < discreteAtomsNumber; ++i)
     {
-      Point currentY(y);
       // Get the discrete values
-      const Point discreteNode(discreteNodes_[i]);
       for (UnsignedInteger j = 0; j < discreteDimension; ++j)
-        currentY[discreteMarginalsIndices_[j]] = discreteNode[j];
-      const Scalar w = conditioningDistribution.computePDF(currentY);
+        currentY(i, discreteMarginalsIndices_[j]) = discreteNodes_(i, j);
+    }
+    const Sample yPDF(conditioningDistribution.computePDF(currentY));
+    const Sample parameters(linkFunction_(currentY));
+    for (UnsignedInteger i = 0; i < discreteAtomsNumber; ++i)
+    {
       Distribution dist(conditionedDistribution);
-      dist.setWeight(w);
-      dist.setParameter(linkFunction_(currentY));
+      dist.setWeight(yPDF(i, 0));
+      dist.setParameter(parameters[i]);
       atoms[atomIndex] = dist;
       ++atomIndex;
     } // Discrete measure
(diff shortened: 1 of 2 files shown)


dbarbier:
`t_PosteriorDistribution_std` runs 30% faster on my machine with this commit (15s instead of 22s), I guess that speed-up on computeExpectation itself is quite important.
It has not been fully tested though, this is almost a proof-of-concept.

lib/src/Uncertainty/Distribution/ConditionalDistribution.cxx
```diff
@@ -528,33 +553,42 @@ Point ConditionalDistribution::computeExpectation(const Function & f,
...
     // Skip the current integration point if the current sub-manifold is outside of the integration region
-    if (rejectNode) continue;
-    // Now, complete by the discretization of the continuous part using Gauss integration
-    for (UnsignedInteger j = 0; j < continuousAtomsNumber; ++j)
+    if (!rejectNode)
     {
-      // Complete the filling of theta using the Gauss integration node
-      const Point continuousNode(continuousNodes_[j]);
-      for (UnsignedInteger k = 0; k < continuousDimension; ++k)
-        currentTheta[continuousMarginalsIndices_[k]] = continuousLowerBounds_[k] + 0.5 * (1.0 + continuousNode[k]) * (subPoint[k] - continuousLowerBounds_[k]);
```
dbarbier: @regislebrun I rewrite this portion of code to perform the same computations as before (if there are changes, it is a bug), but I find it very suspicious.  I would expect to reset currentTheta on its value before loop on j, something like
```
 const Point discreteTheta(currentTheta);
 for (UnsignedInteger j = 0; j < continuousAtomsNumber; ++j)
 {
   const Point continuousNode(continuousNodes_[j]);
   for (UnsignedInteger k = 0; k < continuousDimension; ++k)
     discreteTheta[continuousMarginalsIndices_[k]] = continuousLowerBounds_[k] + 0.5 * (1.0 + continuousNode[k]) * (subPoint[k] - continuousLowerBounds_[k]);
     const Scalar w = conditioningDistribution_.computePDF(discreteTheta) * continuousWeights_[j];
   ...
 }
```

regislebrun: @dbarbier Ok I will check this tomorrow. The Bayesian part of the distributions (ConditionalDistribution, BayesDistribution and PosteriorDistribution) is at an early stage of validation, and should be refactored. The main lines of the algorithms are ok in my view, but there is still a lot to do to get something efficient and robust. Thanks for the effort.

regislebrun: @dbarbier The ConditionalDistribution::computeConditionalDistribution() method is strange as a whole. To be honest, I don't even remember its precise role, the comment being much too laconic. A good thing would be to allow the user to choose what kind of WeightedExperiment she want to use for the discretization. The default GaussProductExperiment is good only when the conditioned distribution depends smoothly on its parameters, which is not always the case (eg: a Uniform with random bounds). In this case, a LowDiscrepancyExperiment would do a better job.

sofianehaddad: @regislebrun yes but in an other PR. The objective here is to improve only performance and this improvement is required whatever the used WeightedExperiment

sofianehaddad:
@dbarbier could you please rebase ?

dbarbier:
Right, my point is that we should write code with performance in mind.  We all know that Sample indexing is convenient but may have poor performance, thus we should always favor `Sample(i,j)` over `Sample[i][j]`.  And likewise it is always faster to compute a function on a Sample rather than evaluating it for each Point.

regislebrun:
@dbarbier It looks like it would be a good idea to replace each occurrence of sample[i][j] by sample(i,j) in the library. A quick count using grep -rl '\]\[' lib/src | wc -l gives a number of 168 files to change, which is large. I can do it but I suspect it will be a pain to review then to merge. Do you have an advice on how to manage this kind of change?

regislebrun:
otherwise, +1 for the PR

PR #573, opened 2017-10-16
Chaos enhancements

jschueller:
- Added NormInfEnumerateFunction that enumerates basis terms according to the infinite norm (|a|inf=max a_i) for use with quadrature that is supposed to give exact decomposition for these additional terms compared to linearenumeratefunction
- Moved FunctionalChaosRV sobol services to FCSobolIndices, with added summary method that displays some sensitivity infos, and draw method equivalent to the Sobolindicesalgorithm one
- Added a simplified ctor to OrthogonalProductPolynomialFactory from a list of marginals and that uses StandardDistributionPolynomialFactory internally.


ChangeLog
```diff
@@ -27,6 +28,7 @@
  * Removed deprecated SobolIndicesAlgorithm::Generate method
  * Removed deprecated NumericalScalar, NumericalComplex types
  * Removed deprecated Function constructors
+ * Deprecated FunctionalChaosRandomVector in favor of FunctionalChaosSensitivity
```
regislebrun: @jschueller This class should not be removed, as it is not only a (bad) interface to sensitivity analysis based on chaos. It also provides a sampler based on the orthogonal measure, ie of Y=g(Z) instead of Y=f(T(Z)) where T is the transformation mapping Z into X, where Z is distributed according to the orthogonality measure. It makes a huge difference eg when T relies on RosenblattEvaluation.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSensitivity.cxx
```diff
@@ -47,38 +48,160 @@ FunctionalChaosRandomVector::FunctionalChaosRandomVector(const FunctionalChaosRe
...
+    {
+      for (UnsignedInteger j = basisSize - 1; j >= i; -- j)
+      {
+        if (varianceRatio[varianceOrder[j]] < varianceRatio[varianceOrder[j + 1]])
+        {
+          UnsignedInteger tmp = varianceOrder[j];
+          varianceOrder[j] = varianceOrder[j + 1];
+          varianceOrder[j + 1] = tmp;
+        }
+      }
```
regislebrun: Ouch! Are you REALLY implementing a bubble sort?

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSensitivity.cxx
```diff
@@ -47,38 +48,160 @@ FunctionalChaosRandomVector::FunctionalChaosRandomVector(const FunctionalChaosRe
...
+      oss << st << String(22-st.size(), ' ') << " | ";
+
+      st = OSS() << getSobolTotalIndex(i, m);
+      oss << st << String(22-st.size(), ' ') << "\n";
+    }
+    oss << String(60, '-') << "\n";
+    oss << "\n";
+
+  }
+  return oss;
```
regislebrun: Definitely NO! Too much complex, and too much inefficient. It should rely on getSobolIndex(), should you improve its interface for multiple indices computation (and even introducing parallelism).

lib/src/Base/Func/EnumerateFunctionImplementation.cxx
```diff
@@ -66,7 +66,12 @@ Indices EnumerateFunctionImplementation::operator() (const UnsignedInteger index
 /* The inverse of the association */
 UnsignedInteger EnumerateFunctionImplementation::inverse(const Indices & indices) const
 {
-  throw NotYetImplementedException( HERE ) << "In EnumerateFunctionImplementation::inverse(const Indices & indices) const";
+  UnsignedInteger index = 0;
```
regislebrun: Why not, but can't be the default implementation for NormInfEnumerateFunction as it is terribly inefficient in the general case (and inverse() is almost trivial in the case of NormInfEnumerateFunction)

lib/src/Base/Func/NormInfEnumerateFunction.cxx
```diff
@@ -0,0 +1,161 @@
...
+UnsignedInteger NormInfEnumerateFunction::getMaximumDegreeStrataIndex(const UnsignedInteger maximumDegree) const
+{
+  return maximumDegree;
+}
+
+/* Method save() stores the object through the StorageManager */
+void NormInfEnumerateFunction::save(Advocate & adv) const
+{
+  EnumerateFunctionImplementation::save(adv);
+}
```
regislebrun: No specific attribute to save here?

lib/src/Base/Func/NormInfEnumerateFunction.cxx
```diff
@@ -0,0 +1,161 @@
...
+
+
+/* Method load() reloads the object from the StorageManager */
+void NormInfEnumerateFunction::load(Advocate & adv)
+{
+  EnumerateFunctionImplementation::load(adv);
+}
```
regislebrun: and to load here?

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSensitivity.cxx
```diff
@@ -387,6 +388,43 @@ Scalar FunctionalChaosSensitivity::getSobolGroupedIndex(const Indices & variable
   else return 0.0;
 }
 
+
+
+/** Method that draw the sensitivity graph of a fixed marginal */
+Graph FunctionalChaosSensitivity::draw(UnsignedInteger marginalIndex) const
```
regislebrun: This method should be merged with SobolIndicesAlgorithm::draw(). BTW, it was a really BAD idea to add a draw() method to an algorithm. An algorithm is supposed to produce a result, let's say a SobolIndicesResult, with drawing capabilities and so on. Maybe FunctionalChaosSensitivity should be renamed FunctionalChaosSobolIndices, with an inheritance wrt SobolIndicesResultImplementation. Something has to be refactored in details here.

lib/src/Base/Func/EnumerateFunctionImplementation.cxx
jschueller: I did not found a direct way to compute the inverse, any clues here ?

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSensitivity.cxx
jschueller: For now I only moved the getSobol* methods from FCRV to FunctionalChaosSensitivity.

mbaudin47:
Is this work integrated or not yet, because of the failing tests ?

jschueller:
@mbaudin47 I rebased the commits so as to avoid the red flags you saw. I has not been approved yet.

jschueller:
@regislebrun You objected that there should be a direct way of iterating over the multi-indices of my NormInfEnumerateFunction, ie on the indices located at the oriented "peel" of an hypercube. I cant seem to find a better method than to iterate on the cube multi-indices and filter out indices not at the "peel", with some optimizations.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -0,0 +1,362 @@
...
+  return a.second > b.second;
+}
+
+String FunctionalChaosSobolIndices::summary() const
+{
+  const UnsignedInteger inputDimension = functionalChaosResult_.getDistribution().getDimension();
+  const UnsignedInteger outputDimension = functionalChaosResult_.getMetaModel().getOutputDimension();
+  OSS oss;
+  FunctionalChaosRandomVector randomVector(functionalChaosResult_);
+  CovarianceMatrix covariance(randomVector.getCovariance());
```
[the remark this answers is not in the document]
jschueller: ok

PR #577, opened 2017-10-20
Fixed ticket #924 (Problem with GeneralizedParetoFactory)

regislebrun:
The problem was in the method of exponential regression, where the case of a repeated minimum value in the sample was not correctly managed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -119,6 +119,7 @@
  * #919 (Wrong simplification mechanism in MarginalTransformationEvaluation for Exponential distribution)
  * #921 (Cannot print a FixedExperiment when built from sample and weight)
  * #923 (Fix ExponentialModel::getParameter for diagonal correlation)
+ * #924 (Probleme with the factory of a Generalized Pareto distribution)
  * #927 (Functional chaos is memory hungry)
  * #929 (The labels of the sensitivity analysis graphics are poor)
  * #930 (The getMean method has a weird behavior on parametrized distribution)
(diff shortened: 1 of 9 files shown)


ChangeLog
```diff
@@ -32,7 +32,7 @@
 === Miscellaneous ===
 
 === Bug fixes ===
-
+ * #924 (Probleme with the factory of a Generalized Pareto distribution)
```
jschueller: could you move this to the 1.10 section ? It will be easier for me to backport to the 1.10 branch ?

jschueller:
Side note: the doc adertises the classical method of moments:
"The classical moments method is proposed but not used by default."
but the buildMethodOfMoments method is not available.
Maybe it should be either removed, available with an argument, resourcemap key...

ChangeLog
regislebrun: @jschueller You mean, the 1.10rc1 section? Or do I have to create a new 1.10 section? There are currently only two sections: 1.11 and 1.10rc1.

jschueller: Yes, 1.10rc1

lib/src/Uncertainty/Distribution/GeneralizedParetoFactory.cxx
```diff
@@ -137,8 +149,13 @@ struct GeneralizedParetoFactoryParameterConstraint
...
-      if ((xNm1mJ == xMin) || (xNm2mJ == xMin)) sampleY_[j][0] = (j + 1.0) * SpecFunc::LogMaxScalar;
-      else sampleY_[j][0] = (j + 1.0) * std::log((xNm1mJ - xMin) / (xNm2mJ - xMin));
+      if ((xNm1mJ == xMin) || (xNm2mJ == xMin))
+	{
+	  const String message("Cannot use exponential regression to estimate a GeneralizedPareto distribution when the mininmum value of the sample is repeated");
+	  LOGINFO(message);
+	  throw InvalidArgumentException(HERE) << message;
+	}
+      sampleY_[j][0] = (j + 1.0) * std::log((xNm1mJ - xMin) / (xNm2mJ - xMin));
     }
```
sofianehaddad: sampleY_(j, 0)

jschueller:
super!

PR #578, opened 2017-10-20
Sample::BuildFromTextFile to handle CRLFs

jschueller:
allows to read windows files from unix

### DIFF
--- a/lib/src/Base/Stat/SampleImplementation.cxx
+++ b/lib/src/Base/Stat/SampleImplementation.cxx
@@ -52,6 +52,22 @@
 int csvparse (OT::CSVParserState & theState, yyscan_t yyscanner, FILE * theFile, OT::SampleImplementation &impl, OT::UnsignedInteger & theDimension, const char * separator);
 #endif
 
+namespace {
+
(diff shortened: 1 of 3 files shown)


lib/src/Base/Stat/SampleImplementation.cxx
```diff
@@ -445,6 +445,18 @@ Bool SampleImplementation::ParseComment(const String & line, const String & mark
 }
 
 
+Bool SampleImplementation::GetLine(std::ifstream & file, String & line)
+{
+  const Bool ok = std::getline(file, line);
+  // deal with DOS EOLs from UNIX
+  if (line.size() && (line[line.size() - 1] == '\r'))
+  {
+    line = line.substr(0, line.size() - 1);
```
dbarbier: No need to copy, you can write for instance `line.erase(line.size() - 1, 1)`.

dbarbier: Define this function in an anonymous namespace instead of a static method.

jschueller: ok

PR #579, opened 2017-10-20
[RFC] Parallelize EvaluationImplementation::operator(Sample)

dbarbier:
When playing with `TensorApproximationAlgorithm::run`, I found that 90% of the time is spent in
```
  transformedInputSample_ = transformation_(inputSample_);
```
with large input samples.

This is a proof-of-concept, it has not been tested at all.


jschueller:
cant wait to test that
the new accessors are not defined in EvaluationImplementation.cxx

regislebrun:
Good idea, now it remains to list all the transformations that cannot be evaluated in parallel and set the flag accordingly, or set the default value to false and change it to true only if 100% sure of the parallelization. For example, SymbolicFunction is not parallel in this context, and MarginalTransformationEvaluation is parallel only if all the distributions are parallel. For this specific transformation, one has essentially to choose between the simplification mechanism and the parallelism. Maybe it should be a good thing to separate the harmless simplifications (an affine transformation, coded using a LinearFunction instead of a SymbolicFunction) from the others in this class.

dbarbier:
@regislebrun I prefer to have isParallel=true by default, otherwise we won't find bugs; users can change ResourceMap if there are trouble.
I made the following classes non-parallel:
  * ReducedLogLikelihoodEvaluation
  * MethodBoundEvaluation
  * InverseNatafEllipticalCopulaEvaluation, NatafEllipticalCopulaEvaluation
  * InverseNatafIndependentCopulaEvaluation, NatafIndependentCopulaEvaluation
  * InverseRosenblattEvaluation, RosenblattEvaluation

Method `operator()(const Sample&) const` is implemented for:
  * MarginalTransformationEvaluation
  * TrendEvaluation

jschueller:
Meaning it's ok for Python & muParser ???

dbarbier:
@jschueller SymbolicEvaluation and PythonEvaluation override `operator()(const Sample &)` method, thus `EvaluationImplementation::operator()(const Sample&)` is not called.

dbarbier:
@jschueller I implemented evaluation on a Sample in CanonicalTensorEvaluation, but did not perform any benchmark yet.

lib/src/Base/Func/EvaluationImplementation.cxx
```diff
@@ -162,22 +163,62 @@ Bool EvaluationImplementation::isActualImplementation() const
...
 {
   const UnsignedInteger inputDimension = getInputDimension();
   if (inSample.getDimension() != inputDimension) throw InvalidArgumentException(HERE) << "Error: the given sample has an invalid dimension. Expect a dimension " << inputDimension << ", got " << inSample.getDimension();
 
-  const UnsignedInteger size = inSample.getSize();
-  Sample outSample(size, getOutputDimension());
-  // Simple loop over the evaluation operator based on point
-  // The calls number is updated by these calls
-  for (UnsignedInteger i = 0; i < size; ++i) outSample[i] = operator()(inSample[i]);
+  Sample outSample((isParallel_ && !isCacheEnabled() && !isHistoryEnabled()) ? evaluateParallel(*this, inSample) : evaluateSequential(*this, inSample));
```
regislebrun: @dbarbier Is there any reason why parallelism is activated only if cache and history are disabled? Also, what is the motivation of using free functions instead of methods for the parallel versions of the operator() method? Do we have to generalize this practice to other classes like Distribution?

lib/src/Base/Func/EvaluationImplementation.cxx
```diff
@@ -353,26 +358,6 @@ Point EvaluationImplementation::operator() (const Point & inP) const
   throw NotYetImplementedException(HERE) << "In EvaluationImplementation::operator() (const Point & inP) const";
 }
 
-Point EvaluationImplementation::operator() (const Point & inP,
```
regislebrun: @dbarbier This will break otrobopt. The motivation behind the introduction of all these methods was to provide a symmetrical treatment of arguments and parameters of a function, as the distinction is somewhat arbitrary between them (and context-dependent such as in otbobopt). Do you have an idea on how to speed-up the evaluation of a function at a given point but for a sample of parameters, which is the actual need for otrobopt?

regislebrun:
@dbarbier Good job! I also worked on the transformations yesterday. I will send you a patch and stop working on it in order to avoid duplicate work. The main point is that all the Nataf transformations can be made parallel if the underlying distribution is parallel, and Rosenblatt should work too even if it is not currently the case. For the marginal transformation, it can also be made parallel if no SymbolicFunction is used, but this restriction may be outdated by your recent developments.

lib/src/Base/Func/EvaluationImplementation.cxx
dbarbier: Most implementations of `operator()(Point&)` contains
```
    if (isHistoryEnabled_) {
      inputStrategy_.store(inP);
      outputStrategy_.store(outP);
    }
```
There is no guarantee that `inputStrategy_` and `outputStrategy_` remain consistent when there are multiple threads, one thread may add inP, then another thread adds inP and outP, then the first thread adds outP.

The only solution is to disable history, compute the whole Sample and fill in history if needed.  And it should be done for all classes implementing `operator()(const Sample&)`.

Cache seems to be unused, so I did not try hard to make it work, and decided to be conservative.

IMHO we should favor free function over method member when possible (ie. when do not need access to non-public members), yes.  For instance there is no need to recompile everything when you introduce free function, unlike a member.

lib/src/Base/Func/EvaluationImplementation.cxx
dbarbier: Okay, I was not aware of that, and will think about it.  But only ParametricEvaluation redefined this method, is the gain worth the trouble?  Otherwise we can call setParameter in otropopt.

dbarbier:
For Nataf, it would be better to perform linear algebra on Sample directly, IMO.  This is not trivial, this is why I chose to declare these methods as being non-parallel for now.  For Rosenblatt, you are right, I willl fix it.

lib/src/Uncertainty/Algorithm/Transformation/NatafEllipticalCopula/InverseNatafEllipticalCopulaEvaluation.cxx
```diff
@@ -105,6 +104,32 @@ Point InverseNatafEllipticalCopulaEvaluation::operator () (const Point & inP) co
...
+  SampleImplementation result(size, dimension);
+  result.setData(*resultMatrix.getImplementation());
+  const Distribution standardMarginal(standardDistribution_.getMarginal(0));
+  // Second, apply the commmon marginal distribution
+  for (UnsignedInteger i = 0; i < dimension; ++i)
+  {
+    const Point resultI(standardMarginal.computeCDF(result.getMarginal(i)).asPoint());
+    for (UnsignedInteger j = 0; j < size; ++j)
+      result(j, i) = resultI[j];
+  }
```
regislebrun: @dbarbier Here you can apply standardMarginal.computeCDF() over ALL the values contained in result instead of iterating over the dimensions. Something like:
```
SampleImplementation resultFlatten(size*dimension, 1);
resultFlatten.setData(*resultMatrix.getImplementation());
SampleImplementation result(size, dimension);
result.setData(standardMarginal.computeCDF(resultFlatten).asPoint());
```
It may speed things a little bit if data copy through setData() is faster than SampleImplementation::getMarginal() and pointwise copy.

dbarbier: Thanks, updated.

lib/src/Base/Func/EvaluationImplementation.cxx
dbarbier: PR sent to otrobopt to not use these functions anymore.

jschueller:
This squares many things in the Function API, great job.

I also want to add that custom evaluation classes declared in headers like GeneralLinearModelAlgorithm.hxx should be moved to the definition file.

Do you get a full speedup with the low rank tensor metamodel evaluation ?

dbarbier:
@jschueller No, this work is on hold, because I now believe that adding an `isParallel_` member is not the way to go.  I would like to fix Distribution classes first, but this needs some discussion.

regislebrun:
@dbarbier @jschueller How do you want to have the discussion? On github? On openturns-dev? By phone?

dbarbier:
Multithreading problems come from Functions not being pure. TODO list:
- [ ] Remove history and cache from Function/Evaluation/etc, and move these features into new classes which extend these ones
- [ ] Remove `operator()` with parameters
- [ ] Redesign all Algorithm classes to be pure too; in order to keep backward compatibility, `run()` method becomes a no-op and `getResult()` computes and returns a Result, without caching any state.
- [ ] Redesign Event to handle history itself
- [ ] HistoryStrategy must be redesigned too, it does not make sense to have different strategies for input and output Sample

jschueller:
I understand all of the items except: "Redesign Event to handle history itself", could you explain ?

dbarbier:
@jschueller Item dropped, I do not remember.

dbarbier:
Closing, I won't work on it anymore.

PR #581, opened 2017-10-24
[viewer] Do not ignore contour even if it has one level

sofianehaddad:
This might be useless for minimum volume set

### DIFF
--- a/python/src/viewer.py
+++ b/python/src/viewer.py
@@ -368,9 +368,6 @@ class View(object):
                 self._ax[0].pie(x, **pie_kwargs)
 
             elif drawableKind == 'Contour':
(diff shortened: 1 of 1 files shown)


jschueller:
xref https://github.com/openturns/openturns/pull/562/commits/a866af107e358b60bc762f6be89dbd3718be7fa0

sofianehaddad:
@jschueller the View class is here to "interpret" drawables into matplotlib objects. 
A Contour object with one level is drawn (Graph.draw) but with your fix could not be visualized. It is inconsistent

IMO we should let matplotlib handle its own issues.

jschueller:
see #605

PR #582, opened 2017-10-24
 Multivariate Pearson & Spearman correlations

jschueller:
CorrelationAnalysis::PearsonCorrelation and CorrelationAnalysis::SpearmanCorrelation
can accept a multivariate first sample and return a Point,
for usability and consistency with the other methods.


lib/src/Base/Stat/CorrelationAnalysis.cxx
```diff
@@ -32,32 +32,30 @@ BEGIN_NAMESPACE_OPENTURNS
...
   if (outputSample.getDimension() != 1) throw InvalidDimensionException(HERE) << "Error: output sample must be 1D";
   if (inputSample.getSize() != outputSample.getSize()) throw InvalidArgumentException(HERE) << "Error: input and output samples must have the same size";
-  const UnsignedInteger size = inputSample.getSize();
-  Sample pairedSample(size, 2);
-  for (UnsignedInteger i = 0; i < size; ++i)
+  Sample pairedSample(inputSample);
+  pairedSample.stack(outputSample);
+  const UnsignedInteger dimension = inputSample.getDimension();
+  Point result(dimension);
+  CorrelationMatrix correlation(pairedSample.computePearsonCorrelation());
```
regislebrun: @jschueller Looks like too many correlations are computed here. You compute a full matrix to extract only its last row!

jschueller: ok

regislebrun:
+1

PR #584, opened 2017-10-26
Covariance model: only activate scale & amplitude parameters by default

jschueller:
Consistently only enable scale & amplitude whatever ctor is used
see http://trac.openturns.org/ticket/926


regislebrun:
+1

PR #585, opened 2017-10-27
Set parametrized distribution range

jschueller:
Fixes http://trac.openturns.org/ticket/930


regislebrun:
-1 The actual bug is that computeMean() is not redefined in ParameterizedDistribution. Ths class is just a wrapper around another DistributionImplementation, and as such it should redefine all the methods of a DistributionImplementation and delegate the computation to the corresponding method of the wrapped DistributionImplementation. By doing this for computeRange(), the getMean() method calls the generic implementation of computeMean(), which is able to compute something more or less accurate thanks to a meaningful value for range_, but it cannot be as fast and as accurate as the computeMean() method of the underlying DistributionImplementation.

jschueller:
ok, will do

PR #586, opened 2017-10-30
Fix chaos memory usage

jschueller:
Add accessors to the ProjectionStrategy interface to fix the reference counting in FunctionalChaosAlgorithm.
Fixes http://trac.openturns.org/ticket/927 (Functional chaos is memory hungry)

Note that now the projection strategy passed as argument of FCA is not modified in-place hence the modification to t_FunctionalChaos_ishigami_database.expout
I kept the friendship of ProjectionStrategyImplementation to FCA only for the updateBasis method and not to copy the Proxy at that moment. But this remains a bad design.

I also think that we shouldnt build metamodel algorithms without passing the actual input/output data: in FCA we do lots of complicated stuff to override the data.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -111,6 +111,7 @@
  * #919 (Wrong simplification mechanism in MarginalTransformationEvaluation for Exponential distribution)
  * #921 (Cannot print a FixedExperiment when built from sample and weight)
  * #923 (Fix ExponentialModel::getParameter for diagonal correlation)
+ * #927 (Functional chaos is memory hungry)
  * #930 (The getMean method has a weird behavior on parametrized distribution)
 
 == 1.9 release (2017-04-18) == #release-1.9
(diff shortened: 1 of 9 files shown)


dbarbier:
> I kept the friendship of ProjectionStrategyImplementation to FCA only for the updateBasis method and not to copy the Proxy at that moment.

I do not understand your statement, and also commit fec6867.  IMO DesignProxy must be copied.  Anyway this is more general, this can be discussed for next OT version.

dbarbier:
>  Note that now the projection strategy passed as argument of FCA is not modified in-place hence the modification to t_FunctionalChaos_ishigami_database.expout

Did you try to print `algo_.getProjectionStrategy()` instead?  This should work.

regislebrun:
+1

jschueller:
@dbarbier Indeed, using getProjection fixes this, thanks

PR #587, opened 2017-10-30
Allow multivariate TruncatedDistribution

jschueller:
The internal parameter is now an Interval instead of the scalar/bool attributes for the each of the two bounds.

```
dimension = 2
size = 70
sample = ot.Normal(dimension).getSample(size)
ks = ot.KernelSmoothing().build(sample)
truncatedKS = ot.TruncatedDistribution(ks, ot.Interval([-2.0]*dimension, [2.0]*dimension))
```

truncatedks


regislebrun:
-1 It is an excellent idea to extend the TruncatedDistribution class to multivariate distributions, but the proposed implementation is wrong in many places. A more detailed review is coming soon.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -135,37 +132,30 @@ TruncatedDistribution::TruncatedDistribution(const Distribution & distribution,
...
     const Scalar thresholdRealization)
   : DistributionImplementation()
+  , bounds_(truncationInterval)
 {
   setName("TruncatedDistribution");
-  if (distribution.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: can truncate only distribution with dimension=1, here dimension=" << distribution.getDimension();
   distribution_ = distribution;
   setThresholdRealization(thresholdRealization);
-  setDimension(1);
+  setDimension(distribution.getDimension());
```
regislebrun: We should check the interval dimension against the distribution dimension

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -135,37 +132,30 @@ TruncatedDistribution::TruncatedDistribution(const Distribution & distribution,
...
-  cdfLowerBound_ = distribution.computeCDF(Point(1, lowerBound_));
-  cdfUpperBound_ = distribution.computeCDF(Point(1, upperBound_));
+  cdfLowerBound_ = distribution.computeCDF(bounds_.getLowerBound());
+  cdfUpperBound_ = distribution.computeCDF(bounds_.getUpperBound());
   if (cdfLowerBound_ >= cdfUpperBound_) throw InvalidArgumentException(HERE) << "Error: the truncation interval does not contain a non-empty part of the support of the distribution";
-  pdfLowerBound_ = distribution.computePDF(Point(1, lowerBound_));
-  pdfUpperBound_ = distribution.computePDF(Point(1, upperBound_));
+  pdfLowerBound_ = distribution.computePDF(bounds_.getLowerBound());
+  pdfUpperBound_ = distribution.computePDF(bounds_.getUpperBound());
   normalizationFactor_ = 1.0 / (cdfUpperBound_ - cdfLowerBound_);
```
regislebrun: The normalization factor is the inverse of the probability of the truncation interval, so the formula is wrong for multivariate distributions

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -135,37 +132,30 @@ TruncatedDistribution::TruncatedDistribution(const Distribution & distribution,
...
-  finiteUpperBound_ = truncationInterval.getFiniteUpperBound()[0] == 1;
-  cdfLowerBound_ = distribution.computeCDF(Point(1, lowerBound_));
-  cdfUpperBound_ = distribution.computeCDF(Point(1, upperBound_));
+  cdfLowerBound_ = distribution.computeCDF(bounds_.getLowerBound());
+  cdfUpperBound_ = distribution.computeCDF(bounds_.getUpperBound());
   if (cdfLowerBound_ >= cdfUpperBound_) throw InvalidArgumentException(HERE) << "Error: the truncation interval does not contain a non-empty part of the support of the distribution";
-  pdfLowerBound_ = distribution.computePDF(Point(1, lowerBound_));
-  pdfUpperBound_ = distribution.computePDF(Point(1, upperBound_));
+  pdfLowerBound_ = distribution.computePDF(bounds_.getLowerBound());
+  pdfUpperBound_ = distribution.computePDF(bounds_.getUpperBound());
```
regislebrun: In the general case, we need the PDF values at all the bounding interval vertices

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -135,37 +132,30 @@ TruncatedDistribution::TruncatedDistribution(const Distribution & distribution,
...
   normalizationFactor_ = 1.0 / (cdfUpperBound_ - cdfLowerBound_);
   // Adjust the truncation interval and the distribution range
   computeRange();
-  lowerBound_ = getRange().getLowerBound()[0];
-  upperBound_ = getRange().getUpperBound()[0];
-  finiteLowerBound_ = getRange().getFiniteLowerBound()[0] == 1;
-  finiteUpperBound_ = getRange().getFiniteUpperBound()[0] == 1;
+  bounds_ = getRange();
```
regislebrun: could we use range_ instead of bounds_ everywhere?

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +204,82 @@ Point TruncatedDistribution::getRealization() const
...
 }
 
 
 /* Get the PDF of the distribution: PDF_trunc = 1[a, b] * PDF / (CDF(b) - CDF(a)) */
 Scalar TruncatedDistribution::computePDF(const Point & point) const
 {
-  if (point.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
+  if (point.getDimension() != getDimension()) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=" << getDimension() << ", here dimension=" << point.getDimension();
+
+  Interval offset(Point(getDimension(), -quantileEpsilon_), Point(getDimension(), quantileEpsilon_));
```
regislebrun: Ouch! it will hurt performance a lot! This variable does not depend on point and is no modified within the method, so it should not be constructed at each call.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +204,82 @@ Point TruncatedDistribution::getRealization() const
...
 Scalar TruncatedDistribution::computeCDF(const Point & point) const
 {
-  if (point.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
+  if (point.getDimension() != getDimension()) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=" << getDimension() << ", here dimension=" << point.getDimension();
+
+  for (UnsignedInteger k = 0; k < getDimension(); ++ k)
+  {
+    const Scalar xK = point[k];
+    if (xK <= bounds_.getLowerBound()[k]) return 0.0;
+    if (xK >= bounds_.getUpperBound()[k]) return 1.0;
```
regislebrun: No. The CDF is zero if at least one of the components of the argument is less than the corresponding component of the lower bound, but is 1 only if ALL the components are greater or equal to the corresponding components of the upper bound. If only a part of the component is greater than the upper bound, then the CDF is obtained from the marginal distribution corresponding to the other components multiplied by the normalization factor. See NormalCopula::computeCDF to have an example of such a truncation.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +204,82 @@ Point TruncatedDistribution::getRealization() const
...
+  }
 
-  const Scalar x = point[0];
-  if (x <= lowerBound_) return 0.0;
-  if (x >= upperBound_) return 1.0;
   // If tail=true, don't call distribution_.computeCDF with tail=true in the next line!
   return normalizationFactor_ * (distribution_.computeCDF(point) - cdfLowerBound_);
```
regislebrun: No. The CDF of a truncated distribution is equal to the probability of the interval [lower, point] divided by the probability of the truncation interval, ie distribution_.computeProbability(Interval(range_.getLowerBound(), point)) * normalizationFactor_ or marginal_.computeProbability(Interval(marginalLowerBound, marginalPoint)) * normalizationFactor_ if a part of the components of point are greater than their associated upper bounds.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +204,82 @@ Point TruncatedDistribution::getRealization() const
...
+    if (xK <= bounds_.getLowerBound()[k]) return 1.0;
+    if (xK >= bounds_.getUpperBound()[k]) return 0.0;
+  }
 
-  const Scalar x = point[0];
-  if (x <= lowerBound_) return 1.0;
-  if (x > upperBound_) return 0.0;
   // If tail=true, don't call distribution_.computeCDF with tail=true in the next line!
   return normalizationFactor_ * (cdfUpperBound_ - distribution_.computeCDF(point));
 }
```
regislebrun: Wrong. See computeCDF() comments

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +204,82 @@ Point TruncatedDistribution::getRealization() const
...
   const Point pdfGradientX(distribution_.computePDFGradient(point));
-  const Point cdfGradientLowerBound(finiteLowerBound_ ? distribution_.computeCDFGradient(Point(1, lowerBound_)) : Point(distribution_.getParametersCollection()[0].getDimension()));
-  const Point cdfGradientUpperBound(finiteUpperBound_ ? distribution_.computeCDFGradient(Point(1, upperBound_)) : Point(distribution_.getParametersCollection()[0].getDimension()));
+  const Point cdfGradientLowerBound(bounds_.getFiniteLowerBound()[0] ? distribution_.computeCDFGradient(bounds_.getLowerBound()) : Point(distribution_.getParametersCollection()[0].getDimension()));
+  const Point cdfGradientUpperBound(bounds_.getFiniteUpperBound()[0] ? distribution_.computeCDFGradient(bounds_.getUpperBound()) : Point(distribution_.getParametersCollection()[0].getDimension()));
   const Scalar pdfPoint = distribution_.computePDF(point);
   Point pdfGradient(normalizationFactor_ * pdfGradientX - pdfPoint * normalizationFactor_ * normalizationFactor_ * (cdfGradientUpperBound - cdfGradientLowerBound));
   // If the lower bound is finite, add a component to the gradient
-  if (finiteLowerBound_)
+  if (bounds_.getFiniteLowerBound()[0])
```
regislebrun: Wrong. All the finite lower bounds should be taken into account, not only the first one. The PDF values at the truncation interval vertices of the original distribution are needed here, see the comment for the constructor.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +204,82 @@ Point TruncatedDistribution::getRealization() const
...
   {
     pdfGradient.add(pdfLowerBound_ * pdfPoint * normalizationFactor_ * normalizationFactor_);
   }
   // If the upper bound is finite, add a component to the gradient
-  if (finiteUpperBound_)
+  if (bounds_.getFiniteUpperBound()[0])
```
regislebrun: Same remark as for the lower bound.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +204,82 @@ Point TruncatedDistribution::getRealization() const
...
 
 /* Get the PDFGradient of the distribution */
 Point TruncatedDistribution::computePDFGradient(const Point & point) const
 {
-  if (point.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
+  if (point.getDimension() != getDimension()) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
 
   const Scalar x = point[0];
-  if ((x <= lowerBound_) || (x > upperBound_)) return Point(distribution_.getParametersCollection()[0].getDimension() + finiteLowerBound_ + finiteUpperBound_);
+  if ((x <= bounds_.getLowerBound()[0]) || (x > bounds_.getUpperBound()[0])) return Point(distribution_.getParametersCollection()[0].getDimension() + bounds_.getFiniteLowerBound()[0] + bounds_.getFiniteUpperBound()[0]);
```
regislebrun: Should check ALL the components of point, not only x

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -304,22 +289,22 @@ Point TruncatedDistribution::computePDFGradient(const Point & point) const
...
   Point cdfGradient(normalizationFactor_ * (cdfGradientX - cdfGradientLowerBound) - (cdfPoint - cdfLowerBound_) * normalizationFactor_ * normalizationFactor_ * (cdfGradientUpperBound - cdfGradientLowerBound));
   // If the lower bound is finite, add a component to the gradient
-  if (finiteLowerBound_)
+  if (bounds_.getFiniteLowerBound()[0])
   {
     cdfGradient.add(pdfLowerBound_ * normalizationFactor_ * ((cdfPoint - cdfLowerBound_) * normalizationFactor_ - 1.0));
   }
   // If the upper bound is finite, add a component to the gradient
-  if (finiteUpperBound_)
+  if (bounds_.getFiniteUpperBound()[0])
```
regislebrun: essentially the same comments as for computePDFGradient(). These methods should throw a NotYetImplementedException if dimension>1, or the correct formulas should be implemented (and written in the comments)

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -338,54 +325,72 @@ Scalar TruncatedDistribution::computeScalarQuantile(const Scalar prob,
...
-  point.add(finiteLowerBound_ ? 1.0 : 0.0);
-  point.add(lowerBound_);
-  point.add(finiteUpperBound_ ? 1.0 : 0.0);
-  point.add(upperBound_);
+  for (UnsignedInteger k = 0; k < getDimension(); ++ k)
+    point.add(bounds_.getFiniteLowerBound()[k] ? 1.0 : 0.0);
+  point.add(bounds_.getLowerBound());
+  for (UnsignedInteger k = 0; k < getDimension(); ++ k)
+    point.add(bounds_.getFiniteUpperBound()[k] ? 1.0 : 0.0);
+  point.add(bounds_.getUpperBound());
```
regislebrun: There was a bug in the 1D truncated distribution. Either the infinite bounds are part of the parameters (ie flags+values) and compute[PDF|CDF]Gradient is wrong, or the values of the infinite bounds are not part of the parameters, and they should not appear in the list of parameters.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -338,54 +325,72 @@ Scalar TruncatedDistribution::computeScalarQuantile(const Scalar prob,
...
 {
-  return distribution_.isElliptical() && finiteLowerBound_ && finiteUpperBound_ && (std::abs(distribution_.getRange().getLowerBound()[0] - lowerBound_ + distribution_.getRange().getUpperBound()[0] - upperBound_) < ResourceMap::GetAsScalar("Distribution-DefaultQuantileEpsilon"));
+  Bool elliptical = distribution_.isElliptical();
+  for (UnsignedInteger k = 0; elliptical && (k < getDimension()); ++ k)
+  {
+    elliptical = elliptical && bounds_.getFiniteLowerBound()[k] && bounds_.getFiniteUpperBound()[k];
+    const Scalar delta = std::abs(distribution_.getRange().getLowerBound()[k] - bounds_.getLowerBound()[k] + distribution_.getRange().getUpperBound()[k] - bounds_.getUpperBound()[k]);
+    elliptical = elliptical && (delta < ResourceMap::GetAsScalar("Distribution-DefaultQuantileEpsilon"));
+  }
+  return elliptical;
```
regislebrun: Wrong.
  + For dimension==1, a distribution is elliptical if and only if it is symmetrical wrt a given point. If the underlying distribution is not elliptical, the truncated one cannot be elliptical, and if the underlying distribution is elliptical, the truncated one is elliptical if and only if the truncation interval is symmetrical wrt to symmetry point of the underlying distribution.
  + For dimension>1, a truncated distribution CANNOT be elliptical if there is an actual truncation (ie the range of the initial distribution is not included into the truncation interval), even if the underlying distribution is elliptical. It is due to the fact that the iso-PDF of an elliptical distribution are ellipses, which is not compatible with a truncation wrt a box.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -414,31 +454,34 @@ Scalar TruncatedDistribution::getThresholdRealization() const
 /* Lower bound accessor */
 void TruncatedDistribution::setLowerBound(const Scalar lowerBound)
 {
-  if ((finiteUpperBound_) && (lowerBound > upperBound_)) throw InvalidArgumentException(HERE) << "Error: the lower bound must be strictly less than the upper bound, here lower bound=" << lowerBound << " and upper bound=" << upperBound_;
+  LOGWARN("TruncatedDistribution::setLowerBound is deprecated, use setBound..");
+  if ((bounds_.getFiniteUpperBound()[0]) && (lowerBound > bounds_.getUpperBound()[0])) throw InvalidArgumentException(HERE) << "Error: the lower bound must be strictly less than the upper bound, here lower bound=" << lowerBound << " and upper bound=" << bounds_.getUpperBound()[0];
```
regislebrun: The test should compare all the components

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -414,31 +454,34 @@ Scalar TruncatedDistribution::getThresholdRealization() const
...
+  bounds_.setLowerBound(Point(1, lowerBound));
+  bounds_.setFiniteLowerBound(BoolCollection(1, true));
   normalizationFactor_ = 1.0 / (cdfUpperBound_ - cdfLowerBound_);
   computeRange();
 }
 
 Scalar TruncatedDistribution::getLowerBound() const
 {
-  return lowerBound_;
+  LOGWARN("TruncatedDistribution::getLowerBound is deprecated, use getRange.");
```
regislebrun: No. Either we call getBound() if we want to keep track of the bound given by the user (ie before its restriction wrt the range of the underlying distribution) or we call getRange() but use range_ everywhere in the code.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -498,7 +545,32 @@ void TruncatedDistribution::setFiniteUpperBound(const Bool finiteUpperBound)
...
+    isAlreadyComputedMean_ = false;
+    isAlreadyComputedCovariance_ = false;
+    isAlreadyCreatedGeneratingFunction_ = false;
+    computeRange();
+  }
+}
+
+Interval TruncatedDistribution::getBounds() const
+{
+  return bounds_;
```
regislebrun: bounds_ or range_?

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -534,8 +606,8 @@ Point TruncatedDistribution::getSingularities() const
   for (UnsignedInteger i = 0; i < size; ++i)
   {
     const Scalar x = nontruncatedSingularities[i];
-    if (x >= upperBound_) return singularities;
-    if (x > lowerBound_) singularities.add(x);
+    if (x >= bounds_.getUpperBound()[0]) return singularities;
+    if (x > bounds_.getLowerBound()[0]) singularities.add(x);
```
regislebrun: As this method is defined only for 1D distributions, it should throw a NotDefinedException if dimension>1, and the upper and lower bounds should be stored as scalars before the loop to speed-up comparisons.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: I propose to exclude the finite flags and the infinite values so that the number of parameters is consistent with pdfgradient.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: ok, done

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +201,144 @@ Point TruncatedDistribution::getRealization() const
...
+  {
+    Point partialPoint(activeDimension);
+    Point partialBound(activeDimension);
+    for (UnsignedInteger i = 0; i < activeDimension; ++ i)
+    {
+      partialPoint[i] = point[indices[i]];
+      partialBound[i] = getRange().getLowerBound()[indices[i]];
+    }
+    // In the other case, we must use the appropriate marginal distribution
+    return distribution_.getMarginal(indices).computeProbability(Interval(partialBound, partialPoint));
```
jschueller: @regislebrun is that part ok ?

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: I throw here if dimension is >1

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: ok

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: I keep bounds_ as the bounds provided by the user.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -96,37 +96,30 @@ TruncatedDistribution::TruncatedDistribution(const Distribution & distribution,
...
       throw InvalidArgumentException(HERE) << "Error: invalid side argument for bounds, must be LOWER or UPPER, here side=" << side;
   } /* end switch */
-  normalizationFactor_ = 1.0 / (cdfUpperBound_ - cdfLowerBound_);
-  // Adjust the truncation interval and the distribution range
   computeRange();
-  lowerBound_ = getRange().getLowerBound()[0];
-  upperBound_ = getRange().getUpperBound()[0];
-  finiteLowerBound_ = getRange().getFiniteLowerBound()[0] == 1;
-  finiteUpperBound_ = getRange().getFiniteUpperBound()[0] == 1;
+  bounds_ = getRange();
```
regislebrun: Should be removed as bounds_ != range_ in the other constructors

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -137,35 +130,20 @@ TruncatedDistribution::TruncatedDistribution(const Distribution & distribution,
...
-  pdfUpperBound_ = distribution.computePDF(Point(1, upperBound_));
-  normalizationFactor_ = 1.0 / (cdfUpperBound_ - cdfLowerBound_);
-  // Adjust the truncation interval and the distribution range
-  computeRange();
-  lowerBound_ = getRange().getLowerBound()[0];
-  upperBound_ = getRange().getUpperBound()[0];
-  finiteLowerBound_ = getRange().getFiniteLowerBound()[0] == 1;
-  finiteUpperBound_ = getRange().getFiniteUpperBound()[0] == 1;
+  setBounds(truncationInterval);
+  bounds_ = getRange();
```
regislebrun: should remove bounds_ = getRange() here too.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +194,144 @@ Point TruncatedDistribution::getRealization() const
   for (;;)
   {
     Point realization(distribution_.getRealization());
-    if ((!finiteLowerBound_ || realization[0] >= lowerBound_) && (!finiteUpperBound_ || realization[0] <= upperBound_)) return realization;
+    if (bounds_.contains(realization)) return realization;
```
regislebrun: There are several problems here:
  + The CDF inversion works only for 1D distributions (and for any distribution, not only continuous ones)
  + The test must be done wrt the probability of the bounding interval according to the underlying non-truncated distribution. As this probability is not stored but its inverse, the test becomes normalizationFactor_ * thresholdRealization_ > 1.0
  + line 196 Point is const
  + The multivariate counterpart of CDF inversion is InverseRosenblattEvaluation feeded with DistFunc::rNormal(getDimension()). In fact, Rosenblatt/InverseRosenblatt should map to or from the uniform distribution on the unit cube instead of the Gaussian standard space and DistFunc::rNormal(getDimension()) should be replaced by RandomGenerator::Generate(getDimension())
  + If one uses the InverseRosenblattEvaluation class, one should implement the computeConditionalQuantile() and computeConditionalCDF() methods. I don't know if it is easy or not to get these methods as simple modifications of the non-truncated counterparts.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +194,144 @@ Point TruncatedDistribution::getRealization() const
...
   }
 }
 
 /* Get the DDF of the distribution: DDF_trunc = 1[a, b] * DDF / (CDF(b) - CDF(a)) */
```
regislebrun: The comment should be updated to reflect the multidimensional extension

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +194,144 @@ Point TruncatedDistribution::getRealization() const
...
+  if (point.getDimension() != getDimension()) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=" << getDimension() << ", here dimension=" << point.getDimension();
 
-  const Scalar x = point[0];
-  if ((x <= lowerBound_) || (x > upperBound_)) return Point(1, 0.0);
+  if (!getRange().contains(point)) return Point(getDimension(), 0.0);
   return normalizationFactor_ * distribution_.computeDDF(point);
 }
 
 
 /* Get the PDF of the distribution: PDF_trunc = 1[a, b] * PDF / (CDF(b) - CDF(a)) */
```
regislebrun: Same here

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +194,144 @@ Point TruncatedDistribution::getRealization() const
...
+  else
+    if (!epsilonRange_.contains(point)) return 0.0;
 
-  const Scalar x = point[0];
-  if ((x < lowerBound_ - quantileEpsilon_) || (x > upperBound_ + quantileEpsilon_)) return 0.0;
   return normalizationFactor_ * distribution_.computePDF(point);
 }
 
 
 /* Get the CDF of the distribution: CDF_trunc = 1[a, b] * (CDF - CDF(a)) / (CDF(b) - CDF(a)) + 1]b, inf] */
```
regislebrun: same here

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +194,144 @@ Point TruncatedDistribution::getRealization() const
...
+  if (getDimension() == 1)
+  {
+    const Scalar x = point[0];
+    if (x <= getRange().getLowerBound()[0]) return 0.0;
+    if (x >= getRange().getUpperBound()[0]) return 1.0;
+
+    // If tail=true, don't call distribution_.computeCDF with tail=true in the next line!
+    return normalizationFactor_ * (distribution_.computeCDF(point) - cdfLowerBound_);
+  }
+
```
regislebrun: While being correct, the next lines should be dropped and replaced by
```
return normalizationFactor_ * distribution_.computeProbability(Interval(range_.getLowerBound(), point));
```
as distribution_.computeProbability() should take care of all the corner cases and do the corresponding optimizations.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -227,74 +194,144 @@ Point TruncatedDistribution::getRealization() const
...
+  if (getDimension() == 1)
+  {
+    const Scalar xI = point[0];
+    if (xI <= getRange().getLowerBound()[0]) return 1.0;
+    if (xI >= getRange().getUpperBound()[0]) return 0.0;
+
+    // If tail=true, don't call distribution_.computeCDF with tail=true in the next line!
+    return normalizationFactor_ * (cdfUpperBound_ - distribution_.computeCDF(point));
+  }
+
```
regislebrun: Same here. Just keep:
```
return normalizationFactor_ * distribution_.computeProbability(Interval(point, range_.getUpperBound()));
```

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
regislebrun: ok

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -338,54 +378,71 @@ Scalar TruncatedDistribution::computeScalarQuantile(const Scalar prob,
...
+    elliptical = distribution_.isElliptical() && bounds_.getFiniteLowerBound()[0] && bounds_.getFiniteUpperBound()[0]
+               && (std::abs(distribution_.getRange().getLowerBound()[0] - getRange().getLowerBound()[0] + distribution_.getRange().getUpperBound()[0] - getRange().getUpperBound()[0]) < ResourceMap::GetAsScalar("Distribution-DefaultQuantileEpsilon"));
+  }
+  return elliptical;
 }
 
 /* distribution accessor */
 void TruncatedDistribution::setDistribution(const Distribution & distribution)
 {
-  if (distribution.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: can truncate only distribution with dimension=1, here dimension=" << distribution.getDimension();
```
regislebrun: Here the distribution dimension should be checked wrt the bounds dimension

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: That's dim 1, the multivariate path does contain the right formula.

jschueller:
@regislebrun seems github lost your comment about Rosenblatt transformations.

jschueller:
@regislebrun You lost me in your comment about Rosenblatt transformations, maybe you can call me?

jschueller:
Ok!

PR #589, opened 2017-10-31
Rename Field properties

jschueller:
see http://trac.openturns.org/ticket/928

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -29,6 +29,7 @@
  * CorrelationAnalysis::(Pearson|Spearman)Correlation accepts a multivariate sample and returns a Point
  * Deprecated Field::getDimension, getSpatialDimension, getSpatialMean, getTemporalMean
  * Deprecated Process::getDimension, getSpatialDimension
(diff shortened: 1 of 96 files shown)


dbarbier:
Slightly related, what do you think of renaming `CovarianceModel::partialGradient` method to `spatialGradient`?

jschueller:
I thought we wanted to get rid of the 'spatial' term :question:

regislebrun:
@dbarbier the name comes from the fact that it returns dC(s,t)/ds and not [dC(s,t)/ds, dC(s,t)/dt], so only half of the gradient is returned (which is enough due to the symmetry of C wrt s and t). Do you want to contrast this method from the parameter gradient with the new name?

dbarbier:
@regislebrun Ok, if there is an explanation for partialGradient, this is fine by me.

jschueller:
If it's ok for you I like to merge that before #584 & #590.

dbarbier:
Okay for me; I edited my PR, these branches are only Requests For Comments (RFC) for now.

jschueller:
@regislebrun is this ok for you ?

regislebrun:
+1

PR #590, opened 2017-10-31
[RFC] Transpose CovarianceModelImplementation::partialGradient

dbarbier:
It is more logical to have elements transposed.  See for instance TensorizedCovarianceModel::partialGradient.


dbarbier:
In fact all gradients should be transposed to speed-up computations.  Is it okay to proceed?

regislebrun:
ok for me!

jschueller:
ok

dbarbier:
It is too much work to transpose all gradients.  I am closing this PR for now, and will get back to it later.

PR #594, opened 2017-11-03
Let GLMAlgo be less verbose

dbarbier:
INFO should give reasonable amount of output, user may change log level to DEBUG to have full output.

### DIFF
--- a/lib/src/Base/Optim/NLopt.cxx
+++ b/lib/src/Base/Optim/NLopt.cxx
@@ -309,11 +309,11 @@ void NLopt::run()
   {
     // Here we catch the roundoff_limited exception as the result
     // of the optimization may be useful even if not at the requested precision
-    LOGWARN(OSS() << "NLopt raised a roundoff-limited exception");
+    LOGWARN("NLopt raised a roundoff-limited exception");
   }
   catch (nlopt::forced_stop)
   {
-    LOGWARN(OSS() << "NLopt was stopped by user");
+    LOGWARN("NLopt was stopped by user");
   }
   catch (std::exception & exc)
   {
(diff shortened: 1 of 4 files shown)


regislebrun:
+1

lib/src/Base/Optim/NLopt.cxx
```diff
@@ -348,6 +348,7 @@ void NLopt::run()
     result.store(inP, outP, absoluteError, relativeError, residualError, constraintError);
   }
 
+  result.setIterationNumber(size);
```
jschueller: That's not a really number of iteration but the number of evaluations.

jschueller:
These changes are fine, but maybe e4fa2b3 can be discussed a little bit ?

lib/src/Base/Optim/NLopt.cxx
dbarbier: IIRC I already requested to merge these numbers, maybe we can agree now?

See how they are used in OT, it does not make sense to define two numbers.

jschueller: I agree one could use only one var.

But note that TNC uses getMaximumIterationNumber whereas NLopt uses getMaximumEvaluationNumber as maximum number of evaluations used by their solver.

dbarbier: and Cobyla uses getMaximumIterationNumber() to define the maximum number of evaluations ;-)

jschueller: So maybe NLopt could use getMaximumIterationNumber too, and we deprecated get/setMaximumIterationNumber, update the doc of getMaximumIterationNumber to mention it also applies to the number of evaluations depending on the solver, what do you think ?

jschueller: Ok I guess it can be part of another PR

PR #597, opened 2017-11-07
Maintenance fixes

jschueller:
It's true in the config file.


jschueller:
@dbarbier I added a few commits here after your review

dbarbier:
Fine by me.

PR #598, opened 2017-11-08
Compute the exact log-likelihood in computeReducedLogLikelihood

dbarbier:
We are only interested to find its maximum value, and thus we did not care to add the constant part, but user may be interested by the exact value.

Let loglikelihood function really uses its cache, it was enabled but not used.


regislebrun:
+1

PR #599, opened 2017-11-09
Speed-up computeStandardRepresentative() when called on a Sample

dbarbier:
Overload CovarianceModelImplementation::computeStandardRepresentative() method with arguments being iterators on Collection<Scalar>, instead of Point. This avoids the creation of temporary Point instances.

### DIFF
--- a/lib/src/Base/Stat/CovarianceModelImplementation.cxx
+++ b/lib/src/Base/Stat/CovarianceModelImplementation.cxx
@@ -307,7 +307,10 @@ struct CovarianceModelDiscretizePolicy
 struct CovarianceModelDiscretizeKroneckerPolicy
 {
   const SampleImplementation & input_;
-  CovarianceMatrix & output_;
+  // output_ is a CovarianceMatrix, but since we fill only its half part,
+  // we can directly store results in the underlying MatrixImplementation.
+  // This avoids comparing row and column numbers.
+  MatrixImplementation & output_;
   const CovarianceModelImplementation & model_;
   const UnsignedInteger inputDimension_;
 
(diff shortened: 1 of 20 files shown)


dbarbier:
With this PR, method `StationaryCovarianceModel::operator() (const Point & tau) const` becomes harmful and should be removed.  We do not want to create another Point by substracting two Point, but instead derived classes should call this new method from `operator() (const Point & s, const Point &t) const`, as in `ProductCovarianceModel`.

regislebrun:
+1

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -290,26 +296,31 @@ struct CovarianceModelDiscretizePolicy
       const UnsignedInteger iLocal = i - (jLocal * (jLocal + 1)) / 2;
       const UnsignedInteger iBase = iLocal * outputDimension_;
       const CovarianceMatrix localCovariance(model_(input_[iLocal], input_[jLocal]));
-      for (UnsignedInteger ii = 0; ii < outputDimension_; ++ii)
-        for (UnsignedInteger jj = 0; jj < outputDimension_; ++jj)
-          output_(iBase + ii, jBase + jj) = localCovariance(ii, jj);
+      for (UnsignedInteger jj = 0; jj < outputDimension_; ++jj)
+        for (UnsignedInteger ii = 0; ii < outputDimension_; ++ii)
+          output_(iBase + ii, iBase + jj) = localCovariance(ii, jj);
```
jschueller: @dbarbier is this line ok ? it seems jBase is no more used.

dbarbier: Good catch, thanks.

dbarbier: I am going to fix master

PR #600, opened 2017-11-09
SymbolicFunction to throw on errors

jschueller:
By default muParser is not compiled with MUP_MATH_EXCEPTIONS enabled and does not throw on domain/division errors
To get this behavior (as the Python function) we can throw if nan/inf are found in the output,
although this will also throw when the evaluation itself is ok but invalid inputs are propagated.

cc @aurelieladier


dbarbier:
Merge the two loops:
```
for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
{
  const Scalar value = parsers_[outputIndex].get()->Eval();
  if (!SpecFunc::IsNormal(value))
    throw InternalException(HERE) << "Cannot evaluate " << formulas_[outputIndex] << " at " << inputVariablesNames_.__str__() << "=" << inP.__str__();
  result[outputIndex] = value;
}
```

jschueller:
@dbarbier ok

regislebrun:
+1

PR #601, opened 2017-11-09
Disable multithreaded BLAS in GLMAlgo::maximizeReducedLogLikelihood

dbarbier:
OpenBLAS performs badly when called from TBB.
In order to have the best performance, multithreading must be disabled from high-level routines and not TBB::ParallelFor.

This is hackish, since we do not want to include cblas.h and/or omp.h just to change the number of threads.  Add a CMake option DISCARD_BLAS_MULTITHREADING (ON by default) to disable this change, in case it causes trouble on some systems.

If some computations run faster when OMP_NUM_THREADS environment variable is set to 1, this means that TBB::{Disable,Enable}ThreadedBlas calls should be added to this method.


dbarbier:
Side note: maybe we should run OT tests with `OMP_NUM_THREADS=1`; this should not hurt performance (out tests are small by design), and may help if OpenBLAS decides for any reason to enable multithreading on a loaded machine.

CMakeLists.txt
```diff
@@ -302,6 +304,30 @@ else()
...
+  if(CMAKE_VERSION VERSION_GREATER 2.8.8)
+    include(CMakePushCheckState)
+    cmake_push_check_state()
+  else()
+    set(save_CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES})
+    set(save_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  endif()
+  list(APPEND CMAKE_REQUIRED_LIBRARIES ${LAPACK_LIBRARIES} ${CMAKE_CXX_IMPLICIT_LINK_LIBRARIES})
+  list(APPEND CMAKE_REQUIRED_FLAGS ${OpenMP_CXX_FLAGS})
+  check_function_exists(openblas_get_num_threads HAVE_OPENBLAS_GET_NUM_THREADS )
```
jschueller: include (CheckFunctionExists)

regislebrun:
@dbarbier If I understand correctly the problem, TBB and OpenBLAS threads interact badly. But TBB is used only if the log-likelihood function is evaluated over a sample, which is the case if one use a gradient-based optimization algorithm such as TNC. During my experiments in Chorus, I found Cobyla to be quite effective when their are several local optima. It is a gradient-free algorithm and in this case it may be a good thing to  use multithreaded blas. Is there a way to allow the user to choose if the multithreaded blas have to be used at runtime, eg through ResourceMap?

dbarbier:
> If I understand correctly the problem, TBB and OpenBLAS threads interact badly.

Yes

> But TBB is used only if the log-likelihood function is evaluated over a sample, which is the case if one use a gradient-based optimization algorithm such as TNC.

No, for two reasons:
  1. function evaluation over a sample does currently not use TBB, and
  1. the covariance matrix is always computed inside TBB threads, and BLAS is involved afterwards to compute its Cholesky factor.

I tried first to change the number of BLAS threads inside TBB::ParallelFor, but results were poor, it looks like changing the number of threads is not a light operation.  I guess that you will have the same poor performance if you disable threaded BLAS only when computing covariance matrix.

You can easily make tests on master, play with parallel-threads resource key and OMP_NUM_THREADS environment variable to find what is the best solution in your case.

> [...] Is there a way to allow the user to choose if the multithreaded blas have to be used at runtime, eg through ResourceMap?

Does not make sense to me, user can disable multithreading in BLAS by setting OMP_NUM_THREADS environment variable to 1.

dbarbier:
I fact I am wrong too, the problem does not come from BLAS being called from TBB, this is not the case here.  So I will close this PR, it seems to be an openBLAS issue alone.

PR #602, opened 2017-11-09
Fix build without muParser

dbarbier:
Fix trac#933.


jschueller:
I already edited the ticket, as I pushed
https://github.com/openturns/openturns/pull/600/commits/034d616286b1b4bea76b948a83a6d2bd81ad9394

jschueller:
I'm canceling my commit.

dbarbier:
Oh sorry, I did not notice.  I jumped in because I have rewritten Cache.hxx and this may be a good benchmark to see if my implementation is better or not.

PR #603, opened 2017-11-09
Initialize scale_ in TensorizedCovarianceModel constructor

dbarbier:
Fix trac#870.


ChangeLog
```diff
@@ -38,6 +38,7 @@
 === Miscellaneous ===
 
 === Bug fixes ===
+ * #870 (Problem with TensorizedCovarianceModel with spatial dim > 1)
```
jschueller: could you move this change to the 1.10 section ? I'll likely backport that fix

dbarbier: IMO it is too late, we do not know whether kriging works in this case, this can wait for 1.11.

jschueller: ok

lib/src/Base/Stat/TensorizedCovarianceModel.cxx
```diff
@@ -46,6 +46,7 @@ TensorizedCovarianceModel::TensorizedCovarianceModel(const CovarianceModelCollec
   : CovarianceModelImplementation()
 {
   setCollection(collection);
+  scale_ = Point(inputDimension_, 1.0);
   activeParameter_ = Indices(getScale().getSize() + getAmplitude().getSize());
```
jschueller: pyinstallcheck_CovarianceModel_std's output must be updated

regislebrun:
+1

PR #604, opened 2017-11-10
ImportFromTextFile to read last line

jschueller:
This is a regression in cfe3591; we exit the getline loop if we do not encounter a newline character, now we continue until eof or failure.

### DIFF
--- a/lib/src/Base/Stat/SampleImplementation.cxx
+++ b/lib/src/Base/Stat/SampleImplementation.cxx
@@ -57,13 +57,14 @@ namespace {
 // wraps std::getline to handle EOLs across systems
 OT::Bool GetLine(std::ifstream & file, OT::String & line)
 {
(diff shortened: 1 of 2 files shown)


dbarbier:
Looks confusing, why don't we check eof()?  Something like
```
OT::Bool GetLine(std::ifstream & file, OT::String & line)
{
  if (file.eof() || file.bad()) return false;
  std::getline(file, line);
  if (file.bad()) return false;
  // deal with DOS EOLs from UNIX
  if (line.size() && (line[line.size() - 1] == '\r'))
  {
    line.erase(line.size() - 1, 1);
  }
  return true;
}
```

jschueller:
The api is complex: http://en.cppreference.com/w/cpp/io/ios_base/iostate
the rationale of using !fail() is that it's equivalent to operator bool, unlike good()

PR #605, opened 2017-11-10
Catch pyplot.clabel exception

jschueller:
Instead of ignoring contours with too few levels we catch the clablel exception.
See https://github.com/matplotlib/matplotlib/issues/9742
That way function plots that contain a contour per level are not ignored.


regislebrun:
+1

PR #606, opened 2017-11-10
Alternate implementation of Cache

dbarbier:
With the previous implementation, when the Cache is full, the least recently used element is searched by traversing all elements whenever a new element is inserted.

This new implementation is more standard, it stores keys in a doubly-linked list, and uses a map to store the position of each key in this iterator. Key is moved at the beginning at the list when it is accessed, so tail list always contains the last recently used key.

Since the cache size is fixed, it would be better to use a hash map, but OT does not use hash map anywhere.

### DIFF
--- a/lib/src/Base/Type/openturns/Cache.hxx
+++ b/lib/src/Base/Type/openturns/Cache.hxx
@@ -22,6 +22,7 @@
 #define OPENTURNS_CACHE_HXX
 
 #include <map>
+#include <list>
 #include <sstream>
 #include "openturns/PersistentObject.hxx"
 #include "openturns/StorageManager.hxx"
@@ -32,23 +33,6 @@
 #include "openturns/OStream.hxx"
 
 BEGIN_NAMESPACE_OPENTURNS
-
-
-template <typename K_, typename V_, typename U_>
-inline static
-std::ostream & operator << ( std::ostream & os, const std::pair< K_, std::pair< V_, U_ > > & val)
-{
-  os << val.first << "->" << val.second.first << "/" << val.second.second;
-  return os;
-}
-
-template <typename K_, typename V_, typename U_>
-inline static
-OStream & operator << ( OStream & OS, const std::pair< K_, std::pair< V_, U_ > > & val)
-{
-  OS << val.first << "->" << val.second.first << "/" << val.second.second;
-  return OS;
-}
 
 
 /**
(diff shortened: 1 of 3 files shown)


dbarbier:
Here is a simple benchmark:
```
import openturns as ot
import time

dim = 20
size = 100000
cacheSize = 1024      # Default
cacheSize = size

ot.ResourceMap.SetAsUnsignedInteger("cache-max-size", cacheSize)
X = ot.Normal(dim).getSample(size)
Y = ot.Normal(dim).getSample(size)
tic = time.time()
f = ot.DatabaseFunction(X,Y)
print("Insertion: ", time.time() - tic)

s = ot.Point(dim)
tic = time.time()
for i in xrange(size):
  s += f(X[i])
s /= size
print("Access: ", time.time() - tic)
print(s)
print(Y.computeMean())
```

Tests have been run without the 2nd commit, to test element insertion when cache is full.

Old implementation:
 * size=100000 cacheSize=1024 Insertion=24.6s Access=1.86s
 * size=100000 cacheSize=size Insertion=0.26s Access=1.37s

New implementation:
 * size=100000 cacheSize=1024 Insertion=0.24s Access=1.88s
 * size=100000 cacheSize=size Insertion=0.28s Access=1.41s

These figures come from a single run in each case, it would be nice to have more figures, if you have ideas on real use cases.  It is pretty clear that insertion performs much better when cache is full (the 2nd commit makes this point moot when creating a DatabaseFunction, but there are other cases where elements are added into a full cache), but it is not clear whether access time is comparable or not.

regislebrun:
@dbarbier The gain looks impressive, but I am not sure it will change our typical user's life (using MonteCarlo with continuous distributions). The cache is expected to be activated for costly functions for which the cache overhead is negligible and when a significant portion of the evaluations is expected to be repeated. It has been implemented to speed-up algorithms like finite difference, optimization in relation with FORM/SORM and other places I can't remember. Nevertheless, I may miss some situations where the cache is used even if a lot of redundant evaluations are expected, in which case your PR is a must.

dbarbier:
@regislebrun DatabaseEvaluation::setSample contains this comment:
> // Don't activate the cache systematically as it can take a significant amount of time for large samples
So I guess that this performance penalty had been noticed.

dbarbier:
I am moving the 2nd commit into a separate PR for clarity.

regislebrun:
@dbarbier For DatabaseEvaluation, IMO the cache should be deactivated for this class, as the KDTree search should be as efficient as the search in the cache. BTW, the evaluation operator based on a sample has no cache at all. But you are right, it has been noticed that the cache was a real pain for this specific function!

regislebrun:
+1

jschueller:
This should go in, but there's a conflict in Cache.hxx, could you rebase ?

PR #608, opened 2017-11-10
Speed-up EvaluationImplementation::addCacheContent

dbarbier:
If sample size is larger than cache size, there is no need to store first points, they will be thrown away.


regislebrun:
+1

PR #610, opened 2017-11-13
Release 1.10

jschueller:


dbarbier:
@jschueller I just noticed https://wiki.debian.org/AutomaticDebugPackages and would like to get rid of libopenturns*-dbg Debian binary package, which is really huge and takes a lot of time to upload.  If you are okay with this proposal, I can make these changes.

jschueller:
Yes, its huge for me too when I have to transfer it from opensuse server to our mirror. Do you want to make the changes here or on the debian git ?

jschueller:
Note that I have to maintain compatibilty with debhelper 8 on my end, and that I patch the control section to "debhelper (>= 8)".

dbarbier:
I do not have access to debian git just now, have pushed a commit on my github repo.  Will test it tonight.  It is not clear to me if we have to implement the transition mentioned in above URL, and since it relies on debhelper 9, I chose to not implement it.

jschueller:
Dope! I'm testing it too now.

jschueller:
Maybe we can change the CMAKE_BUILD_TYPE to Release ?

jschueller:
Should I remove the libopenturns0.11-dbg.install file, or will it be used by the new automatic process ?

dbarbier:
Ah yes, libopenturns0.11-dbg.install should be removed.  I will try to move libOT*-gdb.py elsewhere, but I guess that nobody uses it, so it is not a real problem IMO.

jschueller:
It's building fine for me on all debian/ubuntu versions.

jschueller:
@dbarbier a debug package is built, from debian 9: libopenturns0.11-dbgsym_1.10-1_amd64.deb
https://download.opensuse.org/repositories/science:/openturns/Debian_9.0/amd64/

PR #612, opened 2017-11-16
Add PythonDistribution parameters accessors

jschueller:


dbarbier:
You could easily implement setParameter method for SciPyDistribution.

jschueller:
How ?

dbarbier:
This should work:
```
    def setParameter(self, args):
        self._dist.args = args
```

jschueller:
Ok, this is totally working.

regislebrun:
+1

PR #613, opened 2017-11-17
Speed-up Piecewise{Hermite,Linear}Evaluation on a Sample

dbarbier:
When interpolating on a Sample, it is likely than points are looked at in ascending order, bisection search can then be greatly fastened.

Use Sample::operator() instead of operator[].

### DIFF
--- a/lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
+++ b/lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
@@ -102,6 +102,46 @@ String PiecewiseHermiteEvaluation::__str__(const String & offset) const
          << ", derivatives=" << derivatives_ << ")";
 }
 
+namespace {
+Bool computeRegularHermite(const Point & locations)
+{
+  const UnsignedInteger size = locations.getSize();
+  const Scalar step = locations[1] - locations[0];
+  const Scalar relativeEpsilon = ResourceMap::GetAsScalar("PiecewiseHermiteEvaluation-EpsilonRegular") * std::abs(step);
+  for (UnsignedInteger i = 2; i < size; ++i)
+  {
+    if (!(std::abs(locations[i] - locations[0] - i * step) < relativeEpsilon))
+      return false;
+  }
+  return true;
+}
+
+// Find the segment containing value by bisection
+UnsignedInteger findSegmentIndexHermite(const Point & locations, const Scalar value, const UnsignedInteger start)
+{
+  UnsignedInteger iRight = locations.getSize() - 1;
+  UnsignedInteger iLeft  = start;
+  if (value >= locations[start])
+  {
+    // Shortcuts for the most common cases when looping over a Sample
+    if (start == iRight || value < locations[start+1])
+      return start;
+    else if (start + 1 == iRight || value < locations[start+2])
+      return start + 1 ;
+  }
+  else
+    iLeft = 0;
+
+  while (iRight > iLeft + 1)
+  {
+    const UnsignedInteger im = (iRight + iLeft) / 2;
+    if (value < locations[im]) iRight = im;
+    else iLeft = im;
+  }
+  return  iLeft;
+}
+} // anonymous namespace
+
 
 /* Evaluation operator */
 Point PiecewiseHermiteEvaluation::operator () (const Point & inP) const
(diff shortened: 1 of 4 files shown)


lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
```diff
@@ -113,34 +153,58 @@ Point PiecewiseHermiteEvaluation::operator () (const Point & inP) const
...
   const Scalar beta = theta * alpha;
   const Scalar gamma = 2.0 * theta - 1.0;
-  for (UnsignedInteger i = 0; i < dimension; ++i) value[i] = alpha * vLeft[i] + theta * vRight[i] + beta * (gamma * (vRight[i] - vLeft[i]) + h * (alpha * dvLeft[i] - theta * dvRight[i]));
+  for (UnsignedInteger i = 0; i < dimension; ++i) value[i] = alpha * values_(iLeft, i) + theta * values_(iLeft + 1, i) + beta * (gamma * (values_(iLeft + 1, i) - values_(iLeft, i)) + h * (alpha * derivatives_(iLeft, i) - theta * derivatives_(iLeft + 1, i)));
   return value;
 }
 
+Sample PiecewiseHermiteEvaluation::operator () (const Sample & inSample) const
+{
+  if (inSample.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: expected an input point of dimension 1, got dimension=" << inSample.getDimension();
```
regislebrun: expected an input *sample* of dimension...

lib/src/Base/Func/PiecewiseLinearEvaluation.cxx
```diff
@@ -102,32 +142,58 @@ Point PiecewiseLinearEvaluation::operator () (const Point & inP) const
...
   const Scalar alpha = (x - xRight) / dx;
   const Scalar beta = (xLeft - x) / dx;
-  for (UnsignedInteger i = 0; i < dimension; ++i) value[i] = alpha * vLeft[i] + beta * vRight[i];
+  for (UnsignedInteger i = 0; i < dimension; ++i) value[i] = alpha * values_(iLeft, i) + beta * values_(iLeft + 1, i);
   return value;
 }
 
+Sample PiecewiseLinearEvaluation::operator () (const Sample & inSample) const
+{
+  if (inSample.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: expected an input point of dimension 1, got dimension=" << inSample.getDimension();
```
regislebrun: same comment here

regislebrun:
+1

lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
```diff
@@ -113,34 +153,58 @@ Point PiecewiseHermiteEvaluation::operator () (const Point & inP) const
...
+
+    const Scalar h = locations_[iLeft + 1] - locations_[iLeft];
+    const Scalar theta = (x - locations_[iLeft]) / h;
+    const Scalar alpha = 1.0 - theta;
+    const Scalar beta = theta * alpha;
+    const Scalar gamma = 2.0 * theta - 1.0;
+    for (UnsignedInteger j = 0; j < dimension; ++j) output(i, j) = alpha * values_(iLeft, i) + theta * values_(iLeft + 1, i) + beta * (gamma * (values_(iLeft + 1, i) - values_(iLeft, i)) + h * (alpha * derivatives_(iLeft, i) - theta * derivatives_(iLeft + 1, i)));
+    return output;
+  }
+}
```
jschueller: @dbarbier I didn't notice before but the return should be outside the for loop otherwise all components are not computed.

lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
```diff
@@ -192,10 +243,9 @@ void PiecewiseHermiteEvaluation::setLocations(const Point & locations)
   if (locations.getSize() != values_.getSize()) throw InvalidArgumentException(HERE) << "Error: the number of locations=" << size << " must match the number of previously set values=" << values_.getSize();
   const Scalar step = locations_[0] - locations_[0];
   const Scalar epsilon = ResourceMap::GetAsScalar("PiecewiseHermiteEvaluation-EpsilonRegular") * std::abs(step);
```
jschueller: epsilon is no more used

dbarbier: Err, sorry, fixed in master :-/

PR #614, opened 2017-11-21
[RFC] Smaller allocation of DesignProxy

dbarbier:
If the maximum number of columns is known, it can be used.
Default cache size is 128MB, which is *huge* and can stress
kernel allocator if many instances are created and destroyed.

### DIFF
--- a/lib/src/Base/Algo/DesignProxy.cxx
+++ b/lib/src/Base/Algo/DesignProxy.cxx
@@ -43,7 +43,8 @@ DesignProxy::DesignProxy ()
 
 /* Parameters constructor */
 DesignProxy::DesignProxy(const Sample & x,
-                         const Basis & basis)
+                         const Basis & basis,
+                         const UnsignedInteger maximumDimension)
   : Object()
   , x_(x)
   , basis_(basis)
@@ -53,11 +54,11 @@ DesignProxy::DesignProxy(const Sample & x,
   , weight_(0)
 {
   // keep initialization in the ctor for designCache_ to be shared among DesignProxy copies
-  initialize();
+  initialize(maximumDimension);
 }
 
 
-void DesignProxy::initialize() const
+void DesignProxy::initialize(const UnsignedInteger maximumDimension) const
 {
   // allocate cache
   UnsignedInteger cacheSize = ResourceMap::GetAsUnsignedInteger("DesignProxy-DefaultCacheSize");
@@ -67,6 +68,7 @@ void DesignProxy::initialize() const
   UnsignedInteger nbCols = cacheSize / nbRows;
   // The cache stores at least the first function values
   if (nbCols <= 0) nbCols = 1;
+  if (maximumDimension > 0 && nbCols > maximumDimension) nbCols = maximumDimension;
   designCache_ = Matrix(nbRows, nbCols);
   alreadyComputed_ = Indices(nbCols, 0);
 }
--- a/lib/src/Base/Algo/FittingAlgorithmImplementation.cxx
+++ b/lib/src/Base/Algo/FittingAlgorithmImplementation.cxx
@@ -21,6 +21,7 @@
 
 #include "openturns/PersistentObjectFactory.hxx"
 #include "openturns/FittingAlgorithmImplementation.hxx"
+#include <algorithm>
 
 BEGIN_NAMESPACE_OPENTURNS
 
@@ -48,7 +49,8 @@ Scalar FittingAlgorithmImplementation::run(const Sample & x,
     const Basis & basis,
     const Indices & indices) const
 {
-  const DesignProxy proxy(x, basis);
+  const UnsignedInteger maximumDimension = *std::max_element(indices.begin(), indices.end());
+  const DesignProxy proxy(x, basis, maximumDimension);
   return run(y, weight, indices, proxy);
 }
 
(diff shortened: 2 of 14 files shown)


dbarbier:
This PR is opened for discussion only, it may be buggy.

dbarbier:
Benchmarks of functional chaos expansion showed that lots of time was spent in Linux kernel, because of DesignProxy allocations.  There are several ways to fix it, for instance:
1. Drastically lower ResourceMap value.
1. Let user allocate its own DesignProxy, and pass it to algorithms.
1. Reuse DesignProxy instances (which means that we need setters for input sample and basis)
1. When possible, limit the number of columns

Solution 1 has been tested by changing ResourceMap in our script, which led to a nice 6x speed-up, and performance profile looked much better.

This PR shows how to implement the last solution.  Since basis may be infinite, I am not sure that it is correct though.

@jschueller It is very likely that TensorApproximationAlgorithm could be fixed in the same way.

IMO the 2nd solution is the best one wrt performance, but is intrusive.

Comments?

jschueller:
I would leave out option 2; chaos is complex enough as it is.

regislebrun:
Option 1 is straightforward to implement, so go for it. Option 4 is essentially the same but expressed in more mathematical terms. I don't know what is the best for the user, to specify the cache capacity in terms of columns or memory footprint. Option 3 looks good to me, setters with a test to check if something has actually changed. I agree with Julien for option 2, for most of the users it is just black magic.

When I first implemented the DesignProxy class, I had in mind to bound the total amount of memory by the value given in ResourceMap, and I can't remember why I actually allocated all this memory in the constructor.

So +1 for any of the options but the second one.

jschueller:
Which size did you need to reduce to to avoid to stress the memory system ?

dbarbier:
@jschueller Timings of my application for different versions of DesignProxy-DefaultCacheSize:
 * 4000 140s
 * 10000 140s
 * 100000 140s
 * 262144  142s ( number of double precision scalars which fit in a 2MB page size)
 * 1000000 150s
 * 10000000 277s
 * 16777216 369s (default value)

So in this case, anything smaller than 262144 is a good choice, maybe there are different scenarios.

jschueller:
That's not surprising and may be related to huge page allocation.
Maybe we can settle for half of that ? or 250000 ?

dbarbier:
Yes, 131072 or 250000, no strong opinion about the best choice.

jschueller:
250000 ?

regislebrun:
@dbarbier @jschueller Let's go for 250000: it allows to store the whole design matrix for 50 functions over 5000 points, something one may encounter quite frequently. It is not a round number as 131072 is, but at least it is hidden in a configuration file ;-)!
At one point we will have to list all the ResourceMap keys of interest in the documentation of the main algorithms, such as FunctionalChaosAlgorithm. Pamphile will love it.

jschueller:
Actually setting it to 250k yields a performance loss on medium problems: I just tried cppcheck_FunctionalChaos_ishigami_sparse with samplingSize = 175, degree = 20: time goes from 11.5s to ~30s.

dbarbier:
Indeed, so I guess that we should go with option 4.

regislebrun:
To sum up, we have the use-case of Pamphile where the many instances of DesignProxy exhausted the available memory with the default value of 128MB, the application of Denis where anything les than 2MB gives the best result and for Julien's experiment, things are not clear as the test runs in less than half a second on my laptop regardless of the size of the DesignProxy (and I changed both the sampling size and the maximum marginal degree to 175 and 20). What kind of computer are you using? I don't believe you are able to compile OT on such a computer!

dbarbier:
@regislebrun You made a mistake somewhere, with a cache size smaller than 310000 (roughly), the whole design does not fit into memory.

regislebrun:
Ok I get roughly the same figures as yours. The good news is that DesignProxy is useful: I get 57s with a size of zero, and 5.3s with enough room for the whole design. What do you think of something like:
```
void DesignProxy::initialize() const
{
  // allocate cache
  UnsignedInteger cacheSize = ResourceMap::GetAsUnsignedInteger("DesignProxy-DefaultCacheSize");
  UnsignedInteger nbRows = x_.getSize();
  if (nbRows == 0) throw InvalidArgumentException(HERE) << "Cannot initialize a DesignProxy with an empty sample";
  UnsignedInteger nbCols = cacheSize / nbRows;
  // The cache stores at least the first function values
  if (nbCols <= 0) nbCols = 1;
  // Now, try to adapt to the basis size
  try
  {
    // May throw if the basis is unbounded. Maybe should return 0 as a flag?
    nbCols = std::min(nbCols, basis_.getSize());
  }
  catch (NotYetImplementedException & ex)
  {
    //Nothing to do
  }
  designCache_ = Matrix(nbRows, nbCols);
  alreadyComputed_ = Indices(nbCols, 0);
}
```
This way, we reserve only the needed memory if the default buffer size is too large, and we insure that the memory consumption will not explode if the basis size is too large. A sort of mix between option 1 and 4, as the current default value is way too large! Your opinion?

dbarbier:
Looks good, but exceptions should deal with unexpected cases, I dislike using NotYetImplementedException to handle the case of non finite basis.  Maybe add an isFinite() method to Basis?

jschueller:
+1 for isFinite

regislebrun:
+1. Who makes the change? Denis?

jschueller:
Turns out the Basis::isFunctional method already does that.

dbarbier:
`Basis::isFunctional` does not make sense to me, it always returns false, and is never used.

jschueller:
Yes, adding interfaces for a property was a mistake (like copula interface/isCopula). Is isFunctional method working or not ? Maybe isFinite would be more explicit ?

jschueller:
Seems this can be closed as we dont exceed the basis size in DesignProxy:
```
if (nbCols > basis_.getSize()) nbCols = basis_.getSize();
designCache_ = Matrix(nbRows, nbCols);
```

dbarbier:
@jschueller Right, thanks.

PR #615, opened 2017-11-23
If input is normalized, also normalize optimization bounds

dbarbier:
Optimization bounds are computed by setCovarianceModel(), thus inputTransformation_
must be set before this method is called.

### DIFF
--- a/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx
@@ -55,6 +55,7 @@ GeneralLinearModelAlgorithm::GeneralLinearModelAlgorithm()
   , reducedCovarianceModel_()
   , solver_()
   , optimizationBounds_()
+  , normalizedOptimizationBounds_()
   , beta_(0)
   , rho_(0)
   , F_(0, 0)
@@ -90,6 +91,7 @@ GeneralLinearModelAlgorithm::GeneralLinearModelAlgorithm(const Sample & inputSam
   , reducedCovarianceModel_()
   , solver_()
   , optimizationBounds_()
+  , normalizedOptimizationBounds_()
   , beta_(0)
   , rho_(0)
   , F_(0, 0)
@@ -146,6 +148,7 @@ GeneralLinearModelAlgorithm::GeneralLinearModelAlgorithm(const Sample & inputSam
   , reducedCovarianceModel_()
   , solver_()
   , optimizationBounds_()
+  , normalizedOptimizationBounds_()
   , beta_(0)
   , rho_(0)
   , F_(0, 0)
@@ -214,6 +217,7 @@ GeneralLinearModelAlgorithm::GeneralLinearModelAlgorithm(const Sample & inputSam
   , reducedCovarianceModel_()
   , solver_()
   , optimizationBounds_()
+  , normalizedOptimizationBounds_()
   , beta_(0)
   , rho_(0)
   , F_(0, 0)
@@ -270,6 +274,7 @@ GeneralLinearModelAlgorithm::GeneralLinearModelAlgorithm(const Sample & inputSam
   , reducedCovarianceModel_()
   , solver_()
   , optimizationBounds_()
+  , normalizedOptimizationBounds_()
   , beta_(0)
   , rho_(0)
   , F_(0, 0)
@@ -328,6 +333,7 @@ GeneralLinearModelAlgorithm::GeneralLinearModelAlgorithm(const Sample & inputSam
   , reducedCovarianceModel_()
   , solver_()
   , optimizationBounds_()
+  , normalizedOptimizationBounds_()
   , beta_(0)
   , rho_(0)
   , F_(0, 0)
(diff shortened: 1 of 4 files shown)


lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx
```diff
@@ -431,8 +436,13 @@ void GeneralLinearModelAlgorithm::setCovarianceModel(const CovarianceModel & cov
...
   {
-    const Point lowerBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationLowerBound"));
-    const Point upperBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationUpperBound"));
+    Point lowerBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationLowerBound"));
+    Point upperBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationUpperBound"));
+    if (normalize_)
+    {
+      lowerBound = inputTransformation_(lowerBound);
+      upperBound = inputTransformation_(upperBound);
+    }
```
sofianehaddad: @dbarbier what about  GeneralLinearModelAlgorithm::setOptimizationBounds ?

dbarbier: It depends whether we consider that set/get works on normalized or original bounds.  With this commit, we consider that accessors deal with normalized bounds.  I push another commit to do what you suggest.

dbarbier:
Okay, it does not work since parameters must be positive, not sure how to deal with this constraint.

sofianehaddad:
@dbarbier only upper bound should be transformed

dbarbier:
Good idea, thanks!

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx
```diff
@@ -431,8 +436,13 @@ void GeneralLinearModelAlgorithm::setCovarianceModel(const CovarianceModel & cov
   const UnsignedInteger optimizationDimension = reducedCovarianceModel_.getParameter().getSize();
   if (optimizationDimension > 0)
   {
-    const Point lowerBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationLowerBound"));
-    const Point upperBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationUpperBound"));
+    Point lowerBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationLowerBound"));
+    Point upperBound(optimizationDimension, ResourceMap::GetAsScalar( "GeneralLinearModelAlgorithm-DefaultOptimizationUpperBound"));
+    if (normalize_)
```
regislebrun: @dbarbier Be careful! Only the bounds corresponding to scale parameters should be transformed! They are at locations 0,...,inputTransformation().getInputDimension()-1. BTW, using the same bounds for all the parameters may be dangerous. A more robust default choice should be done, in particular for amplitude parameters, but it can be dne in another PR. A last remark: when the optimal amplitude is deduced from the other optimal parameters, its value is not checked against the associated bound, which is a good thing for the algorithm but rather strange from an optimization point of view.

dbarbier: Okay, done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx
```diff
@@ -724,11 +724,23 @@ Scalar GeneralLinearModelAlgorithm::maximizeReducedLogLikelihood()
...
-      const Interval::BoolCollection finiteLower(optimizationBounds_.getFiniteLowerBound());
-      const Interval::BoolCollection finiteUpper(optimizationBounds_.getFiniteUpperBound());
-      normalizedOptimizationBounds_ = Interval(lowerPoint, upperPoint, finiteLower, finiteUpper);
+      normalizedOptimizationBounds_ = optimizationBounds_;
+      // Normalize upper bounds.  This can be done only if all scale parameters are activated, and scale
+      // parameters are the first ones.
+      const Description activeParametersDescription(reducedCovarianceModel_.getParameterDescription());
+      UnsignedInteger count = 0;
+      for (UnsignedInteger i = 0; i == count && i < activeParametersDescription.getSize(); ++i)
+        if (activeParametersDescription[i].compare(0, 6, "scale_") == 0)
```
regislebrun: @dbarbier Would it be better to check if the active parameter has an index less than reducedCovarianceModel_.getInputDimension() instead of a comparison with a string? All the covariance models should have first a scale parameter of dimension the input dimension, then an amplitude parameter of dimension the output dimension, then the correlation etc. The DiracCovarianceModel does not follow this rule but it is a mistake. Maybe we could leave the code as is and change it once the DiracCovarianceModel would be fixed.

Otherwise, +1 for me.

jschueller: Better how ? That doesnt cost anything and is done only once. Also I thought we discussed ProductCovarianceModel that must have the scale several times or something ?

dbarbier: @regislebrun 
> Would it be better to check if the active parameter has an index less than reducedCovarianceModel_.getInputDimension() instead of a comparison with a string?

Handling of parameters must be improved; at the moment Product and Tensorized covariance models do not take into accound specific parameters of other models, if any.  I do not know how this will change, so I chose this solution, which is more reliable IMO.

regislebrun: @dbarbier I totally agree with you: the handling of parameters must be improved. The meaning of my remark is the following:
  + either we are able to define a set of common parameters among the different covariance models, with a well-defined meaning and size, such as scale, amplitude, output correlation (for a total of n+p+p(p-1)/2 scalars) and additional specific parameters, and in this case I am in favor of enforcing it throughout the code by using indices
  + or we cannot define such common parameters and we get the semantic of what is a scale base on the content of a string. It is not a matter of performance here, but more on semantic.

IMO these points should be discussed in another PR, the current code is ok for me.

PR #616, opened 2017-11-24
Rename UserDefined to Categorical

jschueller:
I find UserDefined not specific enough, while Categorical being more explicit it is also found elsewhere:

- https://en.wikipedia.org/wiki/Categorical_distribution
- https://juliastats.github.io/Distributions.jl/latest/univariate.html#Distributions.Categorical
- https://pypi.python.org/pypi/categorical/0.1.4


regislebrun:
-1 I agree with you that ths distribution should be renamed, but Categorical is definitely NOT a good choice. It implies that the values taken by a random vector with such a distribution takes values into categories, ie non-numerical unordered elements, while it is explicitely taking values in a numerical linear space. What is the meaning of getMean() for a categorical variable? And computeCDF()?
Maybe ArbitrayDiscrete[Distribution] or anything else underlying the fact that the distribution is discrete and takes numerical values (and has finite support?). My English is too bad to propose such names ;-)!

jschueller:
Indeed the support of UserDefined is free whereas the support of Categorical is [[1,K]].
But is there any interest in UserDefined over Categorical other than remapping easily to custom labels ?

regislebrun:
@jschueller Yes indeed. It allows to resample in a given sample, it allows to define count statistics etc. Everytime you call computeCDF() or computeProbability() on a UserDefined you are using the partial order of the underlying value space. Think about a Multinomial or Poisson distribution. Are they categorical or not? Same answer for UserDefined.

regislebrun:
@jschueller Maybe CustomDiscrete as a new name?

sofianehaddad:
@jschueller @regislebrun That means also all UserXXX classes should be renamed for consistency (UserDefinedSpectralModel, UserDefinedCovarianceModel, UserDefinedStationaryCovarianceModel, UserDefinedFactory)
IMHO, `Custom` is a good prefix candidate

jschueller:
Custom is a bit shorter yes, but not so meaningful either.

sofianehaddad:
@jschueller @regislebrun Thinking about it, we had on the one hand Database{Evaluation, Function}; on the other hand we get UserDefined{Distribution, SpectralModel, CovarianceModel, StationaryCovarianceModel, Factory}
All these class allow **defining models** using **data**.
Instead of `Custom`, `Database` prefix seems a better solution. What do you think ?

regislebrun:
@sofianehaddad If you go this way, many things are built based on data in OT! I disagree with you regarding UserDefined{SpectralModel, CovarianceModel, StationaryCovarianceModel}. They are not built "from data" ie based on a raw Sample, but as a user-provided collection of matrices. In the same way, the current UserDefined distribution is built based on a Sample AND a probability table, which is not deduced from the sample. On the other hand, the Database{Evaluation, Function} are built from two samples and an implicite rule on how to interpolate between the points. A DatabaseDistribution would at best reduce to the discrete uniform distribution over a given sample, very restricted wrt to the current distribution.

PR #617, opened 2017-11-27
Really disable sphinx build when missing deps

jschueller:
It seems that the cache value is not taken into account.

cc @aurelieladier


regislebrun:
+1

PR #618, opened 2017-11-29
Reproducible doc builds on CircleCI

jschueller:
Allows to better detect what changes on openturns.github.io, which is also gettting quite huge.

- We set SOURCE_DATE_EPOCH to 1/1/17 which is handled by sphinx, for master builds.
- Do not install the sphinx doctree dir anymore.
- ProductDistribution does not yield reproducible cdf graphs with TBB enabled, so disable it in Distribution graph template.

### DIFF
--- a/lib/src/Uncertainty/Distribution/RandomMixture.cxx
+++ b/lib/src/Uncertainty/Distribution/RandomMixture.cxx
@@ -100,6 +100,8 @@ RandomMixture::RandomMixture(const DistributionCollection & coll,
   // the setDistributionCollection() method that do it for us.
   // This call set also the range.
   setDistributionCollection( coll );
+  cdfEpsilon_ = ResourceMap::GetAsScalar("RandomMixture-DefaultCDFEpsilon");
+  pdfEpsilon_ = ResourceMap::GetAsScalar("RandomMixture-DefaultPDFEpsilon");
 }
 
 /* Default constructor */
@@ -135,6 +137,8 @@ RandomMixture::RandomMixture(const DistributionCollection & coll,
   weights_ = Matrix(1, coll.getSize(), weights);
   // This call set also the range.
   setDistributionCollection(coll);
+  cdfEpsilon_ = ResourceMap::GetAsScalar("RandomMixture-DefaultCDFEpsilon");
+  pdfEpsilon_ = ResourceMap::GetAsScalar("RandomMixture-DefaultPDFEpsilon");
 }
 
 /* Parameter constructor - nD */
@@ -172,6 +176,8 @@ RandomMixture::RandomMixture(const DistributionCollection & coll,
   // This call set also the range.
   weights_ = weights;
   setDistributionCollection(coll);
+  cdfEpsilon_ = ResourceMap::GetAsScalar("RandomMixture-DefaultCDFEpsilon");
+  pdfEpsilon_ = ResourceMap::GetAsScalar("RandomMixture-DefaultPDFEpsilon");
 }
 
 /* Parameter constructor - nD */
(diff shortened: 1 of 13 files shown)


regislebrun:
+1

@jschueller Any clue on why the CDF graph of ProductDistribution is not reproducible with TBBs?

jschueller:
I guess it's simply the reordering of the computations.
The testcase if very simple:
```ot.ProductDistribution().drawCDF().getDrawable(0).getData()```

**Comment by jschueller** [comment]
BTW, the isParallel_ flag seems not correctly handled in ProductDistribution, as in setLeft it depends on left_, and in setRight it depends on right_.

**Comment by dbarbier** [comment]
The problem is that ProductDistribution::computeCDF(Point) modifies DistributionImplementation::cdfEpsilon_, and thus must not be run in parallel.

This also applies to KernelMixture (computeSurvivalFunction), Normal (computeCDF), RandomMixture (computePDF, computeProbability), RatioDistribution (computePDF) and Student (computeCDF).

DistributionImplementation::{quantileEpsilon_,quantileIterations_} members are flagged mutable without reason.
Modifying DistributionImplementation::[cp]dfEpsilon_  members as side-effect is a very bad idea.

**Comment by jschueller** [comment]
Well spoted, I guess I'll propose to set isParallel to false for these unless I can think of something better.

**Comment by regislebrun** [comment]
@jschueller @dbarbier I introduced these members to drive some algorithms such as computeQuantile() or computePDF(): if the CDF is known up to epsilon, the finite difference leading to the PDF should be adapted as well as the Newton iteration for the quantile. The key point here is that these algorithms have changed, and the management of these epsilon is very poorly implemented. In my view it should be removed or only used as a cache to avoid a systematic call to ResourceMap, but should not be modified after the creation of the distribution. Well done Denis.

**Comment by jschueller** [comment]
@regislebrun do you mean setting cdfEspilon_ is useless now ?

**Comment by regislebrun** [comment]
@jschueller yes. It should be used as a read-only member. If something goes wrong doing that, it is a bug to be fixed by another mean.

**Comment by jschueller** [comment]
@regislebrun indeed, using cdfepsilon_ & pdfepsilon_ as read-only seem ok, it only changes a decimal in the RandomMixture projections test but maybe you'd want to check this last commit.

File: lib/src/Uncertainty/Distribution/RandomMixture.cxx [hunk 1]
```diff
@@ -2241,7 +2240,6 @@ Scalar RandomMixture::computeProbability(const Interval & interval) const
   // Special case for combination containing only one contributor
   if (isAnalytical_)
   {
-    cdfEpsilon_ = ResourceMap::GetAsScalar( "RandomMixture-DefaultCDFEpsilon" );
```
dbarbier: This key equals 1e-10 whereas default CDF epsilon equals 1e-14, you may move this initialization in constructor.

jschueller: It was already initialized in the ctor, I guess it was to reinitialize the value after altering it.

jschueller: you're right

jschueller: now I see RandomMixture also has its own pdfPrecision, cdfPrecision as mutable and use the same pattern.

dbarbier: IMHO you can fix [pc]dfEpsilon_ in this PR, and handle [pc]dfPrecision later.

jschueller: Yes, done, cdfPresision_ is already read-only so I could removed the mutable keyword.

regislebrun:
+1

PR #619, opened 2017-12-01
add python and ot version in windows build

adumasphi:
This allows to have multiple build folder automatically for different OT version and python version.

### DIFF
--- a/distro/windows/Makefile
+++ b/distro/windows/Makefile
@@ -16,11 +16,6 @@
 ARCH=i686
 TARGET = $(ARCH)-w64-mingw32
 
-# OpenTurns paths
-OT_SRC    ?= $(PWD)/../..
-OT_BUILD  ?= $(OT_SRC)/build-$(TARGET)
-OT_PREFIX ?= $(OT_BUILD)/install
-
 OT_VERSION ?= $(shell cat ../../VERSION)
 
 WINDEPS = $(PWD)/openturns-developers-windeps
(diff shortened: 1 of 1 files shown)


regislebrun:
+1

jschueller:
This would also require targets per arch/pyversion/otversion.
Otherwise this would also break many of my scripts :(
Maybe all that you need is a shell script that calls the Makefile for several flavors? like this:
https://github.com/openturns/build/blob/master/.travis.yml

dbarbier:
> This would also require targets per arch/pyversion/otversion.

@jschueller No, targets must not be modified since each build directory deals with only one arch/pyversion/otversion.

@adumasphi   But I do not see the point of having different versions of OT, there is only one source directory.  And why several build directories if there is only one install dir?  Can you please explain what you want to achieve by this PR?

adumasphi:
I use this in order to have several versions of openturns on my desktop when I am developing modules based on openturns with different versions. Also if I want to propose a pull request for openturns, I need to have a build of the current code under development. But when developing a module, I am using a stable version.

dbarbier:
@adumasphi  You can already do that if you install OT in different directories.  This means then that you are using build directories instead of installation directories, but one build does then not match sources, and I do not see why you do not use installation directories then.

adumasphi:
@dbarbier  Yes you are right, I use the installation directories but it was located in the build directory. I will change it so it should solve my problem. We can close the pull request.

PR #621, opened 2017-12-04
[CircleCI] Use texlive backport

jschueller:


jschueller:
I wanted to use newer texlive binaries that are patched for reprocible builds, but I think the their repo is too slow now.

PR #624, opened 2017-12-04
Speed-up conversion from 2-d Python sequence into Sample

dbarbier:


regislebrun:
+1

jschueller:
Did you test this with python3 ?

dbarbier:
No, this is mostly copy/paste thus I did not expect any breakage.  And I thought that this was tested by CI, we should definitely test with Python3.

jschueller:
Python3 is tested on the autobuilder, sorry.

PR #625, opened 2017-12-05
Copulas related enhancements

jschueller:
- ComposedCopula::getMarginal to handle non-contiguous dependency blocs
- MarginalDistribution::isCopula correctly handles the isCopula flag

use case:
https://gist.github.com/jschueller/7585cf999a0f1104d451e23c43024768

### DIFF
--- a/lib/etc/openturns.conf.in
+++ b/lib/etc/openturns.conf.in
@@ -367,7 +367,7 @@
   <!-- OT::TrapezoidalFactory parameters -->
   <TrapezoidalFactory-RhoBeg           value="0.1"    />
   <TrapezoidalFactory-RhoEnd           value="1.0e-5" />
-  <TrapezoidalFactory-MaximumIteration value="10000"  />
+  <TrapezoidalFactory-MaximumIteration value="2000"  />
 
   <!-- OT::TruncatedDistribution parameters -->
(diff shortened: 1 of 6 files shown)


regislebrun:
+1 A further improvement would be to check if several contiguous copulas are normal copulas, in which case it would be possible to deal with indices mixing different contiguous blocks. The benefit would be to avoid to use MarginalDistribution, which is a kind of "last chance" option. I was too lazy to implement it when I proposed the ComposedCopula class, shame on me. It can wait for another PR.

jschueller:
If several copulas are Normal, this could even be dealt with at the construction of the object. This is also true for the independent copula.

regislebrun:
@jschueller The simplification has to be done with care: the merge of two independent copulas is harmless in terms of speed and memory, but the merge of two normal copulas can have a bad impact on both. Think about a ComposedCopula made of N NormalCopula instances of dimension 2. The composed copula stores (and computes) 4N correlations while the full normal copula uses 4N^2 correlations. My point was to avoid the use of MarginalDistribution as much as possible, because it implies the evaluation of the full underlying distribution even if the marginal is of significantly lower dimension.

jschueller:
ok I understand.

jschueller:
@regislebrun I added the example in this branch

jschueller:
@dbarbier are you okay with those changes ?

PR #630, opened 2017-12-07
MaximumLikelihoodFactory enhancements

jschueller:
- the constraint was not taken into account
- factories using MLF did not take their solver attribute into account, removed it in favor of an RM key to set the MLF solver

### DIFF
--- a/lib/src/Uncertainty/Distribution/TrapezoidalFactory.cxx
+++ b/lib/src/Uncertainty/Distribution/TrapezoidalFactory.cxx
@@ -19,7 +19,7 @@
  *
  */
 #include "openturns/TrapezoidalFactory.hxx"
-#include "openturns/SymbolicFunction.hxx"
+#include "openturns/LinearFunction.hxx"
 #include "openturns/SpecFunc.hxx"
 #include "openturns/ResourceMap.hxx"
 #include "openturns/Log.hxx"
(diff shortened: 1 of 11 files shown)


lib/src/Uncertainty/Distribution/TrapezoidalFactory.cxx
```diff
@@ -131,7 +110,17 @@ Trapezoidal TrapezoidalFactory::buildAsTrapezoidal(const Sample & sample) const
...
+  input[0] = "a";
+  input[1] = "b";
+  input[2] = "c";
+  input[3] = "d";
+  Description formula(3);
+  formula[0] = "b - a";// a <= b
+  formula[1] = OSS() << "c - b - " << SpecFunc::ScalarEpsilon;// b < c
+  formula[2] = "d - c";// c <= d
+  SymbolicFunction constraint(input, formula);
+  factory.setOptimizationInequalityConstraint(constraint);
```
sofianehaddad: hmm what about if Muparser is missing ?

jschueller: I guess I could use a LinearFunction.

jschueller: done

sofianehaddad: ok

regislebrun:
+1

PR #631, opened 2017-12-07
Speed-up PiecewiseLinearEvaluation::setLocations

dbarbier:
Values are unchanged if locations are already sorted, and sort is faster.
Fix PiecewiseHermiteEvaluation::setLocations, it did not reorder values and derivatives.

### DIFF
--- a/lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
+++ b/lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
@@ -239,10 +239,35 @@ Point PiecewiseHermiteEvaluation::getLocations() const
 void PiecewiseHermiteEvaluation::setLocations(const Point & locations)
 {
   const UnsignedInteger size = locations.getSize();
-  if (size < 2) throw InvalidArgumentException(HERE) << "Error: there must be at least 2 points to build a piecewise Hermite interpolation function.";
   if (locations.getSize() != values_.getSize()) throw InvalidArgumentException(HERE) << "Error: the number of locations=" << size << " must match the number of previously set values=" << values_.getSize();
-  locations_ = locations;
-  std::stable_sort(locations_.begin(), locations_.end());
+  Bool isSorted = true;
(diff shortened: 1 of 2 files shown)


regislebrun:
+1 There is a Point::isMonotonic() method to be added soon, which is very close to the loop testing the increasing nature of the locations. Would it be a good thing to extend these classes to monotonic locations and not only increasing locations? On the other hand, would it be a good thing to add Point::isIncreasing() and Point::isDecreasing()?

dbarbier:
> Would it be a good thing to extend these classes to monotonic locations and not only increasing locations?

Does not seem useful to me.

> On the other hand, would it be a good thing to add Point::isIncreasing() and Point::isDecreasing()?

Yes.

dbarbier:
Rebased to use new functions

lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
```diff
@@ -239,10 +239,35 @@ Point PiecewiseHermiteEvaluation::getLocations() const
 void PiecewiseHermiteEvaluation::setLocations(const Point & locations)
 {
   const UnsignedInteger size = locations.getSize();
-  if (size < 2) throw InvalidArgumentException(HERE) << "Error: there must be at least 2 points to build a piecewise Hermite interpolation function.";
   if (locations.getSize() != values_.getSize()) throw InvalidArgumentException(HERE) << "Error: the number of locations=" << size << " must match the number of previously set values=" << values_.getSize();
-  locations_ = locations;
-  std::stable_sort(locations_.begin(), locations_.end());
+  Bool isSorted = true;
+  for (UnsignedInteger i = 1; isSorted && i < size; ++i)
+    isSorted = isSorted && locations[i-1] <= locations[i];
```
jschueller: are you sure the branch is in sync ?

dbarbier: Oops, fixed.

regislebrun:
+1

PR #632, opened 2017-12-08
Remove SubSquareCopula

dbarbier:
This class has never been compiled, but appears in doc.
Also remove lib/test/res file, it has likely been added by accident.


regislebrun:
I agree on the fact that only part of this class has been implemented, but it is already a significant amount of work and I don't want to loose it. As it has been developed prior to the use of git, I have no branch dedicated to this specific development. Is it possible to store it in a dedicated branch before to remove these files? I don't know how to do it by myself.
BTW, RungeKutta and Fehlberg are there to implement the SubSquareCopulaFactory class.

jschueller:
It's not lost, you can still restore it by something like ```git checkout <commit> <file>```.

dbarbier:
For instance
```
git checkout tags/v1.10 -- lib/src/Uncertainty/Distribution/SubSquareCopula.cxx
```

jschueller:
@regislebrun what about the TracyWidom distributions (http://trac.openturns.org/ticket/690) ?
They seem to implement moment and are not documented, imho they should go to.

regislebrun:
@jschueller Exact. You can remove them too.

PR #635, opened 2017-12-11
ComposedCopula::getMarg. to merge Gaussian copulas

jschueller:
This allows to avoid using the costly MarginalDistribution

### DIFF
--- a/lib/src/Uncertainty/Distribution/ComposedCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/ComposedCopula.cxx
@@ -466,7 +466,33 @@ ComposedCopula::Implementation ComposedCopula::getMarginal(const Indices & indic
   if (!indices.check(dimension)) throw InvalidArgumentException(HERE) << "Error: the indices of a marginal distribution must be in the range [0, dim-1] and must be different";
   CopulaCollection marginalCopulas;
   const UnsignedInteger indicesSize = indices.getSize();
-  const UnsignedInteger size = copulaCollection_.getSize();
+  // Merge contiguous Gaussian copulas
+  CopulaCollection copulaCollection(1, copulaCollection_[0]);
+  UnsignedInteger size = copulaCollection.getSize();
+  for (UnsignedInteger ii = 1; ii < copulaCollection_.getSize(); ++ ii)
+  {
+    if ((copulaCollection_[ii].getImplementation()->getClassName() == "NormalCopula")
+      && (copulaCollection[size - 1].getImplementation()->getClassName() == "NormalCopula"))
+    {
+      const CorrelationMatrix corr1(copulaCollection[size - 1].getShapeMatrix());
+      const UnsignedInteger dim1 = corr1.getDimension();
+      const CorrelationMatrix corr2(copulaCollection_[ii].getShapeMatrix());
+      const UnsignedInteger dim2 = corr2.getDimension();
+      CorrelationMatrix aggregatedCorrelation(dim1 + dim2);
+      for (UnsignedInteger j = 0; j < dim1; ++ j)
+        for (UnsignedInteger i = 0; i < j; ++ i)
+          aggregatedCorrelation(i, j) = corr1(i, j);
+      for (UnsignedInteger j = 0; j < dim2; ++ j)
+        for (UnsignedInteger i = 0; i < j; ++ i)
+          aggregatedCorrelation(dim1 + i, dim1 + j) = corr2(i, j);
+      copulaCollection[size - 1] = NormalCopula(aggregatedCorrelation);
+    }
+    else
+    {
+      copulaCollection.add(copulaCollection_[ii]);
+      ++ size;
+    }
+  }
   // For each copula, see if there is something to extract
   UnsignedInteger currentPosition = 0;
   UnsignedInteger currentIndex = indices[currentPosition];
@@ -476,7 +502,7 @@ ComposedCopula::Implementation ComposedCopula::getMarginal(const Indices & indic
   UnsignedInteger upperIndex = 0;
   for (UnsignedInteger i = 0; i < size; ++i)
   {
-    const Copula copula(copulaCollection_[i]);
+    const Copula copula(copulaCollection[i]);
     // Update index range for the current copula
     lowerIndex = upperIndex;
     upperIndex += copula.getDimension();
@@ -491,7 +517,7 @@ ComposedCopula::Implementation ComposedCopula::getMarginal(const Indices & indic
       currentIndex = indices[currentPosition];
     }
     // If there is something to extract
-    if (copulaIndices.getSize() > 0) marginalCopulas.add(copulaCollection_[i].getMarginal(copulaIndices));
+    if (copulaIndices.getSize() > 0) marginalCopulas.add(copulaCollection[i].getMarginal(copulaIndices));
     // All the indices have been taken into account
     if (currentPosition == indicesSize) break;
     // non-contiguous dependency blocs case
--- a/lib/src/Uncertainty/Distribution/MarginalDistribution.cxx
+++ b/lib/src/Uncertainty/Distribution/MarginalDistribution.cxx
@@ -401,7 +401,7 @@ void MarginalDistribution::load(Advocate & adv)
   Distribution distribution;
   Indices indices;
   adv.loadAttribute( "distribution_", distribution );
-  adv.loadAttribute( "indices_", indices_ );
+  adv.loadAttribute( "indices_", indices );
   setDistributionAndIndices(distribution, indices);
 }
 
--- a/python/test/t_ComposedCopula_std.expout
+++ b/python/test/t_ComposedCopula_std.expout
@@ -43,3 +43,5 @@ isoprobabilistic transformation (single contributor)= (class=LinearEvaluation na
  [ 0        0.333333 ]])o(| y0 = Uniform(a = 0, b = 1) -> y0 : Student(nu = 3, mu = 1, sigma = 3)
 | y1 = Uniform(a = 0, b = 1) -> y1 : Student(nu = 3, mu = 1, sigma = 3)
 )
+ComposedCopula(NormalCopula(R = [[ 1 0 ]
+ [ 0 1 ]]))
--- a/python/test/t_ComposedCopula_std.py
+++ b/python/test/t_ComposedCopula_std.py
@@ -1,154 +1,149 @@
 #! /usr/bin/env python
 
 from __future__ import print_function
-from openturns import *
+import openturns as ot
 
-TESTPREAMBLE()
-RandomGenerator.SetSeed(0)
+ot.TESTPREAMBLE()
 
-try:
-    # Instanciate one distribution object
-    R = CorrelationMatrix(3)
-    R[0, 1] = 0.5
-    R[0, 2] = 0.25
-    collection = [FrankCopula(3.0), NormalCopula(R), ClaytonCopula(2.0)]
-    copula = ComposedCopula(collection)
+# Instanciate one distribution object
+R = ot.CorrelationMatrix(3)
+R[0, 1] = 0.5
+R[0, 2] = 0.25
+collection = [ot.FrankCopula(3.0), ot.NormalCopula(R), ot.ClaytonCopula(2.0)]
+copula = ot.ComposedCopula(collection)
 
-    print("Copula ", repr(copula))
-    print("Copula ", copula)
+print("Copula ", repr(copula))
+print("Copula ", copula)
 
-    # Is this copula elliptical ?
-    print("Elliptical distribution= ", copula.isElliptical())
+# Is this copula elliptical ?
+print("Elliptical distribution= ", copula.isElliptical())
 
-    # Is this copula continuous ?
-    print("Continuous = ", copula.isContinuous())
+# Is this copula continuous ?
+print("Continuous = ", copula.isContinuous())
 
-    # Is this copula elliptical ?
-    print("Elliptical = ", copula.hasEllipticalCopula())
+# Is this copula elliptical ?
+print("Elliptical = ", copula.hasEllipticalCopula())
 
-    # Is this copula independent ?
-    print("Independent = ", copula.hasIndependentCopula())
+# Is this copula independent ?
+print("Independent = ", copula.hasIndependentCopula())
 
-    # Test for realization of copula
-    oneRealization = copula.getRealization()
-    print("oneRealization=", repr(oneRealization))
+# Test for realization of copula
+oneRealization = copula.getRealization()
+print("oneRealization=", repr(oneRealization))
 
-    # Test for sampling
-    size = 10000
-    oneSample = copula.getSample(size)
-    print("oneSample first=", repr(
-        oneSample[0]), " last=", repr(oneSample[size - 1]))
-    print("mean=", repr(oneSample.computeMean()))
-    print("covariance=", repr(oneSample.computeCovariance()))
+# Test for sampling
+size = 10000
+oneSample = copula.getSample(size)
+print("oneSample first=", repr(
+    oneSample[0]), " last=", repr(oneSample[size - 1]))
+print("mean=", repr(oneSample.computeMean()))
+print("covariance=", repr(oneSample.computeCovariance()))
 
-    # Define a point
-    point = Point(copula.getDimension(), 0.6)
-    print("Point= ", repr(point))
+# Define a point
+point = ot.Point(copula.getDimension(), 0.6)
+print("Point= ", repr(point))
 
-    # Show PDF and CDF of point
-    # Scalar eps(1e-5)
-    DDF = copula.computeDDF(point)
-    print("ddf     =", repr(DDF))
-    PDF = copula.computePDF(point)
-    print("pdf     =%.6f" % PDF)
-    CDF = copula.computeCDF(point)
-    print("cdf=%.6f" % CDF)
-    #    Point PDFgr = copula.computePDFGradient( point )
-    #    print "pdf gradient     =", PDFgr
-    #    Point CDFgr = copula.computeCDFGradient( point )
-    #    print "cdf gradient     =", CDFgr
-    quantile = copula.computeQuantile(0.95)
-    print("quantile=", repr(quantile))
-    print("cdf(quantile)=%.6f" % copula.computeCDF(quantile))
-    # Get 95% survival function
-    inverseSurvival = Point(copula.computeInverseSurvivalFunction(0.95))
-    print("InverseSurvival=", repr(inverseSurvival))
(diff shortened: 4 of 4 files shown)


dbarbier:
I am not opposed to this PR, but there is no reason for MarginalDistribution to be costly.  Maybe you should define computeCDF(Sample).

lib/src/Uncertainty/Distribution/ComposedCopula.cxx
```diff
@@ -466,7 +466,33 @@ ComposedCopula::Implementation ComposedCopula::getMarginal(const Indices & indic
   if (!indices.check(dimension)) throw InvalidArgumentException(HERE) << "Error: the indices of a marginal distribution must be in the range [0, dim-1] and must be different";
   CopulaCollection marginalCopulas;
   const UnsignedInteger indicesSize = indices.getSize();
-  const UnsignedInteger size = copulaCollection_.getSize();
+  // Merge contiguous Gaussian copulas
```
dbarbier: What does "contiguous" mean here?  I am confused since you check copulas at indices ii and size-1 below.

jschueller: The new copulaCollection variable stores the merged copulas. At the beginning it stores only the first copula from the original copulaCollection_ attribute. In the for loop I check if the next item in copulaCollection_ and the last item in copulaCollection can be merged into a new NormalCopula else I append the copulaCollection_[ii] into copulaCollection.

At the end we have one NormalCopula instance in copulaCollection for all sets of "contiguous" NormalCopula instances in copulaCollection_. For example [NormalC, NormalC, ClaytonC, NormalC, NormalC] leads to [NormalC, ClaytonC, NormalC].

dbarbier:
Another idea, `MarginalDistribution::getMarginal(const UnsignedInteger i)` would be much faster it it did not call `MarginalDistribution::getMarginal(const Indices & indices)` but instead return the right marginal distribution.

jschueller:
I suspect this is written this way from fear some distribution does not implement getMarginal.

jschueller:
I think the point at the time was not to rely on getMarginal.

dbarbier:
As explained by @regislebrun in #625, merging copulas may decrease performances, so I am not sure that this is the way to go.  Do you have a use case which exhibits performance gains?

I was mixing ComposedCopula and MarginalDistribution in my previous comment, sorry.  In fact, I do not understand why `MarginalDistribution::getMarginal(UnsignedInteger)` is defined this way.  It is worse than `DistributionImplementation::getMarginal(UnsignedInteger)`, so it should either not be redefined, or should return `distribution_.getMarginal(i)`.

And `ComposedCopula::getMarginal(i)` should be defined by finding the copula and returning the right marginal.

jschueller:
I did not measure the performance. I trusted Regis comment. Another argument is that the resulting structure involves less objects and may be subject to less bugs (unimplemented services or so).

Maybe I'll let @regislebrun comment the implementation of ``MarginalDistribution::getMarginal```.

I think ```ComposedCopula::getMarginal(i)``` is fine as it inherits the one from CopulaImplementation, as and should be equivalent to an Uniform(0,1).

regislebrun:
First of all, MarginalDistribution is here to provide a generic implementation of the getMarginal() method for distributions for which there is no closed-form for the marginal distributions but the definition: its CDF function is equal to the original CDF function in which all the removed indices are replaced by the upper bound of the range of the component. So MarginalDistribution::computeCDF() will have the cost of the original CDF even is the dimension has been greatly reduced by the marginal extraction. This small benchmark shows the point:
```
from openturns import *
from time import time

size = 100000

dim = 3
R = CorrelationMatrix(dim)
for i in range(dim-1):
    R[i,i+1] = 0.5
copula = ComposedCopula([NormalCopula(R), NormalCopula(R)])
sample = copula.getSample(size)

# Full distribution
print("copula=", copula)
t0 = time()
cdf = copula.computeCDF(sample)
t1 = time()
print("t=", t1 - t0, "s")

# 4-d marginal in nice order
copula_marginal1 = copula.getMarginal([dim-1, dim-2, dim+1, dim])
print("copula_marginal1=", copula_marginal1)
sample_marginal = sample.getMarginal([0, 1, 2, 3])
t0 = time()
cdf = copula_marginal1.computeCDF(sample_marginal)
t1 = time()
print("t=", t1 - t0, "s")

# 4-d marginal in reversed order
copula_marginal2 = MarginalDistribution(copula, [dim+1, dim, dim-1, dim-2])
print("copula_marginal2=", copula_marginal2)
t0 = time()
cdf = copula_marginal2.computeCDF(sample_marginal)
t1 = time()
print("t=", t1 - t0, "s")
```
It gives:
```
copula= ComposedCopula(NormalCopula(R = [[ 1   0.5 0   ]
 [ 0.5 1   0.5 ]
 [ 0   0.5 1   ]]), NormalCopula(R = [[ 1   0.5 0   ]
 [ 0.5 1   0.5 ]
 [ 0   0.5 1   ]]))
t= 0.61726975440979 s
copula_marginal1= ComposedCopula(NormalCopula(R = [[ 1   0.5 ]
 [ 0.5 1   ]]), NormalCopula(R = [[ 1   0.5 ]
 [ 0.5 1   ]]))
t= 0.2098245620727539 s
copula_marginal2= MarginalDistribution(distribution=ComposedCopula(NormalCopula(R = [[ 1   0.5 0   ]
 [ 0.5 1   0.5 ]
 [ 0   0.5 1   ]]), NormalCopula(R = [[ 1   0.5 0   ]
 [ 0.5 1   0.5 ]
 [ 0   0.5 1   ]])), indices=[4,3,2,1])
t= 3.2062911987304688 s
```
So you see the effect of using MarginalDistribution instead of the more specific ComposedCopula class.

dbarbier:
hmmm your script seems to hang forever with this branch :-/

jschueller:
Indeed, the cost of compute CDF explodes here even for size=100

dbarbier:
With master, my output is:
```
('copula=', class=ComposedCopula name=ComposedCopula dimension=6 copula[0]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1])
('t=', 0.3222510814666748, 's')
('copula_marginal1=', class=ComposedCopula name=ComposedCopula dimension=4 copula[0]=class=NormalCopula name=NormalCopula dimension=2 correlation=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0.5,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=2 correlation=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0.5,0.5,1])
('t=', 0.2615690231323242, 's')
('copula_marginal2=', class=MarginalDistribution name=MarginalDistribution dimension=4 distribution=class=ComposedCopula name=ComposedCopula dimension=6 copula[0]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] indices=[4,3,2,1])
('t=', 6.475864887237549, 's')
```

If I define
```
Sample MarginalDistribution::computeCDF(const Sample & sample) const
{
  if (sample.getDimension() != dimension) throw InvalidArgumentException(HERE) << "Error: expected a Sample of dimension=" << dimension << ", got dimension=" << sample.getDimension();
  const UnsignedInteger size = sample.getSize();
  Sample distributionSample(size, lowerBound_);
    for (UnsignedInteger j = 0; j < dimension; ++j)
      distributionSample(i, indices_[j]) = sample(i, j);
  return distribution_.computeCDF(distributionSample);
}
```
output becomes
```
('copula=', class=ComposedCopula name=ComposedCopula dimension=6 copula[0]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1])
('t=', 0.15528011322021484, 's')
('copula_marginal1=', class=ComposedCopula name=ComposedCopula dimension=4 copula[0]=class=NormalCopula name=NormalCopula dimension=2 correlation=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0.5,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=2 correlation=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0.5,0.5,1])
('t=', 0.09446978569030762, 's')
('copula_marginal2=', class=MarginalDistribution name=MarginalDistribution dimension=4 distribution=class=ComposedCopula name=ComposedCopula dimension=6 copula[0]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] indices=[4,3,2,1])
('t=', 0.0064220428466796875, 's')
```

Please review carefully, I did not check that results are right.

dbarbier:
Oops, should be upperBound_ instead of lowerBound_, and output becomes then:
```
('copula=', class=ComposedCopula name=ComposedCopula dimension=6 copula[0]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1])
('t=', 0.15160417556762695, 's')
('copula_marginal1=', class=ComposedCopula name=ComposedCopula dimension=4 copula[0]=class=NormalCopula name=NormalCopula dimension=2 correlation=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0.5,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=2 correlation=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0.5,0.5,1])
('t=', 0.09290313720703125, 's')
('copula_marginal2=', class=MarginalDistribution name=MarginalDistribution dimension=4 distribution=class=ComposedCopula name=ComposedCopula dimension=6 copula[0]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] copula[1]=class=NormalCopula name=NormalCopula dimension=3 correlation=class=CorrelationMatrix dimension=3 implementation=class=MatrixImplementation name=Unnamed rows=3 columns=3 values=[1,0.5,0,0.5,1,0.5,0,0.5,1] indices=[4,3,2,1])
```

jschueller:
By changing lb_ to ub_ the time goes from 0.006 to 8s ?

dbarbier:
Yes

dbarbier:
The problem is that NormalCopula::computeCDF creates a new marginal distribution for each point of the sample; we should define NormalCopula::computeCDF(Sample) to compute marginal distribution only once.

dbarbier:
See my branch db/computeCDF

regislebrun:
@dbarbier The idea is very nice, but it should be placed in another location. For example, when I wrote NormalCopula::computeCDF(), I took care of the border cases but it would have been more readable to delegate to Normal::computeCDF() the treatment of all the border cases (which is done indeed). So if your code move to Normal::computeCDF() we should have the gain for both Normal and NormalCopula.
In fact, this treatment of the input sample is generic of all the multivariate distributions: if at least one of the components is less than its corresponding lower bound, the result is zero, and if at least one of the components is grater than its corresponding upper bound but not all, a marginal distribution has to be extracted (otherwise the value is 1.0). The problem is that I can't see how to implement it in DistributionImplementation::computeCDF() without creating an infinite loop with DistributionImplementation::getMarginal(), which relies on MarginalDistribution!

regislebrun:
@jschueller @dbarbier BTW it looks strange to introduce this kind of optimization in all the other multivariate distributions to speed-up a unique distribution. I don't say we shouldn't do it, but I have the feeling that we miss something here...

dbarbier:
@regislebrun Basically, in computeCDF(Point), every distribution builds some internal objects, and then use them to provide a result.  Most of the time, computeCDF(Sample) could reuse the same objects, this is why implementing this method is dramatically faster.  

A first option would be to cache these objects, but as explained in #579, this is a very bad idea.  So the only remaining option is to implement it as in this branch.  Of course, generic code should be provided when possible to avoid duplication, I will update my branch.

dbarbier:
Branch updated.

jschueller:
The last 2 commits of the computeCDF branch ?

dbarbier:
@jschueller Yes

jschueller:
Your commits seem more interesting than my initial change, maybe you'd want to submit it ?

regislebrun:
The point is that MarginalDistribution tries to make use of the specific getMarginal() method of the underlying distribution by setting some components of the input point to its upper bound. In this case, the underlying computeCDF() method will be called with the same saturation pattern every time. But if you look at the underlying distribution in a generic situation, its input point is most of the time inside of its range, it is why I didn't tried to reuse the marginal distribution built on the fly. I agree with @dbarbier  that this kind of optimization (ie checking the saturated bounds) is here essentially to cope efficiently with the situation corresponding to MarginalDistribution, so the solution he proposed is  not a hack but a better understanding of the problem that I had ;-)!

Concerning ComposedDistribution::getMarginal(), there is a pattern which could be checked but does not seem to be so easy to implement: the case where the marginal indices are provided in blocks of marginals of atomic copulas.
Imagine that the ComposedCopula is built using [Cop1, Cop2] with dim(Cop1)=d1, dim(Cop2)=d2.
The current implementation is efficient (ie produces a ComposedCopula) if the indices [i1,...,in] are such that there is a partition [i1,...,ik,ik+1,...,in] such that [i1,...,ik] is a valid set of indices to extract a marginal of Cop1 and [ik+1-d1,...,in-d1] a valid set of indices to extract a marginal of Cop2.
But if [i1,...,in] can be partitioned into [i1,...,ik,ik+1,...,in] such that [i1-d1,...,ik-d1] is a valid set of indices to extract a marginal of Cop2 and [ik+1,...,in] a valid set of indices to extract a marginal of Cop1, then the resulting marginal copula is a ComposedCopula corresponding to [Cop2.getMarginal([i1-d1,...,ik-d1]), Cop1.getMarginal([ik+1,...,in])]. The idea is to detect if there is a block permutation of the given marginal indices such that after the block permutation the indices are given in increasing order, in which case we recover the nice situation.
A possible implementation could be to go through the given marginal indices, to flag them according to the atomic copula they belong, and to see if the flags are contiguous, eg [0,0,1,1,1,2,2] (the situation we cope with) or [1,1,1,0,0,2,2] (the new situation to consider) but not [1,1,0,0,1,2,2].
The optimization I proposed (to merge contiguous atomic copulas if they are NormalCopula of IndependentCopula as @jschueller noticed) will reduce the number of atomic copulas hence it should increase the occurrence of the new pattern.

dbarbier:
@regislebrun I do not understand your point in the first paragraph.  Anyway this discussion tackles different issues, I suggest that we discuss MarginalDistribution when the dedicated PR is submitted.

About ComposedCopula, as said initially, I am not opposed to the changes proposed by @jschueller, it is just that it is not clear to me if merging copulas will improve performance, thus I would like to see some benchmark.

jschueller:
Using Régis's computeCDF benchmark, we see that a 4d normal copula is really slow compared to a composedcopula of 3d+1d, even on only 100 pts, so this PR is not really useful atm.

```
from openturns import *
from time import time

size = 100000
size = 100

# original 6-d copula
R = CorrelationMatrix(dim)
for i in range(dim-1):
    R[i,i+1] = 0.5
copula = ComposedCopula([NormalCopula(R), NormalCopula(R)])
print(copula.getDimension())
sample = copula.getSample(size)


# 3d + 1d blocks
copula_marginal1 = copula.getMarginal([0, 1, 2, 3])
print("copula_marginal1=", copula_marginal1)
sample_marginal = sample.getMarginal([0, 1, 2, 3])
t0 = time()
cdf = copula_marginal1.computeCDF(sample_marginal)
t1 = time()
print("t=", t1 - t0, "s")

# equivalent 4-d block
R = CorrelationMatrix(dim+1)
for i in range(dim-1):
    R[i,i+1] = 0.5
copula1 = NormalCopula(R)
print("copula1=", copula1)
sample_marginal = sample.getMarginal([0, 1, 2, 3])
t0 = time()
cdf = copula1.computeCDF(sample_marginal)
t1 = time()
print("t=", t1 - t0, "s")
```

yields:
```
copula_marginal1= ComposedCopula(NormalCopula(R = [[ 1   0.5 0   ]
 [ 0.5 1   0.5 ]
 [ 0   0.5 1   ]]), NormalCopula(R = [[ 1 ]]))
t= 0.0014340877532958984 s
copula1= NormalCopula(R = [[ 1   0.5 0   0   ]
 [ 0.5 1   0.5 0   ]
 [ 0   0.5 1   0   ]
 [ 0   0   0   1   ]])
t= 34.16700577735901 s
```

PR #636, opened 2017-12-11
Do not clone newly built objects

dbarbier:
When cloning newly built objects, constructor is called twice, this is useless.


dbarbier:
Please note that some classes were already using this pattern, but not all, so this is mostly refactoring to replace `return Foo(bar, baz).clone()` by `return new Foo(bar, baz)`.  Distribution factories also build objects twice, but the same solution cannot be applied.

regislebrun:
+1 So when do we have to use the clone() method?

dbarbier:
@regislebrun When you do not know at compilation time the class of the object being cloned.

PR #639, opened 2017-12-13
Add logical operators

regislebrun:
The exclusive or and the negation operators were missed somewhere between OT 1.2 and OT 1.3 when the sources of MuParser were removed from OT source tree. These operators are needed to explore system reliability using symbolic functions.
It also increases the minimum version of MuParser to 2.2.3, as other logical operators have been renamed since version 2.0.0 and others have been removed. Several bug fixes and new desirable features have appeared in version 2.2.3, which is old enough (22.12.2012, the "ternary" release) to match the version embedded in many distributions.


jschueller:
Could you repush without the min version change just to see what happens on the autobuilder ?

There's  a bug in findmuparser.cmake: remove _MUPARSER_VERSION_MATCH from the 2nd find_package_handle_standard_args call (same thing in findtbb btw).

regislebrun:
@jschueller  I removed the commit. Concerning CMake, do you mean I have something to change in a CMakeList, or in a detection script on my computer?

jschueller:
Yes, in cmake/FindmuParser.cmake, you could remove the _MUPARSER_VERSION_MATCH from the 2nd find_package_handle_standard_args call, there is the same problem in FindTBB but we dont use the version argument so the bug does not show up.

jschueller:
In fact this change is useless, sorry. I misinterpreted the autobuilder log.

jschueller:
You can change back the minimum version requirement in Cmakelists, also it must be changed in python/doc/developer_guide/architecture.rst.

jschueller:
Please update the Dependencies section in python/doc/developer_guide/architecture.rst

regislebrun:
@jschueller Ouppss! Sorry! Done.

PR #640, opened 2017-12-13
Speed-up conversion of Python sequences into OT objects

dbarbier:
Let SWIG use Python buffer protocol when possible.


regislebrun:
@dbarbier I don't undestand exactly how it works, but the result is impressive. Well done.

jschueller:
Great!

Also, you missed a warning from a previous PR:
```
PythonWrappingFunctions.hxx:771:59: warning: comparison of integers of different signs: 'Py_ssize_t' (aka 'long') and 'const UnsignedInteger' (aka 'const unsigned long')
      [-Wsign-compare]
      if (PySequence_Fast_GET_SIZE( newPyPointObj.get() ) != dimension) throw;
```

PR #645, opened 2018-01-03
RandomVector parameters

jschueller:
Add RandomVector::[sg]etParameter, getParameterDescription (same as Function, Distribution)

For example, allows to propagate deterministic parameters through an event:
https://gist.github.com/jschueller/fa282ebdfe89b40eb951b198f02f2abf


regislebrun:
+1

PR #647, opened 2018-01-03
Remove FunctionalBasis/FiniteBasis

jschueller:
The FunctionalBasis interface is removed, see #614.
isFunctional is replaced by isFinite

FiniteBasis is removed, while adding a BasisImplementation::collection_ attribute, not sure how to update it for now when the basis is not finite, leaving for later.
Add BasisImplementation::getCurrentSize which returns the size of collection_.


lib/src/Base/Func/TensorizedUniVariateFunctionFactory.cxx
```diff
@@ -45,7 +45,7 @@ typedef ProductUniVariateFunctionEvaluation::UniVariateFunctionCollection UniVar
 
 /* Default constructor */
 TensorizedUniVariateFunctionFactory::TensorizedUniVariateFunctionFactory()
-  : FunctionalBasisImplementation()
+  : BasisImplementation()
   , coll_()
```
dbarbier: Is coll_ member still needed here, now that BasisImplementation contains a collection?

dbarbier:
Your commit message is misleading, we did not rename isFunctional by isFinite, they have opposite semantics.

lib/src/Base/Func/TensorizedUniVariateFunctionFactory.cxx
jschueller: Yes; coll_ is a collection of UniVariateFunctionFamily:
```typedef Collection<UniVariateFunctionFamily>           FunctionFamilyCollection;```

jschueller:
Indeed, fixed.

dbarbier:
* `TensorizedUniVariateFunctionFactory::build()` method should use `BasisImplementation::coll_` to cache built function
* `BasisImplementation::build()` is a placeholder to ensure that derived classes do override this method; it should only contain the throw statement
* `BasisImplementation::operator[]` works only if it is called with index being incremented from 0 to N

jschueller:
Ok, this is clearer now.

dbarbier:
In fact, I am unsure whether we want to apply these changes, or make much more invasive changes and replace `Basis` by `Collection<Function>` where needed.

jschueller:
The justification of this interface is that some algorithms need to be applied on some basis, whether they're finite or not.

dbarbier:
I am not convinced by your argument, see my branch `db/functionalbasis`.  I naively replaced Basis by `Collection<Function>` in few places, and from there had to change it in a lot of places.  The only problem was in FunctionalChaosAlgorithm.cxx, see FIXME.  I just tested that there is no error with cppcheck.

jschueller:
I think you're right, we introduced Basis everywhere, seems it's only needed in a few place.
I was trying to rebase your patch on master, without my initial changes.

jschueller:
superseded by #715

PR #651, opened 2018-01-29
Multiply a matrix by a sample

dbarbier:
New method `MatrixImplementation::genSampleProd()` to compute `matrix*sample`
or `sample*matrix` multiplications as if sample was a `MatrixImplementation`
stored in row-major order.  Result is directly stored into a `Sample`, thus
two copies are avoided.

Add `Matrix::operator*(Sample)`.

Deprecate `Sample{,Implementation}::operator*(SquareMatrix)` and
`operator/(SquareMatrix)`.

### DIFF
--- a/TODO
+++ b/TODO
@@ -6,3 +6,5 @@ Remove deprecated SecondOrderModel::getDimension, getSpatialDimension in 1.12
 Remove deprecated SpectralModel::getDimension, getSpatialDimension, getSpatialCorrelation in 1.12
 Remove deprecated TruncatedDistribution single bound accessors in favor of setBound/getRange in 1.12
 Remove deprecated Function ctors in 1.12
+Remove deprecated Sample,SampleImplementation::operator*(SquareMatrix) and operator/(SquareMatrix) (and in-place operators) in 1.12
+Remove deprecated SampleImplementation::scale(SquareMatrix) in 1.12
--- a/lib/src/Base/Func/LinearEvaluation.cxx
+++ b/lib/src/Base/Func/LinearEvaluation.cxx
@@ -131,16 +131,8 @@ Sample LinearEvaluation::operator() (const Sample & inS) const
   if (inS.getDimension() != center_.getDimension()) throw InvalidArgumentException(HERE) << "Invalid input dimension";
   const UnsignedInteger size = inS.getSize();
   if (size == 0) return Sample(0, getOutputDimension());
-  SampleImplementation temporary(inS.getSize(), getOutputDimension());
-  // Some OT black magic
-  // + We use the parallelized translation of the input sample inS - center_
-  // + We cast the resulting sample into a matrix
-  // + We perform a matrix/matrix multiplication, using potentially
-  //   high-performance BLAS
-  // + Then the resulting matrix is converted into a sample and the final
-  //   translation is parallelized
-  temporary.setData(*(linear_ * Matrix(getInputDimension(), inS.getSize(), (inS - center_).getImplementation()->getData())).getImplementation());
-  const Sample result(temporary + constant_);
+  const Sample centered(inS - center_);
+  const Sample result(linear_.getImplementation()->genSampleProd(centered, true, false, 'R') + constant_);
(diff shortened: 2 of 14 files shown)


regislebrun:
@dbarbier Strange, using this benchmark I get over 13M mults/s using TBBs and only 11.5M mults/s using parallel OpenBLAS. A loss of 11% is not such a big deal but is unexpected as OpenBLAS is supposed to peak at almost the theoretical maximum performance on dgemm with large matrices...

dbarbier:
@regislebrun What is your test case?

jschueller:
There a small conflict.

regislebrun:
@dbarbier I used this code:
```
from openturns import *
from time import time

size = 10000000
M = SquareMatrix(3, [1.0,2.0,3.0,4.0,5.0,6.0,7.0,8.0,9.0])
sample = Normal(3).getSample(size)
t0 = time()
res = sample*M
t1 = time()
print("speed=", size / (t1 - t0), "mult/s")
```
I am mostly interested into 3D points for 3D mesh drawing. Things may be different for higher dimensions.

dbarbier:
@regislebrun Changes in this branch have zero impact on your benchmark, I did not touch that code.  Current multiplication `Sample*SquareMatrix` must be dropped, this is why I deprecated it:
 1. it returns sample`*`transpose(M) and not sample`*`M, if you see sample as a matrix stored in row-major order
 1. we can multiply by a Matrix if dimensions do match, there is no reason to restrict to a SquareMatrix

My main motivation is to change current C++ code when we perform some Sample and Matrix multiplications.  I had been facing this problem several times in the past, the solution was to copy
Sample into a Matrix, perform Matrix-Matrix multiplications, and copy result into a Sample.  
Unfortunately I do not remember where I had to make these computations, this is why I changed only RandomMixture and LinearEvaluation.

dbarbier:
@regislebrun BTW if you want to compare with code added in this branch, you can add to your script
```
import numpy as np
array = np.array(sample)
sampleT = Sample(array.transpose())
t0 = time()
res2 = M*sampleT
t1 = time()
print("speed=", size / (t1 - t0), "mult/s")
# Check result
array2 = np.array(res2)
res2T = ot.Sample(array2.transpose())
print((res - res2T).getMax())
```
But again, I am not advocating for using Matrix*Sample in Python, since Sample*SquareMatrix is broken, I am primarily focused on C++ code.

dbarbier:
Branch rebased to fix conflict in TODO, no other change.

regislebrun:
@dbarbier I fully understand and share your motivations: the semantic of the multiplication between a sample and a matrix is much clearer this way, it solves a bug and is more general.

My point was on the raw speed of the new implementation of SampleImplementation::operator *(const SquareMatrix & scaling), and yes, your change HAS an impact on my benchmark. See line 2206 of SampleImplementation.cxx. The good new is that I messed up my first measurements, the new implementation is not less than 3.42x faster than the previous one! Impressive, good job. To merge ASAP!

dbarbier:
@regislebrun Sorry, but this branch does not modify SampleImplementation.cxx and can thus not have any impact on `SampleImplementation::operator *(const SquareMatrix & scaling)`.

regislebrun:
@dbarbier Ouppss! Sorry, I was playing with your db/matrix-sample-product branch... But you have so many branches... BTW, is there any connection between these two branches?

dbarbier:
Okay, it is clear now, the branch you were looking for is a preliminary version of this PR, I forgot to delete it.
Since I eventually want to get rid of `SampleImplementation::operator * (const Point & scaling)`, I decided in this final version to not modify it ;-)
But `Mesh::draw3D` does use the new method, so it should address your concern.

dbarbier:
@regislebrun BTW if you know some portion of code which could use this new method, please let me know, I will change it.

regislebrun:
@dbarbier At least the LinearEvaluation class. Its evaluation over a sample is a real mess.

dbarbier:
LinearEvaluation is already fixed by this PR.

PR #652, opened 2018-01-31
New Mesh::computeKDTree method

dbarbier:
In Mesh constructor, do not build KDTree by default anymore.
Building a KDTree is useful only if getNearest* methods are called.


regislebrun:
+1

PR #653, opened 2018-02-01
DistFactory::buildEstimator to handle Exceptions

jschueller:
Fixes http://trac.openturns.org/ticket/938

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -49,6 +49,7 @@
  * #926 (Covariance model active parameter set behavior)
  * #928 (Covariance model/ Field/Process properties naming)
  * #933 (Compilation failed when Muparser is missing)
+ * #938 (DistributionFactory::buildEstimator to handle Exception)
  * #939 (The help of the getMaximumDegreeStrataIndex method is wrong.)
 
 == 1.10 release (2017-11-13) == #release-1.10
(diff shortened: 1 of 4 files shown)


regislebrun:
-1 Disagree with the ticket, hence with the fix. If the bootstrap estimator throw an exception, it is exactly because it is not defined for the distribution or not compatible with the given sample. You may imagine an estimator of the parameters of a continuous distribution based on the increase between its sample statistics, and undefined as soon as some values are repeated (which should occur with probability zero in theory). For this case, each resample of the sample will throw an exception and the factory will enter an infinite loop. It is not just a matter of controlling a loop, it is because the estimator is NOT mathematically defined on bootstrap samples.

If this feature is really needed, I propose to control it through a key in ResourceMap, with a default value leading to exceptions. It may be useful in some circumstances (eg the estimator is well defined as soon as the maximum or minimum value of the sample is unique), but the user has to check if the proportion of failure is negligible wrt the sample size. The number of rejected samples should at least be recorded and printed if not zero.

jschueller:
I propose to add a tolerance of failed samples (rm key BootstrapErrorTolerance=0.1).
I the proportion of failed samples is above we throw else we warn if > 0.

regislebrun:
Let's go for it. And as the Dirichlet distribution belongs to the exponential family, it is asymptotically normal with asymptotic covariance the inverse of Fisher's information matrix. We should use it to provide the asymptotic distribution of the estimate either using the generic implementation in MaximumLikelihoodFactory or by doing the maths.

regislebrun:
+1

PR #654, opened 2018-02-01
Mesh does no more derive from DomainImplementation

dbarbier:
The aim is to let Mesh use the standard Interface/Implementation idiom.
This class is heavy, thus all methods returning a Mesh perform a copy
of an object which may be huge.

A MeshDomain class is added as an adaptor to see a Mesh as a
DomainImplementation.

The inheritance between Mesh and DomainImplementation was never used
in C++ code.  In Python, automatically convert a Mesh into a Domain
thanks to MeshDomain.


dbarbier:
Please carefully review the services provided by Mesh.  I removed `computeVolume()` to make it clear that Mesh is not a Domain, but maybe we could keep it, and/or drop `contains()` etc.

jschueller:
Have you considered using the Domain interface to make these copies instead ? (And add Domain::isMesh() etc ...)

dbarbier:
I do not understand your question, my aim is to eliminate and not to make copies.

dbarbier:
@jschueller What is your suggestion?  In order to avoid copies, we need interface+implementation.  Do you suggest to have Mesh inherit from Domain, and MeshImplementation from DomainImplementation?  This would work, but I find it weird to let Mesh not inherit from `TypedInterfaceObject<MeshImplementation>`, and Python bindings would also to be adjusted as well.

Anyway this is a fundamental question, do we want to see a Mesh as a Domain?  AFAICT this feature was never used except in a notebook, so in my opinion, a Mesh is not a Domain, hence this PR.

jschueller:
ok, good for me

dbarbier:
@regislebrun  And because of that, you duplicated most methods of Domain: getVolume/getNumericalVolume, isEmpty/isNumericallyEmpty, contains/numericallyContains.  From the user point of view, these duplicates do not make sense.

There are many cases where an Interval is used as a geometric interval (for instances by using join or intersect methods) and not as a Domain.  Likewise for Mesh.  So these classes are not Domains, they may behave like Domains, this is very different.
Not so clear for LevelSet.  Anyway in my opinion, all current derived classes of DomainImplementation should be modified to drop this inheritance, and new classes IntervalDomain and LevelSetDomain be created.
This way, consumers have to decide if they use these classes for their intrinsic geometrical properties or as a Domain, in which case they must use the appropriate *Foo*Domain class.

regislebrun:
@dbarbier It is a bug if the methods are redefined instead of overloaded. I forgot the relevant virtual qualifiers in DomainImplementation.hxx, sorry. I may also have used the term 'Domain' in an improper way, as a bad translation of the french word. In my mind, it is a geometrical region for which one can check if it is empty or not, one can test if a point is inside or outside of this region, one can compute the intersection of two such regions... So in my mind it was a full geometrical concept, and I understand that I am wrong when you write "an Interval is used as a geometric interval (for instances by using join or intersect methods) and not as a Domain". Domain *should* have an intersect() method!

dbarbier:
@regislebrun You did not the answer the question I did not ask ;-)  Why are there duplicate methods in Domain: getVolume/getNumericalVolume, isEmpty/isNumericallyEmpty, contains/numericallyContains?

regislebrun:
@dbarbier Ok, I missed your point. Domain tries to encode both a pure mathematical behavior and a numerical behavior. As it sounds strange, let's explain things a little bit. From a pure mathematical point of view, Normal(0, 1) is defined over R so its range is an Interval with both XXXFiniteBound() returning false. So the volume of such an interval should be +inf. But from a numerical point of view, taking int,  account that real numbers are in fact floating point numbers, the actual range of the distribution is [-8.5, 8.5]. I should have used two intervals for that, but I chose to use only one and to put -8.5 as the value of the lower bound, 8.5 for the upper bound. The numerical volume is then 17. I know that the code is wrong in Interval::getVolume() (it doesn't check against the finite flags), but it is a bug. Compare contains() and numericallyContains() to better see my point.
On the other side, I don't understand your remark about Interval being sometimes a geometric interval. It is *always* a geometric interval, it is why Mesh.cxx is in Base/Geom! Same remark for Domain: it is a geometric concept and nothing else, so Domain.cxx is also in Base/Geom. So I still miss your point wrt LevelSet, Interval, Mesh and Domain being not related through inheritance. I remember my course in object design: if you can use any instance of B where an instance of A is expected, then it is likely that B is a specialization of A, and a way to express it is to use inheritance. It was in 1998 so things may have change since then...

dbarbier:
>  In my mind, it is a geometrical region for which one can check if it is empty or not, one can test if a point is inside or outside of this region, one can compute the intersection of two such regions...

Okay, let me try to explain my point of view.  Mesh class should define a tessellation, and nothing more.  It contains vertices and simplices, as is done currently, but without the services you mention above.  These services are implemented by the new MeshDomain class, which takes a Mesh in its constructor, and which inherits from DomainImplementation.

Advantages:
 1. Less coupling, Mesh can now use our pimpl idiom and be a light object
 1. Better separation between Domain generic services and the more specific ones.  It is an error to try to put everything into Domain.  For instance numericallyContains, isNumericallyEmpty and getNumericalVolume should be defined in Interval class and nowhere else.
 1. Good luck to let Mesh implement intersect and join ;-)

regislebrun:
Ok let's go for it (my opinion, to be agreed by others). BTW, it is not that difficult to implement meshes intersection if you see them as domains: first you implement an IntersectionDomain based on two Domain instances, then you implement its contains() method as return leftDomain_.contains(point) && rightDomain_.contains(point). This way, you can implement the intersect() method in a generic way in DomainImplementation, hence you are able to build the intersection between two meshes (but also between a Mesh and a LevelSet, an Interval and so on). Of course it will not be a Mesh, but will be good enough to define complex events. If you plan to build the intersection as a Mesh, it is another story...

dbarbier:
That is interesting.  If you only consider the `contains()` method, it is indeed easy to implement these operations, and even the complementary and all common boolean operations.  But Domains must also implement `isEmpty()` and `getVolume()` methods, and these ones are almost impossible to implement.

So, if `isEmpty()` and `getVolume()` methods have been added by convenience because you thought that all Domains will implement these methods, this was a mistake, because you put constraints on this class which prevent some interesting usages.  If those methods are needed, then join and intersect cannot be implemented.

dbarbier:
I just checked and everything builds just fine if Domain only implements `contains()` and `getDimension()` methods, except KarhunenLoeveQuadratureAlgorithm.cxx which also needs getLower/UpperBound.

Thus it would be much better to let Domain contain these 2 methods only, introduce a new class (say BoundingBox) which provides getLower/UpperBound methods, and add methods in Mesh and Interval (and maybe LevelSet?) to return a BoundingBox.

Thus I believe that this PR can be closed, we should first do this change, which seems more important to me.

dbarbier:
About Interval, my point is that there are cases when an Interval is used only for its endpoints, and the fact that it inherits from Domain is  ignored.  For instance when computing numerical integration.  Likewise for Mesh.  Thus by letting Interval and Mesh inherit from Domain, we introduce a coupling which is too strong for some use cases.

regislebrun:
LevelSet already implements the getLowerBound()/getUpperBound() methods by solving several optimization problems. It is not robust enough to be used blindly, but it works very well for smooth convex domains. Domain::getDimension() should be called at least in EventDomain and EventProcess, to check if the domain is of the same dimension as the element to be checked (value of a Process or a RandomVector). Another bug in the wall...

regislebrun:
Concerning the intersection of two meshes as a third mesh, one possibility is to define the indicator function of one of the meshes (in order to see it as a LevelSet) and to select from the other mesh the simplices fully inside the first mesh and to deform the simplices partly included. Something like LevelSetMesher::build() but based on another mesh than the mesh of the bounding box.

dbarbier:
I see, instead of simple and extensible code,  we have to make it complex, fragile and less maintainable, good idea.

regislebrun:
@dbarbier *please* keep cool. My last comment was just an idea to be discussed. Of course nobody prefer to have complex, fragile and less maintainable code instead of simple and extensible code! I am more and more convinced that these discussions in English on github are regularly a source of misunderstanding. My objective is by no means to have the last word, but I have to understand why a given construction is better than another. Most of the time I implement something for a specific purpose, then its usage goes far beyond the initial intention and things start to deteriorate: a given class becomes fat, its purpose is no more clear and your work on these points is *really* appreciated.
What are the precise points in my last comments that make you so upset? Because of the way bounds are computed in LevelSet? We could move this code in another place and allow the user to give a bounding box if you prefer. Because some calls to getDimension() are missing in some places? It is mostly my fault, and will be fixed asap. For the intersection of two meshes, it is mostly a joke: there must be some open source library with all we need concerning mesh intersection & co. I am not convinced at all that we really need it, a boolean arithmetic over events should be much more useful in our context. I am in Nantes on Monday and Tuesday, give me a call Wednesday afternoon if you can/want.

PR #656, opened 2018-02-01
New classes: TranslationEvaluation, TranslationFunction

dbarbier:
Use TranslationFunction in LevelSetMesher::build, it is faster than LinearFunction.


jschueller:
Needs entries in rst doc, else +1

regislebrun:
+1

PR #657, opened 2018-02-02
Improve mesh draw3d

regislebrun:
This PR provides several improvements:
  + the speed of Mesh::draw3d() has been improved
  + the shading code has been fixed wrt the specular part
  + the method LevelSetMesher::build() is now able to mesh only the boundary of the level set. It is not a true surface mesh but rather a thin set of simplices bordering the level set, for visualization purpose.


lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -501,7 +501,11 @@ void ResourceMap::loadDefaultConfiguration()
   // Mesh parameters
   setAsScalar( "Mesh-VertexEpsilon", 1.0e-12 );
   setAsUnsignedInteger( "Mesh-LargeSize", 5000 );
-  setAsUnsignedInteger( "Mesh-UseKDTree", 1 );
+  setAsBool( "Mesh-UseKDTree", true );
```
dbarbier: It must also be changed in resource file, or better yet, remove it, this entry is never used.

dbarbier:
There is a complex conflict with #654, I would like to merge it first.

lib/src/Base/Geom/LevelSetMesher.cxx
```diff
@@ -177,17 +181,41 @@ Mesh LevelSetMesher::build(const LevelSet & levelSet,
...
+	  if (numGood <= dimension) goodSimplices.add(currentSimplex);
+	  else
+	    {
+	      Bool addSimplex = false;
+	      Point localVertex(dimension);
+	      // For each vertex of the current simplex
+	      for (UnsignedInteger j = 0; j <= dimension && !addSimplex; ++j)
+		{
+		  const UnsignedInteger index = currentSimplex[j];
+		  localVertex = boundingVertices[index];
```
dbarbier: There is still a Point copy here, so your code is equivalent to using
```
  const Point localVertex(boundingVertices[currentSimplex[j]]);
```
which is more readable IMHO.
But it would be better to not use localVertex at all, and instead check `boundingVertices(index, k)`.

regislebrun: ok I will do that as well as all the other suggestions after the merge of PR#654.

lib/src/Base/Common/ResourceMap.cxx
regislebrun: @dbarbier It was there to prevent the mesh to build its kd-tree on a systematic basis, but forgot to propagate it in the code. Your solution is much better than this hack, so I will drop this entry.

dbarbier:
I do not understand the change in `LevelSetMesher::build`; if the only reason is to speed-up `Mesh::draw3D`, then this method should be fixed instead.  There are at least 2 easy improvements:
1. Display  only triangles which are oriented toward the front, this removes about half triangles
1. Do not display inner triangles (which allows rendering much larger meshes).

These changes will modify current rendering though if elements are shrunk.
See commit 81ce43b4.

dbarbier:
With your original mesh from `plot_mesh_3D.py` there are 274964 simplices and thus 1099856 triangles are drawn. With this PR and surface=true, there are 446016 triangles. 
With my commit (and surface reset to false):
 * !isInnerFace: 40244 triangles
 * isVisible: 549969 triangles
 * !isInnerFace && isVisible: 20163 triangles

regislebrun:
@dbarbier Thanks for your contribution which looks very promising. I was too lazy to implement back face culling because I imagined I would have to check if the orientation was consistent within a simplex in all the places (in fact, only in IntervalMesher). The primary purpose of the new flag in LevelSetMesher is... to be able to mesh the surface of a level set ;-)! We support lower dimensional meshes in OT (ie curves in 2D, curves and surfaces in 3D and so on) with the convention that a simplex is a lower dimensional element if its last vertices are repeated (see Mesh::exportToVTKFile()). I used this convention to avoid the implementation of a full-featured meshing library in OT while preserving most of the capabilities we need to define processes over arbitrary geometries.
BTW, the implementation is not complete as I would have to remove a face and to replace the remaining faces by simplices with a repeated vertex.
Do you keep this commit for a future branch or do you want to integrate it to this PR? If you recall me the relevant git command I do it ASAP (the next 2 days I will have a limited access to the internet).

jschueller:
Hi, one solution can be to cherry-pick his commit into your branch:
```
git remote add dbarbier https://github.com/dbarbier/openturns.git
git fetch dbarbier
git cherry-pick 81ce43b
```

dbarbier:
> [...] I imagined I would have to check if the orientation was consistent within a simplex in all the places [...]

I wrote 2 distinct methods; isVisible can be dropped if we accept ill-oriented simplices (which is a very bad idea IMO), it only removes half simplices, the real killer is isInnerFace.

> The primary purpose of the new flag in LevelSetMesher is... to be able to mesh the surface of a level set ;-)! We support lower dimensional meshes in OT (ie curves in 2D, curves and surfaces in 3D and so on)

I was not aware of that, and am pretty sure that this is useless, all methods that deal with simplices will only work if 3D meshes are composed of tetrahedra.

>  BTW, the implementation is not complete as I would have to remove a face and to replace the remaining faces by simplices with a repeated vertex.

IMO this new surface argument is useless until this implementation is finished, one may want a surface mesh (if there is a real justification for using surface meshes) but not a volumic mesh of the surface.

dbarbier:
There are glitches I do not understand with isVisible; with
```
view = View(graph, (800, 800), polygoncollection_kwargs={'edgecolors': 'yellow', 'linewidth': 0.1})
```
there are visual artifacts as if some foreground triangles were transparent.  Thus I suggest to drop it for now, and only keep isInnerFace, which seems to work very well.

regislebrun:
@dbarbier I agree with you concerning the flag in LevelSetMesher. It should probably be a distinct algorithm to build such meshes and trying to merge this algorithm with the other one would probably result into a terrible mess.
Concerning surface meshes, I developed this feature for the LOCOMACHS project, where I add to build a Karhunen-Loeve decomposition of processes defined over 3D surfaces. I needed it to load such meshes from ABAQUS, to export fields over such meshes in Paraview, and to build the KL decomposition of a process given by a process sample. It works quite well, as the following example demonstrates:
```
from openturns import *

Log.Show(Log.INFO)

size = 1000
N = 25
threshold = 1.0e-6

f = SymbolicFunction(["x", "y"], ["sin(_pi*x)*cos(_pi*y)"])
mesh2D = IntervalMesher([N]*2).build(Interval([0.0]*2, [4.0]*2))
vertices2D = mesh2D.getVertices()
values = f(vertices2D)
vertices3D = Sample(vertices2D.getSize(), 3)
for i in range(vertices2D.getSize()):
    vertices3D[i, 0] = vertices2D[i, 0]
    vertices3D[i, 1] = vertices2D[i, 1]
    vertices3D[i, 2] = values[i, 0]
simplices2D = mesh2D.getSimplices()
simplices3D = IndicesCollection(simplices2D.getSize(), Indices(4))
for i in range(simplices2D.getSize()):
    simplex3D = [simplices2D[i][0], simplices2D[i][1], simplices2D[i][2], simplices2D[i][2]]
    simplices3D[i] = simplex3D
    
mesh3D = Mesh(vertices3D, simplices3D)
mesh3D.exportToVTKFile("surface.vtk")

covariance = SquaredExponential([1.0]*3)
process = GaussianProcess(covariance, mesh3D)
sample = process.getSample(size)
# Here we have to provide the vertices weights as the automatic values are for true 3D meshes
weights = Point(mesh3D.getVerticesNumber(), 1.0 / mesh3D.getVerticesNumber())
algo = KarhunenLoeveSVDAlgorithm(sample, weights, threshold)
algo.run()
result = algo.getResult()
print(result.getEigenValues())
modes = result.getScaledModesAsProcessSample()
for i in range(10):
    modes[i].exportToVTKFile("mode" + str(i) + ".vtk")
```
so I will not say that the feature is useless. But I agree with you if you think that some algorithms should be revisited with this case in mind.

dbarbier:
Simplices are not used by KL, the result is the same if mesh3D is built without simplices, so you could achieve the same result with an extra argument (simplices) to exportToVTKFile.

regislebrun:
@dbarbier Right, but Mesh seemed to be a nice place to put those simplices. At one time I imagined to make a difference between what is currently named the mesh dimension, which is its embedding dimension (the dimension of the vertices) and its topological dimension and adapt the code (first, Mesh::contains()) to take this information into account. Do you think it is a good idea? Would it be better to differentiate meshes by their topological dimension and add a 3D surface mesh, a 3D curve mesh, a 2D curve mesh and so on? In my view we should be able to support meshes of arbitrary embedding and topological dimension as it is currently the case when they are equal, but I am no more sure it is a so good idea...

dbarbier:
Code dealing with simplices is buggy if dimensions are different (`contains`, `computeVolume`, `getNearestVertexAndSimplexIndicesWithCoordinates`, etc).  And since this feature does not have any useful usage, it should be avoided until a correct implementation is proposed.

regislebrun:
You mean, I should remove the exportToVTK() method too? Or change its implementation to take the degenerate simplices as an additional argument? Or add a check of the simplices (only the first indeed) in all the methods to throw an exception if the two dimensions are not equal? Many methods of Mesh are already used in the C++ prototype version of the Anatole software dealing with elastic parts, I would prefer to keep the current interfaces as much as possible...

dbarbier:
Add an overload `exportToVTK(const String & fileName, const IndicesCollection & simplices)`.  Simplices do not have to be degenerated then, you could even export simplices of different types.

dbarbier:
Still +1, please mention API changes in ChangeLog

regislebrun:
done.

jschueller:
The resource map keys are not added to openturns.conf, also you left some warnings in Mesh:
```
/home/schueller/projects/openturns/schueller2/lib/src/Base/Geom/Mesh.cxx:862:19: warning: declaration shadows a local variable [-Wshadow]
      const Point center((visuVertices[i0] + visuVertices[i1] + visuVertices[i2]) / 3.0);
                  ^
/home/schueller/projects/openturns/schueller2/lib/src/Base/Geom/Mesh.cxx:737:15: note: previous declaration is here
  const Point center(noRotation ? Point(3) : vertices_.computeMean());
              ^
/home/schueller/projects/openturns/schueller2/lib/src/Base/Geom/Mesh.cxx:909:13: warning: declaration shadows a local variable [-Wshadow]
      Point Ilight(3);
            ^
/home/schueller/projects/openturns/schueller2/lib/src/Base/Geom/Mesh.cxx:849:9: note: previous declaration is here
  Point Ilight(3);
        ^
/home/schueller/projects/openturns/schueller2/lib/src/Base/Geom/Mesh.cxx:919:15: warning: declaration shadows a local variable [-Wshadow]
      Polygon triangle(face);
              ^
/home/schueller/projects/openturns/schueller2/lib/src/Base/Geom/Mesh.cxx:743:11: note: previous declaration is here
  Indices triangle(3);
          ^
/home/schueller/projects/openturns/schueller2/lib/src/Base/Geom/Mesh.cxx:716:6: warning: unused function 'isVisible' [-Wunused-function]
Bool isVisible(const Point & visuVertex0, const Point & visuVertex1, const Point & visuVertex2)
```
Could you fix those please ?

PR #659, opened 2018-02-06
Deprecate several Domain methods

dbarbier:
Deprecate numericallyContains, isEmpty, isNumericallyEmpty, getVolume,
getNumericalVolume and computeVolume.

Some of these methods are defined in derived classes, when they make sense:
  * Interval: isEmpty, isNumericallyEmpty, numericallyContains,
              getVolume, computeVolume
  * Mesh: isNumericallyEmpty, getVolume, computeVolume

Rename DomainImplementation::volume_ member into volumeOld_ to avoid name
conflict with derived classes, likewise for isAlreadyComputedVolume_.

In order to support these deprecated methods, protected method
DomainImplementation::computeVolume has been modified to return a Scalar.
When it is dropped, {Mesh,Interval}::computeVolume can be modified again
to not return a value.


regislebrun:
+1

PR #662, opened 2018-02-08
[DO NOT MERGE] Drop KDTree services from Mesh, RegularGrid, Field

dbarbier:
<s>This is a fake pull request to show what I tried to explain in #654 and #661.

This branch is based on PR #661, only the last commit is relevant.  This is a work in progress, there are no new tests (but old tests use these new classes, and I did not change object representation to show that tests pass with minimal changes) and Python bindings are not finished yet, but C++ classes are almost complete.</s>

In #652, I added a method Mesh::computeKDTree to explicitly build a KDTree to speed-up neighbour searches.  The reason is that building this KDTree takes some time when mesh is large.  We could build it on demand, but then KDTree must be a mutable member, which may cause multithreading issues. Thus the reason for computeKDTree.

But in fact things become much simpler if these services are managed outside of Mesh.  If some code needs to find the nearest point within a Mesh, it builds a KDTree and queries it.

Code is much cleaner, we do not have to worry whether mesh is modified or not, this is the caller responsability to update its NearestNeighbor instance if it modifies mesh.
The only downside is that two pieces of code may build their own KDtree instead of reusing the same one, but I guess that in this case, a KDTree could likely be passed as argument.

And of course this is the same for simplex queries discussed in #661, it should go into its own class (maybe with several implementations).  And at the end, Mesh would only contain vertices and simplices, and provide only accessors, draw and export methods, this is all that is needed for this class.

Update: code is complete now, with tests and docstrings.


regislebrun:
@dbarbier It looks interesting but I can't see how far it will go. For example, the volume computation will go into a specific class? Something like: for each query, there is a dedicated structure to build and it is why it is delegated to a specific algorithm? It is a new way of building things in OT, do you have in mind some probabilistic classes to transform this way? Maybe the iso-probabilistic transformations?

dbarbier:
@regislebrun It is not new, it is exactly like other algorithms already in Algo: IntegrationAlgorithm integrates a Function onto an Interval, and different algorithms are provided.  Same here, nearest-neighbour search operates on an input Sample, and several algorithms can be provided.  Simplex query operates on a Mesh, and different algorithms can be provided.

Volume computations do not need any helper data structure, thus it is out of scope.

dbarbier:
Maybe the key point is that these algorithms *require* an appropriate data structure, otherwise computations are deadly slow. This was hidden in Mesh, but as a result, it was very hard to decide whether this helper data structure had to be built or not.  The situation is much clearer when it is explicit as is done here, and IMO it is still easy to understand for users.

jschueller:
RegularGridNearestNeighbour must be listed as new class in Changelog

lib/src/Base/Algo/openturns/RegularGridNearestNeighbour.hxx
```diff
@@ -0,0 +1,115 @@
...
+                              const Bool sorted = false) const;
+
+  /** Get the index of the nearest neighbour of the given scalar */
+  UnsignedInteger getNearestScalarNeighbourIndex(const Scalar & x) const;
+
+  /** Get the index of the nearest neighbour of the given scalars */
+  Indices getNearestScalarNeighbourIndex(const Point & x) const;
+
+  /** Get the nearest neighbour of the given scalar */
+  Scalar getNearestScalarNeighbour(const Scalar & x) const;
```
jschueller: We usually pass without reference (&) for single types, look at UnsignedInteger, Bool etc just above

dbarbier:
@jschueller Okay; there are conflicts because of #664 even if github does not detect them.  This was expected, I will push commits to fix your remarks, but this branch won't build for now.

dbarbier:
I pushed changes from db/BVH in this branch, here are two benchmarks.

First benchmark: Mesh based on an Interval.  Everything works fine since
points not contained in this mesh are instantaneously detected by bounding
box checks.

Test code for master

```
import openturns as ot
from time import time

N = 4000
size = 1000

mesh = ot.IntervalMesher([N]*2).build(ot.Interval([0.0]*2, [10.0]*2))
f = ot.SymbolicFunction(["x", "y"], ["x + 0.5*sin(y)", "y-0.1*x*sin(x)"])
mesh.setVertices(f(mesh.getVertices()))
sample = ot.ComposedDistribution([ot.Uniform(-2.0, 12.0)]*2).getSample(size)
print(mesh.getSimplicesNumber(), 'simplices', mesh.getVerticesNumber(), 'vertices')

t0 = time()
mesh.computeKDTree()
t1 = time()
print("build KDTree t=", t1 - t0, "s")

t0 = time()
res = mesh.contains(sample)
t1 = time()
print("res=", sum(res), "t=", t1 - t0, "s")
```

Test code for db/BVH branch
```
import openturns as ot
from time import time

N = 50
size = 100000

mesh = ot.IntervalMesher([N]*2).build(ot.Interval([0.0]*2, [10.0]*2))
f = ot.SymbolicFunction(["x", "y"], ["x + 0.5*sin(y)", "y-0.1*x*sin(x)"])
mesh.setVertices(f(mesh.getVertices()))
sample = ot.ComposedDistribution([ot.Uniform(-2.0, 12.0)]*2).getSample(size)

vertices = mesh.getVertices()
simplices = mesh.getSimplices()
print(mesh.getSimplicesNumber(), 'simplices')

print("NaiveEnclosingSimplex")
t0 = time()
naive = ot.NaiveEnclosingSimplex(mesh.getVertices(), mesh.getSimplices())
print("build speed=", time() - t0)

t0 = time()
result1 = naive.getEnclosingSimplexIndex(sample)
print("speed=", time() - t0)

print("BoundingVolumeHierarchy")
t0 = time()
bvh = ot.BoundingVolumeHierarchy(vertices, simplices)
print("build speed=", time() - t0)

t0 = time()
result2 = bvh.getEnclosingSimplexIndex(sample)
print("speed=", time() - t0)
print(result1 == result2)
```

Results:

|  N      | #simplices | build(s) | get(s) | build(s) | get(s) | build(s) | get(s) | get(s) size=100000
| - | - | - | - | - | - | - | - | -
| | <th colspan=2 align=center> master <th colspan=2 align=center> db/BVH Naive <th colspan=3 align=center> db/BVH BVH
| 50      |   5K       |  2.1e-3  |   0.07 |  5.1e-3  |  0.015 |  2e-3    | 5e-3   |   0.50
| 500     | 500K       |  0.31    |   3.35 |  0.63    |  1.03  |  0.25    | 5.8e-3 |   0.58
| 1000    |   2M       |  2.1     |  13.8  |  3.1     |  4.20  |  1.03    | 6.1e-3 |   0.58
| 2000    |   8M       | 11.2     |  50.9  | 14.6     | 16.8   |  4.39    | 6.7e-3 |   0.64
| 4000    |  32M       | 51.9     | 192.4  | 66.7     | 70.7   | 18.6     | 0.01   |   0.64


Second benchmark: complex Mesh, not convex.

Test code for master

```
import openturns as ot
from time import time

N = 62
size = 100

indices = [N]*3
factory = ot.LevelSetMesher(indices)
f = ot.SymbolicFunction(["x0", "x2", "x1"], ["(x0^8+x2^30+x1^8-(x0^4+x2^50+x1^4-0.3))*((x0^2+x1^2+x2^2+0.5^2-0.2^2)^2-4*0.5^2*(x0^2+x1^2))"])
ls = ot.LevelSet(f, 0.0)
mesh = factory.build(ls, ot.Interval([-1.2, -1.3, -1.5], [1.2, 1.3, 1.5]), True)

sample = ot.ComposedDistribution([ot.Uniform(-1.2, 1.2), ot.Uniform(-1.3, 1.3), ot.Uniform(-1.5, 1.5)]).getSample(size)
print(mesh.getSimplicesNumber(), 'simplices', mesh.getVerticesNumber(), 'vertices')

t0 = time()
mesh.computeKDTree()
t1 = time()
print("build KDTree t=", t1 - t0, "s")

t0 = time()
res = mesh.contains(sample)
t1 = time()
print("res=", sum(res), "t=", t1 - t0, "s")
```

Test code for db/BVH

```
import openturns as ot
from time import time

N = 62
size = 1000

indices = [N]*3
factory = ot.LevelSetMesher(indices)
f = ot.SymbolicFunction(["x0", "x2", "x1"], ["(x0^8+x2^30+x1^8-(x0^4+x2^50+x1^4-0.3))*((x0^2+x1^2+x2^2+0.5^2-0.2^2)^2-4*0.5^2*(x0^2+x1^2))"])
ls = ot.LevelSet(f, 0.0)
mesh = factory.build(ls, ot.Interval([-1.2, -1.3, -1.5], [1.2, 1.3, 1.5]), True)

sample = ot.ComposedDistribution([ot.Uniform(-1.2, 1.2), ot.Uniform(-1.3, 1.3), ot.Uniform(-1.5, 1.5)]).getSample(size)
print(mesh.getSimplicesNumber(), 'simplices', mesh.getVerticesNumber(), 'vertices')

vertices = mesh.getVertices()
simplices = mesh.getSimplices()

print("NaiveEnclosingSimplex")
t0 = time()
naive = ot.NaiveEnclosingSimplex(vertices, simplices)
print("build speed=", time() - t0)

t0 = time()
result1 = naive.getEnclosingSimplexIndex(sample)
print("speed=", time() - t0)

print("BoundingVolumeHierarchy")
t0 = time()
bvh = ot.BoundingVolumeHierarchy(vertices, simplices)
print("build speed=", time() - t0)

t0 = time()
result2 = bvh.getEnclosingSimplexIndex(sample)
print("speed=", time() - t0)
print(result1 == result2)
```

Results:

|  N    | #simplices | build(s) | get(s) | build(s) | get(s) | build(s) | get(s) | get(s) size=100000
| - | - | - | - | - | - | - | - | -
| | <th colspan=2 align=center> master <th colspan=2 align=center> db/BVH Naive <th colspan=3 align=center> db/BVH BVH
| 62    | 489K       |   0.10   |   1.12 |  0.51  |  0.75  |  0.31   | 7.4e-4 | 0.73
| 81    |   1M       |   0.26   |   2.44 |  1.19  |  1.63  |  0.72   | 7.6e-4 | 0.74
| 101   |   2M       |   0.57   |   4.63 |  2.31  |  3.10  |  1.39   | 8.0e-4 | 0.75
| 161   |  7.9       |   3.53   |  18.4  | 10.4   | 12.6   |  5.79   | 9.6e-4 | 0.77
| 231   |  23M       |  11.4    |  52.6  | 30.3   | 35.1   | 18.0    | 9.1e-4 | 0.79

dbarbier:
FYI I do not plan to work on this branch in a near future.  There is still room for improvements, but there are design choices which should be changed first:
* Mesh simplices are stored as a Collection of Indices, which is a very poor choice.  They should be stored contiguously, as has been done in EnclosingSimplexImplementation
* It is likely that Mesh::verticesToSimplices_ should be flattened too, but I did not benchmark.
* Mesh::contains from master prevents this branch from being merged.  I already told that this method should be dropped, maybe it becomes more obvious now.

Here are few comments about results above.  Code from NaiveEnclosingSimplex is very similar to Mesh::contains from master.  It takes longer to build for 2 reasons:
1. I wrote the scripts above before adding constructor with a Mesh.  Simplices are thus passed via SWIG, which checks each element of the Collection.
2. Simplices are flattened and copied into EnclosingSimplexImplementation.  If Mesh did store simplices already flattened, it is likely that build speed would be improved.

On the other hand, query location is faster with NaiveEnclosingSimplex.  Speed-up is 2.7 in the first case, but only 1.5 in the 2nd case.  I did not check exactly why it is faster, my guess is that reusing SquareMatrix in checkPointInSimplex makes the difference.

The winner is obviously BoundingVolumeHierarchy, query location is about 10000 times faster, and allows querying very large samples.  Build time is not impressive though.

regislebrun:
@dbarbier Do you want me to make the changes you suggest in PR#657? I had a look at EnclosingSimplexImplementation.hxx in db/BVH, and simplices_ is declared as an IndicesPersistentCollection. Is this the flattened storage you are talking about? I may have miss something here.

dbarbier:
@regislebrun Which changes?  PR#657 can be merged as is, IMO.
I forgot to push db/BVH, fixed now, but in fact db/BVH and this branch are the same now.

regislebrun:
@dbarbier I was talking about your proposal to remove Mesh::contains() and to use a flat storage for simplices. Where should these changes be made?

regislebrun:
and of course +1. The performances are astonishing.

dbarbier:
BTW performance of KDTree was suboptimal, query complexity was O(N) rather than O(log N).  This should be fixed by d534d54a.

dbarbier:
Superseded by #714

PR #664, opened 2018-02-11
Speed-up Mesh::contains

dbarbier:
Several tests have been added to Mesh::contains in order to exit early if
result is known to be false:
 * If point is outside Mesh bounding box, result is false
 * Simplices containing the nearest vertex are searched first; if point is
   not contained in any such simplices, all simplices are searched for.
 * checkPointInSimplexWithCoordinates() is modified too to return false
   early if point is not contained in simplex bounding bounds.  Those
   bounding boxes are precomputed.

### DIFF
--- a/lib/src/Base/Geom/Mesh.cxx
+++ b/lib/src/Base/Geom/Mesh.cxx
@@ -45,6 +45,8 @@ Mesh::Mesh(const UnsignedInteger dimension)
   , simplices_()
   , tree_()
   , verticesToSimplices_(0)
+  , lowerBoundingBoxSimplices_(0, dimension)
+  , upperBoundingBoxSimplices_(0, dimension)
   , isAlreadyComputedVolume_(false)
   , volume_(0.0)
 {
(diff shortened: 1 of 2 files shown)


regislebrun:
+1

jschueller:
It seems the doctest segfaults on the autobuilder

dbarbier:
@jschueller It has already been fixed, restart the autobuilder.

jschueller:
ok

regislebrun:
@dbarbier you don't like the parallel version of `Mesh::contains(const Sample & sample)`? It works quite well on my side.

dbarbier:
@regislebrun I do not like it because different threads (from TBB or user application for instance) may call getVerticesToSimplicesMap at the same time.  Until this is fixed, I prefer to not add parallelism here.  The solutions are either to add a mutex or move contains() into an helper data structure.

Moreover, your should also benchmark when this code is called from a TBB region, to make sure that parallelism does not hinder performance too much.

regislebrun:
ok.

PR #666, opened 2018-02-13
[RFC] Rename ProcessSample to FieldSample

jschueller:
Rationale: This object is more associated to the notion of Field (a collection of evaluations on a Field) rather than attached to the notion of Process (at least less directly)
Brought up by Michael

TODO: python alias+deprecation message for (deprecated in 1.10, removed in ot 1.12 [or 1.13]), changelog entry, deprecate buildAsPS methods too


regislebrun:
-1 As for Sample, Matrix or Point, I think we should make a clear distinction between the data container and its applicative interface. A Point is a linear algebra interface to a collection of scalars with an 1d access and a contiguous memory allocation. It allows to compute various norms, to add another point etc. A Sample is a statistical interface to a collection of scalars with a 2d access and a contiguous memory allocation, providing methods to compute moments etc. A Matrix is another interface to te same scalar container as Sample, but providing linear algebra services.
For the ProcessSample class, it is used both as a collection of Field and as a statistical interface to such a collection, providing different mean value computations (wrt the vertices or the fields), quantile computation etc. For now, there is no constraint on the memory pattern, and it is internally represented as a unique Mesh and a collection of Sample. The Process prefix is here to recall that it is a statistical interface, while a FieldCollection should be a more appropriate name for the underlying container. I am not against a name change but IMO it should be done on a broader range.

jschueller:
ok for the stat interface explanation, closing

PR #669, opened 2018-02-15
Predefined TimerCallback callback

jschueller:
Allows to define a maximum duration for simulation, optimization algorithms.

Uses system time (wall time), not cpu clock time.

### DIFF
--- a/python/src/Simulation.i
+++ b/python/src/Simulation.i
@@ -52,3 +52,45 @@ void setStopCallback(PyObject * callBack) {
 
 } // Simulation
 } // OT
+
+
+%pythoncode %{
+def TimerCallback(duration):
+    """
+    Time stopping criterion.
+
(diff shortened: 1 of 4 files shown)


regislebrun:
+1

python/src/Simulation.i
```diff
@@ -52,3 +52,45 @@ void setStopCallback(PyObject * callBack) {
...
+    >>> algo.setMaximumOuterSampling(int(1e9))
+    >>> algo.setMaximumCoefficientOfVariation(-1.0)
+    >>> algo.setStopCallback(TimerCallback(1.5))
+    >>> algo.run()
+    """
+    from time import time
+    try:
+        float(duration)
+    except ValueError:
+        raise ValueError('duration must be a float')
```
dbarbier: try/catch seems useless to me here since you catch/raise the same exception.

PR #670, opened 2018-02-16
Fix Saltelli estimator

jschueller:
The numerator constant of the Saltelli first order index was computed as mean of product of G(A), G(B) whereas it should be the product of means as recommended by saltelli02
This was not consistent with the documentation either where noted as G_0^2, so we explicited it also.
The variance is larger (the estimator is worse after), but it is not a duplicate of Mauntz anymore.

before:
g-sobol-saltellisensitivityalgorithm

after:
g-sobol-saltellisensitivityalgorithm


regislebrun:
@jschueller You mean, after your change the is estimator is worse than before and worse than another estimator available in OT? So why should we keep it in the library? If you remember well, I developed several orthonormalization algorithms and at the end I removed all but the best. We should probably do the same thing here...

jschueller:
After the change it's a bit worse, yes.
Some are best for first order indices some for total, etc.. there is no absolute best one.

regislebrun:
so +1

PR #672, opened 2018-02-16
Add TypedInterfaceObject::operator!=

dbarbier:
This operator is already defined in PersistentObject.hxx, but
interface classes inherit from TypedInterfaceObject.

Without this operator, we could have have (a==b) and (a!=b) both
return true (either in C++ or Python).


regislebrun:
+1

PR #675, opened 2018-02-16
Fix pr657 glitches

regislebrun:
Removed warnings on shadowed variables, added missing entries in openturns.conf and fixed unused function. Sorry.

### DIFF
--- a/lib/etc/openturns.conf.in
+++ b/lib/etc/openturns.conf.in
@@ -238,6 +238,7 @@
   <!-- OT::Mesh parameters -->
   <Mesh-VertexEpsilon   value="1.0e-12" />
   <Mesh-LargeSize       value="5000"    />
+  <Mesh-BackfaceCulling value="false"   />
   <Mesh-DiffuseFactor   value="0.7"     />
   <Mesh-AmbientFactor   value="0.1"     />
   <Mesh-SpecularFactor  value="0.2"     />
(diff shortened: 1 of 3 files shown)


regislebrun:
@dbarbier The code looks good and I suspect a bug in matplotlib, as the R output is fine. The use of isVisible() is deactivated by default, but I found sad to throw away good code because of the graphical backend troubles...

dbarbier:
> The use of isVisible() is deactivated by default

It is not, isVisible() is always called.

dbarbier:
I was wrong, but isInnerFace() is also not called by default, IMO it should.

regislebrun:
@dbarbier There is a typo in openturns.conf.in, Mesh-BackfaceCulling was false in ResourceMap and true in openturns.conf.in. I just fixed it. Otherwise, the test to check if a given face has to be drawn is correct. Either backfaceCulling is false, and isVisible is not evaluated because || is true, and the face is selected only if it is not an inner face, or backfaceCulling is true, so isVisible is evaluated and if it is true, then || is true and the face is selected only if it is not an inner face.

dbarbier:
@regislebrun Your code is equivalent to
```
  if ((!backfaceCulling) || (isVisible(visuVertex0, visuVertex1, visuVertex2) && !isInnerFace(simplicesVertex0, simplicesVertex1, simplicesVertex2)))
```

regislebrun:
@dbarbier Thanks for the explanation. I was thinking that the negation operator had a greater priority than || and &&, but that || and && had the same priority, with a left to right evaluation order. I added enough parentheses to make things as clear as an old lisp macro in emacs ;-)!

dbarbier:
Okay, looks good now.

PR #676, opened 2018-02-17
Add SmoothedUniform distribution

regislebrun:
The random mixture of a Uniform distribution and a Normal distribution is a classical way to model uncertainties resulting from both variability and lack of knowledge. This new distribution is exactly made for that: it is an efficient implementation of such a RandomMixture. As a side effect, it allows for more simplifications in the RandomMixture class.
This PR also provides an improved implementation of RandomMixture::computeScalarQuantile(), with a typical speed-up of 6x~10x when the random mixture is reasonably close to a unimodal distribution. This is the typical situation due to the smoothing effect of the convolution of PDFs.


lib/test/t_SmoothedUniform_std.expout
```diff
@@ -0,0 +1,64 @@
...
+Distribution class=SmoothedUniform name=SmoothedUniform dimension=1 a=-0.5 b=1.5 sigma=0.5
+Distribution SmoothedUniform(a = -0.5, b = 1.5, sigma = 0.5)
+Elliptical = true
+Continuous = true
+oneRealization=class=Point name=Unnamed dimension=1 values=[1.15307]
+oneSample first=class=Point name=Unnamed dimension=1 values=[1.3731] last=class=Point name=Unnamed dimension=1 values=[1.75997]
+mean=class=Point name=Unnamed dimension=1 values=[0.990093]
+covariance=class=CovarianceMatrix dimension=1 implementation=class=MatrixImplementation name=Unnamed rows=1 columns=1 values=[0.581443]
+Kolmogorov test for the generator, sample size=100 is rejected
+Kolmogorov test for the generator, sample size=1000 is rejected
```
jschueller: Maybe something wrong with getRealization ?

lib/src/Uncertainty/Distribution/SmoothedUniform.cxx
```diff
@@ -0,0 +1,436 @@
...
+}
+
+/* Parameters value accessor */
+Point SmoothedUniform::getParameter() const
+{
+  Point point(4);
+  point[0] = a_;
+  point[1] = b_;
+  point[2] = sigma_;
+  return point;
```
jschueller: Maybe you changed it to 3 parameters ?

lib/src/Uncertainty/Distribution/SmoothedUniform.cxx
```diff
@@ -0,0 +1,436 @@
...
+}
+
+/* Parameters description accessor */
+Description SmoothedUniform::getParameterDescription() const
+{
+  Description description(2);
+  description[0] = "a";
+  description[1] = "b";
+  description[2] = "sigma";
+  return description;
```
jschueller: 3 here too

lib/test/t_SmoothedUniform_std.expout
sofianehaddad: @jschueller agreed. getRealization and getSample differ (missing a_ in getSample)

lib/src/Uncertainty/Distribution/SmoothedUniform.cxx
```diff
@@ -0,0 +1,436 @@
...
+SmoothedUniform * SmoothedUniform::clone() const
+{
+  return new SmoothedUniform(*this);
+}
+
+/* Get one realization of the distribution */
+Point SmoothedUniform::getRealization() const
+{
+  return Point(1, a_ + (b_ - a_) * RandomGenerator::Generate() + sigma_ * DistFunc::rNormal());
+}
```
sofianehaddad: Seems that the translation factor a_ should be removed ? compare with getSample/computeMean

lib/src/Uncertainty/Distribution/SmoothedUniform.cxx
```diff
@@ -0,0 +1,436 @@
...
+
+/* Get a sample of the distribution */
+Sample SmoothedUniform::getSample(const UnsignedInteger size) const
+{
+  SampleImplementation result(size, 1);
+  const Point values(RandomGenerator::Generate(size) * (b_ - a_) + DistFunc::rNormal(size) * sigma_);
```
sofianehaddad: missing a_

regislebrun:
@sofianehaddad @jschueller Thanks for the review. There are many problems here:
+ a copy/paste from Uniform.cxx for getParameterDescription(), which should be of size 3 and not 2
+ a change of mind concerning SmoothedUniform, which was a 4 parameters distribution at the beginning and ended as a 3 parameters distribution (no need to have a nonzero mean for the Normal component), so the Point of dimension 4 in getParameter() and the missing a_ in getSample (removed when I removed mu_)
I fixed everything on my side but have troubles to interact with github from EDF...

sofianehaddad:
Could you also please add a test to `RandomMixture` that illustrate the changes ? Thanks

regislebrun:
@sofianehaddad The current test already illustrate this change ;-)!

jschueller:
t_Distribution_arithmetic fails on the autobuilder

PR #677, opened 2018-02-18
Fix wrong performance in KDTree

dbarbier:
Tree pruning was wrongly done, and as a result query complexity was 
O(N) instead of O(log N). 

Replace Point::normSquare() by explicit computations; in expression
  const Scalar localSquaredDistance = (x - sample_[localIndex]).normSquare();
one Point is created for sample_[localIndex], and a second one for (x - sample_[localIndex]).
Explicit norm computation avoids these two Point creations.

Replace Sample::operator[] by Sample::operator().

Remove KDTree::insert method.

Add KDTree::getNearestNeighbour on a Sample.

Also add an optional 'sorted' argument to getNearestNeighbours.


dbarbier:
I am not sure if/when #662 is going to be merged, so let us merge this change in a separate PR.

dbarbier:
Below are comparisons between scipy KDTree (pure Python), scipy cKDTree, OT master and this branch (via Python scripts).
We give the time to build KDtree (samples contain in both cases 1M, 2M, 4M and 8M points),
to query nearest neighbor for each point of a Sample of size 100000, and the time
to query the 1000 nearest neighbors of a point (in ms).

In dimension 2, tree contains N² elements

|               | N=1000 | N=1414 | N=2000 | N=2828
|-|-|-|-|-
|scipy KDTree   |   x1   |   x2   |   x4   |   x8
| build(s)      |  2.200 |  4.826 |  9.891 | 22.528
| query(s) 10^5 | 10.504 | 11.340 | 11.663 | 26.584
| 1000-NN(ms)   |  4.193 |  5.421 |  4.764 |  1.663
| | | | |
|scipy cKDTree  |   x1   |   x2   |   x4   |   x8
| build(s)      |  1.650 |  5.032 | 14.380 | 42.440
| query(s) 10^5 |  0.106 |  0.117 |  0.127 |  0.140
| 1000-NN(ms)   |  0.229 |  0.246 |  0.265 |  0.267
| | | | |
|OT master      |   x1   |   x2   |   x4   |   x8
| build(s)      |  1.974 |  4.863 | 10.119 | 24.571
| query(s) 10^5 |  3.869 |  7.624 | 14.172 | 28.447
| 1000-NN(ms)   |  1.024 |  1.085 |  1.845 |  1.626
| | | | |
|OT This PR     |   x1   |   x2   |   x4   |   x8
| build(s)      |  1.952 |  4.599 | 10.239 | 23.259
| query(s) 10^5 |  0.322 |  0.365 |  0.411 |  0.455
| k-NN(ms)      |  0.527 |  0.491 |  0.857 |  0.508

In dimension 3, tree contains N³ elements

|               | N=100  |  N=126 |  N=159 |  N=200
|-|-|-|-|-
|scipy KDTree   |   x1   |   x2   |   x4   |   x8
| build(s)      |  2.857 |  4.423 | 10.934 | 24.529
| query(s) 10^5 | 14.020 | 13.205 | 14.170 | 16.802
| 1000-NN(ms)   |  7.133 |  5.662 |  7.447 |  8.504
| | | | |
|scipy cKDTree  |   x1   |   x2   |   x4   |   x8
| build(s)      | 0.611  | 1.792  | 6.899  | 8.395
| query(s) 10^5 | 0.159  | 0.176  | 0.192  | 0.199
| 1000-NN(ms)   | 0.261  | 0.258  | 0.279  | 0.300
| | | | |
|OT master      |   x1   |   x2   |   x4   |   x8
| build(s)      |  2.122 |  4.572 | 10.607 | 25.795
| query(s) 10^5 | 12.749 | 21.640 | 38.066 | 64.329
| 1000-NN(ms)   |  1.858 |  2.230 |  3.109 |  5.321
| | | | |
|OT This PR     |   x1   |   x2   |   x4   |   x8
| build(s)      |  2.115 |  4.852 | 10.628 | 25.095
| query(s) 10^5 |  0.736 |  0.821 |  0.955 |  1.114
| 1000-NN(ms)   |  0.786 |  0.787 |  1.111 |  1.554

This benchmark is highly volatile, figures must be considered with care, but they clearly show that query time was problematic, and is much better now.

regislebrun:
+1 good job. Does scipy give te same answer as OT? There are many implementations of approximate nearest neighbors much faster than the exact ones... At one point, if this feature is critical we could consider the use of an external implementation such as this one: [https://github.com/jlblancoc/nanoflann](url), even if it is a C++11 implementation it should be possible to expose a pure C interface or something similar.

dbarbier:
@regislebrun OT KDtree returns exact nearest neighbors, not approximate, I modified t_KDTree_std.py in this PR to ensure that this is the case.  Same for scipy.KDTree.query method, unless eps argument is set to a positive value (of course I kept 0 for these tests). Thus results are the same, yes.

python/test/t_KDTree_std.py
```diff
@@ -2,16 +2,40 @@
...
+def nearest_debug_indices(x):
+    global sample_np
+    a = np.sum(np.square(sample_np - x), axis=1)
+    return a.argsort()
+
+neighbourIndices = tree.getNearestNeighbourIndex(test)
+
+neighbourIndices_np = map(nearest_debug, test)
+
+if neighbourIndices != neighbourIndices_np:
```
jschueller: On the autobuilder:
TypeError: Object passed as argument is not convertible to an Indices

This must be because map does not return a list in python3
I suggest `neighbourIndices_np = [nearest_debug_indices(x) for x in test]` which is also more readable than map

dbarbier:
Indeed, I always forgot about that :-/  Will fix it.

python/test/t_KDTree_std.py
dbarbier: Indeed, I always forgot about that :-/ Will fix it.

jschueller:
It still fails on the autobuilder, dont know why exactly, maybe the type returned by np.argmin.

dbarbier:
No idea, I added an explicit cast to Indices, please retry.

jschueller:
argmin returns numpy.int64 and it's not recognized, try to put an int(x) somewhere

dbarbier:
@jschueller Did you test my fix?

jschueller:
Yes, the cast to indices is useless, use `return int(np.argmin(a))`

dbarbier:
Okay, thanks, I will make this change.  Does nearest_debug_indices also have to be fixed?

jschueller:
No, I dont think I did see any further error.

PR #680, opened 2018-02-20
Try to convert Python exception value

jschueller:
Dont try to convert the exception value directly,
look if it has an str(obj) operator.
Allows to retrieve joblib exceptions.


jschueller:
cc @aurelieladier

regislebrun:
+1

jschueller:
TODO: backport to 1.10 branch

PR #681, opened 2018-02-21
New NearestNeighbourAlgorithm class

dbarbier:
This interface provides fast nearest neighbour lookup.
NearestNeighbourAlgorithmImplementation is an abstract implementation,
and 3 concrete implementations are provided:

  * NaiveNearestNeighbour is a brute-force implementation, it may be
    used if input sample size is very small, or in very high dimension.

  * KDTree now derives from NearestNeighbourAlgorithmImplementation.

  * RegularGridNearestNeighbour provides the same service for a
    RegularGrid, which is of course much lighter and faster.

All classes which previously were using a KDTree have now a
NearestNeighbourAlgorithm member, and algorithm can be changed by
calling setNearestNeighbourAlgorithm() method.

KDTree interface is simplified, there are now 3 main methods:
 * setSample(sample) builds the KDTree for this sample
 * query() on a Point or Sample returns nearest neighbour index
 * query_k() on a Point returns a collection of k nearest neighbour
   indices.  query_k() on a Sample is not provided yet.

With this framework, it is now possible to bind third-party libraries
for fast nerest-neighbour lookup, as has been done with OptimizationAlgorithm.

This commit also fix handling of cache in DatabaseEvaluation.


python/test/t_Mesh_std.expout
```diff
@@ -10,8 +10,8 @@ volume= 2.2
 First simplex volume= 1.0
 is p= [1.3]  in mesh?  True
 Nearest index( [1.8] )= 1
-Nearest index( [1.8] ), simplex and coordinates= [[1,1], class=Point name=Unnamed dimension=2 values=[0.5,0.5]]
-Nearest index( [-1.8] ), simplex and coordinates= [[0], class=Point name=Unnamed dimension=0 values=[]]
+Nearest index( [1.8] ), simplex and coordinates= [True, 1L, class=Point name=Unnamed dimension=2 values=[0.5,0.5]]
+Nearest index( [-1.8] ), simplex and coordinates= [False, 0L, class=Point name=Unnamed dimension=0 values=[]]
```
jschueller: 1L, 0L: You may want to cast to int here this is the formatting of the long type, which was removed in Python3

lib/src/Base/Algo/openturns/RegularGridNearestNeighbour.hxx
```diff
@@ -0,0 +1,91 @@
...
+  UnsignedInteger queryScalar(const Scalar x) const;
+
+  /** Get the index of the nearest neighbour of the given scalars */
+  Indices queryScalar(const Point & x) const;
+
+  /** Get the indices of the k nearest neighbours of the given point */
+  Indices query_k(const Point & x, const UnsignedInteger k, const Bool sorted  = false) const;
+
+  /** Get the indices of the k nearest neighbours of the given scalar */
+  Indices queryScalar_k(const Scalar x, const UnsignedInteger k, const Bool sorted  = false) const;
```
jschueller: Maybe queryK, queryScalarK rather than query_k, queryScalar_k

dbarbier: Good idea, thanks

regislebrun: + 1 with these changes

python/doc/user_manual/base_objects.rst
```diff
@@ -141,4 +141,6 @@ K-d tree
     :toctree: _generated/
     :template: class.rst_t
 
+    NearestNeighbourAlgorithm
+    RegularGridNearestNeighbour
```
jschueller: NaiveNearestNeighbour may be missing here

dbarbier: Indeed.  This section must also be renamed, it is currently "K-d tree", but this is no more accurate with these changes.  And EnclosingSimplex classes will be added there too.  I am thinking about "Spatial lookup", any other idea?

jschueller: Nope, in Scipy it's "Spatial algorithms and data structures":
https://docs.scipy.org/doc/scipy-0.14.0/reference/spatial.html

regislebrun: Spatial lookup sounds good.

dbarbier:
@jschueller Any problem with this PR?  @regislebrun approved these changes, but his comment disappeared after I rebased commits to fix conflicts.

jschueller:
There seems to be an issue on the autobuilder:
http://autobuilder.openturns.org/openturns/pull/681/squeeze-i386-ot-linux/LastTest.log

dbarbier:
This should be fixed by d27fb7c6, no idea why we did not encounter this problem earlier.  Do you want me to move this commit into a separate PR?

dbarbier:
BTW this is why I would like to remove mutable members from Mesh.

jschueller:
You can move d27fb7c here if it fails again.

regislebrun:
+1 one more time!

PR #682, opened 2018-02-21
Replace Function by SymbolicFunction in Python scripts

dbarbier:


dbarbier:
I did not realize that SymbolicFunction does not define outputVar, and this change is broken.  I do not have time to check whether 2nd argument can be blindly dropped or not, thus I am closing this PR.

PR #683, opened 2018-02-21
[RFC] New class IndicesFrozenCollection

dbarbier:
This is going to replace `Collection<Indices>` almost everywhere.
Collection is frozen, which means that Indices cannot be extended.

There is no accessor to Indices.  This in on purpose to avoid
creation of Indices inside loops.  Iterators must be used instead,
or values changed in-place.


dbarbier:
Name sucks, but I could not find a better one.  On the other hand, IndicesFrozenCollection has several advantages over `Collection<Indices>`:
1. It uses pimpl idiom, so when used as a return value, copy is avoided
1. Data locality is improved
1. There is on purpose no accessor to individual Indices, so that no Indexes are created when iterating over lists.  

Hopefully this class will avoid many allocations when dealing with large data structures.

dbarbier:
If there is no objection, the next step is to change `Mesh::simplices_` to this class, and I should then be able to make a PR for BoundingVolumeHierarchy.

regislebrun:
@dbarbier Frozen->FixedSize? Not a big deal otherwise (and it will recall us a well-known Disney cartoon). Ok for the change in Mesh from my side of course.

dbarbier:
An alternative is to fully drop IndicesCollection, and rename IndicesFrozenCollection into IndicesCollection.

regislebrun:
@dbarbier Up to now, all the XXXCollection exposed to the user in Python were able to be resized. I can see at least these options:
  * the IndicesCollection is an exception: it is the only Collection that cannot be resized
  * all the XXXCollection are hidden from the user, in favor of Python sequences. I don't know if it is possible to do that but it would be nice. The conversion to the fixed size collections is done internally.
  * all the XXXCollection now have a fixed size, so IndicesCollection is not an exception
I don't know if it has to be done right now. Maybe we could check that this change gives the expected performance boost and then decide to do make the change or not?

jschueller:
That's a non-issue: in Python we dont need any XXXCollection types (nor Points, Description, ...) as it's just easier to use native sequences (list, tuples).

lib/src/Base/Type/IndicesFrozenCollection.cxx
```diff
@@ -0,0 +1,137 @@
...
+  if (values.getSize() <= 1) return true;
+  const UnsignedInteger firstStride = values[0].getSize();
+  for(Collection<Indices>::const_iterator it = values.begin(), guard = values.end(); it != guard; ++it)
+  {
+    if (firstStride != it->getSize())
+      return false;
+  }
+  return true;
+}
+}
```
jschueller: Just a warning here, we use unity builds (we concatenate several .cxx in order to reduce build time) on the CIs, and functions in anonymous namespace might be merged from another .cxx file, so we have to keep in mind not to create conflicts.

dbarbier: Indeed, I'll prefix this function, thanks.

python/src/openturns/PythonWrappingFunctions.hxx
```diff
@@ -978,6 +978,125 @@ convert< Indices, _PySequence_ >(Indices inP)
...
+            {
+              indices( i, j ) = checkAndConvert<_PyInt_, UnsignedInteger>(elt.get());
+            }
+          }
+        }
+      }
+      catch (InvalidArgumentException &)
+      {
+        throw;
+      }
```
jschueller: The catch emits the same exception here, what did you have in mind ?

python/src/openturns/PythonWrappingFunctions.hxx
```diff
@@ -978,6 +978,125 @@ convert< Indices, _PySequence_ >(Indices inP)
...
+    // Check that object is a sequence
+    try
+    {
+      check<_PySequence_>( indicesObj );
+    }
+    catch (InvalidArgumentException &)
+    {
+      throw;
+    }
+    const UnsignedInteger dimension = PySequence_Fast_GET_SIZE( newPyIndicesObj.get() );
```
jschueller: here too

python/src/openturns/PythonWrappingFunctions.hxx
dbarbier: Copy/paste mistake; this was needed for MatrixImplementation and the like because a pointer had to be deleted, but this is not the case here, will fix it.

jschueller:
Would it be only used in Mesh, or do you think of other places that this could be useful ?

dbarbier:
IMO it should be used almost everywhere, for the same reasons than we use Sample instead of `Collection<Point>`.  I am still playing with this idea to evaluate it, and just pushed branch db/IndicesCollection in case you are intersted, this looks quite good to me.

dbarbier:
All suggested changes have been applied.  I will add commits to fully replace IndicesCollection, but not squashed with these ones, so that we can decide which version we adopt eventually.

dbarbier:
Here it is, now let us discuss if we want cba16bc or not.

jschueller:
So all classes making use of IndicesCollection are now frozen, like BipartiteGraph, and if we do want to make a class behave like before we have to explicitely use of `Collection<Indices>`
I dont see yet which impact this could have on our modules, but if it passes the change is fine I guess.

dbarbier:
@jschueller This is exactly the idea, yes.  Very few classes really need to change data size, thus it is better IMO to optimize for the most common use case.

dbarbier:
>  I dont see yet which impact this could have on our modules

According to https://github.com/search?q=org%3Aopenturns+IndicesCollection&type=Code IndicesCollection is only referenced from few scripts in doc, so hopefully there should be no breakage.

jschueller:
A rebase is needed here, I hope I fixed the autobuilder ssl error completely.

dbarbier:
Branch rebased, commits are squashed.
I still believe that IndicesCollection as  defined in the last commit is better, but it can be dropped  if we want to use IndicesFixedSizeCollection instead.

regislebrun:
+1 For the name I have no strong opinion, do what you think is the best.

jschueller:
IndicesCollection is ok for me too

dbarbier:
Just to be very clear, the problem I am trying to tackle in this PR is to avoid costful copies induced by methods dealing with `Collection<Indices>`.  This is why I first defined `IndicesFixedSizeCollection` as a `TypedInterfaceObject<IndicesFixedSizeCollectionImplementation>`.

But then the question was: do we change the return type of these methods to IndicesFixedSizeCollection, or do we add new methods, eg. `Mesh::GetFixedSizeCollectionSimplices`?  I tried both solutions, but at the end it looked simpler to me to just rename IndicesFixedSizeCollection into IndicesCollection and replace `Collection<Indices>` by IndicesCollection everywhere.

As explained by @jschueller in a previous comment, there is no reason to use XXXCollection in Python scripts.

jschueller:
Clear for me. I've rerun the autobuilder. Once it passes it should be merged.

regislebrun:
It was also what I had in mind. Ok for me, I will adapt my scripts afterward.

PR #684, opened 2018-02-23
Add SymbolicParser::operator()(Sample)

dbarbier:
This speeds up SymbolicEvaluation::operator()(Sample).

On a script Python with a Sample of size 32M and dimension 3, evaluation takes 1.5s instead of 5.2s.  I did not make more benchmarks, I expect gain to be larger with C++, but smaller with smaller size, which is more typical.

### DIFF
--- a/lib/src/Base/Func/SymbolicEvaluation.cxx
+++ b/lib/src/Base/Func/SymbolicEvaluation.cxx
@@ -111,11 +111,15 @@ Point SymbolicEvaluation::operator() (const Point & inP) const
 /* Operator () */
 Sample SymbolicEvaluation::operator() (const Sample & inS) const
 {
-  UnsignedInteger size = inS.getSize();
-  Sample outSample(size, getOutputDimension());
-  for (UnsignedInteger i = 0; i < size; ++ i) outSample[i] = operator()(inS[i]);
-  outSample.setDescription(getOutputDescription());
-  return outSample;
+  Sample result(parser_(inS));
+  result.setDescription(getOutputDescription());
+  callsNumber_ += inS.getSize();
+  if (isHistoryEnabled_)
+  {
+    inputStrategy_.store(inS);
+    outputStrategy_.store(result);
+  }
+  return result;
 }
 
 /* Accessor for input point dimension */
--- a/lib/src/Base/Func/SymbolicEvaluation.cxx
+++ b/lib/src/Base/Func/SymbolicEvaluation.cxx
@@ -112,13 +112,13 @@ Point SymbolicEvaluation::operator() (const Point & inP) const
 Sample SymbolicEvaluation::operator() (const Sample & inS) const
 {
   Sample result(parser_(inS));
-  result.setDescription(getOutputDescription());
   callsNumber_ += inS.getSize();
   if (isHistoryEnabled_)
   {
     inputStrategy_.store(inS);
     outputStrategy_.store(result);
   }
+  result.setDescription(getOutputDescription());
   return result;
 }
 
--- a/lib/src/Base/Func/SymbolicParser.cxx
+++ b/lib/src/Base/Func/SymbolicParser.cxx
@@ -81,8 +81,8 @@ void SymbolicParser::initialize() const
 {
   const UnsignedInteger inputDimension = inputVariablesNames_.getSize();
   const UnsignedInteger outputDimension = formulas_.getSize();
-  inputStack_ = Point(inputDimension);
   if (parsers_.getSize() == outputDimension) return;
+  inputStack_ = Point(inputDimension);
   parsers_ = Collection<Pointer<MuParser> >(outputDimension);
   try
   {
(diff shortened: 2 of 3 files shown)


lib/src/Base/Func/SymbolicEvaluation.cxx
```diff
@@ -112,6 +112,7 @@ Point SymbolicEvaluation::operator() (const Point & inP) const
 Sample SymbolicEvaluation::operator() (const Sample & inS) const
 {
   Sample result(parser_(inS));
+  result.setDescription(getOutputDescription());
   ++ callsNumber_;
```
jschueller: I just noticed that everywhere else we use callsNumber_ as the number of point evaluations:
```
$ grep -nr 'callsNumber_ +=' lib/src/
lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingGradient.cxx:158:  callsNumber_ += p;
lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx:241:  callsNumber_ += size;
lib/src/Uncertainty/Algorithm/OrthogonalBasis/SoizeGhanemFactorEvaluation.cxx:141:  callsNumber_ += size;
lib/src/Base/Algo/KarhunenLoeveProjection.cxx:109:  callsNumber_ += outSample.getSize();
lib/src/Base/Func/FieldToPointFunctionImplementation.cxx:147:  callsNumber_ += size;
lib/src/Base/Func/BoxCoxEvaluation.cxx:141:  callsNumber_ += size;
lib/src/Base/Func/IdentityEvaluation.cxx:91:  callsNumber_ += inS.getSize();
lib/src/Base/Func/InverseBoxCoxEvaluation.cxx:136:  callsNumber_ += size;
lib/src/Base/Func/ProductPolynomialEvaluation.cxx:178:  callsNumber_ += size;
lib/src/Base/Func/LinearEvaluation.cxx:136:  callsNumber_ += size;
lib/src/Base/Func/P1LagrangeEvaluation.cxx:229:  callsNumber_ += size;
lib/src/Base/Func/ProductEvaluation.cxx:103:  callsNumber_ += size;
...
```

dbarbier:
There are still problems with result Description, I do not understand why this is not fixed now.

dbarbier:
Commits are squashed together to help reviewing.

dbarbier:
> I was also thinking about using one single parser for all output as it supports multiple return values

If I understand correctly, bulk evaluation only works for scalar functions, we should try to use that instead of multiple outputs.

regislebrun:
+1 There is also a bulk mode in muparser [http://beltoforion.de/article.php?a=muparser&hl=en&p=eval&s=idEvalBulk#idEvalBulk](url), which, coupled with openmp [http://beltoforion.de/article.php?a=muparser&p=compiler_switches](url), is supposed to provide a significant speed-up on complex formulas. I never figured out how to use it but if you feel you can get it work it should give another boost...

jschueller:
@dbarbier Maybe I was unclear ; I was not talking about evaluating a Sample, but to evaluating the multiple ouputs of a single point ; Here are the apis:
```
1    value_type  Eval() const;
2   value_type* Eval(int &nStackSize) const;
3    void Eval(value_type *results, int nBulkSize);
```
Current one is 1, I was talking about 2, Régis and you about 3

dbarbier:
@jschueller Yes, it was clear for me.  I am not sure that 2 provides any speed up, the same operations are performed.  And 3 is wrongly implemented in muparser, it should use SSE to perform the same instructions of an array of inputs, and not openmp threads. 

And there is indeed a muparserSSE project, maybe we could try it.

jschueller:
I've not tried 2, I don't know if using a single parser is more efficient but multiple outputs are nice.

regislebrun:
If we want to improve the SymbolicFunction performance, one way is to change the back-end library. I started to separate SymbolicParser from MuParser in an old branch (exptk) because of this benchmark:
[https://github.com/ArashPartow/math-parser-benchmark-project](url)
The author of the benchmark is also the author of the winning library, but the comparison looks fair enough to make our opinion.

dbarbier:
I played with your branch, it seems to work fine.  There is a failure with derivative of abs, no idea why yet.

regislebrun:
I didn't worked on this branch for a long time, I am not sure of its state. My main concerns are:
  * how to deal with a third-party dependency which is a one-file header with no official release and no packaging (at least in my linux distro)
  * how to enhance exptk with additional elementary functions as I did with MuParser?
Concerning the last point, maybe we could stick on the default functions of each back-end? It should cover most of the needs.
Another point is the interaction with Ev3: some elementary functions are present with different names in different parsers, so how Ev3 should deal with these functions?

dbarbier:
I added these functions, the only remaining issues are:
1. Constant names cannot start with an underscore, thus I added e_ and pi_.
1. Boolean operators on floats are not supported.

IMO these are very minor.  OTOH compilation time may be an issue, exprtk.hpp is huge.

PR #685, opened 2018-02-26
New analytical parser based on ExprTk

dbarbier:
This branch provides an analytical parser based on ExprTk.
It is a proof of concept, its integration must be discussed:

 * For convenience, exprtk.hpp has been included in sources, but it is surely better to use a CMake variable to define its location
 * Do we want to provide both muParser and ExprTk parsers, as has been done here, or do we want to provide a single backend?
 * If we provide both backends, which one do we use by default?


jschueller:
Reopened to test the autobuilder got the webhook trigger .

regislebrun:
When I started this development I had three points in mind:
  * To insure that OT has a SymbolicFunction in order to use it in all the tests and in some specific algorithms. Today SymbolicFunction depends on MuParser, which is an optional feature;
  * To provide the most efficient implementation for SymbolicFunction, as it allows to define a significant subset of functions defined using PythonFunction, for a 10x more efficient evaluation;
  * To allow for the use of exotic parsers, such as a CUDA based parser

All these points were discussed during technical committees. The use of exprtk looks like a good choice as it is the winner of many benchmarks, is a single file to include and thanks to your job is now as functional as the MuParser binding. If we agree on the first point, maybe we should simply add exprtk to our source tree as we did with many other dependencies (cobyla, tnc, dsfmt, ev3, faddeeva...).

regislebrun:
What I propose:
  * we embed exprtk.hpp into our source tree and is used as the default backend if muparser is not available
  * we keep muparser at least because there are small differences which could break user's code
  * muparser is deprecated as the default backend

Do you agree on this? Any comment? I will do some benchmarks ASAP.

dbarbier:
@regislebrun I implemented what you suggest, and will have to rebase this branch to fix conflict in ChangeLog.  But I do not understand your last point, how can we deprecate muParser as the default backend, but still make it the default backend?

regislebrun:
@dbarbier My objective is to deprecate muparser in favor of exprtk, but we cannot do this in one shot. Usually, we have the two features in parallel for a full release then we remove the old feature. In this case, I suggest to have both exprtk and muparser for the 1.11 release, with muparser as the default parser and a warning saying that it will be replaced by exprtk as a default in the 1.12 release. If you think it is not worth the burden we can silently replace muparser by exprtk as the default parser.

jschueller:
I'd prefer to rely on the system providing the header, specially when development seems to continue, also it's huge.
I'll have to make sure it compiles fine on both MinGW flavours.

dbarbier:
Do we want to remove muParser in 1.12?
IMO we can keep it to provide an alternative if ExprTk chokes on some formulas, these SymbolicParser* wrappers are easy to maintain.  ExprTk and muParser give the exact same results (modulo bugs), so users should not notice this change, except for _e and _pi which must be replaced by e\_ and pi\_.

@jschueller If exprtk.hpp is not stored in our tree, how do we manage these analytical parsers?  Do we let CMake check for ExprTk and muParser?  And which one is the default?

jschueller:
We must use a FindExprtk, for the default I don't know.

regislebrun:
Excepted its size, what is the difference between exprtk.hpp and, say, incgamma.cxx, which is included into our source tree? A CMake macro is fine, but to detect what? An official release of exprtk? There is no such thing in this project!

dbarbier:
I pushed commits to implement multiple return values, this topic had been raised several times in the past.

This works pretty well for ExprTk, but is almost useless for muParser, which is not designed to handle complex statements.

My first try was to use return statement, but it is very slow.  The last commit is much more interesting.

dbarbier:
@regislebrun When we introduce new features and deprecate old ones, we always switch to the new one by default, this is why I prefer to have ExprTk by default.  If things broke, users will have to switch back to muParser via ResourceMap.

dbarbier:
In newer commits, I dropped support for ExprTk return statement, it is too slow, and likely to be never used.  I also removed support for muParser multiple evaluations, muParser only works for simple statements.

Thus at the end, this PR only adds a SymbolicFunction constructor
```
  SymbolicFunction (const Description & inputVariablesNames,
                    const Description & outputVariablesNames,
                    const String & formula);
```
which can be instantiated by ExprTk, but not muParser.  But I do not know how to document its usage, it is rather arcane.

CMake files have been modified:
* new option to use ExprTk, ON by default
* muParser option is kept
* if both muParser and ExprTk are available, ExprTk is set as the default parser

For now, there is no deprecation warning when using muParser, i will add one if we decide to remove muParser in a future release.

regislebrun:
@dbarbier Concerning the usage of the new interface, you could briefly describe this feature (the ability to return multiple output) and say that the syntax is parser dependent, with a link to the ExprTK documentation for the exact syntax.
At one point, we will have to reconsider the scope of the simplified Python wrapper. With the ExprTK extended syntax, most of the simple codes able to be wrapped into a single Python function can be expressed using a SymbolicFunction for a much more efficient evaluation.

I will love to see this feature in OT!

ArashPartow:
@regislebrun 

> An official release of exprtk? There is no such thing in this project!

ExprTk does have an official release. It is mirrored in several locations.  It can always be downloaded from the website or the tip of the [github](https://github.com/ArashPartow/exprtk/)

The library's interface has been locked in for nearly a decade now. Furthermore I guarantee any changes/updates/fixes will not effect user's expressions - aka if an expression was valid before, it will be valid and compilable after

In short I would recommend using a git submodule, like so: https://github.com/LMMS/lmms/tree/master/plugins/Xpressive

ArashPartow:
@regislebrun  

> Concerning the usage of the new interface, you could briefly describe this feature (the ability to return multiple output) and say that the syntax is parser dependent, with a link to the ExprTK documentation for the exact syntax.


There is a project that provides the user with either a Python or ExprTk based scripting interface. 

The project is called Natron and the relevant documentation can be found here: http://natron.readthedocs.io/en/master/guide/compositing-exprs.html

Perhaps something similar to this approach?...

lib/src/Base/Func/SymbolicParserExprTk.cxx
```diff
@@ -120,7 +121,7 @@ void SymbolicParserExprTk::initialize() const
   symbol_table.add_function("rint", ExprTk_rint);
   symbol_table.add_function("ln", std::log);
   symbol_table.add_function("lngamma", lgamma);
-  symbol_table.add_function("gamma", tgamma);
```
ArashPartow: **Note:** Free functions should be viable option, in this case tgamma should work.

Documentation: https://github.com/ArashPartow/exprtk/blob/master/readme.txt#L2347

Example of a free function being used: https://github.com/ArashPartow/exprtk/blob/master/exprtk_simple_example_05.cpp#L64

dbarbier: @ArashPartow Many thanks for your comments (and for ExprTk which is really great), I will have a closer look.  I had to replace tgamma because of build failures on Windows, see https://ci.appveyor.com/project/openturns/openturns/build/1.0.1785/job/62kb7qp5jil9b1fe

dbarbier:
@regislebrun Python functions can be almost as fast as ExprTk, the problem is that we are converting several times between OT and Python objects.

Try my branch db/PythonFunction with Michael's example:
```
def functionCrue(X) :
    Q, Ks, Zv, Zm, Hd, Zb, L, B = X
    alpha = (Zm - Zv)/L
    H = (Q/(Ks*B*sqrt(alpha)))**(3.0/5.0)
    Zc = H + Zv
    Zd = Zb + Hd
    S = Zc - Zd
    return [H,S]

def functionCrueSample(X) :
    a = np.array(X, copy=False)
    Q, Ks, Zv, Zm, Hd, Zb, L, B = a.T
    alpha = (Zm - Zv)/L
    H = (Q/(Ks*B*np.sqrt(alpha)))**(3.0/5.0)
    Zc = H + Zv
    Zd = Zb + Hd
    S = Zc - Zd
    return np.concatenate((H[:,np.newaxis],S[:,np.newaxis]),axis=1)

myFunctionPython = ot.PythonFunction(8, 2, functionCrue, functionCrueSample)
```

regislebrun:
Ouppsss! Very sorry for the mistake and thanks again for your good job! I was relying on the 'releases' tab in github, which is empty.
To make things clear: I hope we will gain more than an order of magnitude in performance wrt the binding of Python functions with your parser, which is GREAT!
A++
Régis
    Le dimanche 4 mars 2018 à 23:47:34 UTC+1, Arash Partow <notifications@github.com> a écrit :  
 
 
@regislebrun


An official release of exprtk? There is no such thing in this project!


ExprTk does have an official release. It is mirrored in several locations. It can always be downloaded from the website or the tip of the github

The library's interface has been locked in for nearly a decade now. Furthermore I guarantee any changes/updates/fixes will not effect user's expressions - aka if an expression was valid before, it will be valid and compilable after

In short I would recommend using a git submodule, like so: https://github.com/LMMS/lmms/tree/master/plugins/Xpressive

COPYING.exprtk
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included
+in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```
jschueller: This file should be installed with the others, see bottom of root CMakeLists.

dbarbier: Oops, fixed now, thanks

dbarbier:
Does adding  `PlatformInfo::SetNumericalPrecision(4)` fix this issue?

jschueller:
I cannot test it easily myself, but it seems like a good idea.

dbarbier:
Too bad, it doesn't work.

jschueller:
Maybe try to just comment the print(graph1d) line.

jschueller:
Still fails. Maybe use a simpler function, polynomial?

dbarbier:
This one should work, there is no 0 except a trivial one.

jschueller:
The autobuilder passes, congrats! This is ready to merge, @regislebrun ?

PR #686, opened 2018-02-27
Fix {KDTree,NaiveNearestNeighbour}::__str__ definition

dbarbier:
It must match `NearestNeighbourAlgorithmImplementation::__str__`.

Also fix a compiler warning about unused argument in
`NearestNeighbourAlgorithmImplementation::__str__`.


jschueller:
+1

PR #688, opened 2018-02-27
New EnclosingSimplexAlgorithm class

dbarbier:
EnclosingSimplexAlgorithm is a new interface to provide fast point
location; the query() method returns the index of the simplex
enclosing the point(s) passed as argument.

EnclosingSimplexAlgorithmImplementation is an abstract implementation,
and two concrete classes are provided:

 * NaiveEnclosingSimplex uses a brute force algorithm; it creates
   a NearestNeighbourAlgorithm to find the nearest vertices, and
   check simplices in their vicinity.  If not found, check all
   simplices.
   This method is very efficient for Interval, but it can be very
   slow if all simplices have to be looked for.

 * BoundingVolumeHierarchy is much more robust.  It is based on
   Bounding Volume Hierarchy.  Only mean and median splitting
   are implemented, not Surface Area Heuristic, which is more
   complex.


dbarbier:
These  are the same classes  as in #662; performance should be better thanks to the changes with IndicesCollection, but I do not have access to my desktop machine to provide benchmarks.

dbarbier:
@jschueller No, at the moment I cherry-picked uncontroversial changes from my first proposals, by adding NearestNeighbourAlgorithm and EnclosingSimplexalgorithm classes.

My initial proposal was to remove entirely those services from Mesh.  Maybe we can keep  nearestNeighbour_ and enclosingSimplex_ members in this class, I will check.

regislebrun:
@dbarbier In the spirit of your changes, we could remove Mesh::computeP1Gram() and add a new hierarchy to provide such computations (P0Gram, P2Gram, binding with FE libraries)
Otherwise +1 for this PR

PR #689, opened 2018-02-28
Added Distribution::computeEntropy()

regislebrun:
The main point of this PR is to provide a new method to compute the entropy of any distribution (the true entropy for discrete distribution, the differential entropy for the continuous ones). This service is needed eg to detect conditional independence based on 3 points relative information.
The PR is organized as follow:
  * a commit provides a reasonably efficient generic implementation of computeEntropy() in DistributionImplementation
  * a (huge) commit add a specific implementation of this method for most of the usual distributions
  * a series of minor commits improve or fix some bugs encountered during the preceding developments. The most important fix is in Wishart::computeLogPDF(), where the log-normalization factor was added before a scaling instead of after, resulting in erroneous values.


lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
```diff
@@ -1334,6 +1337,79 @@ protected:
     const DistributionImplementation * p_distribution_;
   }; // class ConditionalCDFWrapper
 
+  // Class used to compute entropy
+  class EntropyKernel: public FunctionImplementation
```
jschueller: This should go to the cxx file, or swig will try to wrap it (or at least be private).

lib/src/Uncertainty/Distribution/ChiSquare.cxx
```diff
@@ -185,6 +185,12 @@ Scalar ChiSquare::computeComplementaryCDF(const Point & point) const
   return DistFunc::pGamma(0.5 * nu_, 0.5 * x, true);
 }
 
+/* Compute the entropy of the distribution */
+Scalar ChiSquare::computeEntropy() const
+{
+  return 0.5 * nu_ + M_LN2 + SpecFunc::LogGamma(0.5 * nu_) + (1.0 - 0.5 * nu_) * SpecFunc::Psi(0.5 * nu_);
```
dbarbier: Sign error?
```
return 0.5 * nu_ + M_LN2 + SpecFunc::LogGamma(0.5 * nu_) - (1.0 - 0.5 * nu_) * SpecFunc::Psi(0.5 * nu_);
```

lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
```diff
@@ -195,6 +195,12 @@ void FisherSnedecor::computeMean() const
   isAlreadyComputedMean_ = true;
 }
 
+/* Compute the entropy of the distribution */
+Scalar FisherSnedecor::computeEntropy() const
+{
+  return std::log(d2_ / d1_) + SpecFunc::LogBeta(0.5 * d1_, 0.5 * d2_) + (1.0 - 0.5 * d1_) * SpecFunc::Psi(0.5 * d1_) + (0.5 * (d1_ + d2_)) * SpecFunc::Psi(0.5 * (d1_ + d2_)) - (1.0 + 0.5 * d2_) * SpecFunc::Psi(0.5 * d2_);
```
dbarbier: Shouldn't it be
```
std::log(d1_  / d2_)
```
instead?

jschueller: You did not add the numerical test using computeLogPDF for every distributions, I guess that's why these were not checked.

lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
regislebrun: @jschueller You mean, with all the other methods bindings? Fun...

lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
regislebrun: @jschueller I agree that some methods are not fully tested, but IMO it should go into another PR: this one is large enough and focused on computeEntropy(). BTW, it could be a shared work...

lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
jschueller: Or at the very least not be public. My concern is swig trying to wrap it.

lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
jschueller: I'm only referring to the test of computeEntropy:
```
fullprint << "entropy=" << distribution.computeEntropy() << std::endl;
fullprint << "entropy (MC)=" << -distribution.computeLogPDF(distribution.getSample(1000000)).computeMean()[0] << std::endl;
```

lib/src/Uncertainty/Distribution/ChiSquare.cxx
regislebrun: @dbarbier It all depends on which wikipedia page you read ;-). If you read this:
[https://en.wikipedia.org/wiki/Differential_entropy#Differential_entropies_for_various_distributions](url)
you have a minus sign, and if you read this:
[https://en.wikipedia.org/wiki/Chi-squared_distribution](url)
you have a plus sign!
The best way I found to check if it is correct is to compare the entropy of a distribution d with:
```
-d.computeLogPDF(d.getSample(100000000)).computeMean()[0]
```
If the two values are close enough (5 figures or more) it is probably correct. In this case, the winner is the plus version, so no error here.

lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
regislebrun: @dbarbier There is an error... but in the formula given by wikipedia! One more... If you want a reliable source to check the formulas, you could have a look here:
[https://books.google.fr/books?id=k2fOBQAAQBAJ](url)
and if you look here:
[https://books.google.fr/books?id=k2fOBQAAQBAJ&printsec=frontcover&hl=fr&source=gbs_atb#v=onepage&q=fisher&f=false](url)
you get the formula in the code.

regislebrun: @dbarbier note that I really appreciate your careful reading. Even if I did my best to check things, mistakes are likely to happen in so many formulas.

regislebrun: @jschueller Ok I got it. I will add this test in all the C++ tests. I skipped this check in the Python tests in order to reduce the testing time. If you think I should add these checks here too just let me know.

dbarbier:
@regislebrun In fact my concern is the same as written by @jschueller, all tests with computeEntropy() should also display output of
```
-d.computeLogPDF(d.getSample(100000000)).computeMean()[0]
```

lib/src/Uncertainty/Distribution/FisherSnedecor.cxx
dbarbier: No, C++ is enough to check computations, thanks.
BTW in the book about Fisher distribution, there is a log2 factor; we really need some automated tests to have more confidence in these formulas.

regislebrun: @dbarbier It is because the entropy can be computed according to any positive scale a, the base of the logarithm used in the definition of the entropy. The two most common choices are a=2, mainly in information theory (the entropy is given in bits), and a=e in probabilities (the entropy is given in nats, ie NATural logarithm).

jschueller:
All the outputs were added to the cxx tests, thanks.

regislebrun:
Sorry, I pushed this morning in a hurry and forgot to comment the changes. I found two bugs on the way: the VonMises distribution was wrongly supported by [-\pi, \pi] instead of [\mu-\pi, \mu+\pi] and the contant value of the log-PDF of KPermutationsDistribution was equal to -log(N!/K!) instead of -log(N!/(N-K)!). This time, it was the entropy which helped to debug other methods!
I will make the other changes this afternoon.

lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
regislebrun: @jschueller I don't understand why SWIG is trying to do something here. The definition is in the protected section of the definition of the class, and is supposed to be hidden to SWIG by a preprocessor conditional directive (#ifndef SWIG / #endif). Do you have an explanation why SWIG still try to do something here? It should never see this code!

jschueller: Sorry, I thought it was in the public section.
SWIG does not parse protected sections, so the ifdefs should be useless here.

jschueller:
On windows, vc10 a tests fails:
```
t_KPermutationsDistribution_std.py <type 'exceptions.IndexError'> Error in function boost::math::lgamma<long double>(long double): Evaluation of lgamma at 0.
```

regislebrun:
I am working on it. Computing multivariate quantiles is tricky for discrete distributions.

regislebrun:
I added a specific computeQuantile() method to KPermutationsDistribution, which avoids totally the construction of 1D marginal distributions (the only way for KPermutationsDistribution to call SpecFunc::LogGamma). Not clear yet why it was called with argument==0.

jschueller:
A last problem: t_NoncentralChiSquare fails on the autobuilder, its result may be rounded up, also you did not add the numerical value for this one.

jschueller:
The test still fails :/

regislebrun:
I hope this time it will work!

jschueller:
Good to merge, @dbarbier ?

dbarbier:
Sorry, I thought I already approved this PR.

PR #690, opened 2018-02-28
Improve viewer

regislebrun:
This PR allows to force orthonormal axes and to draw the axes according to the value of Graph::getAxes(). It also provides a better implementation of color conversion in Drawable.

### DIFF
--- a/lib/src/Base/Graph/DrawableImplementation.cxx
+++ b/lib/src/Base/Graph/DrawableImplementation.cxx
@@ -1324,6 +1324,23 @@ void DrawableImplementation::setPalette(const Description & palette)
   throw NotDefinedException(HERE) << "Error: no palette in " << getClassName();
 }
 
+Sample DrawableImplementation::getPaletteAsNormalizedRGBA() const
+{
+  const Description palette(getPalette());
+  const UnsignedInteger size = palette.getSize();
+  Sample normalizedRGBA(size, 4);
+  Indices rgba(4);
+  for (UnsignedInteger i = 0; i < size; ++i)
+    {
+      (void) ScanColorCode(palette[i], rgba);
+      normalizedRGBA(i, 0) = rgba[0] / 255.0;
+      normalizedRGBA(i, 1) = rgba[1] / 255.0;
+      normalizedRGBA(i, 2) = rgba[2] / 255.0;
+      normalizedRGBA(i, 3) = rgba[3] / 255.0;
+    }
+  return normalizedRGBA;
+}
+
 /* Accessor for origin */
 Scalar DrawableImplementation::getOrigin() const
 {
--- a/lib/src/Base/Graph/DrawableImplementation.cxx
+++ b/lib/src/Base/Graph/DrawableImplementation.cxx
@@ -816,37 +816,20 @@ String DrawableImplementation::ConvertFromName(const String & name)
 /* Convert an hexadecimal code into an RGB triplet */
 Indices DrawableImplementation::ConvertToRGB(const String & key)
 {
-  Indices rgb(3, 0);
-  UnsignedInteger code = 0;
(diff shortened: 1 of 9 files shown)


python/src/viewer.py
```diff
@@ -524,6 +527,8 @@ def convRGBA(c):
             legend_kwargs.setdefault('prop', {'size': 10})
 
             self._ax[0].legend(**legend_kwargs)
+        if square_axes:
+            self._ax[0].axis('square')
```
jschueller: Seems this is not supported by older matplotlib (see the autobuilder log), maybe use something like this:
```
try:
    self._ax[0].axis('square')
except ValueError:
    plt.gca().set_aspect('equal', adjustable='box')
```

lib/src/Base/Graph/DrawableImplementation.cxx
```diff
@@ -816,37 +816,20 @@ String DrawableImplementation::ConvertFromName(const String & name)
...
-    code = code >> 8;
-    rgba[1] = code % 256;
-    code = code >> 8;
-    rgba[0] = code;
-  }
-  else LOGWARN(OSS() << "Code " << key << " is an invalid color code. Default to black.");
+  Indices rgba;
+  if (!ScanColorCode(key, rgba))
+    LOGWARN(OSS() << "Code " << key << " is an invalid color code. Default to black.");
   return rgba;
```
jschueller: Maybe this could be reused to factorize some code with the PolygonArray section where color conversion seems partially rewritten:
```
elif drawableKind == 'PolygonArray':

                polygonsNumber = drawable.getPalette().getSize()
                verticesNumber = drawable.getData().getSize() // polygonsNumber
                colors = drawable.getPalette()
                def convRGB(c):
                    return (int(c[1:3], 16) / 255.0, int(c[3:5], 16) / 255.0, int(c[5:7], 16) / 255.0, 1.0)
                def convRGBA(c):
                    return (int(c[1:3], 16) / 255.0, int(c[3:5], 16) / 255.0, int(c[5:7], 16) / 255.0, int(c[7:9], 16) / 255.0)
                if colors[0][0] == '#' and len(colors[0]) == 7:
                    colorsRGBA = [convRGB(color) for color in colors]
                elif colors[0][0] == '#' and len(colors[0]) == 9:
                    colorsRGBA = [convRGBA(color) for color in colors]
                else:
                    colorsRGBA = []
                    for i in range(polygonsNumber):
                        hex_code = ot.Drawable.ConvertFromName(colors[i])
                        rgba = ot.Drawable.ConvertToRGBA(hex_code)
                        colorsRGBA.append(
                            (rgba[0] / 255.0, rgba[1] / 255.0, rgba[2] / 255.0, rgba[3] / 255.0))
                if 'facecolors' not in polygoncollection_kwargs_default:
                    polygoncollection_kwargs['facecolors'] = colorsRGBA

                if 'edgecolors' not in polygoncollection_kwargs_default:
                    polygoncollection_kwargs['edgecolors'] = colorsRGBA
                self._ax[0].add_collection(
                    matplotlib.collections.PolyCollection(np.array(data).reshape((polygonsNumber, verticesNumber, 2)), **polygoncollection_kwargs))
```

dbarbier: The problem is that no C++ method is provided to convert the whole palette with a single call.

regislebrun: I added Drawable::getPaletteAsNormalizedRGBA() to solve this problem. Any other issue here?

jschueller:
A rebase is needed.

regislebrun:
done.

jschueller:
Could you rebase one last time?

regislebrun:
done.

PR #691, opened 2018-03-01
Optimized checkPointInSimplex()

regislebrun:
The new implementation is more than 3x faster than the generic one on the following benchmark (I kept the naive algorithm to check the result):
```
import openturns as ot
from time import time

N = 50
size = 1000000

mesh = ot.IntervalMesher([N]*2).build(ot.Interval([0.0]*2, [10.0]*2))
f = ot.SymbolicFunction(["x", "y"], ["x + 0.5*sin(y)", "y-0.1*x*sin(x)"])
mesh.setVertices(f(mesh.getVertices()))
sample = ot.ComposedDistribution([ot.Uniform(-2.0, 12.0)]*2).getSample(size)

vertices = mesh.getVertices()
simplices = mesh.getSimplices()
print(mesh.getSimplicesNumber(), 'simplices')

print("NaiveEnclosingSimplex")
t0 = time()
naive = ot.NaiveEnclosingSimplex(mesh.getVertices(), mesh.getSimplices())
print("build time=", time() - t0, "s")

t0 = time()
result1 = naive.query(sample)
print("speed=", size / (time() - t0), "queries/s")

print("BoundingVolumeHierarchy")
t0 = time()
bvh = ot.BoundingVolumeHierarchy(vertices, simplices)
print("build time=", time() - t0, "s")

t0 = time()
result2 = bvh.query(sample)
print("speed=", size / (time() - t0), "queries/s")
print(result1 == result2)
```
The 3x factor is a lower bound on my laptop when I vary N from 20 to 200 and size from 1000 to 1000000.

I also added the 3D case, and implemented the different improvements suggested by Denis.

Concerning the mesh orientation, now the mesh generated by OpenTURNS are positively oriented (IntervalMesher, LevelSetMesher, ImportFromGMSHFile()) and a new Mesh::fixOrientation() method is provided. Nevertheless, the positivity of the meshes is not enforced in the Mesh constructors, so I kept the point location algorithms as they are instead of simplifying them assuming a positive orientation.

I also fixed a bug in IntervalMeshser: the sample of vertices was pre-allocated with too much vertices when diamond==true, resulting in a concentration of nodes at the origin in the mesh.


regislebrun:
@dbarbier I inlined the function and added references for a full 2.5% performance boost ;-)! Ok, not worth the obfuscation...

dbarbier:
@regislebrun 2.5% is not noticeable, I doubt that your timings are reproducible.

BTW your code only works if triangles are well oriented, is this intentional?  Otherwise you have to compute the sign of (v1,v2,v3) and write 2 branches whether this sign is positive or negative.

Non-optimized code is also wrong, we should check
```
    if ((coordinates[i] < - SpecFunc::ScalarEpsilon) || (coordinates[i] > 1.0))
```

dbarbier:
This code should be fast and handle all corner cases (but mostly untested):
```
namespace {
inline Bool EnclosingSimplexAlgorithm_checkPointInSimplex2D(const Scalar * const v1, const Scalar * const v2, const Scalar * const v3, const Scalar * const pt, const UnsignedInteger index)
{
  const Scalar x = *pt;
  const Scalar y = *(pt+1);
  const Scalar v1x = *v1;
  const Scalar v1y = *(v1+1);
  const Scalar v2x = *v2;
  const Scalar v2y = *(v2+1);
  const Scalar v3x = *v3;
  const Scalar v3y = *(v3+1);
  const Scalar totalVolume = (v3x - v2x) * (v1y - v2y) - (v1x - v2x) * (v3y - v2y);
  const Scalar volume1 = (x - v2x) * (v1y - v2y) - (v1x - v2x) * (y - v2y);
  const Scalar volume2 = (x - v3x) * (v2y - v3y) - (v2x - v3x) * (y - v3y);
  const Scalar volume3 = (x - v1x) * (v3y - v1y) - (v3x - v1x) * (y - v1y);
  if (totalVolume > SpecFunc::ScalarEpsilon)
  {
    return volume1 > - SpecFunc::ScalarEpsilon && volume2 > - SpecFunc::ScalarEpsilon && volume3 > - SpecFunc::ScalarEpsilon;
  }
  else if (totalVolume < - SpecFunc::ScalarEpsilon)
  {
    return volume1 < SpecFunc::ScalarEpsilon && volume2 < SpecFunc::ScalarEpsilon && volume3 < SpecFunc::ScalarEpsilon;
  }
  else
  {
    // Degenerate case
    if (volume1 > SpecFunc::ScalarEpsilon || volume1 < - SpecFunc::ScalarEpsilon ||
        volume2 > SpecFunc::ScalarEpsilon || volume2 < - SpecFunc::ScalarEpsilon ||
        volume3 > SpecFunc::ScalarEpsilon || volume3 < - SpecFunc::ScalarEpsilon)
      return false;
    // Point is on the same line
    return x <= std::max(std::max(v1x, v2x), v3x) && x >= std::min(std::min(v1x, v2x), v3x) &&
           y <= std::max(std::max(v1y, v2y), v3y) && y >= std::min(std::min(v1y, v2y), v3y);
  }
}
}

/* Check if the given point is in the given simplex */
Bool EnclosingSimplexAlgorithmImplementation::checkPointInSimplex(const Point & point, const UnsignedInteger index, SquareMatrix & simplexMatrix) const
{
  // Special case for dimension==2
  // It is more efficient to skip the tests against both the global bounding box
  // and the triangle bounding box
  const UnsignedInteger dimension = vertices_.getDimension();
  if (dimension == 2)
  {
    IndicesCollection::const_iterator cit2d = simplices_.cbegin_at(index);
    return EnclosingSimplexAlgorithm_checkPointInSimplex2D(&vertices_(*cit2d, 0), &vertices_(*(cit2d+1), 0), &vertices_(*(cit2d+2), 0), &point[0], index);
  }
...
```

regislebrun:
@dbarbier The 2.5% are a joke ;-)! Concerning the orientation, you suggested some times ago that we allowed only simplices with positive orientation. It was a good idea and I already changed IntervalMesher in order to produce only positively oriented simplices. I suspect that LevelSetMesher has also to be checked as some vertices are moved during the generation of the mesh.
I could add a Mesh::fixOrientation() method and call it from the constructor, but it would slow the construction a lot. Do you have a better idea? A flag in the constructor to tell if the orientation has to be checked?

dbarbier:
Ok, I did not remember the outcome of this discussion, and your return condition was rather confusing; if simplices are oriented, it should be written as `return b1 && b2 && b3;`, and in this case it is surely better to check b1 just after it is computed, to return early if it is false.  Likewise for b2.

I indeed prefer if simplices have a consistent orientation, it simplifies most algorithms.  To keep things simple, we can add a fixOrientation() method, but there is no reason to add an argument in constructors, users will have to call this method explicitly if dealing with wrongly oriented meshes.

regislebrun:
I added the 3D special case to checkPointInSimplex, but some corner cases are not tested (flat tetrahedron). The speedup is significant (4x). I also added a function to fix the orientation of a given simplex or of the whole mesh and used it into LevelSetMesher. The documentation is still missing. Comments are welcomed!

regislebrun:
@dbarbier  Is there any reason why I should not parallelize EnclosingSimplexAlgorithmImplementation::query(Sample)? What is your feeling about the use of TBB in general?

lib/src/Base/Geom/Mesh.cxx
```diff
@@ -555,6 +555,53 @@ Point Mesh::getUpperBound() const
...
+      const Scalar v3x = vertices_(*cit + 2, 0);
+      const Scalar v3y = vertices_(*cit + 2, 1);
+      const Scalar v3z = vertices_(*cit + 2, 2);
+      const Scalar v4x = vertices_(*cit + 3, 0);
+      const Scalar v4y = vertices_(*cit + 3, 1);
+      const Scalar v4z = vertices_(*cit + 3, 2);
+      if ((v1x - v4x) * ((v2y - v4y) * (v3z - v4z) - (v3y - v4y) * (v2z - v4z)) + (v3x - v4x) * ((v1y - v4y) * (v2z - v4z) - (v2y - v4y) * (v1z - v4z)) < (v2x - v4x) * ((v1y - v4y) * (v3z - v4z) - (v3y - v4y) * (v1z - v4z)) ) std::swap(*cit, *(cit + 1));
+      return;
+    }
+  SquareMatrix matrix(buildSimplexMatrix(index));
```
dbarbier: In checkPointInSimplex, I inlined call to buildSimplexMatrix to avoid SquareMatrix allocation.  The same idea could be used here, but it is surely better to add SquareMatrix& argument to buildSimplexMatrix (and to fixOrientation of course).  This is uncommon, but these methods are not meant for users, and almost always called when looping over simplices.

dbarbier:
@regislebrun I prefer to first benchmark and see which portion of code takes time.

lib/src/Base/Geom/Mesh.cxx
regislebrun: @dbarbier If I understand correctly your proposal, the signature of Mesh::buildSimplexMatrix() should be `void Mesh::buildSimplexMatrix(const UnsignedInteger index, SquareMatrix & matrix)`, with the caller in charge of the allocation? Maybe with a test to see if the given matrix has the correct dimension?

dbarbier: Yes and yes, and this method should then not be wrapped by SWIG, and not documented.

regislebrun:
@dbarbier I parallelized the query() method over a sample. It is in a separate commit to be easily removed if you think it is not useful. On the following benchmark:
```
import openturns as ot
from time import time

N = 100
size = 100000

mesh = ot.IntervalMesher([N]*2).build(ot.Interval([0.0]*2, [10.0]*2))
f = ot.SymbolicFunction(["x", "y"], ["x + 0.5*sin(y)", "y-0.1*x*sin(x)"])
mesh.setVertices(f(mesh.getVertices()))
mesh.exportToVTKFile("mesh2d.vtk")
sample = ot.ComposedDistribution([ot.Uniform(-2.0, 12.0)]*2).getSample(size)

vertices = mesh.getVertices()
simplices = mesh.getSimplices()
print(mesh.getSimplicesNumber(), 'simplices')

print("NaiveEnclosingSimplex")
t0 = time()
naive = ot.NaiveEnclosingSimplex(mesh.getVertices(), mesh.getSimplices())
print("build time=", time() - t0, "s")

t0 = time()
result1 = naive.query(sample)
print("speed=", size / (time() - t0), "queries/s")

print("BoundingVolumeHierarchy")
t0 = time()
bvh = ot.BoundingVolumeHierarchy(vertices, simplices)
print("build time=", time() - t0, "s")

M = 20
t0 = time()
for rep in range(M):
    result2 = bvh.query(sample)
t1 = time()
print(result1 == result2)
print("mean speed=", M * size / (t1 - t0), "queries/s")

# 3D

N = 35
size = 10000

mesh = ot.IntervalMesher([N]*3).build(ot.Interval([0.0]*3, [10.0]*3))
f = ot.SymbolicFunction(["x", "y", "z"], ["x + 0.5*sin(y) - 0.1 * z", "y-0.1*x*sin(x) + 0.1 * z", "z - 0.05 * y * sin(x)"])
mesh.setVertices(f(mesh.getVertices()))
mesh.exportToVTKFile("mesh3d.vtk")
sample = ot.ComposedDistribution([ot.Uniform(-2.0, 12.0)]*3).getSample(size)

vertices = mesh.getVertices()
simplices = mesh.getSimplices()
print(mesh.getSimplicesNumber(), 'simplices')

print("NaiveEnclosingSimplex")
t0 = time()
naive = ot.NaiveEnclosingSimplex(mesh.getVertices(), mesh.getSimplices())
print("build time=", time() - t0, "s")

t0 = time()
result1 = naive.query(sample)
print("speed=", size / (time() - t0), "queries/s")

print("BoundingVolumeHierarchy")
t0 = time()
bvh = ot.BoundingVolumeHierarchy(vertices, simplices)
print("build time=", time() - t0, "s")

M = 20
M = 20
t0 = time()
for rep in range(M):
    result2 = bvh.query(sample)
t1 = time()
print(result1 == result2)
print("mean speed=", M * size / (t1 - t0), "queries/s")
```
With the sequential implementation I get:
```
20000 simplices
NaiveEnclosingSimplex
build time= 0.010172605514526367 s
speed= 26099.344303058006 queries/s
BoundingVolumeHierarchy
build time= 0.004305839538574219 s
True
mean speed= 548634.4420269566 queries/s
257250 simplices
NaiveEnclosingSimplex
build time= 0.08294558525085449 s
speed= 743.0214632607123 queries/s
BoundingVolumeHierarchy
build time= 0.08256340026855469 s
True
mean speed= 333892.2212890617 queries/s
```
and with the parallel implementation:
```
20000 simplices
NaiveEnclosingSimplex
build time= 0.010472774505615234 s
speed= 114023.33747727651 queries/s
BoundingVolumeHierarchy
build time= 0.004132986068725586 s
True
mean speed= 2457577.2002115217 queries/s
257250 simplices
NaiveEnclosingSimplex
build time= 0.08402204513549805 s
speed= 3108.9697951453036 queries/s
BoundingVolumeHierarchy
build time= 0.08608770370483398 s
True
mean speed= 1679226.1866632502 queries/s
```
so for NaiveEnclosingSimplex it is a 4.3x speedup in 2D, 4.2x in 3D, and for BoundingVolumeHierarchy it is a 4.5x speedip in 2D, 5x in 3D.

PR #692, opened 2018-03-02
Asymptotic Sobol' variance estimators

jschueller:
This is a preview of our work on Sobol' estimators using the Janon2014 idea (https://arxiv.org/pdf/1303.6451.pdf)
I'll elaborate more on this on our 14th march meeting, but as some code works why not share it now.

The idea is to use the central-limit theorem on a function of the output of the A, B and E output samples, more details in:
[RT-PMFRE-01116-001C_-_Rapport_loi_estimateur_sobol.pdf](https://github.com/openturns/openturns/files/1775335/RT-PMFRE-01116-001C_-_Rapport_loi_estimateur_sobol.pdf)
The estimators can handle aggregated indices (when the output dimension is >1).

The API basically remains the same while the variance estimator is implemented in all Saltelli, Jansen, MauntzKucherenko estimator classes in addition to the existing Martinez estimator.
I decided to replace the previous Martinez asymptotic Fisher-based estimator as it was said to be valid only for gaussian output.
Also a setUseAsymptotic accessor is added to SobolIndicesAlgorithm instead of the Martinez RM key as now the methods are available for all four classes. This simplifies the mecanism to discard invalid results.
[gs]etBootstrapConfidenceLevel is deprecated in favor of [gs]etConfidenceLevel as that same variable  is used for both bootstrap and asymptotic methods.


lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
```diff
@@ -321,11 +326,51 @@ void SobolIndicesAlgorithmImplementation::computeIndicesInterval() const
...
+
+void SobolIndicesAlgorithmImplementation::computeAsymptoticInterval() const
+{
+  throw NotYetImplementedException(HERE) << "SobolIndicesAlgorithmImplementation::computeAsymptoticInterval";
+}
+
+
+Scalar SobolIndicesAlgorithmImplementation::computeVariance(const Sample & u, const Function & psi) const
+{
+  Point gradient(psi.gradient(u.computeMean()) * Point(1, 1.0));
```
regislebrun: or
```
Point gradient(*psi.gradient(u.computeMean()).getImplementation());
```

lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
```diff
@@ -517,6 +583,19 @@ Point SobolIndicesAlgorithmImplementation::computeSumDotSamples(const Sample & s
   return value;
 }
 
+/* Multiplication of two sub-samples */
+Sample SobolIndicesAlgorithmImplementation::ComputeProdSample(const Sample & sample,
+                                                              UnsignedInteger marginalIndex,
+                                                              UnsignedInteger size,
+                                                              UnsignedInteger indexX,
+                                                              UnsignedInteger indexY)
+{
```
regislebrun: Missing const for UnsignedInteger arguments

jschueller:
Thank you for the review.
- Yes, the distribution is even explicitely built as I had this change in mind, I did not want to change the API too much at first.
- The next step is to provide simulation algorithms based on these estimators, and we also want to add moments simulation algorithm. For now it's not very clear to me if stuff can be mutualized or not. I also have a very naive moment simulation algorithm of a function by Monte Carlo. We also talked about quantiles simulations this summer, maybe it's a larger scope, that's why we wanted to have a specific meeting.
- The use of symbolic function is mostly for code readability. Its expression is always very simple so I doubt it has a numerical impact. I first implemented the estimators without it and it passed the same tests.

jschueller:
I added distribution accessors. Now we dont store the intervals but only the distribution, and the interval accessors call computeMinimumVolumeInterval. But in the case of boostrap, the interval computed on the kernel smoothing distribution differs now from the one computed with the formula ```[2*aggregatedIndice-qupper; 2*aggregatedIndice-qlower] ```, what do you think about that ?

regislebrun:
I don't know from where these formulas are coming. Any reference?

jschueller:
I could find the formula he used in wikipedia:
https://en.wikipedia.org/wiki/Bootstrapping_(statistics) (Methods for bootstrap confidence intervals)

sofianehaddad:
@jschueller @regislebrun when I made the methods, I wanted to map R/Sensitivity within the "Basic Bootstrap"
We could however investigate other methods. See http://www.andreasaltelli.eu/file/repository/Archer.pdf for more details

jschueller:
If we want to keep that result, the interval could be saved, but it'd be weird to get an interval that does not match the distribution's interval withe the same confidence level. Or else we could still store the intervals and provide the distribution only for asymptotic estimators.

lib/src/Uncertainty/Algorithm/Sensitivity/JansenSensitivityAlgorithm.cxx
```diff
@@ -115,6 +115,73 @@ Sample JansenSensitivityAlgorithm::computeIndices(const Sample & sample,
...
+  Function psiFO = SymbolicFunction(XY, Description(1, "1 - 0.5*" + sumX + "/" + sumY));
+  Function psiTO = SymbolicFunction(XY, Description(1, "0.5*" + sumX + "/" + sumY));
+
+  Point varianceFO(inputDimension);
+  Point varianceTO(inputDimension);
+
+  for (UnsignedInteger p = 0; p < inputDimension; ++ p)
+  {
+    Sample uFO(size_, 0);
+    Sample uTO(size_, 0);
```
dbarbier: For performance reasons, you should avoid Sample::stack.  Allocate uFO and uTO only once outside p-loop, and fill these Sample in these loops; yemB and yAmB are then no more needed.

dbarbier: If this makes code much less readable, an alternative is to work on transposed Sample, and replace  Sample::stack by Sample::add.

jschueller: I prefer the code to be maintainable and readable here. Not sure if option 2 would be very readable too, besides one still have to transpose to compute the covariance.

jschueller: I started to implement your solution, and it looks good. I can even swap the i and q loops to follow the sample memory layout.

jschueller: Thanks

dbarbier:
@jschueller Looks much better, thanks.  If you can, benchmark it on large samples, you should see a huge speed-up.

jschueller:
Yes, not so bad, I got 1m38 instead of 2m28 with N=1e6, 10 outputs.

dbarbier:
I was hoping for a much better speed-up :-(

jschueller:
I did not separate the computation of indices themselves, maybe they take longer than the variance.

regislebrun:
+1

ArashPartow:
@regislebrun 

>  I am not 100% sure that the optimization pass made by exprtk will keep intact the numerical stability of the gradient formulas...

One can instantiate a version of the parser that has the various strength reduction optimisations turned-off. 

For more information have a read of **Section 11 - Compilation Options** of the readme.txt sub-section 7:  https://github.com/ArashPartow/exprtk/blob/master/readme.txt#L1134

lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
```diff
@@ -134,6 +134,15 @@ SobolIndicesAlgorithmImplementation::SobolIndicesAlgorithmImplementation(const D
...
 
+  if (computeSecondOrder && (inputDimension == 2))
+  {
+    // Special case: the experiment does not contain the C=[E_2, E_1] sample
+    Sample E1(outputDesign_, size * 2, size * 3);
+    Sample E2(outputDesign_, size * 3, size * 4);
+    outputDesign_.add(E2);
+    outputDesign_.add(E1);
+  }
+
```
jschueller: @sofianehaddad, ok fixed

PR #693, opened 2018-03-06
Check NLopt starting point is inside bounds

jschueller:
Seems NLopt checks it and throws as none of the algorithms succeed with a point outside bounds.
Cobyla, TNC allow starting points outside bounds.


jschueller:
cc @aurelieladier

regislebrun:
+1

jschueller:
@regislebrun @dbarbier I propose to check TNC & Cobyla too for consistency.

regislebrun:
@jschueller Not sure it's a good idea. Sometimes it is hard to find a feasible point, so it is a good thing to have an algorithm doing this hard job for us. At most a warning IMO.

jschueller:
I dont understand your argument, as both bounds and starting points are parameters of the algorithm you can make sure the point is inside, and if you know the problem will be difficult you can specify larger bounds.

regislebrun:
@jschueller Sorry, I was in the fog! I was thinking about a starting point not satisfying the inequality constraints. Of course, in the case of bound constraints it is trivial to check if the given starting point is admissible or not. So +1

dbarbier:
I agree with @regislebrun, a warning is better if algorithms are able to find a solution.

PR #695, opened 2018-03-06
Speed-up evaluation of a PythonFunction on a Sample

dbarbier:
Here is what happens when a PythonFunction is evaluated
on a Sample:
 1. If argument X is a 2d Python sequence, it is converted by SWIG
    into a Python Sample X1.  Otherwise X is a Python Sample
 2. `PythonEvaluation::operator()(const Sample &inS)` is called on the
    Sample wrapped by X (or X1).
 3. This method converts inS into a Python sequence P1.
 4. `OpenTURNSPythonFunction.__call__(P1)` is called
 5. It tries to convert P1 into a Point (with a copy), this
    fails and it converts P1 into a Python Sample S1 (with another
    copy).
 6. `OpenTURNSPythonFunction._exec_sample(S1)` is called and evaluated.

Dispatch at step 5 is already performed by `PythonEvaluation::operator()`
overloads, this part of this commit can be applied independantly.

With this commit, a reference is added at step 1 into the Sample object
to point to the underlying Python object.  Step 3 can then be avoided,
we call _exec_sample directly on the Python object.

This works only when called directly from Python.  If this PythonFunction
is called from C++ classes, there is no Python object associated to argument,
and step 3 is called as before.

An optional 'copy' argument is added to PythonFunction, to revert to old
code if something goes wrong.  Normally this should not happen, but who
knows?  It also allows to compute the cost of step 3.

Cache is removed, it prevents this optimization.

### DIFF
--- a/lib/src/Base/Stat/Sample.cxx
+++ b/lib/src/Base/Stat/Sample.cxx
@@ -67,6 +67,7 @@ String Sample::streamToRFormat() const
 /* Default constructor */
 Sample::Sample()
   : TypedInterfaceObject<SampleImplementation>(new SampleImplementation(0, 1))
+  , pyObjSample_(0)
 {
   // Nothing to do
 }
@@ -75,6 +76,7 @@ Sample::Sample()
 Sample::Sample(const UnsignedInteger size,
                const UnsignedInteger dim)
   : TypedInterfaceObject<SampleImplementation>(new SampleImplementation(size, dim))
+  , pyObjSample_(0)
 {
   // Nothing to do
 }
@@ -82,6 +84,7 @@ Sample::Sample(const UnsignedInteger size,
 /* Constructor from implementation */
 Sample::Sample(const SampleImplementation & implementation)
   : TypedInterfaceObject<SampleImplementation>(implementation.clone())
+  , pyObjSample_(0)
 {
   // Nothing to do
 }
@@ -89,6 +92,7 @@ Sample::Sample(const SampleImplementation & implementation)
 /* Constructor from implementation */
 Sample::Sample(const Implementation & implementation)
   : TypedInterfaceObject<SampleImplementation>(implementation)
+  , pyObjSample_(0)
 {
   // Nothing to do
 }
@@ -97,6 +101,7 @@ Sample::Sample(const Implementation & implementation)
 Sample::Sample(const UnsignedInteger size,
                const Point & point)
   : TypedInterfaceObject<SampleImplementation>(new SampleImplementation(size, point))
+  , pyObjSample_(0)
 {
   // Nothing to do
 }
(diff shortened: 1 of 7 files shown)


dbarbier:
Here is a benchmark:
```
import openturns as ot
import numpy as np
from math import cos
from time import time

sampleSize=10000000

inputSample = ot.ComposedDistribution([ot.Uniform(-10.0, 10.0)]*2).getSample(sampleSize)

mySymbolicFunction = ot.SymbolicFunction(['x1', 'x2'],
          ['2 * x1^2 + x1 + 8 * x2 + 4 * cos(x1) * x2 + 6'])

def function(X) :
    x1, x2 = X
    return [2 * x1**2 + x1 + 8 * x2 + 4 * cos(x1) * x2 + 6]

def functionSample(X) :
    Xarray = np.array(X, copy=False)
    x1, x2 = Xarray.T
    y = 2 * np.square(x1) + x1 + 8 * x2 + 4 * np.cos(x1) * x2 + 6
    return y[:,np.newaxis]

t0 = time()
outputSampleSymbolicFunction = mySymbolicFunction(inputSample)
print("SymbolicFunction : ", (time() - t0))

myFunctionPythonExecSample = ot.PythonFunction(2, 1, function, functionSample)

t0 = time()
outputSampleExecSample = myFunctionPythonExecSample(inputSample)
print("Python exec_sample: ", (time() - t0))
print(outputSampleExecSample == outputSampleSymbolicFunction)

myFunctionPythonExecSample2 = ot.PythonFunction(2, 1, function, functionSample, copy=True)

t0 = time()
outputSampleExecSample2 = myFunctionPythonExecSample2(inputSample)
print("Python exec_sample with copy=True: ", (time() - t0))
print(outputSampleExecSample2 == outputSampleSymbolicFunction)

myFunctionPythonExec = ot.PythonFunction(2, 1, function)

t0 = time()
outputSampleExec = myFunctionPythonExec(inputSample)
print("Python exec: ", (time() - t0))
print(outputSampleExec == outputSampleSymbolicFunction)
```

Output:
```
SymbolicFunction :  0.748180627822876
Python exec_sample:  0.3889744281768799
True
Python exec_sample with copy=True:  4.442799091339111
True
Python exec:  118.09509897232056
True
```

regislebrun:
@dbarbier The result is impressive as usual. I am not an expert user of numpy. Is it the standard way of dealing with numpy objects? It remembers me the old time when I had to figure out strange constructions in matlab in order to recover the performance I would have had in a straightforward way in C.
I hope that all our users will be aware of these constructions, otherwise they will stay on the Python exec usage, a full 150x slower solution than the symbolic function...
Anyway, +1

dbarbier:
I did not benchmark current master, I expect it to run in ~5s for SymbolicFunction and exec_sample.

Most of the time is spent in creating a 2d Python object.  Unfortunately I do not see how to improve that.  We could use numpy API to create arrays, but Python OT module would then depend on a specific numpy version, this is problematic.

This PR works around this issue when PythonFunction is evaluated from Python, but this is not the normal use case.  So contrary to what I say previously, SymbolicFunction with ExprTK seemms to be the only viable solution to provide full performance.

dbarbier:
I found a better solution, will provide another PR later.

PR #696, opened 2018-03-08
Improve kl algorithms

regislebrun:
The selection of the cut-off eigenvalue is now based on the full sum of eigenvalues, as they are already computed before the selection.

Several improvements have been made in the various Karhunen-Loeve algorithms:
  * now the sorting is done without moving the eigenvectors
  * the reuse of computations has been improved in the quadrature method
  * the log messages are more consistent from one algorithm to the other

I also propose to switch to a more conventional notation for vectors and matrices, using boldface instead of underline. This commit can be dropped if it causes more trouble than others.


dbarbier:
Why does `lib/test/t_PointToPointConnection_std.expout` change?  By reading commit messages, one would expect that the only change is in the first eigenvalue being counted twice during cut-off.

regislebrun:
On the other hand, now the current eigenvalue is compared to the sum of all the eigenvalues and not only the sum of the preceding ones...

PR #697, opened 2018-03-08
Improved KarhunenLoeveResult

regislebrun:
Now the projection of a field defined over the same mesh as the result is optimized.


dbarbier:
Tests do not pass.  BTW do you use a use case for benchmarking?

regislebrun:
I remove this PR as it corresponds to an old branch no more interesting

PR #698, opened 2018-03-08
Optim algos max number of evaluation stop criteria

jschueller:
This tries to address the inconsistencies mentioned in http://trac.openturns.org/ticket/932.

Here I chose to add support for the maximum of function calls in all optim algos.
As a consequence TNC, Cobyla, EGO are parametrized by maximumEvaluationNumber instead of maximumIterationNumber, this criteria is added to SQP & AbdoR. NLopt stays unchanged.
I also added the maximum number of evaluations in the result, previously the number of iterations was used for the number of evaluations even if the algo doesnt have the concept of iteration (only SQP & AdboR have).

BUT, one also can decide to parametrize all optimization algorithms only by the maximum number of evaluations (and only report that in the result).


regislebrun:
+1 Not sure that we will are able to stop exactly when the maximum number of evaluation has been reached, eg in Cobyla, but it can be a best effort stopping criterion.

jschueller:
@regislebrun I just checked, Cobyla does stop exactly at the right number of evaluations. For SQP & AbdoRackwitz I figured it'd be reasonable to check between iterations, ie it can make several extra line search or gradient evaluation above the limit to complete the current iteration.

jschueller:
I had to set DefaultMaximumEvaluationNumber=100 instead of 1e5 to get the previous behavior on all tests. SQP and AbdorAckwitz override this default. I think its best to have a lower value anyways.

dbarbier:
IMO this issue deserves some online discussion to make sure that we all agree on the solution to implement, can we discuss it next week?

jschueller:
Totally.

PR #699, opened 2018-03-08
New openturns.MemoryView extension module

dbarbier:
With this commit, a new openturns.MemoryView extension module is added.
It provides a Buffer class which supports Python protocol buffer
(aka PEP 3118); with this class, we can wrap OT data containers into
Python objects without copy, as is done for instance in numpy.

Commit 65cc96a allowed to efficiently copy Python buffers into OT
Sample, so conversion between OT and Python should now be quite fast.

The Buffer class supports indexing (but not slicing for the moment) and
iterators, but nothing more.  To use this class for numerical computations,
it must be passed to a Point or Sample (which makes a copy), or numpy
(with or without copy).

This example illustrates these capabilities; there is no reason for
users or developers to use this Buffer class as in this example, it
should only be used by SWIG to wrap OT containers into Python objects.

```
  >>> import openturns as ot
  >>> import openturns.MemoryView
  >>> import numpy as np
  >>> from __future__ import print_function
  >>> ot.RandomGenerator.SetSeed(0)
  >>> s = ot.Normal(4).getSample(2)
  >>> print(s)
      [ X0        X1        X2        X3        ]
  0 : [  0.608202 -1.26617  -0.438266  1.20548  ]
  1 : [ -2.18139   0.350042 -0.355007  1.43725  ]
  >>> x = openturns.MemoryView.Buffer(int(s.__baseaddress__()), (s.getSize(), s.getDimension()))
  >>> print(x)
  <read-only buffer at 0x1594440 shape=(2,4)>
  # Indexing returns a new Buffer, without copy
  >>> print(x[0])
  <read-only buffer at 0x1594440 shape=(4)>
  >>> print(x[1])
  <read-only buffer at 0x1594460 shape=(4)>
  # A 0-d Buffer is a scalar
  >>> print(x[1][0])
  -2.1813852346165143
  #  Buffer is read-only
  >>> x[1][0] = 3
  TypeError: 'openturns.MemoryView.Buffer' object does not support item assignment
  >>> for p in x:
  ...     print([q for q in p])
  [0.6082016512187646, -1.2661731022166567, -0.43826561996041397, 1.2054782008285756]
  [-2.1813852346165143, 0.3500420865302907, -0.3550070491856397, 1.437249310140903]
  # Copy into a Sample; our SWIG bindings have been taught recently to use Python
  # buffer protocol, so this is fast.
  >>> y = ot.Sample(x)
  >>> print(y)
  0 : [  0.608202 -1.26617  -0.438266  1.20548  ]
[...]

PR #700, opened 2018-03-08
[RFC] Speed-up evaluation of a PythonFunction on a Sample

dbarbier:
This is a revised version of #695, and is a work in progress.  This PR is there to gather comments.

Here is what happens when a PythonFunction is evaluated
on a Sample:

 1. If argument X is a 2d Python sequence, it is converted by SWIG
    into a Python Sample X1.  Otherwise X is a Python Sample
 2. `PythonEvaluation::operator()(const Sample &inS)` is called on the 
    Sample wrapped by X (or X1).
 3. This method converts inS into a Python sequence P1. 
 4. `OpenTURNSPythonFunction.__call__(P1)` is called
 5. It tries to convert P1 into a Point (with a copy), this
    fails and it converts P1 into a Python Sample S1 (with another
    copy).
 6. `OpenTURNSPythonFunction._exec_sample(S1)` is called and evaluated.

This commit uses the new openturns.MemoryView.Buffer class to wrap OT objects
into Python more efficiently.  In step 2, a Buffer is created, and passed to _exec_sample
method directly.

The downside is that _exec_sample Python function now receives a Buffer object,
and no more a Sample.  For efficiency reasons, I propose to also remove
_exec_point_on_exec_sample and implement it in C via this Buffer.

An optional 'copy' argument is added to PythonFunction, to revert to old 
code if something goes wrong.  Normally this should not happen, but who 
knows?  It also allows to compute the cost of step 3.

Cache is removed, it prevents this optimization.

### DIFF
--- a/python/src/CMakeLists.txt
+++ b/python/src/CMakeLists.txt
@@ -885,6 +885,19 @@ foreach (module ${OPENTURNS_PYTHON_MODULES})
          )
 endforeach ()
 
+
+# Build and install openturns.memoryview
+add_library( memoryview MODULE PythonReadOnlyBuffer.c )
+if (LINK_PYTHON_LIBRARY)
+  target_link_libraries (memoryview ${PYTHON_LIBRARIES})
+endif ()
+set_target_properties(memoryview PROPERTIES NO_SONAME ON PREFIX "")
+if(WIN32 AND NOT CYGWIN)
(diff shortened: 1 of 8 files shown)


dbarbier:
I am closing this PR and will move this commit into #699.

PR #701, opened 2018-03-09
Use library location to find config file

jschueller:
Previously, on windows only, for relocatable layouts, the config file was found relatively to the python executable.
This does not apply if the library is used from another form (Linux wheels, C++ only, ...).
GetConfigDirectoryList now falls back to the location of the shared library libOT using dladdr (Unix) or VirtualQuery (Win32).


lib/src/Base/Common/Path.cxx
```diff
@@ -136,28 +140,55 @@ FileName Path::GetExecutableDirectory()
...
 
-  // get parent dir
+
+FileName Path::GetParentDirectory(const FileName & file)
+{
+  FileName parent;
 #ifdef OPENTURNS_HAVE_LIBGEN_H
-  String pythonDir(dirname(szPath));
+  char * szPath = strdup(file.c_str());
+  parent = String(dirname(szPath));
```
dbarbier: parent is a FileName

dbarbier:
I do not understand why this is needed, IMO it is not OT job to work around installation trouble. We should instead ensure that package managers can easily make relocatable packages.

distro/windows/Makefile
```diff
@@ -100,7 +100,7 @@ ot-config: ot-check-compiler
 	  -DCMAKE_INSTALL_PREFIX=$(OT_PREFIX) \
 	  -DPYTHON_SITE_PACKAGES=Lib/site-packages \
 	  -DUSE_TBB=OFF \
-	  -DUSE_SPHINX=OFF \
+	  -DUSE_SPHINX=OFF -DUSE_EXPRTK=OFF \
```
dbarbier: I fail to see how it is related to this PR

jschueller: Oops!

jschueller:
I see several cases where this is useful:
- On windows, with our nsis installer, there is no easy way to modify the OT binary with the new installation prefix (no package manager here).
- With linux wheels; It handles relocation by isolating the required libraries and their dependencies in a subdir, I'm not aware in any replacement in the binary either.

dbarbier:
Okay, but if GetLibraryDirectory throws for any reason, OT becomes unusable without user being able to fix it.

dbarbier:
Either modify GetLibraryDirectory to not throw, or add try/catch in GetConfigDirectoryList.

jschueller:
I see. As it happens in the initialization of the dll with the pthread stuff we'd better just output a warning message here.

jschueller:
I added a try/catch in which LOGWARN is used.

PR #702, opened 2018-03-12
Fix pip install on CircleCI

jschueller:


dbarbier:
I do not see any failure on CircleCI, why is this needed?

jschueller:
@dbarbier nbsphinx is not importable and the examples doc is disabled

PR #703, opened 2018-03-13
Set up a default NLopt local solver

jschueller:
When not set, AUGLAG, MLSL families of algorithms fail with a nlopt "invalid argument error".


regislebrun:
+1

PR #704, opened 2018-03-13
Always return Interface instead of Pointer<Implementation>

dbarbier:
DistributionImplementation::getCopula -> Distribution
  DistributionImplementation::getMarginal -> Distribution
  DistributionImplementation::getStandardRepresentative -> Distribution
  DistributionImplementation::getStandardDistribution -> Distribution
  DistributionFactoryImplementation::build -> Distribution
  ProcessImplementation::getMarginal -> Process
  RandomVector::getAntecedent -> RandomVector
  RandomVector::getMarginal -> RandomVector


dbarbier:
My aim is eventually to remove SWIG_TypeQuery from TypedInterfaceObject.i to make bindings faster, but all methods which return a Pointer without a good reason must be fixed first.

jschueller:
I hope this works : )

regislebrun:
+1

dbarbier:
Branch updated, everything should now be in shape.  This may have an import on modules, I hope not.

jschueller:
Only otmixmod seems impacted, could you update it ?

dbarbier:
PR sent for otmixmod.

PR #706, opened 2018-03-14
Better default NearestNeighbourAlgorithm

dbarbier:
Prefer RegularGridNearestNeighbour on a RegularGrid, it provides faster
instantiation and queries.

Add missing methods in RegularGridNearestNeighbour.


regislebrun:
+1

dbarbier:
Commits will cause trouble with #662, I need to rewrite them.

dbarbier:
Okay, done, it was easier than expected.  I had to drop Field::[gs]etValueAtNearestPosition.

regislebrun:
+1

PR #710, opened 2018-03-16
Improve KarhunenLoeveResult::project(ProcessSample)

dbarbier:
Computations with TBB were buggy.
Rewrite project() to perform matrix-matrix multiplications without TBB,
this is much faster than TBB wih matrix-vector multiplications, and as
a bonus gives the right result.

Remove project(Basis), it is very likely that it was buggy too because
of TBB usage, and it was also very slow.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -61,6 +61,7 @@
  * Changed RandomVector::getAntecedent,getMarginal to return a RandomVector instead of Pointer<RandomVectorImplementation>
  * Changed all {get,set}{Evaluation,Gradient,Hessian} methods to work on Evaluation/Gradient/Hessian instead of Pointer
  * Removed Field::[gs]etValueAtNearestPosition
+ * Removed KarhunenLoeveResult::project(Basis), use project(Function) on an AggregatedFunction instead
 
 === Python module ===
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -61,7 +61,6 @@
  * Changed RandomVector::getAntecedent,getMarginal to return a RandomVector instead of Pointer<RandomVectorImplementation>
  * Changed all {get,set}{Evaluation,Gradient,Hessian} methods to work on Evaluation/Gradient/Hessian instead of Pointer
  * Removed Field::[gs]etValueAtNearestPosition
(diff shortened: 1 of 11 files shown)


dbarbier:
BTW ProcessSample::operator[] works on Sample in C++ and on Field in Python, this is a very bad idea.

dbarbier:
I reintroduce project(Basis), it cannot be implemented efficiently with project(Function), but rewritten to not use TBB.  @regislebrun Could you please check?  I think it's right, but this method is not  tested.

regislebrun:
I agree on the fact that operator[] should act the same way in Python and C++. IMO it should return a Sample, as a ProcessSample is a collection of Field sharing the same Mesh, so it is more a collection of Sample associated to the same Mesh. A getField() method is here in C++ if one want to see the ProcessSample more like a collection of Field.

regislebrun:
@dbarbier Is there a technical reason why Basis cannot be replaced by a Collection<Function>, or is it because this is too much work to be done now? I can help on this point.

dbarbier:
@regislebrun No technical reason, but this should already be addressed by #647, thus I did not want to add a conflict here.
@jschueller Do you want me to make this change here, or is it already done in your branch?

regislebrun:
ok so +1. I don't know in which order it is better to merge this PR and PR#711, but Nolwen will love to have both ;-)

jschueller:
@dbarbier It's not ready yet, I have to figure out if we can use Psi_k_p_ in chaos, but I fixed the doc, the tests.
See my branch functionalbasis2.

dbarbier:
Ok, my point was moot anyway, I did change project(Basis) implementation, so there was already a conflict.  Replaced by project(FunctionCollection).

regislebrun:
I found a bug with the keepIntact flag when I was working on the Wishart distribution. If you pass a const Matrix as an argument of a method, make a copy of it inside of the method and perform some linear algebra (eg solveLinearSystem) with it with the keepIntact flag set to false, the const Matrix is modified. A copyOnWrite() should be added somewhere, but it would completely cancel the benefit of the flag as it would involve a copy anyway...

Here it is harmless.

dbarbier:
@jschueller I do not plan to change operator[] in this PR, please merge it as is when you approve it, thanks.

PR #715, opened 2018-03-19
Replace Basis by Collection<Function> in low-level classes.

jschueller:
Had to add AdaptiveStrategyImplementation::Psi_ to hold the full max basis.

Workaround a segfault in CanonicalTensorEvaluation when serializing directly
the PersistentCollection<FunctionPersistentCollection> basis_.


jschueller:
Needed by #614

regislebrun:
+1

PR #719, opened 2018-03-19
[RFC] Move cache handling away from Evaluation/Function

dbarbier:
This is now only handled by MemoizeFunction.

Modify `MemoizeEvaluation::__str__` to mimic EvaluationImplementation;
there is no display on cache and history, but this helps fixing
discrepancies with previous code when cache and history were in
EvaluationImplementation.

Make all methods from EvaluationImplementation virtual, this looks
more logical, but changes Function description in many C++ tests.

Let callsNumber_ member be an AtomicInt instead of UnsignedInteger.
This ensures that counts are right even inside a multithreaded block.


dbarbier:
This is a preliminary work, not well tested, to demonstrate how cache could be handled by MemoizeFunction.
The change in lib/test/t_AbdoRackwitz_std.expout is interesting, it seems to show that cache was not working properly in this test case.

jschueller:
The change in t_AbdoRackwitz_std is due to the fact that SymbolicFunction does not use the cache.

For me this is good to go, I dont understand why you want to wait. Maybe you thought we'd have more time to test it during the 1.12 development ?

dbarbier:
Cache was activated at line 84 of t_AbdoRackwitz_std.cxx.

I just wanted to tell that this had not been fully tested, but I am afraid that nobody will test it anyway, so let merge.

jschueller:
SymbolicEvaluation does not use p_cache_ even if the enabled flag is on:
```
Point SymbolicEvaluation::operator() (const Point & inP) const
{
  Point result(parser_(inP));
  ++ callsNumber_;
  if (isHistoryEnabled_)
  {
    inputStrategy_.store(inP);
    outputStrategy_.store(result);
  }
  return result;
}

/* Operator () */
Sample SymbolicEvaluation::operator() (const Sample & inS) const
{
  UnsignedInteger size = inS.getSize();
  Sample outSample(size, getOutputDimension());
  for (UnsignedInteger i = 0; i < size; ++ i) outSample[i] = operator()(inS[i]);
  outSample.setDescription(getOutputDescription());
  return outSample;
}
```

jschueller:
Could you rebase to make sure it still works with the latest changes ?

jschueller:
What do you say @regislebrun @sofianehaddad ?

dbarbier:
Waow, so cache was activated to check that it was not used?  If this is really what was intended, I can drop cache from this test.
Rebase in progress.

regislebrun:
Sorry for the silence, I was struggling with a bug in my branch improve_bernstein_copula_factory. For me it looks ok. I remember that it was decided a long time ago not to use the cache for symbolic functions as they were intended to implement only very light functions for which it was more efficient to recompute things instead of checking the cache (which was on by default). It was one of these bad ideas we rediscover from time to time.

So +1

PR #720, opened 2018-03-20
Speed-up RankMCovarianceModel::operator()(s,t)

dbarbier:
Functions on t were evaluated several times, and remove matrix scaling.

Fix RankMCovarianceModel for input or output dimension > 1.

### DIFF
--- a/lib/src/Base/Stat/RankMCovarianceModel.cxx
+++ b/lib/src/Base/Stat/RankMCovarianceModel.cxx
@@ -116,8 +116,8 @@ CovarianceMatrix RankMCovarianceModel::operator() (const Point & s,
   {
     MatrixImplementation phiS(outputDimension_, size);
     MatrixImplementation phiT(outputDimension_, size);
-    Collection<Scalar>::iterator itPhiS = phiS.begin();;
-    Collection<Scalar>::iterator itPhiT = phiT.begin();;
+    Collection<Scalar>::iterator itPhiS = phiS.begin();
+    Collection<Scalar>::iterator itPhiT = phiT.begin();
     for (UnsignedInteger i = 0; i < size; ++i)
     {
       const Point evalPhiS(functions_[i](s));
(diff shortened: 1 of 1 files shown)


lib/src/Base/Stat/RankMCovarianceModel.cxx
```diff
@@ -114,15 +114,32 @@ CovarianceMatrix RankMCovarianceModel::operator() (const Point & s,
     }
   else
   {
+    MatrixImplementation phiS(outputDimension_, size);
+    MatrixImplementation phiT(outputDimension_, size);
+    Collection<Scalar>::iterator itPhiS = phiS.begin();;
+    Collection<Scalar>::iterator itPhiT = phiT.begin();;
```
jschueller: double semicolons

lib/src/Base/Stat/RankMCovarianceModel.cxx
```diff
@@ -114,15 +114,32 @@ CovarianceMatrix RankMCovarianceModel::operator() (const Point & s,
...
+      const Point evalPhiS(functions_[i](s));
+      std::copy(evalPhiS.begin(), evalPhiS.end(), itPhiS);
+      itPhiS += outputDimension_;
+      const Point evalPhiT(functions_[i](t));
+      std::copy(evalPhiT.begin(), evalPhiT.end(), itPhiT);
+      itPhiT += outputDimension_;
+    }
+    for (UnsignedInteger i = 0; i < size; ++i)
+    {
+      itPhiT = phiT.begin();;
```
jschueller: here too

dbarbier: Thanks, fixed

lib/src/Base/Stat/RankMCovarianceModel.cxx
```diff
@@ -114,27 +114,23 @@ CovarianceMatrix RankMCovarianceModel::operator() (const Point & s,
...
       itPhiT += outputDimension_;
     }
     for (UnsignedInteger i = 0; i < size; ++i)
     {
+      const Point phiS(functions_[i](s));
       itPhiT = phiT.begin();
       for (UnsignedInteger j = 0; j < size; ++j)
       {
         const Point ptPhiT(Point(Collection<Scalar>(itPhiT, itPhiT + outputDimension_)) * covariance_(i, j));
         dger_(&dim, &dim, &plusOne,
```
regislebrun: Not very fond of direct calls to LAPACK, but when the speed-up is significant and there is no other option...

regislebrun:
+1

PR #721, opened 2018-03-20
Deprecate Evaluation::operator()(inP,parameter)

dbarbier:
Likewise for gradient and hessian.

Changing parameters has a cost, functions must be cloned, this should not
be hidden.


regislebrun:
+1

PR #723, opened 2018-03-21
Remove FunctionalBasis

jschueller:
These two commits from #647 are still relevant:
- Remove the useless FunctionalBasis interface
- Replace Basis::isFunctional by isFinite


dbarbier:
AFAICT we do not need Basis::isFinite anymore.  IIRC we said that algorithms which deal with orthogonal basis know how to generate new functions when needed, and everywhere else we use a Collection of Function thanks to #715.

jschueller:
Then we could could as well drop FiniteBasis. But there are still methods which require a Basis, and I dont want to look into this right now, maybe in another PR.

regislebrun:
+1 Basis could be replaced by Collection<Function> in FunctionalBasisProcess too, but it can wait for another PR

PR #724, opened 2018-03-21
Use Sample::operator(i, j)

jschueller:

PR #725, opened 2018-03-22
 Revert Field::operator(i,j) modification 

jschueller:
Revert erroneous FieldImplementation modification in #724 that results in a recursive call

Also add Field[i,j] missing operator in Python, as well as len(TimeSeries) operator.


jschueller:
It seems getValueAtIndex, setValueAtIndex are redundant with operator[]

dbarbier:
@jschueller No, previous changes are right, you only have to change
```
const Scalar & FieldImplementation::operator () (const UnsignedInteger i,
    const UnsignedInteger j) const
{
#ifdef DEBUG_BOUNDCHECKING
  return at(i, j);
#else
  return values_(i, j);
#endif /* DEBUG_BOUNDCHECKING */
}

```

dbarbier:
Okay, C++ is good.  By reading Python code it looks like we may rename Field into Sample :-D

PR #726, opened 2018-03-23
Add a test for Sobol indices distribution

jschueller:


regislebrun:
No C++ counterpart?

jschueller:
Ok, added

PR #633, opened 2017-12-08
Add content to the examples section

jschueller:
This imports contents from the old UseCasesGuide to the examples section.
Sometimes new notebooks are created, and often it consists in completing the existing notebooks from the first wave of examples with contents from the UseCasesGuide with additional part of scripts, text details, figures, sometimes math but kept to a minimum.
This basically covers all the UCG except the last section (5) on stochastic processes.

Here is the list I used to track advancement:
```
1.0.1 UC : Creation of the random input vector from a distribution > Random vector manipulation
1.1.1 UC : List of usual distributions > cf Python API
1.1.2 UC : Creation of a truncated distribution > Create a truncated distribution
1.1.3 UC : List of Copula > cf Python API
1.1.4 UC : Creation of nD distribution from (marginals, copula) > Create a composed distribution
1.1.5 UC : Creation of a nD distribution from a Mixture > Create a mixture of PDFs
1.1.6 UC : Creation of 1D distribution from a 1D distribution > Create a composite distribution
1.1.7 UC : Manipulation of a distribution > Distribution manipulation
1.1.8 UC : Creation of a custom distribution or copula from the python script > Creation of a custom distribution or copula
1.1.9 UC : Creation of a custom random vector from the python script > Creation of a custom random vector
1.2.1 UC : Import / Export data from a file at format CSV (Comma Separated Value) > Import / export a sample via a CSV file
1.2.2 UC : Drawing Empirical CDF, Histogram, Clouds / PDF or superposition of two clouds from data > Visualize clouds/empirical CDF/histogram/cobweb/pairs
1.2.3 UC : Do two samples have the same distribution : QQ-plot visual test, Smirnov numerical test > Compare samples using Smirnov, QQ-plot
1.2.4 UC : Are two scalar samples independent : ChiSquared test, Pearson test, Spearman test > Sample independence test
1.2.5 UC : Particular manipulations of the Pearson and Spearman tests, when the first sample is of dimension superior to 1. > cf Python API
1.2.6 UC : Regression test between two scalar numerical samples > Test samples independence using regression test
1.2.7 UC : Distribution fitting tests, numerical and visual validation tests : Chi-squared test, Kolmogorov test, QQ-plot graph > Distribution fitting test using chi2/kolmogorov/qq
[...]

PR #729, opened 2018-03-28
Fix Sample html conversion

jschueller:
See http://trac.openturns.org/ticket/943

Introduced in https://github.com/openturns/openturns/pull/633/commits/fbafb39fd43967cf2d38d44b548f4d981ace030e


regislebrun:
+1

jschueller:
@dbarbier @regislebrun this adds build dependency to swig>=2.0.9

dbarbier:
@jschueller No problem for me with this dependency

regislebrun:
@jschueller Ok too if it doesn't cause trouble with the different OT packages

PR #730, opened 2018-03-28
Fix TranslationFunction constructor

dbarbier:
Add tests.


lib/src/Base/Func/openturns/TranslationFunction.hxx
```diff
@@ -44,7 +44,7 @@ public:
   /* Some typedefs for easy reading */
 
   /** Default constructor */
-  explicit TranslationFunction(const Point constant);
+  explicit TranslationFunction(const Point & constant);
```
jschueller: In setConstant too

dbarbier: Thanks, fixed

regislebrun:
+1

PR #731, opened 2018-03-28
Fix EnclosingSimplexAlgorithm constructor

dbarbier:
When there are no simplices, Mesh::isRegular() does not check
vertex position, and thus always return true.


regislebrun:
+1

PR #732, opened 2018-03-29
Include internal OT directories first

dbarbier:
A similar change had already been performed in 6f2986e to force SWIG
to include files from source directories and not installed ones.

But the same fix is also needed for lib when one sets CMAKE_INSTALL_PREFIX
to a path which is used by some system library.

This commit ensures that all local paths are searched before system paths.


lib/src/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ endif()
 foreach (built_source ${BUILT_SOURCEFILES})
   set_source_files_properties (${built_source} PROPERTIES GENERATED TRUE)
 endforeach (built_source)
-include_directories (${INTERNAL_INCLUDE_DIRS})
+include_directories (BEFORE ${INTERNAL_INCLUDE_DIRS})
 link_directories (${CMAKE_Fortran_IMPLICIT_LINK_DIRECTORIES})
```
jschueller: Could you remove the `link_directories (${CMAKE_Fortran_IMPLICIT_LINK_DIRECTORIES})` line ?

dbarbier: Are you sure?  In lib/src/CMakeLists.txt we have
```
if (NOT DEFINED LAPACK_LIBRARIES)
  # workaround for cmake issue #0009976
  if (CMAKE_VERSION VERSION_LESS 2.8.4)
    enable_language (Fortran)
  endif ()
  find_package (LAPACK REQUIRED)
  if (LAPACK_FOUND)
    set (OPENTURNS_HAVE_LAPACK TRUE)
  endif ()
endif ()
```
so Fortran libs may be needed.

jschueller: We used CMAKE_CXX_IMPLICIT_LINK_DIRECTORIES at one point and it was useless, so I wonder if this one is really useful.

dbarbier: Okay, but maybe it would be better to remove it just after 1.11, to have more time to detect and fix breakage?

jschueller: Ok

PR #733, opened 2018-03-29
Fix warnings

jschueller:


jschueller:
I had to pin nbsphinx because of https://github.com/spatialaudio/nbsphinx/issues/174

PR #734, opened 2018-03-29
Use PNG format for inline notebook plot 

jschueller:
Avoids to send a huge svg file when the data is big:
http://trac.openturns.org/ticket/944

### DIFF
--- a/python/src/Graph.i
+++ b/python/src/Graph.i
@@ -41,12 +41,12 @@ OTTypedInterfaceObjectHelper(Graph)
 %include openturns/Graph.hxx
 
 %pythoncode %{
-def Graph__repr_svg_(self):
+def Graph__repr_png_(self):
     """Get the SVG representation."""
-    from .viewer import ToSVGString
-    return ToSVGString(self)
+    from .viewer import ToImageString
+    return ToImageString(self)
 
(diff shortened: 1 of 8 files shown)


jschueller:
SVG is still cool (no pixels, always right size), maybe I should try to render to svg first and if the length of the string exceeds a limit fallback to png. Thoughts ?

jschueller:
It is not possible to check the length, and the generation of the svg itself is costly too, so this is the best option.

regislebrun:
+1

jschueller:
@regislebrun I added a resourcemap key to switch png/svg (still defaults to png)

python/src/Graph.i
```diff
@@ -3,58 +3,32 @@
...
   {
-    typedef _PySequence_ Type;
+    typedef _PyObject_ Type;
   };
 
   template <>
   inline
   OT::Graph
-  convert<_PySequence_,OT::Graph>(PyObject * pyObj)
+  convert<_PyObject_, OT::Graph>(PyObject * pyObj)
```
dbarbier: Is this block still needed?  It looks like it could be dropped.

jschueller: Yes

regislebrun:
still +1

PR #737, opened 2018-04-03
Allow to serialize nan/inf values

jschueller:
Nan/inf values are correctly written to xml, but converted to 0 when reading back.


dbarbier:
If you want to support reading inf/nan, this must be done for all numeric types, not only Scalar.

jschueller:
@dbarbier ok

regislebrun:
+1

jschueller:
cc @aurelieladier

PR #740, opened 2018-04-03
New P1LagrangeInterpolation class

dbarbier:
This class inherits from FieldFunctionImplementation,
and allow interpolating a Field on another Mesh.

This class is somewhat special, since it can deal with Fields
of any dimension (and output Field has the same dimension as
input Field), whereas input and output dimensions are usually
fixed by FieldFunction.
In order to provide basic checks during evaluation, a dimension
is declared in constructor, can be changed afterwards with the
setDimension() method.


lib/src/Base/Func/openturns/P1LagrangeInterpolation.hxx
```diff
@@ -0,0 +1,118 @@
+//                                               -*- C++ -*-
+/**
+ *  @brief P1 Lagrange piecewise linear function.
```
jschueller: The description can be updated

regislebrun:
+1

python/test/t_P1LagrangeInterpolation_std.py
```diff
@@ -0,0 +1,11 @@
...
+
+from __future__ import print_function
+import openturns as ot
+
+mesh = ot.RegularGrid(0.0, 1.0, 4)
+values = [(x, 2.0 * x, x*x) for x in mesh.getValues()]
+outPoint = [2.3]
+interpolation = ot.P1LagrangeInterpolation(mesh, ot.Mesh([outPoint]), len(values[0]))
+print("Interpolation=", interpolation)
+print("Values at", outPoint, "=", interpolation(ot.Field(mesh, values)))
```
jschueller: http://autobuilder.openturns.org/openturns/pull/740/squeeze-x86_64-ot-linux/LastTest.log:
-Values at [2.3] = 0 : [ 2.3 2.3 4.6 5.5 ]
+Values at [2.2999999999999998] = 0 : [ 2.3 2.3 4.6 5.5 ]

PR #741, opened 2018-04-03
Added a flag to tell if the symbolic parsers have to check their output

regislebrun:
In some circumstances it is desirable to let symbolic parsers produce NaNs or Infs when they are evaluated over a sample and then to filter out the result afterward instead of throwing an exception at the first nonnoral result encountered.

### DIFF
--- a/lib/src/Base/Func/SymbolicParserExprTk.cxx
+++ b/lib/src/Base/Func/SymbolicParserExprTk.cxx
@@ -65,27 +65,36 @@ Point SymbolicParserExprTk::operator()(const Point & inP) const
   if (outputVariablesNames_.getSize() == 0)
   {
     // One formula by marginal
-    for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
+    if (checkResult_)
     {
-      const Scalar value = expressions_[outputIndex]->value();
-      // ExprTk does not throw on domain/division errors
-      if (!SpecFunc::IsNormal(value))
-        throw InternalException(HERE) << "Cannot evaluate " << formulas_[outputIndex] << " at " << inputVariablesNames_.__str__() << "=" << inP.__str__();
-      result[outputIndex] = value;
-    }
-  }
+      for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
+      {
+        const Scalar value = expressions_[outputIndex]->value();
+        // ExprTk does not throw on domain/division errors
+        if (!SpecFunc::IsNormal(value))
+          throw InternalException(HERE) << "Cannot evaluate " << formulas_[outputIndex] << " at " << inputVariablesNames_.__str__() << "=" << inP.__str__();
+        result[outputIndex] = value;
+      } // outputIndex
+    } // checkResult
+    else
+      for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
+        result[outputIndex] = expressions_[outputIndex]->value();
+  } // outputVariablesNames_.getSize() == 0
   else
   {
     // Single formula, evaluate expression
     (void) expressions_[0]->value();
 
     std::copy(inputStack_.begin() + inputDimension, inputStack_.end(), result.begin());
-    for (UnsignedInteger outputIndex = 0; outputIndex < outputDimension; ++ outputIndex)
+    if (checkResult_)
     {
-      if (!SpecFunc::IsNormal(result[outputIndex]))
-        throw InternalException(HERE) << "Cannot evaluate " << formulas_[0] << " at " << inputVariablesNames_.__str__() << "=" << inP.__str__();
-    }
-  }
+      for (UnsignedInteger outputIndex = 0; outputIndex < outputDimension; ++ outputIndex)
+      {
+        if (!SpecFunc::IsNormal(result[outputIndex]))
+          throw InternalException(HERE) << "Cannot evaluate " << formulas_[0] << " at " << inputVariablesNames_.__str__() << "=" << inP.__str__();
+      }
+    } // checkResult
+  } // !outputVariablesNames_.getSize() == 0
   return result;
 }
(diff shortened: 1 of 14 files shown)


lib/src/Base/Func/SymbolicParserExprTk.cxx
```diff
@@ -82,7 +82,7 @@ Point SymbolicParserExprTk::operator()(const Point & inP) const
     std::copy(inputStack_.begin() + inputDimension, inputStack_.end(), result.begin());
     for (UnsignedInteger outputIndex = 0; outputIndex < outputDimension; ++ outputIndex)
     {
-      if (!SpecFunc::IsNormal(result[outputIndex]))
+      if (checkResult_ && !SpecFunc::IsNormal(result[outputIndex]))
```
dbarbier: The whole for-loop is useless if `checkResult_` is false, this block should be written instead
```
if (checkResult_)
  for...
```

BTW the previous loop should also be rewritten IMO
```
    if (checkResult_)
    {
      for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
      {
        const Scalar value = expressions_[outputIndex]->value();
        // ExprTk does not throw on domain/division errors
        if (!SpecFunc::IsNormal(value))
          throw InternalException(HERE) << "Cannot evaluate " << formulas_[outputIndex] << " at " << inputVariablesNames_.__str__() << "=" << inP.__str__();
        result[outputIndex] = value;
      }
    }
    else
    {
      for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
        result[outputIndex] = expressions_[outputIndex]->value();
    }
```

This applies of course to both parsers.

regislebrun: @dbarbier good idea

jschueller:
The tests fail on both mingw & msvc.

lib/src/Base/Common/XMLStorageManager.cxx
```diff
@@ -407,12 +407,30 @@ void fromStringConverter(const String & st, String & value)
...
+    value = -std::numeric_limits<_Tp>::quiet_NaN();
+  }
+  else if (st == "inf")
+  {
+    value = std::numeric_limits<_Tp>::infinity();
+  }
+  else if (st == "-inf")
+  {
+    value = -std::numeric_limits<_Tp>::infinity();
+  }
```
dbarbier: Have nothing to do in this PR

lib/src/Base/Func/SymbolicParserExprTk.cxx
```diff
@@ -65,27 +65,36 @@ Point SymbolicParserExprTk::operator()(const Point & inP) const
...
-      result[outputIndex] = value;
-    }
-  }
+    if (checkResult_)
+      {
+	for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
+	  {
+	    const Scalar value = expressions_[outputIndex]->value();
+	    // ExprTk does not throw on domain/division errors
+	    if (checkResult_ && !SpecFunc::IsNormal(value))
```
dbarbier: Remove `checkResult_` here

lib/src/Base/Func/SymbolicParserExprTk.cxx
```diff
@@ -128,7 +137,7 @@ Sample SymbolicParserExprTk::operator() (const Sample & inS) const
       std::copy(inputStack_.begin() + inputDimension, inputStack_.end(), &result(i, 0));
       for (UnsignedInteger outputIndex = 0; outputIndex < outputDimension; ++ outputIndex)
       {
-        if (!SpecFunc::IsNormal(result(i, outputIndex)))
+        if (checkResult_ && !SpecFunc::IsNormal(result(i, outputIndex)))
```
dbarbier: `if(checkResult_)` should be moved outside of the for-loop

dbarbier:
Same as #743, we should not print inf/nan into output, but use isinf/isnan to check values.

lib/src/Base/Common/XMLStorageManager.cxx
regislebrun: I probably messed a merge or a rebase, I never touched this class by myself in this PR...

lib/src/Base/Func/SymbolicParserMuParser.cxx
```diff
@@ -148,18 +156,30 @@ Sample SymbolicParserMuParser::operator() (const Sample & inS) const
...
-      }
-    }
+	for (UnsignedInteger i = 0; i < size; ++i)
+	  {
+	    std::copy(&inS(i, 0), &inS(i, inputDimension), inputStack_.begin());
+	    for (UnsignedInteger outputIndex = 0; outputIndex < outputDimension; ++ outputIndex)
+	      {
+		const Scalar value = parsers_[outputIndex].get()->Eval();
+		// By default muParser is not compiled with MUP_MATH_EXCEPTIONS enabled and does not throw on domain/division errors
+		if (!SpecFunc::IsNormal(value))
```
jschueller: use checkResult_ here too ?

regislebrun:
done (hopefully)

lib/src/Base/Func/SymbolicParserExprTk.cxx
```diff
@@ -65,27 +65,36 @@ Point SymbolicParserExprTk::operator()(const Point & inP) const
...
-      result[outputIndex] = value;
-    }
-  }
+    if (checkResult_)
+      {
+	for (UnsignedInteger outputIndex = 0; outputIndex < result.getDimension(); ++ outputIndex)
+	  {
+	    const Scalar value = expressions_[outputIndex]->value();
+	    // ExprTk does not throw on domain/division errors
+	    if (!SpecFunc::IsNormal(value))
```
jschueller: use checkResult_ ?

regislebrun:
Updated muparser test.

PR #743, opened 2018-04-03
Standardize nan/inf serialization

jschueller:


jschueller:
@dbarbier are you aware of this kind of problems on msvc with inf/nans ?

dbarbier:
@jschueller No, but it is unclear to me where the problem is; you should add a similar test in C++, maybe this comes from Python?

regislebrun:
I am not sure that this case is specified by the norm (I don't have access to the official document, not free). If it is not, you can expect almost anything, which seems to be the case...

dbarbier:
In fact, it is a very bad idea to check output in this case, AFAICT we do not know how NaNs are printed.  Thus output should be removed, and isnan/isinf (also available in Python from math module) used instead to check value.

jschueller:
Ok. For now I want to print to get the representation in order to modify the string matching code for msvc. Then I can use assert(str(point)==str(point2)) in the final test.

dbarbier:
Err no, `assert(point==point2)`, but it won't work because of positive and negative inf

dbarbier:
No in fact my suggestion do not work either, isinf/isnan must be used.

jschueller:
Why do you think string comparison of the two points won't work (assert(str(point)==str(point2)))?

dbarbier:
Okay I was unclear; it should work, but it is a bad habit to compare string representation.  If you want to check that something is inf or nan, it is better to use isinf/isnan IMO.

regislebrun:
BTW there are SpecFunc::IsNaN() and SpecFunc::IsInf() for this purpose ;-)

jschueller:
Yes, that's a lot more reasonable!

dbarbier:
Yes, looks much better this way.  We specialize fromStringConverter for Bool, so maybe we will have to do the same for toStringConverter, autobuilders will tell.

dbarbier:
They make sense for all numeric types

jschueller:
I think we're there, updated my commit:

```
Standardize nan/inf serialization

The string representation of nan or inf values varies across compilers,
for example gcc: nan/inf, vc<14: 1.#QNAN, 1.#IND, 1.#INF etc.

Now when we detect nan/inf using SpecFunc primitives and serialize it as "nan", "inf" of "-inf".
The tests are updated accordingly.
```

dbarbier:
Tests are not accurate enough IMO, you should print output of the 3 functions IsInf, IsNaN and IsNormal, to make sure that values keep their semantics after write+read.  This does not preserve negative NaNs, but this does not make sense, so there is no problem.  Otherwise it looks good to me.

jschueller:
Ok, updated.

regislebrun:
+1

PR #745, opened 2018-04-04
Document SpecFunc::IsNormal

jschueller:


jschueller:
isinf, isnan are already available in math, so add only IsNormal.

dbarbier:
Docstring is wrong, in fact 'IsNormal(x)' is just 'not (isnan(x) || isinf(x))'

regislebrun:
+1

regislebrun:
As SpecFunc.IsNaN() and SpecFunc.IsInf() are visible by the user IMO they should be implemented. BTW, I don't want them to be hidden from the Python interface as it helps a lot in developing Python code that will be later translated into C++.

jschueller:
SpecFunc.IsNaN() and SpecFunc.IsInf() are not exported in the Python interface (they are available in the math module).

dbarbier:
@jschueller In fact IsNaN is exported since there is a typo in SpecFunc.i, but I agree that we should try to avoid function duplicated with math module

regislebrun:
+1

PR #747, opened 2018-04-05
Allow blank in imported sample description

jschueller:
Improve Sample::ImportFromText to allow a blank space in the description


jschueller:
The code is simpler and shorter too :)

PR #748, opened 2018-04-05
Fix Sample partial copy constructor

dbarbier:


dbarbier:
I guess so, but did not check.

PR #753, opened 2018-04-08
Fix EmpiricalBernsteinCopula tests

dbarbier:
Enable Python test.
Update C++ test, there is a change because of 0620d4a.


regislebrun:
+1

jschueller:
Right, I was a bit too quick too merge.

PR #754, opened 2018-04-09
Release

jschueller:
If you want to submit something before the 1.11rc1, speak now :)


regislebrun:
+1

ChangeLog
```diff
@@ -89,12 +89,10 @@
  * #926 (Covariance model active parameter set behavior)
  * #928 (Covariance model/ Field/Process properties naming)
  * #932 (Optim algo iteration/evaluation number)
- * #933 (Compilation failed when Muparser is missing)
  * #937 (Small bugs in FunctionalChaosResult::get{Residuals, RelativeErrors})
  * #938 (DistributionFactory::buildEstimator to handle Exception)
  * #939 (The help of the getMaximumDegreeStrataIndex method is wrong.)
  * #941 (Duplicate with SobolIndicesExperiment)
- * #943 (Sample don't show up correctly in notebooks)
```
dbarbier: Is this change intentional (removal of two bugs)?

distro/debian/patches/series
```diff
@@ -1,4 +1,2 @@
 tests-disable-coupling_tools.patch
 discard-levelsetmesher-test.patch
-fix-multiprocessing-hurd.patch
-test-NormalGamma.patch
```
dbarbier: I will have to keep test-NormalGamma.patch in Debian

ChangeLog
jschueller: Yes, these two bugs affected the git version only, not a published version, so the end user does not care about that.

distro/debian/patches/series
jschueller: Why is that ? I applied it.

ChangeLog
```diff
@@ -89,12 +89,10 @@
...
  * #944 (Inline plots crash in notebooks if too many points)
  * #942 (Negative optimization bounds in Kriging)
```
sofianehaddad: @jschueller I think 942 also affect git version ?

jschueller: Indeed

PR #756, opened 2018-04-13
Fix build with CMake 3.11

dbarbier:


jschueller:
Seems it fails on OSX after all:
https://github.com/conda-forge/openturns-feedstock/pull/51#partial-pull-merging

dbarbier:
@jschueller Weird, do you know which make version is used on OSX?

dbarbier:
Got it, should work now

PR #757, opened 2018-04-13
Default R executable value

jschueller:
Sets a reasonable default when R is not found instead of an empty value
in the config file if R is found in the PATH at runtime.

Adresses http://trac.openturns.org/ticket/946


dbarbier:
Maybe R${CMAKE_EXECUTABLE_SUFFIX} instead so that it works on Windows too?

jschueller:
Agreed it's cleaner, updated.
But I think the windows shell allows to forget the ".exe". If the os-createprocess rm key is enabled I dont think it will work as the PATH may be ignored though.

PR #758, opened 2018-04-13
FieldFunction to aggregate input/output meshes

jschueller:
I started to add required FieldFunction::inputMesh_, outputMesh_ attributes

A few thoughts:
- I did not find a simple way to get an 1-d mesh from a (sorted) list of values (not regular)
- Is the input mesh of a KLProjection defined ? It is needed if we build a FieldToFieldConnection
- This forces to have meshes for all fieldfunction, including ValueFunction
- Often ctors are a bit redundant with function, evaluation, evaluationimplementation args, got to check how to simplify this

TODO: modify the operators


dbarbier:
Thanks @jschueller.
You added input and output mesh arguments in FieldFunction constructor as discussed, but other functions should be modified too IMO:
 * FieldToPointFunction takes an input Mesh and an output dimension
 * PointToFieldFunction takes an input dimension and an output Mesh

dbarbier:
* In which function do you need that?  IIRC there are several places where we write something like `Sample sample(size, dimension); sample.getImplementation()->setData(values); Mesh mesh(sample);`.  But `data_` is then allocated twice, maybe we should add a `Sample(Point)` constructor.
* The input mesh of a KLProjection is available via getModesAsProcessSample().getMesh() as you did, but  there should be an accessor to avoid building a ProcessSample, which is an heavy object.
* I do not see the problem with ValueFunction

dbarbier:
Also these operators do not make sense:
```
Point FieldFunction::operator() (const Scalar timeStamp, const Point & inP) const;
Point FieldFunction::operator() (const Point & location, const Point & inP) const;
```
FieldFunction should only operate on Field and ProcessSample.
Operators above are used by ODESolver, but a ParametricFunction should be used instead.

jschueller:
- I did not see the mesh(values) constructor, sorry
- I modified FieldToPointFunction, PointToFieldFunction too
Still in progress

dbarbier:
Ok.  This PR will become huge when it is finished, maybe we could merge this one if there is no objection (when CI is fixed), and you continue in other PRs?

jschueller:
Why not, but even this first stage is still not ready yet.

lib/src/Base/Solver/ODESolverImplementation.cxx
```diff
@@ -85,6 +85,13 @@ FieldFunction ODESolverImplementation::getTransitionFunction() const
   return transitionFunction_;
 }
 
+Point ODESolverImplementation::evaluate(const Scalar t, const Point & inP) const
+{
+  Field X(Mesh(Sample(1, Point(1, t))), Sample(1, inP));
+  Field Y(transitionFunction_(X));
+  return Y.getValues()[0];
+}
```
dbarbier: As said previously, IMO transitionFunction_ should be a ParametricFunction instead of a FieldFunction, so that we do not have to create so many objects for each evaluation.

dbarbier:
@jschueller See the fieldfunction branch in my repository, I modified your code to copy transitionFunction_ only once for each call to solve().

jschueller:
Ok, applied, thanks

jschueller:
I changed FieldFunction's `Field operator() (const Field & inFld) const` into
`Sample FieldFunction::operator() (const Sample & inFld) const`
I think it becomes less evident that we need a new type for field values as now we can check for values size against the mesh vertices number : it's harder to confuse, but this can still evolve.

Anyways now this PR goal is complete: the field is known at the initialization, and not passed around at each evaluation, so I propose it for review.

jschueller:
Also I assumed input and output mesh can have different vertices number, is this correct ?

regislebrun:
@jschueller Yes, the input and output meshes can be totally different: different dimensions, different vertices number and position, different simplices. Can the review waits for late afternoon?

jschueller:
Yes, this can even wait 2 weeks of holidays :)

lib/src/Base/Algo/KarhunenLoeveProjection.cxx
```diff
@@ -95,9 +95,10 @@ KarhunenLoeveProjection::Implementation KarhunenLoeveProjection::getMarginal(con
 }
 
 /* Operator () */
-Point KarhunenLoeveProjection::operator() (const Field & inFld) const
+Point KarhunenLoeveProjection::operator() (const Sample & inFld) const
 {
-  const Point outPoint(result_.project(inFld));
+  const Field inputField(getInputMesh(), inFld);
+  const Point outPoint(result_.project(inputField));
```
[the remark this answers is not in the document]
jschueller: Oh, yes, that will avoid to pass a Field.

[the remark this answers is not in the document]
dbarbier: And symmetrically liftAsField should return a Sample so that KarhunenLoeveLifting::operator()(Point) does not have to create a Field.

dbarbier:
@jschueller You forgot to update P1LagrangeInterpolation.

lib/src/Base/Algo/KarhunenLoeveResultImplementation.cxx
```diff
@@ -152,10 +152,15 @@ Point KarhunenLoeveResultImplementation::project(const Function & function) cons
 Point KarhunenLoeveResultImplementation::project(const Field & field) const
 {
   if (field.getMesh() == modesAsProcessSample_.getMesh())
-    return projection_ * field.getValues().getImplementation()->getData();
+    return project(field.getValues());
   return project(Function(new P1LagrangeEvaluation(field)));
```
dbarbier: I see why you kept project as is, but honestly it would be cleaner to state that `KLResult::project` projects values on its input mesh, and user would call P1LagrangeInterpolation if needed.

dbarbier:
Thanks, please consider commit a18fb96 in my repo.  I added getMesh to avoid calling `result.getModesAsProcessSample()` just to get some pieces of information.
Other accessors should be added IMO to replace:
 * getModesAsProcessSample().getDimension()
 * getModesAsProcessSample()[0].getDescription()
 * getModesAsProcessSample().getSize(), can already be changed to getEigenValues().getSize()

dbarbier:
I do not understand the use case for VertexFunction, it should be dropped IMO.

lib/src/Base/Func/P1LagrangeInterpolation.cxx
```diff
@@ -207,18 +190,17 @@ void P1LagrangeInterpolation::computeProjection()
   }
 }
 
-Field P1LagrangeInterpolation::operator()(const Field & field) const
+Sample P1LagrangeInterpolation::operator()(const Sample & values) const
```
dbarbier: @jschueller My bad, I forgot to increment callsNumber_ in this method, can you please fix it?

dbarbier:
More precisely, I more or less see why VertexFunction had been added, but this task can been performed by PointToFieldFunction.  And since this PR now fixes input mesh, this function always gives the same output, this is why it looks useless to me.  Likewise for VertexValueFunction.

lib/src/Base/Func/P1LagrangeInterpolation.cxx
jschueller: done

regislebrun:
The output of a VertexValueFunction depends on both the vertices and the values of the input field. Even if the input mesh is now fixed, the input values may change from call to call (otherwise it looks strange to build a function to call it always on the same input!). Now that the input mesh is fixed, a VertexFunction will always return the same output so it is probably useless now, but I disagree with you when you say that the same task can be done using a PointToFieldFunction. Could you elaborate a little bit on this?

jschueller:
TODO: drop VertexFunction

dbarbier:
@jschueller Why did you drop commit c042c7c?

jschueller:
@dbarbier I did not sync my branch on my laptop, fixed

jschueller:
@dbarbier, @regislebrun @sofianehaddad this is getting huge. I plan on working on sobol simulation stuff now. Maybe this could be merged ?

lib/src/Base/Solver/RungeKutta.cxx
```diff
@@ -80,14 +81,24 @@ Sample RungeKutta::solve(const Point & initialState,
...
+                              const Scalar t,
                               const Point & state,
                               const Scalar h) const
 {
-  const Point k1(transitionFunction_(t, state));
-  const Point k2(transitionFunction_(t + 0.5 * h, state + k1 * (0.5 * h)));
-  const Point k3(transitionFunction_(t + 0.5 * h, state + k2 * (0.5 * h)));
-  const Point k4(transitionFunction_(t + h, state + k3 * h));
+  Point parameter(1, t);
+  if (transitionFunction.getParameter().getDimension() > 0)
```
regislebrun: This check looks strange. The parameter dimension should be equal to 1 and it should be checked only once in the constructor. Same remark for the Fehlberg class

lib/src/Base/Algo/KarhunenLoeveLifting.cxx
```diff
@@ -95,17 +95,17 @@ PointToFieldFunction KarhunenLoeveLifting::getMarginal(const Indices & indices)
...
+Sample KarhunenLoeveLifting::operator() (const Point & inP) const
 {
-  const Field outField(result_.liftAsField(inP));
+  const Sample outField(result_.liftAsSample(inP));
   callsNumber_.increment();
   return outField;
 }
 
 /* Accessor for the output mesh */
 Mesh KarhunenLoeveLifting::getOutputMesh() const
```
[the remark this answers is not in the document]
jschueller: ok

lib/src/Base/Solver/RungeKutta.cxx
jschueller: I thought there was an example with no parameter, but the tests seem to pass, strange.

regislebrun:
+1

PR #761, opened 2018-04-19
[To integrate in 1.11] Relax empirical bernstein copula tests

regislebrun:
This PR fixes [ticket #947](http://trac.openturns.org/ticket/947#ticket). It allows to build an EmpiricalBernsteinCopula object even if the resulting distribution is not a copula, for statistical estimation purpose. It may be strange to name a distribution an EmpiricalBernsteinCopula even when it is not a proper copula, but the name is in the literature since 2004 and the fact that an arithmetic condition is mandatory to get a proper copula has been proved only in 2017!

### DIFF
--- a/lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
@@ -36,7 +36,7 @@ static const Factory<EmpiricalBernsteinCopula> Factory_EmpiricalBernsteinCopula;
 
 /* Default constructor */
 EmpiricalBernsteinCopula::EmpiricalBernsteinCopula()
-  : CopulaImplementation()
+  : ContinuousDistribution()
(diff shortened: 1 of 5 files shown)


dbarbier:
C++ test fails, the problem is that `isCopula_` is true but this distribution does no more derive from CopulaImplementation, and calls to `getKendallTau` or `getShapeMatrix` enter an infinite loop.

dbarbier:
@regislebrun I just pushed commit 269c9bc in my repository to propose a way to solve this issue; all methods in CopulaImplementation are blindly modified, a manual review is needed.

BTW do we really need this in 1.11?  I would very much prefer to have this PR in 1.12.

regislebrun:
I will close this PR as the circular calls have to be solved first.

regislebrun:
I finally succeeded in fixing the circular dependency between CopulaImplementation and DistributionImplementation, so the PR should be ok now...

PR #765, opened 2018-04-30
Add Evaluation::setCallsNumber

jschueller:
For GUI stuff


sofianehaddad:
I do not see the 'interest'/use.

Maybe we could add an ignore Evaluation::setCallsNumber in swig ?

jschueller:
The use-case is we want to reset the number of calls to zero to count easily the number of evaluations in a calculus, without rebuilding the function, but maybe we can manage another way.

PR #766, opened 2018-05-01
Fix compilation with older cmake

jschueller:


regislebrun:
+1

python/src/CMakeLists.txt
```diff
@@ -57,7 +61,7 @@ macro (ot_add_python_module MODULENAME SOURCEFILE)
     swig_add_library (${MODULENAME} LANGUAGE python SOURCES ${SOURCEFILE} ${swig_other_sources})
   endif ()
 
-  add_dependencies (${SWIG_MODULE_${MODULENAME}_REAL_NAME} generate_swig_runtime)
+  add_dependencies (${SWIG_MODULE_${MODULENAME}_REAL_NAME} generate_swig_runtime generate_docstrings)
```
dbarbier: This is only needed for CMake < 3, I prefer (untested)
```
  add_dependencies (${SWIG_MODULE_${MODULENAME}_REAL_NAME} generate_swig_runtime) 
  if (NOT CMAKE_VERSION VERSION_GREATER 3)
   add_dependencies (${SWIG_MODULE_${MODULENAME}_REAL_NAME} generate_docstrings)
  endif ()
```
so that we remember to delete it when support for CMake < 3 is dropped.

jschueller:
This becomes too complicated, I think I'll just install a more recent cmake on the autobuilder.

PR #768, opened 2018-05-02
Set minimum required CMake version to 2.8.4

dbarbier:
According to #766, CMake >= 3 is needed, thus make this versioned dependency explicit.


jschueller:
In fact it's lower, it's just that 2.8.2 does not work.
Also it wont build on travis since it uses trusty (cmake 2.8.12).
But I agree that we should set it to something higher, and that would allow to remove some old wrokarounds.

dbarbier:
Just tested that 2.8.4 works fine, I'll use that.

jschueller:
You can update architecture.rst too.

jschueller:
Also, you can simplify this:
```
if (NOT DEFINED LAPACK_LIBRARIES)
  # workaround for cmake issue #0009976
  if (CMAKE_VERSION VERSION_LESS 2.8.4)
    enable_language (Fortran)
  endif ()
  find_package (LAPACK REQUIRED)
  if (LAPACK_FOUND)
    set (OPENTURNS_HAVE_LAPACK TRUE)
  endif ()
endif ()
list (APPEND OPENTURNS_LIBRARIES ${LAPACK_LIBRARIES})
```

cmake/FindPythonModule.cmake
```diff
@@ -57,7 +59,6 @@ macro (find_python_module module)
                         OUTPUT_VARIABLE _${module}_version
                         ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)
 
-      set (_${module_upper}_VERSION_MATCH TRUE)
```
jschueller: What did you have in mind with this change ? Currently it makes the i386 build fail as numpy as the cmake logic does not see anymore that it's not available with python3 (we configure for python3 in the same location as the python2 build).

dbarbier: Variables used in macros must always be set to a default value, otherwise the value set in a previous call is used.  Anyway I'll revert this change.

dbarbier: Hmmm I also kept the wrong FPHSA call :-/

PR #769, opened 2018-05-03
Relax empirical bernstein copula

regislebrun:
This PR fixes ticket #947. It allows to build an EmpiricalBernsteinCopula object even if the resulting distribution is not a copula, for statistical estimation purpose. It may be strange to name a distribution an EmpiricalBernsteinCopula even when it is not a proper copula, but the name is in the literature since 2004 and the fact that an arithmetic condition is mandatory to get a proper copula has been proved only in 2017!
I finally succeeded in fixing the circular dependency between CopulaImplementation and DistributionImplementation, so the PR should be ok now...

### DIFF
--- a/lib/src/Uncertainty/Model/DistributionImplementation.cxx
+++ b/lib/src/Uncertainty/Model/DistributionImplementation.cxx
@@ -3102,10 +3102,7 @@ Scalar DistributionImplementation::computeDensityGeneratorSecondDerivative(const
 /* Get the i-th marginal distribution */
 Distribution DistributionImplementation::getMarginal(const UnsignedInteger i) const
 {
-  if (!(i < dimension_)) throw InvalidArgumentException(HERE) << "Marginal index cannot exceed dimension";
-  if (dimension_ == 1) return clone();
-  if (isCopula()) return new Uniform(0.0, 1.0);
-  return new MarginalDistribution(*this, i);
+  return getMarginal(Indices(1, i));
 }
 
 /* Get the distribution of the marginal distribution corresponding to indices dimensions */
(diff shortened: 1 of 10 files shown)


lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -2817,19 +2817,84 @@ CorrelationMatrix DistributionImplementation::getSpearmanCorrelation() const
...
+	Function function;
+	if (isCopula())
+	    function = (bindMethod<DistributionImplementationKendallTauWrapper, Point, Point>(functionWrapper, &DistributionImplementationKendallTauWrapper::kernelForCopula, 2, 1));
+	else
+	    function = (bindMethod<DistributionImplementationKendallTauWrapper, Point, Point>(functionWrapper, &DistributionImplementationKendallTauWrapper::kernelForDistribution, 2, 1));
+        tau(rowIndex, columnIndex) = integrator.integrate(function, square)[0];
+      }
+    } // loop over column indices
+  } // loop over row indices
 }
```
jschueller: you forgot to return tau

dbarbier:
@regislebrun You fixed computeKendallTau, but if you want to go that way, you must do the same changes for all methods defined by CopulaImplementation, and remove this class.

jschueller:
The changelog must be updated in the right section 1.11 or 1.12. So which commits do you want to be backported 34faa48 and a88c051 ?

regislebrun:
Ok to modify the changelog, but currently there is no section 1.11. Do I update the 1.11rc1 section? Concerning the commits, if we don't include the one concerning the marginals we will get one of the most expensive Uniform(0,1) as the result of EmpiricalBernsteinCopula::getMarginal(UnsignedInteger)

regislebrun:
@dbarbier Nice to see you in good shape ;-)!

jschueller:
@regislebrun, yes it's the 1.11rc1 section then

PR #770, opened 2018-05-04
Fixes

jschueller:


jschueller:
@dbarbier Mesh has get(Upper|Lower)Bound methods, I was wondering if they should be deprecated and only be visible in MeshDomain ?

dbarbier:
@jschueller I would instead remove them from MeshDomain

PR #775, opened 2018-05-09
Memoize to preserve useDefaultGrad|Hess attributes

jschueller:
We use that to detect if default finite difference method is used


regislebrun:
+1

PR #778, opened 2018-05-09
Fix segfault in LinearCombinationEval::load

jschueller:
Attributes are reset in setFunction... as it's a ref we cannot pass them


jschueller:
Maybe that's the cause of lockups of t_Stydy_saveload on osx ?

regislebrun:
+1

PR #780, opened 2018-05-09
Fix segfault in Cobyla::ComputeObjectiveAndConstraint

jschueller:


lib/src/Base/Optim/Cobyla.cxx
```diff
@@ -219,7 +219,7 @@ int Cobyla::ComputeObjectiveAndConstraint(int n,
 
   /* Convert the input vector in OpenTURNS format */
   Point inPoint(n);
-  memcpy(&inPoint[0], &x[0], n * sizeof(Scalar));
+  memcpy(&inPoint[0], x, n * sizeof(Scalar));
```
dbarbier: Why is this change needed?

jschueller: Not needed, just simpler.

dbarbier: ok

regislebrun:
+1

PR #782, opened 2018-05-14
Large sample covariance estimation

regislebrun:
I changed the implementation of UserDefinedCovarianceModel: now, it is based on a big covariance matrix instead of many small blocks. It speeds up a lot the creation of the model and its discretization over its natural mesh.


lib/src/Base/Stat/NonStationaryCovarianceModelFactory.cxx
```diff
@@ -75,101 +75,52 @@ CovarianceModel NonStationaryCovarianceModelFactory::build(const ProcessSample &
...
+  // Build the design matrix
+  MatrixImplementation designMatrix(fullDimension, size);
+  MatrixImplementation::iterator designBegin = designMatrix.begin();
+  if (isCentered)
+    {
+      for (UnsignedInteger i = 0; i < size; ++i)
+	{
+	  std::copy(&sample[i].getImplementation()->operator()(0, 0), &sample[i].getImplementation()->operator()(0, 0) + fullDimension, designBegin);
+	  designBegin += fullDimension;
+	  // std::cerr << "i=" << i << ", design=\n" << designMatrix << std::endl;
```
dbarbier: Please remove these comments

lib/src/Base/Stat/NonStationaryCovarianceModelFactory.cxx
```diff
@@ -75,100 +75,30 @@ CovarianceModel NonStationaryCovarianceModelFactory::build(const ProcessSample &
...
-  const ComputeCovariancePolicy policy( sample, isCentered, collection );
-  TBB::ParallelFor( 0, size, policy );
+  CovarianceMatrix currentCovariance(dimension);
+  for (UnsignedInteger index = 0; index < size; ++index)
+    {
+      const UnsignedInteger i = static_cast< UnsignedInteger >(sqrt(2 * index + 0.25) - 0.5);
+      const UnsignedInteger j = index - (i * (i + 1)) / 2;
+      for (UnsignedInteger k = 0; k < dimension; ++k)
+	{
+	  std::copy(&sigma(i * dimension, (j + k) * dimension), &sigma(i * dimension, j * dimension) + dimension, &currentCovariance(0, k));
```
dbarbier: 2nd argument is wrong; this error had been fixed in later commits, thus it would be better to squash commits together.

lib/src/Base/Stat/NonStationaryCovarianceModelFactory.cxx
regislebrun: oupss!

regislebrun:
@dbarbier I squashed all the commits, now it should be ok.

sofianehaddad:
@regislebrun the tests are wrong!

sofianehaddad:
@regislebrun Have a look at 03e83694 for the fix !

regislebrun:
done plus additional tests

lib/src/Base/Stat/UserDefinedCovarianceModel.cxx
```diff
@@ -37,41 +37,67 @@ static const Factory<UserDefinedCovarianceModel> Factory_UserDefinedCovarianceMo
...
-// For a non stationary model, we need N x N covariance functions with N the number of vertices in the mesh
+// For a non stationary model, we need N x (N+1)/2 covariance matrices with N the number of vertices in the mesh
 UserDefinedCovarianceModel::UserDefinedCovarianceModel(const Mesh & mesh,
-    const CovarianceMatrixCollection & covarianceFunction)
+    const CovarianceMatrixCollection & covarianceCollection)
   : CovarianceModelImplementation()
-  , covarianceCollection_(0)
+  , covariance_(0)
   , p_mesh_(0)
   , nearestNeighbour_()
```
dbarbier: `p_mesh_` and `covariance_` should be initialized here

lib/src/Base/Stat/UserDefinedCovarianceModel.cxx
```diff
@@ -37,41 +37,67 @@ static const Factory<UserDefinedCovarianceModel> Factory_UserDefinedCovarianceMo
...
+	++index;
+      } // i
+}
+
+// For a non stationary model, we need N x (N+1)/2 covariance matrices with N the number of vertices in the mesh
+UserDefinedCovarianceModel::UserDefinedCovarianceModel(const Mesh & mesh,
+    const CovarianceMatrix & covariance)
+  : CovarianceModelImplementation()
+  , covariance_(0)
+  , p_mesh_(0)
```
dbarbier: `p_mesh_` and `covariance_` should be initialized here

sofianehaddad: same remark for nearestNeighbour_ ? for all constructors...

dbarbier: It seems so, I do not remember why I did not do that when introducing it.

regislebrun:
@sofianehaddad I will keep the old constructor for now as I had no news about the state of the Anatole flexible module, which is using it. But I agree with you, at one point it will have to be removed.

jschueller:
I agree with sofiane, let's deprecate that ctor now before forgetting about it. There is still time (deprecated in 1.12, removed in 1.13, in 1 year). Moreover the documentation becomes ambiguous as the 2 ctors are not listed: it's not a ctor with 3 arguments, but 2 ctors with 2 arguments each. You can just drop the matrx list argument from the doc.

lib/src/Base/Stat/UserDefinedCovarianceModel.cxx
jschueller: Ok for me, our Pointer class initializes to 0 anyways.

PR #783, opened 2018-05-30
Improved Kolmogorov test

regislebrun:
Now, the p-value is computed using bootstrap when the reference distribution is estimated from the tested sample. See https://stats.stackexchange.com/questions/110272/a-naive-question-about-the-kolmogorov-smirnov-test for the details.


jschueller:
Some chaos tests seem to timeout, how bad are testing times impacted on your machine ?

regislebrun:
It somewhat triples the execution time. The problem is that I have to reduce a lot the sampling size to get acceptable run times, but on the other hand I get completely stupid results if I reduce the size too much... Maybe it is not a problem for the tests (to get stupid results)

jschueller:
An idea I had in mind was to test standard distributions first (Normal/Uniform/Beta/Gamma) in FCA, if one one them fits very good (againt a P-value threshod in the RM), then select it, else fallback to testing all the remaining factories. This will be faster, and maybe the iso-transformation too. What do you think ?

regislebrun:
Your proposal sounds interesting. IMO the automatic construction of the distribution should be done in a separate class, with different strategies depending on the context. I propose to add this in a separate PR as it is not the aim of this PR.

sofianehaddad:
@jschueller @regislebrun why not using only non parametric distributions ? Says for example HistogramFactory and drop **all** parametric estimate. 
IMO this ctor will be updated with future `easy chaos` so we can relax it now and wait new changes with upcoming `easy chaos`. 

What do you think ?

jschueller:
The idea was also to make the metamodel faster, and HistogramFactory should make the iso transformation slower (I have not benchmarked this affirmation).

jschueller:
But maybe HIstogramfactory is better option than the kernelsmoothing fallback that we use in FCA.

jschueller:
TODO: in FCA, make a 2-pass marginale estimation: basic then bootstrap

regislebrun:
@jschueller I made the proposed modifications, and the running time of t_FunctionalChaosSbolIndices_std.py droped to 1.65s instead of 1.97s in master. But there remain several tests which timeout...

regislebrun:
I changed my password recently and I still use the old one from time to time... and the transactions fail!

jschueller:
The tests seem to run well :)

jschueller:
@sofianehaddad good for you ?

jschueller:
@regislebrun it needs a rebase

regislebrun:
done

PR #784, opened 2018-05-31
Enhance FCA distribution estimation

jschueller:
I propose to change the distribution estimation in FCA as follows: check the standard distributions, else fallback to HistogramFactory. The old logic: test all univariate distributions including HistogramFactory else fallback to kernel smoothing.
This speeds up the iso-transfo of the metamodel while avoiding scoring all univariate factories through the Kolmogorov test and leads to a speedup in tests, related to #783.

Beta, Gamma yield slower iso transformations than HistogramFactory, so do not actually propose them, ie only test Normal/Uniform (actually HistogramFactory is quite good @sofianehaddad):
- Normal: ~150k eval/s
- Uniform: ~94k eval/s
- HistogramFactory: ~90k eval/s
- Gamma: ~48k eval/s
- Beta: ~20k eval/s
- KernelSmoothing:  ~1500 eval/s  :-(

There was a bug in the KernelSmoothing fallback in the old code, with an invalid getmarginal of an 1-d sample, happily it was never actually chosen because at least HistogramFactory may always give a good fit.

This logic was made public as we want to control more closely the chaos parameters and use the other FCA ctors.

bench:
https://gist.github.com/jschueller/6f3e80e1d77c2b55399aa84b418429bb


regislebrun:
I had a meeting with Michael Baudin this afternoon partly on this subject. We had the following idea to improve the effectiveness of FittingTest::BestModelKolmogorov:
+ first, rank the models using the naive approach (no sampling to get the exact p-value). The computed p-values underestimate the true p-value
+ second, for the largest p-values, update their estimate by sampling in order to confirm the ranking
At the FunctionalChaosAlgorithm level, we could restrict the set of factories to fit our needs the best (a compromise between accuracy and speed)

I am not very fond of the systematic use of the HistogramFactory class. The associated isoprobabilistic transformation maps all the points outside of the range to the nearest extremity of the range, resulting into a meta-model with saturations. It may be a problem if the user needs to explore the meta-model slightly outside of the convex hull of its training dataset.

jschueller:
The current logic selects HistogramFactory too most of the time.

jschueller:
@regislebrun I updated my code, now I test all 4 standard distributions.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosAlgorithm.cxx
```diff
@@ -268,6 +253,39 @@ AdaptiveStrategy FunctionalChaosAlgorithm::getAdaptiveStrategy() const
...
+  Collection< Distribution > marginals(inputDimension);
+  // The strategy is to test first standard distributions, else fallback to HistogramFactory
+  // Beta, Gamma yield slower iso transformations than HistogramFactory
+  // in the other hand HistogramFactory transformation shrinks the range and may results in a metamodel with saturations
+  Collection< DistributionFactory > factories;
+  factories.add(NormalFactory());
+  factories.add(UniformFactory());
+  factories.add(GammaFactory());
+  factories.add(BetaFactory());
+  LOGINFO("In FunctionalChaosAlgorithm, identify marginal distributions");
```
sofianehaddad: why only 4 factories ?
We try to estimate distributions without advanced statistical tests (like in #783 ) thus IMO we can check all available factories ?

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosAlgorithm.cxx
```diff
@@ -268,6 +253,39 @@ AdaptiveStrategy FunctionalChaosAlgorithm::getAdaptiveStrategy() const
...
+  const Description inputDescription(inputSample.getDescription());
+  for (UnsignedInteger i = 0; i < inputDimension; ++ i)
+  {
+    TestResult bestResult;
+    // Here we remove the duplicate entries in the marginal sample using the automatic compaction of the support in the UserDefined class
+    const Sample marginalSample(inputSample.getMarginal(i).sortUnique());
+    const Distribution candidate(FittingTest::BestModelKolmogorov(marginalSample, factories, bestResult));
+    // This threshold is somewhat arbitrary. It is here to avoid expensive kernel smoothing.
+    if (bestResult.getPValue() > ResourceMap::GetAsScalar("FunctionalChaosAlgorithm-PValueThreshold")) marginals[i] = candidate;
+    else marginals[i] = HistogramFactory().build(marginalSample);
```
sofianehaddad: we could set Histogram or KernelSmoothing. Choice could be oriented using a ResouceMap key

regislebrun:
I had a look at your benchmark. First of all, it should be simplified a little bit to be more readable (no need to define a projection strategy or an adaptive strategy if the algorithm is used through samples only). Second, the key problem for the speed of the isoprobabilistic transformation is the systematic use of a NormalCopula with a nondiagonal correlation matrix. This way, the marginal transformation does not simplify as it has to map the selected margina distributions into normal distributions (with a call to computeQuantile() at each evaluation of the meta-model).
If you want to speed-up things, the most important thing to do is to allow the estimation procedure to select the independent copula. For this, you can test if the off-diagonal components of the Spearman correlation matrix are significantly different from 0, using the fact that under the hypothesis of independence, the estimator is approximately distributed as a centered normal distribution with standard deviation 1/sqrt(n), where n is the sample size. As a result, any value less than 1.96/sqrt(n) in absolute value is not significantly different from 0 at the confidence level of 95%.
This way, you should select the independent copula in your benchmark, and the iso-probabilistic transformation should be either the identity function or an affine function.
I will try this on your branch and give you some figures.

regislebrun:
Using the idea I proposed, I get 300k evals/s regardless of the chosen marginal distributions. It is expected, as the polynomial bases are built for the standard representative of each marginal distribution, ie either this marginal distribution or an affine transformation of it. You have to replace:
```
// For the dependence structure, we test first the independent copula, then the normal copula, but not a non-parametric copula as the penalty on the meta-model evaluation speed is most of the time prohibitive
ComposedDistribution distribution(marginals, NormalCopulaFactory().build(inputSample));
return distribution;
```
by
```
  // For the dependence structure, we check if the Spearman correlation matrix is significantly different from the identity matrix. If yes, we use a normal copula, otherwise we use the independent copula.
  // First, the sampling Spearman correlation
  const CorrelationMatrix rho(inputSample.computeSpearmanCorrelation());
  // Second, the threshold at the symmetric 0.95 level
  const Scalar threshold = DistFunc::qNormal(0.9) / std::sqrt(inputSample.getSize());
  // Third, the significance test
  Bool isIndependent = true;
  for (UnsignedInteger j = 0; j < inputDimension && isIndependent; ++j)
    for (UnsignedInteger i = j+1; i < inputDimension && isIndependent; ++i)
      isIndependent = isIndependent && (abs(rho(i, j)) <= threshold);
  // If all the off-diagonal terms are not significantly different from zero use the independent copula
  if (isIndependent) return ComposedDistribution(marginals);
  // Here we use the sampling Spearman correlation instead of its thresholded version as there is no gain in speed or precision in using the thresholded version.
  return ComposedDistribution(marginals, NormalCopula(NormalCopula::GetCorrelationFromSpearmanCorrelation(rho)));
```
in FunctionalChaos::BuildDistribution().

Note that the hard-coded 0.9 value should be replaced by a user-defined value.

regislebrun:
Following the idea of Sofiane, maybe the test on the dependence should be done at the beginning of the procedure, and the list of possible marginal distribution should depend on the copula. If it is the independent copula, there is no penalty in using an arbitrary marginal distribution as at the end, the transformation is affine (the cost is only in the selection of the marginal distributions), and if the copula is not independent, then we use a reduced set. In fact, we should use only distributions that map nicely with the Normal distribution, or with cheap CDF/quantile functions..

sofianehaddad:
@regislebrun Agreed for testing the dependence. However using threshold is not accurate as measure might be sensitive to sample size.

We can easily implement a statistical test using Spearman's `rho`. Such test is already implemented in HpothesisTest (relying on `rot`). p-value for such is given thanks to the following function :

```
def compute_pvalue(x,i, j):
    from math import sqrt
    n = len(x)
    rho = x.computeSpearmanCorrelation()[i,j]
    sqrt_dof = sqrt(n - 2)
    if abs(rho) >= 1:
       statistic = rho / abs(rho) * ot.SpecFunc.MaxScalar
    else :
        statistic = rho * sqrt_dof / sqrt(1.0 - rho * rho)
    pvalue = 2 * ot.Student(n - 2).computeCDF(-abs(statistic))
    return pvalue
```

Hi a small bench : 

```
import openturns as ot

ot.RandomGenerator.SetSeed(1)
x = ot.Normal(2).getSample(10000)
pval = compute_pvalue(x, 0, 1)
pval2 = ot.HypothesisTest.FullSpearman(x[:,0], x[:,1])[0].getPValue()
print(pval - pval2)
```

which returns 
```
-3.1086244689504383e-15
```

regislebrun:
@sofianehaddad Thanks for the remark, I completely overlooked HypothesisTest. Two points should be noted:
  * The Student approximation corresponds to the R code only for large n, and needs n > 2 to work. For n=20 you get a difference of -0.002275134166117576
  * Starting from n=100, you get at least a 2 figures correct value using the normal approximation, and for n=10000 you have 4 correct digits
I had a look at the R documentation and got the source code of AS89, an idea could be to implement the test directly in OT, remove it from rot and use it as you propose.

sofianehaddad:
@regislebrun yes. I had the idea to implement it directly in OT
As I can remember, within the AS89, starting from n=2 to 1290 we get the exact evaluation 
n > 1290, we get the previous approximation
Have a look at https://stats.stackexchange.com/questions/235024/different-methods-for-finding-spearmans-coefficient-produce-diff-p-values-depen and `AS89`

regislebrun:
@sofianehaddad I am currently implementing DistFunc::pSpearmanCorrelation() for this purpose. The only difficulty is to get Edgeworth expansion with more accurate coefficients than the 4 figures given in the reference...

jschueller:
Indeed, filtering out dependencies has much more impact than marginal selection, eventhough I still think that part can be sped up (KS is still the worst thing ever, not shown here):

https://gist.github.com/jschueller/0a3aed7234a563baebd6dba457e401bc:

```
correlation= False marginal= Uniform speed= 251318.45744553363 evals/s t= 0.03979015350341797
correlation= False marginal= Normal speed= 270162.7686778185 evals/s t= 0.03701472282409668
correlation= False marginal= Gamma speed= 128237.5494155739 evals/s t= 0.07798027992248535
correlation= True marginal= Uniform speed= 949.8720630466422 evals/s t= 10.527733564376831
correlation= True marginal= Normal speed= 834.9119337252981 evals/s t= 11.977311134338379
correlation= True marginal= Gamma speed= 492.3374763764982 evals/s t= 20.31127119064331
```

jschueller:
I modified my PR to drop the marginal estimation change and finalize the independence test.

jschueller:
Done, rebased on top of #783

regislebrun:
+1

PR #785, opened 2018-05-31
Misc fixes

jschueller:
- A crash in Cobyla
- A crash in simulation doctest


lib/src/Base/Optim/Cobyla.cxx
```diff
@@ -149,10 +150,11 @@ void Cobyla::run()
     result_.store(inP, Point(1, outP[0]), absoluteError, relativeError, residualError, constraintError);
   }
 
-  result_.setOptimalPoint(x);
-  const UnsignedInteger index = evaluationInputHistory_.find(x);
-  Scalar bestValue = evaluationOutputHistory_(index, 0);
-  result_.setOptimalValue(Point(1, bestValue));
+  // x might not be exactly the best point evaluated, so find the nearest
+  const UnsignedInteger optimalIndex = NaiveNearestNeighbour(evaluationInputHistory_).query(x);
```
dbarbier: Why NaiveNearestNeighbour and not NearestNeighbourAlgorithm?

jschueller: Ok, I see it instanciates KDTree which is more efficient.

lib/src/Base/Optim/Cobyla.cxx
```diff
@@ -149,10 +150,11 @@ void Cobyla::run()
     result_.store(inP, Point(1, outP[0]), absoluteError, relativeError, residualError, constraintError);
   }
 
-  result_.setOptimalPoint(x);
-  const UnsignedInteger index = evaluationInputHistory_.find(x);
-  Scalar bestValue = evaluationOutputHistory_(index, 0);
-  result_.setOptimalValue(Point(1, bestValue));
+  // x might not be exactly the best point evaluated, so find the nearest
+  const UnsignedInteger optimalIndex = NearestNeighbourAlgorithm(evaluationInputHistory_).query(x);
```
dbarbier: But building any NearestNeighbourAlgorithm is costly, I am pretty sure that it is not worth when querying a single point.
You should only do that if `evaluationInputHistory_.find(x)` fails.

jschueller: Ok, done

regislebrun:
+1

lib/src/Base/Geom/Mesh.cxx
```diff
@@ -374,10 +374,10 @@ Bool Mesh::isRegular() const
   Bool regular = true;
   const Scalar epsilon = ResourceMap::GetAsScalar("Mesh-VertexEpsilon");
   const Scalar step = vertices_(simplices_(0, 1), 0) - vertices_(simplices_(0, 0), 0);
-  for (UnsignedInteger i = 1; i < size; ++i)
+  for (UnsignedInteger i = 1; (i < size) && regular; ++ i)
   {
-    regular = regular && (std::abs(vertices_(simplices_(i, 1), 0) - vertices_(simplices_(i, 0), 0) - step) < epsilon);
-    if (!regular) break;
+    const Scalar delta = vertices_(simplices_(i, 1), 0) - vertices_(simplices_(i, 0), 0);
+    regular = regular && ((std::abs(delta - step) / std::abs(step)) < epsilon);
```
dbarbier: I prefer
```
regular = (std::abs(delta - step) <= abs_step * epsilon);
```
where `abs_step` is defined outside of the loop.
Rationale:
* gracefully handle the case where step is zero
* there is no need to use boolean && operator since we exit as soon as regular is false

jschueller: Got it, thanks

dbarbier:
Okay, looks good to me.

regislebrun:
+1

jschueller:
I plan on backporting the cobyla & mesh fixes to the 1.11 branch.

PR #786, opened 2018-06-07
Misc fixes

jschueller:
cc @aurelieladier


jschueller:
@dbarbier I think I figured what happened to our Linux builds failing: https://github.com/swig/swig/issues/1273

dbarbier:
@jschueller We cannot rely on swig version number, IMO you should let CMake run swig and determine how many times backslashes must be escaped.

jschueller:
All versions (<=3) do not write raw docstrings, while swig master (4.0.0) handles docstrings as raw (at least in the current state of master since the identified PR). What do you say we cannot rely on the version number ?

jschueller:
@dbarbier I should add that there doesn't seem to be a 3.1 milestone in the swig tracker, so I think there should not be any 3.x version anymore, as they made significant changes to master (dropping support for old Python version) and there's no maintenance 3.x branch.

dbarbier:
BTW I do not understand your fix; if backslashes do no more have to be escaped, just copy doc file, for instance
```
      if (SWIG_VERSION VERSION_LESS 4)
          add_custom_target (generate_${DOCSTRING_FILE_BASENAME}
                          COMMAND ${CMAKE_COMMAND} -DDOCSTRING_IN_FILE=${DOCSTRING_IN_FILE} -DDOCSTRING_FILE=${CMAKE_CURRENT_BINARY_DIR}/${DOCSTRING_FILE_BASENAME}.i -P ${CMAKE_SOURCE_DIR}/cmake/escape_backslash_x4.cmake
                          DEPENDS ${DOCSTRING_IN_FILE})
          add_dependencies (generate_docstrings generate_${DOCSTRING_FILE_BASENAME})
      else ()
          execute_process (COMMAND ${CMAKE_COMMAND} -E copy ${DOCSTRING_IN_FILE} ${CMAKE_CURRENT_BINARY_DIR}/${DOCSTRING_FILE_BASENAME}.i)
      endif ()
```

dbarbier:
@jschueller If you define

```
set (SWIG_FORCE_ESCAPE_BACKSLASH FALSE CACHE BOOL "Force SWIG to escape backslashes.")
mark_as_advanced(SWIG_FORCE_ESCAPE_BACKSLASH)
```
and then use it
```
      if ((SWIG_VERSION VERSION_LESS 4) OR SWIG_FORCE_ESCAPE_BACKSLASH)
          add_custom_target (generate_${DOCSTRING_FILE_BASENAME}
                          COMMAND ${CMAKE_COMMAND} -DDOCSTRING_IN_FILE=${DOCSTRING_IN_FILE} -DDOCSTRING_FILE=${CMAKE_CURRENT_BINARY_DIR}/${DOCSTRING_FILE_BASENAME}.i -P ${CMAKE_SOURCE_DIR}/cmake/escape_backslash_x4.cmake
                          DEPENDS ${DOCSTRING_IN_FILE})
          add_dependencies (generate_docstrings generate_${DOCSTRING_FILE_BASENAME})
      else ()
          execute_process (COMMAND ${CMAKE_COMMAND} -E copy ${DOCSTRING_IN_FILE} ${CMAKE_CURRENT_BINARY_DIR}/${DOCSTRING_FILE_BASENAME}.i)
      endif ()
```
we will have a workaround in case SWIG developers change their mind.

dbarbier:
> Note that we still have to escape the backslashes once

I don't think so, documentation looks fine with a copy as written above.

jschueller:
It doesnt build with a simple copy.

dbarbier:
Indeed, I do not understand what I tested yesterday, sorry for the noise.
BTW I prefer my solution with a cached variable, so that this is hidden for normal users.

jschueller:
Ok, marking it advanced.

jschueller:
@dbarbier I've got a reply from the swig maintainer in swig/swig#1273: checking the version is the good way to go

regislebrun:
+1

PR #790, opened 2018-06-19
Allow to sum/subtract n-d distributions

jschueller:


lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -148,11 +148,23 @@ Distribution DistributionImplementation::operator + (const Distribution & other)
...
 Distribution DistributionImplementation::operator + (const DistributionImplementation & other) const
 {
-  if ((dimension_ != 1) || (other.dimension_ != 1)) throw NotYetImplementedException(HERE) << "In DistributionImplementation::operator + (const DistributionImplementation & other) const: can add 1D distributions only.";
-  Collection< Distribution > coll(2);
-  coll[0] = *this;
-  coll[1] = other.clone();
-  return new RandomMixture(coll);
+  if (dimension_ != other.dimension_) throw InvalidDimensionException(HERE) << "Can only sum distributions with the same dimension:" << dimension_ <<  " vs " << other.dimension_;
+
+  if ((dimension_ > 1) && (!hasIndependentCopula() || !other.hasIndependentCopula())) throw NotYetImplementedException(HERE) << "Can only sum distributions with independent copulas";
```
regislebrun: If dimension_==1 then hasIndependentCopula() returns true (or it is a bug)

jschueller: Yes, then hasIndependentCopula is not even called and it falls through, what is your point ?

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -148,11 +148,23 @@ Distribution DistributionImplementation::operator + (const Distribution & other)
...
+  Collection< Distribution > marginals(dimension_);
+  for (UnsignedInteger j = 0; j < dimension_; ++ j)
+  {
+    Collection< Distribution > coll(2);
+    coll[0] = getMarginal(j);
+    coll[1] = other.getMarginal(j);
+    marginals[j] = new RandomMixture(coll);
+  }
+
+  if (dimension_ > 1)
```
regislebrun: If you separate the case dimension_==1 from dimension_>1, then you should test it at line 152 and keep the old code for this (very usual) case, with no call to getMarginal().

lib/src/Uncertainty/Model/DistributionImplementation.cxx
regislebrun: My point is that in the univariate case, we can always sum distributions, and in the multivariate case, we have to check the copulas. I would prefer to have first a test to deal with the very common univariate case then to deal with the multivariate case, where the test would be reduced to the independent copulas on both side.

regislebrun:
+1

PR #792, opened 2018-06-19
Change evaluation order in NonCenteredFiniteDifferenceGradient::gradient

dbarbier:
We currently evaluate functions on shifted points, and last on center point.
This commit switches evaluation order to first evaluate on center point.

The reason is that the most common case is to perform evaluation then
gradient at the same location, so new evaluation order is better in order
to benefit from an LRU cache, or from a cache with a very small size.


regislebrun:
There is the same problem in CenteredFiniteDifferenceHessian, it would be a good idea to fix it too.
Otherwise, +1

PR #793, opened 2018-06-22
Add SobolSimulationAlgorithm

jschueller:
The algorithm allows to estimate Sobol indices by simulation.
It samples the distribution of a random vector X through a model f.
At each iteration a fixed number N of replications inputs is generated (N = replicationSize_*(d+2))
These inputs are evaluated by blocks of size blockSize_ through the
model f.
Then we compute the distribution of the indices (first and total order) on this current replication sample.
We use one of the 4 available estimators (Saltelli, Jansen, ...)
At the end of each iteration we update the global distribution of the indices.
Did I caught correctly the principle of the algorithm?

The most sketchy part for me at the moment is the stopping criteria.
We proposed different criteria according if the estimated Sobol index is
small, controlled or large.
For each input component i, we check:

- If S_i^T is small, then stop if Phi_i^T(alpha) <= espilon
- If S_i_F is large, then stop if Phi_i^F(1-alpha) <= epsilon
- If S_i^{T,F} is controlled (nor small nor large), then stop if
Phi^{T,F}(1-alpha/2) - Phi^{T,F}(alpha/2) <= epsilon

I'm not sure if I get that correctly, nor the default values to choose
for alpha (0.95?), epsilon (1e-3?) the threshold to decide if an index
is small/large (1e-2?)

Also, we have already SobolIndicesAlgorithm[Implementation] classes that
contain only the estimators, maybe I should rename these?

TODO: doc


regislebrun:
+1

jschueller:
Michael:
- gsobol/ishigami
- validation scripts

PR #795, opened 2018-07-01
Fix ticket 953

regislebrun:
See [here](http://trac.openturns.org/ticket/953) for a full analysis of the bug. This PR provides several contributions:
  * RandomMixture::computeScalarQuantile() now relies on computeCDF() instead of computeProbability()  when there are only two continuous 1D atoms, to avoid the use of Poisson's summation formula when a direct convolution is more efficient
  * RandomMixture::computeProbability() now relies on computeCDF() instead of computeProbability()  when there are only two continuous 1D atoms for the same reason, but involves two calls to computeCDF() so it is still better to call computeCDF() in computeScalarQuantile() in this case
  * RandomMixture::computeProbability() had a bug in the analytical case, relying on the distribution weight instead of its coefficient in the random mixture. This bug was visible only for coefficients not equal to 1.
  * TruncatedDistribution was suboptimal in its computation of the normalization factor, which was done in computeRange() independently from the call to computeProbability() in setBounds() or computeCDF() in the legacy constructors.
  * Now TruncatedDistribution implements a simplification mechanism through the getSimplifiedVersion() method. This mechanism allows to simplify only some 1D truncated distributions (Uniform, Normal, Exponential, nested TruncatedDistribution, or range fully included into truncation bounds). More simplifications will be added with time.


lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -175,19 +157,82 @@ TruncatedDistribution * TruncatedDistribution::clone() const
...
+      localDistribution.setWeight(w);
+      return localDistribution;
+    }
+  // Intersect the underlying distribution range and the truncation range
+  const Scalar alpha = std::max(lower, a);
+  const Scalar beta = std::min(upper, b);
+  // Actual simplifications
+  if (kind == "Uniform")
+    {
+      const Uniform * uniform(static_cast< const Uniform * >(localDistribution.getImplementation().get()));
```
jschueller: uniform is not used

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -175,19 +157,82 @@ TruncatedDistribution * TruncatedDistribution::clone() const
...
+      Uniform simplified(alpha, beta);
+      simplified.setWeight(getWeight());
+      return simplified;
+    }
+  if (kind == "Normal")
+    {
+      const Normal * normal(static_cast< const Normal * >(localDistribution.getImplementation().get()));
```
jschueller: I'd rather use dynamic_cast instead of static_cast to enable the runtime check, just in case

jschueller:
Seems it segfaults since https://github.com/swig/swig/commit/4b4e0180f42e992de6ff458477b70e4ec63ecb49 (https://github.com/swig/swig/pull/1279)

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
regislebrun: I hope there will be no "case": the introspection mechanism is at the heart of OT!

lib/src/Uncertainty/Distribution/RandomMixture.cxx
```diff
@@ -453,6 +454,12 @@ void RandomMixture::setDistributionCollectionAndWeights(const DistributionCollec
       // Aggregate the atoms
       atomCandidates.add(mixture->getDistributionCollection());
     } // atom is a RandomMixture
+    else if (atomKind == "TruncatedDistribution")
+    {
+      const TruncatedDistribution * truncatedDistribution(static_cast< const TruncatedDistribution * >(atom.getImplementation().get()));
```
jschueller: here too

regislebrun: I checked the sources, and we have static_cast conversions of pointers of complex types in several othr places (HMatrixImplementation, NLopt, MemoizeFunction) in addition to the numerous static_cast of simple types. Do you think we should change all of them? Is the penalty involved in using dynamic_cast instead of static_cast that important? My main point is that I hope we can rely on the information given by getClassName() to perform a cast, it is exactly why it has been introduced in the first place. Not using it is like wearing a condom and have no sex: something which does not carry a good confidence of developers into their own developement...

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -63,22 +63,14 @@ TruncatedDistribution::TruncatedDistribution(const Distribution & distribution,
     const Scalar upperBound,
     const Scalar thresholdRealization)
   : DistributionImplementation()
-  , bounds_(Point(1, lowerBound), Point(1, upperBound))
+  , bounds_(Point(distribution.getDimension(), lowerBound), Point(distribution.getDimension(), upperBound))
```
jschueller: Dont you think it may confuse the user if he does not have to provide explicitely nd bounds ?
Maybe we should only keep only the interval ctor in the long term ?

regislebrun: In fact it is useful for many applications where the margins have been standardized, such as in multivariate extreme values analysis where you put standard Frechet margins to your model and study the convergence of symmetrically truncated distributions. For 1D cases, it is also very convenient to specify the bounds as two scalars.

lib/src/Uncertainty/Distribution/RandomMixture.cxx
jschueller: Imho it's just a good practice, nothing special about your code. You dont have to replace it everywhere.

PR #796, opened 2018-07-02
Various Process enhancements

jschueller:
- Add GaussianProcess::getMarginal 
- Add ProcessSample::drawMarginal interpolate option 
- Add Field._repr_html_ 
-  Fix http://trac.openturns.org/ticket/954


regislebrun:
Not 100% sure that CovarianceModel::getMarginal() is correct (what about the management of active parameters of a  TensorizedCovarianceModel?), otherwise +1 (and thanks for the last commit)

jschueller:
Yes, it should be reimplemented for TensorizedCovarianceModel, I can mark it as notyetimplemented.

Also I noticed TensorizedCovarianceModel(CovarianceModelCollection) initializes with a scale of 1.0, why not take the scale of each cov model in the list like it's done for the amplitude ?

regislebrun:
It is because TensorizedCovarianceModel is here to stack p covariance 1D models sharing the same input dimension to form a p dimensional covariance model. So it is not possible to pick one scale here and another here. Instead of taking the first model scale as the tensorized scale, a default value is used. This behavior can be change if you want.

jschueller:
no, it's ok

lib/src/Uncertainty/Process/CompositeProcess.cxx
```diff
@@ -129,7 +129,16 @@ TimeSeries CompositeProcess::getFuture(const UnsignedInteger stepNumber) const
   /* TimeGrid associated with the possible future */
   const Scalar timeStep = timeGrid.getStep();
   const RegularGrid futurTimeGrid(timeGrid.getEnd(), timeStep, stepNumber);
-  return TimeSeries(futurTimeGrid, function_(antecedent_.getFuture(stepNumber).getValues()));
+  FieldFunction function(function_);
+  try
+  {
+    function.setInputMesh(futurTimeGrid);
```
dbarbier: Even if this is useless here, I suggest to also call setOutputMesh.

jschueller: Did you remove this comment ?

dbarbier: No

jschueller: ok, it's just that it appeared grayed out

jschueller: ok, done

dbarbier:
Thanks, but I am still uncomfortable with setInputMesh, sorry about that.

In fact, ValueFunction and VertexValueFunction classes are designed to have the same mesh as input and output meshes; here we add FieldFunctionImplementation::set{In,Out}putMesh methods, and they are defined only within these two classes.  Maybe we should add a single method setInputOutputMesh instead?

jschueller:
I've no strong opinion about one or the other. I'd say it depends on whether you prefer to reflect the actual storage of the meshes or the fact that they share input/output meshes.

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -791,9 +791,63 @@ Bool CovarianceModelImplementation::isDiagonal() const
...
+  if (outputDimension_ > 1)
+  {
+    const UnsignedInteger oldOutputDimension = outputDimension_;
+    result.getImplementation()->outputDimension_ = indices.getSize();
+    Point newAmplitude;
+    for (UnsignedInteger i = 0; i < indices.getSize(); ++ i)
+    {
+      newAmplitude.add(amplitude_[indices[i]]);
+    }
+    result.getImplementation()->amplitude_ = newAmplitude;
```
sofianehaddad: I think we should use setter. Some models have to calibrate their internal parameters

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -791,9 +791,63 @@ Bool CovarianceModelImplementation::isDiagonal() const
...
+      {
+        // amplitude (assumes ascending order)
+        newActiveParameters.add(inputDimension_ + counter);
+        ++ counter;
+      }
+      else
+      {
+        // others
+        newActiveParameters.add(activeParameters[i] - oldOutputDimension - indices.getSize());
+      }
```
sofianehaddad: No. others might contain output correlation coefficients. In that case, the code here is wrong.

jschueller: Yes, I dont think we can resolve this correctly here, I'll rather just resize the activeParameters.

lib/src/Base/Stat/CovarianceModelImplementation.cxx
jschueller: ok

lib/src/Base/Stat/CovarianceModelImplementation.cxx
jschueller: ok

regislebrun:
IMO we should add a property to FieldFunction to check if it acts pointwise (isPointWise() true for ValueFunction and VertexValueFunction, false otherwise), restrict CompositeProcess::getFuture() to the case where isPointWise() is true and call only setInputMesh().

jschueller:
@regislebrun , @dbarbier I added the FieldFunction::isPointWise property

lib/src/Base/Func/FieldFunctionImplementation.cxx
```diff
@@ -138,14 +141,21 @@ Mesh FieldFunctionImplementation::getInputMesh() const
...
+  outputMesh_ = outputMesh;
 }
 
 Mesh FieldFunctionImplementation::getOutputMesh() const
 {
   return outputMesh_;
 }
 
+Bool FieldFunctionImplementation::isPointWise() const
+{
```
sofianehaddad: why "w" is in capital letters ? isPointwise or isActingPointwise?

jschueller: I agree, isActingPointwise is better

PR #797, opened 2018-07-03
[RFC] Random svd & Robbins-Monro quantiles

regislebrun:
This PR proposes two main enhancements:
  * Two random algebra algorithms to compute SVD decomposition of large ProcessSample wich can be stored in main memory. See [here](https://code.google.com/archive/p/redsvd/wikis/English.wiki) for the redsvd version (based on a random sampling of the columns and the rows of the matrix to decompose), and [here](https://arxiv.org/pdf/1612.08709.pdf) for the tygert version (based on a single random sampling with tiling). The speed-up is impressive while retaining most of the accuracy. On a single benchmark over a 41416x66728 matrix, the LAPACK version takes more than 5h when the redsvd version takes 27s!
  * A stochastic algorithm to estimate fields of pointwise quantiles, based on Robins-Monro algorithm. It can be used as a replacement of ProcessSample::computeQuantilePerComponent() for large samples with essentially the same precision and a much favorable memory access pattern and a linear complexity. On the same example, it allows to compute quantiles in 5s when OT 1.11 takes 1100s and master 270s (partially due to swap).

For now, these algorithms are proposed either as an option for KarhunenLoeveSVDAlgorithm (random svd) or as a new method of ProcessSample. This choice is open for discussion.
**The documentation and the tests are still missing but will be influenced by the discussion**


dbarbier:
Do you have validation cases?  I have not yet looked at RedSVD, but Tygert and Robbins-Monro look bogus to me.

lib/src/Base/Stat/ProcessSampleImplementation.cxx
```diff
@@ -350,6 +350,34 @@ ProcessSampleImplementation ProcessSampleImplementation::computeQuantilePerCompo
...
+  Point quantiles(data_[0].getImplementation()->getData());
+  Point scale(data_[1].getImplementation()->getData() - quantiles);
+  for (UnsignedInteger j = 0; j < scale.getSize(); ++j)
+    scale[j] = std::abs(scale[j]);
+  for (UnsignedInteger i = 1; i < size; ++i)
+    {
+      const Point & Xi = data_[i].getImplementation()->getData();
+      const Scalar tau = 1.0 / std::pow(i, exponent);
+      for (UnsignedInteger j = 0; j < scale.getSize(); ++j)
+	quantiles[j] -= scale[j] * tau * (int(Xi[j] <= quantiles[j]) - prob);
```
dbarbier: Why did you introduce scale?  It is not mentioned in the paper AFAICT, and results look better without it.

regislebrun:
@dbarbier The documentation and the tests are missing because I am not 100% sure of the architecture of the code. Up to now, we provide most of the "light" computations as methods of objects which are primarily data containers, but another choice could be to have pure data containers (eg ProcessSample) and atomic algorithms for all the computations (mean, quantile etc). The documentation and the tests depend strongly on these choices.
For now, I use the following script to check the random svd algorithms:
```
import openturns as ot
from time import time

# k=nb of svd k<n, k<m
k = 100
m = 1000
n = 500
ot.Log.Show(ot.Log.INFO)
t0 = time()
print("generate data")
p = ot.GaussianProcess(ot.SquaredExponential([0.1]), ot.RegularGrid(0.0, 1.0 / (m - 1.0), m))
s = p.getSample(n)
print("t=", time() - t0, "s")

algo = ot.KarhunenLoeveSVDAlgorithm(s, 1e-6)
print("Compute SVD, LAPACK")
t0 = time()
algo.run()
resultLAPACK = algo.getResult()
print(resultLAPACK.getEigenValues()[0:20])
print("t=", time() - t0, "s")

ot.ResourceMap.SetAsBool("KarhunenLoeveSVDAlgorithm-UseRandomSVD", True)
ot.ResourceMap.SetAsUnsignedInteger( "KarhunenLoeveSVDAlgorithm-RandomSVDMaximumRank", 20)
ot.RandomGenerator.SetSeed(1)
print("Compute SVD, TYGERT")
ot.ResourceMap.Set("KarhunenLoeveSVDAlgorithm-RandomSVDVariant", "tygert")
ot.ResourceMap.SetAsUnsignedInteger( "KarhunenLoeveSVDAlgorithm-TygertMargin", 2)
ot.ResourceMap.SetAsUnsignedInteger( "KarhunenLoeveSVDAlgorithm-TygertIterations", 10)
t0 = time()
algo.run()
resultTYGERT = algo.getResult()
print(resultTYGERT.getEigenValues()[0:20])
print("t=", time() - t0, "s")
```
I use it to check the performance and the accuracy of these new algorithms, wit redsvd as a clear winner...

regislebrun:
It is difficult to compare redsvd and tygert in terms of math: redsvd is tygert with TygertMargin=0 and TygertIterations=0 but with an additional random sampling of the rows. Something between TygertIterations=0 and TygertIterations=1...

lib/src/Base/Stat/ProcessSampleImplementation.cxx
regislebrun: The theoretical convergence is conditioned to the convergence of a series related to the decreasing deterministic sequence of step sizes. It is unchanged if a scale is introduced (ie we have an almost sure convergence if scale!=1 iff there is an almost sure convergence if scale==1), but in practice it can change things a lot. If you take scale=1, the update on the quantile estimator is either step_i or step_i * (1-q) depending on X_i being greater or non greater than the current quantile estimate, with step_i = 1/i^alpha, 0.5<alpha<=1. Suppose that your distribution is widely spread or significantly non centered. The quantile will be large and it will be obtained by an accumulation of tiny steps if scale=1. With scale=X_1-X_0, there is a hope to capture the actual spread of the distribution and to accelerate the estimator.
It is more a heuristic than a mathematical result, but it sounds reasonable to me. You (and anybody else) are welcome to confirm/disproof this idea ;-)
Here is a script to check that things are going the right way:
```
import openturns as ot
from time import time

m = 1000
n = 50000
ot.Log.Show(ot.Log.INFO)
t0 = time()
print("generate data")
scale = 0.1
dt = 1.0 / (m - 1.0)
mesh = ot.RegularGrid(0.0, dt, m)
amplitude = 1.0
mean = 10.0
trend = ot.TrendTransform(ot.SymbolicFunction('t', str(mean)), mesh)
p = ot.GaussianProcess(trend, ot.SquaredExponential([scale], [amplitude]), mesh)
s = p.getSample(n)
print("t=", time() - t0, "s")

# Quantiles
q = 0.5
theoretical = ot.Normal(mean, amplitude).computeQuantile(q)[0]
t0 = time()
print("empirical quantile")
empirical = s.computeQuantilePerComponent(q)
print("t=", time() - t0, "s")
t0 = time()
print("Robbins Monro quantile")
robbins_monro = s.computeQuantilePerComponentRobbinsMonro(q)
print("t=", time() - t0, "s")
# Check the distribution of empirical quantiles and robbins_monro quantiles
indep_bw = ot.KernelSmoothing().computeMixedBandwidth(empirical.getValues())
# Adjust bandwidth to take into account dependence
bw = indep_bw * (scale / dt)**(1.0/5.0)
graph = ot.KernelSmoothing().build(empirical.getValues(), bw).drawPDF()
graph.add(ot.KernelSmoothing().build(robbins_monro.getValues(), bw).drawPDF())
yMax = graph.getBoundingBox().getUpperBound()[1]
graph.add(ot.Curve([[theoretical, 0.0], [theoretical, yMax]]))
graph.setColors(["blue", "green", "red"])
graph.setLegends(["empirical", "Rob.Mon.", "theor."])
ot.Show(graph)
```

dbarbier: With scale:
robbins-scale

Without scale:
robbins-woscale

dbarbier: Please check also with q=0.05 or q=0.95.

dbarbier: I added
```
diff = empirical.getValues() - robbins_monro.getValues()
print(diff.getMin(), diff.getMax(), diff.computeMean(), diff.computeStandardDeviation())
```
at the end of your script.
With scale:
```
[-0.852258] [1.09544] [0.0132888] [[ 0.128232 ]]
```
Without scale:
```
[-0.011142] [0.01822] [0.00314498] [[ 0.00544802 ]]
```

regislebrun:
I implemented the nested procedure. It looks quite simple to implement but gives deceiving results. I found some typos in the reference https://arxiv.org/pdf/1510.00967.pdf, namely the fact that k runs up to K instead of K-1 in equations (11) and (15), which is a minor typo, and that equation (15) is written with a_1 instead of a_k, which is more problematic. I would appreciate the help of anybody wanting to read the code and check the algorithm!

regislebrun:
I split this PR into 2 parts: one for the random SVD, one for the stochastic algorithm.

sofianehaddad:
See #832 and #833

PR #798, opened 2018-07-06
Improved Field::asDeformedMesh()

regislebrun:
Now, it is possible to build a Mesh of dimension n from a field of input dimension m<=n and output dimension p<=n by padding the vertices and the values with zeros at given locations then moving the vertices by the values.
To do this, I added the Indices::complement() method. This method allows to produce an Indices given a nonnegative integer n, which is the increasing sequence of elements of integers in {0,...,n-1} not in the base Indices.


dbarbier:
Field::asDeformedMesh should not have been introduced, this method does not make any sense.
One can obtain the same result with a single line of code:
```
deformedMesh = ot.Mesh(field.getMesh().getVertices() + field.getValues(), field.getMesh().getSimplices())
```
If you want padding:
```
vertices = field.getMesh().getVertices()
values = field.getValues()
vertices.stack(ot.Sample(vertices.getSize(), p - vertices.getDimension()))
values.stack(ot.Sample(values.getSize(), p - values.getDimension()))
deformedMesh = ot.Mesh(vertices + values, field.getMesh().getSimplices())
```

regislebrun:
And the simplices are wrong in your 5 lines version with padding. You have to extend the dimension of the simplices if you extended the dimension of the vertices.

dbarbier:
> Does it makes more sense now?

To me, no.

> You have to extend the dimension of the simplices if you extended the dimension of the vertices.

Why?  Export already takes care of duplicating vertices, and I do not see which other method you can do with such meshes.

regislebrun:
You are wrong. Try this:
```
import openturns as ot

mesh0 = ot.IntervalMesher([3]*2).build(ot.Interval([-1.0]*2, [1.0]*2))
f = ot.SymbolicFunction(["x", "y"], ["x * y"])
values0 = f(mesh0.getVertices())
field = ot.Field(mesh0, values0)
field.asDeformedMesh([2], [0,1]).exportToVTKFile("test_new.vtk")

p = 3
vertices = field.getMesh().getVertices()
size = field.getValues().getSize()

vertices.stack(ot.Sample(vertices.getSize(), p - field.getInputDimension()))
values = ot.Sample(size, p - field.getOutputDimension())
values.stack(field.getValues())
deformedMesh = ot.Mesh(vertices + values, field.getMesh().getSimplices())
deformedMesh.exportToVTKFile("test.vtk")
```
If you compare test.vtk and test_new.vtk, you see that the simplices are different, and it **does** matter if you open both files in paraview (test.vtk shows nothing, test_new.vtk shows what is expected). Note that I had to change your script a little bit as you were padding with zeros the values at the wrong locations.

This PR is clearly **NOT** a scientific breakthrough, but provides a very convenient way to visualize 3D  data of lower intrinsic dimension.

dbarbier:
Indeed, I had it backwards, export removes duplicated vertices and does not add them; as already said in #657, IMO using duplicate vertices is a very bad idea.   You could instead fix export to deal with simplices of lower dimensions.
```patch
--- a/lib/src/Base/Geom/Mesh.cxx
+++ b/lib/src/Base/Geom/Mesh.cxx
@@ -933,9 +933,10 @@ String Mesh::streamToVTKFormat(const IndicesCollection & simplices) const
   }
   // There is at least one simplex. Assume homogeneous simplices,
   // ie all the simplices are of the same kind as the first one
+  UnsignedInteger simplexDimension = simplices.cend_at(0) - simplices.cbegin_at(0);
   UnsignedInteger verticesPerSimplex = 1;
   UnsignedInteger lastIndex = simplices(0, 0);
-  while ((verticesPerSimplex <= dimension_) && (simplices(0, verticesPerSimplex) != lastIndex))
+  while ((verticesPerSimplex < simplexDimension) && (simplices(0, verticesPerSimplex) != lastIndex))
   {
     lastIndex = simplices(0, verticesPerSimplex);
     ++verticesPerSimplex;

```

I still stand by my opinion, you are adding a method which seems to be generic but will only work in a few specific cases (or maybe just this one?), and can be emulated in 4 lines of code.

jschueller:
The tests fail

regislebrun:
IMO the actual question is: do we truly support low dimensional meshes embedded into higher dimensional spaces? I plan to add an intrinsic dimension to the Mesh class, for which eg getVolume() or computeWeight() would be based on, but probably not for the next release.
Concerning the interest of the method, everyone can have its own opinion. I disagree with the fact that a 4 lines python code, would these lines being correct, is acceptable by many of the OT users, not talking about the C++ users who would probably implement them as a generic service in their code.
Concerning the usage, in addition to the (2,1)->3 case (mesh of dimension 2, values of dimension 1) which already cover 3 sub-cases, you have the (1,2)->3 case, the (1,1)->2 or 3 cases, plus the overlapping cases where the addition is at least a true translation of at least one component. This method has been introduced for the LOCOMACHS project for a workpackage dedicated to random geometry generation where the random geometry is a deformed mesh. It was mainly used for the (3,3)->3 case corresponding to the one-liner, but now there are applications for plates (2,1)->3 or (2,3)->3 cases, for shells (3,1)->3 or even for beam (1,3)->3. The notation (m,v)->d indicates the dimension m of the mesh the dimension v of the  values and d the dimension of the deformed mesh.

regislebrun:
Any further comment? I am waiting for this development for a study on an EDF use case... Either the code is wrong (and please show me where), or it is correct and let's merge it! It has been proved to be useful, at least for myself and for the EDF guys.

PR #801, opened 2018-07-09
Optimization fix + use swig tag

jschueller:


lib/src/Base/Optim/Cobyla.cxx
```diff
@@ -123,7 +123,7 @@ void Cobyla::run()
    */
   int returnCode = ot_cobyla(n, m, &x[0], rhoBeg_, rhoEnd, message, &maxFun, Cobyla::ComputeObjectiveAndConstraint, (void*) this);
 
-  result_ = OptimizationResult(dimension, 2);
+  result_ = OptimizationResult(dimension, 1);
```
dbarbier: Looks wrong to me, see l.101
```
  // initialize history
  evaluationInputHistory_ = Sample(0, dimension);
  evaluationOutputHistory_ = Sample(0, 2);
```

dbarbier: Okay, got it now

jschueller: in evaluationOutputHistory_ we store the value + constraint so its of dim 2 yes

my modification changes the dimension of the outputHistory_ of OptimResult:
```result_.store(inP, Point(1, outP[0])```

jschueller:
@regislebrun could you review this one ? it moves away from swig master and would fix our circleci builds.

regislebrun:
+1

PR #804, opened 2018-07-09
Fix compilation errors when compiling with -Wall -Werror

dbarbier:
`SpecFunc::IncompleteBetaInverse`, `DistFunc::[pq]DickeyFuller*` and `RandomMixture::computeQuantile` methods have an unused `tail` argument which should be fixed too.


jschueller:
I've also had the new gcc 8 warning Wclass-memaccess (enabled by Wall), any how idea how to fix this one ?
```
In file included from /tmp/openturns/lib/src/Base/Stat/RandomGenerator.cxx:22,
                 from /tmp/openturns/lib/src/cotire/OT_orig_CXX_unity_270_359.cxx:47:
/tmp/openturns/lib/src/Base/Stat/dsfmt.h: In instantiation of ‘void tutils::detail::dsfmt<MEXP>::set_state(OT::UnsignedInteger*) [with int MEXP = 19937; OT::UnsignedInteger = long unsigned int]’:
/tmp/openturns/lib/src/Base/Stat/RandomGenerator.cxx:71:37:   required from here
/tmp/openturns/lib/src/Base/Stat/dsfmt.h:113:21: warning: ‘void* memcpy(void*, const void*, size_t)’ copying an object of non-trivial type ‘class tutils::uint64v2_t’ from an array of ‘uint32_t’ {aka ‘unsigned int’} [-Wclass-memaccess]
               memcpy(&u_[i], &words[0], 16);
               ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
In file included from /tmp/openturns/lib/src/Base/Stat/dsfmt.h:29,
                 from /tmp/openturns/lib/src/Base/Stat/RandomGenerator.cxx:22,
                 from /tmp/openturns/lib/src/cotire/OT_orig_CXX_unity_270_359.cxx:47:
/tmp/openturns/lib/src/Base/Stat/simd.h:40:9: note: ‘class tutils::uint64v2_t’ declared here
   class uint64v2_t
```

dbarbier:
@jschueller Ok, this is fixed now.

I had to come back at these changes, everything should be fine now.

python/src/ResourceMap.i
```diff
@@ -16,7 +16,7 @@ namespace OT {
 
 %extend ResourceMap {
 
-UnsignedInteger __len__() const
+static UnsignedInteger __len__() const
```
jschueller: ResourceMap.i:22: Error: Static function __len__() const cannot have a qualifier.

python/src/IdentityMatrix.i
```diff
@@ -13,7 +13,7 @@ namespace OT {
 %extend IdentityMatrix {
   IdentityMatrix(const IdentityMatrix & other) { return new OT::IdentityMatrix(other); }
 
-PyObject * __setitem__(PyObject * args, Scalar val)
+static PyObject * __setitem__(PyObject * , Scalar )
```
jschueller: Why static ? Is there a warning after swig translates this ?

dbarbier: Yes

python/src/ResourceMap.i
jschueller: Maybe it only fails with swig master here?

dbarbier: Indeed it does not make sense to have a const qualifier for static methods, I'll drop it.

regislebrun:
+1

PR #805, opened 2018-07-12
Added ProductDistribution::getStandardDeviation()

regislebrun:
This fixes ticket #957. This PR also provides a specific implementation of getSkewness(), getKurtosis() and getMomemt() to ProductDistribution. The performance of these methods is greatly improved (30s becomes 5e-5s)


jschueller:
I would add the example from the ticket to the tests.

sofianehaddad:
@jschueller the problem with example is CPU time. It is dificult to illustrate it. Any idea?

@regislebrun you mentionned that we are using ProductDistribution::getCenteredMoment(2) to evaluate default standard deviation, whic is true. Btw I think it is a bad idea : you implemented a standard *computeCovarianceContinuous* and it seems usefull relying on this method (or getCovariance). 

The reason is that current DistributionImplementation::getStandardDeviation does not store the value. You can try this : 

```import openturns as ot
import time
tic = time.time()
p = ot.Uniform() * ot.Uniform() * ot.Uniform()
toc = time.time()
print("Time to initialize dist : ", toc - tic)
tic = time.time()
std = p.getStandardDeviation()
toc = time.time()
print("Time evaluating std : ", toc - tic)
tic = time.time()
std = p.getStandardDeviation()
toc = time.time()
print("Time evaluating std (2nd time) : ", toc - tic)
```

I get around 30 s for each call....

sofianehaddad:
@regislebrun  Have a look at https://github.com/sofianehaddad/openturns/commit/7b498d959ae93b3920d99beeeb45e23756506aa2

regislebrun:
@sofianehaddad Thanks for having a look at this problem. Concerning the test, at least we will see a timeout if a future change break the evaluation path, so I added it to t_RandomMixture_std as it involves a sum of distributions.
Concerning your benchmark, I think there is a problem either with your hardware or your configuration, on my laptop your example run in 5e-5s! It is expected as the standard deviation of both RandomMixture and ProductDistribution are computed through a call to getCovariance (so the computation is cached, reread the code), the covariance of a RandomMixture is a simple algebra based on the variance of its atoms, and for a ProductDistribution atom the covariance is also a simple algebra on mean and variance of the factors of the product.
To summarize:
  * there is no need to call computeCovarianceContinuous() in ProductDistribution
  * the computation of the covariance is cached in both RandomMixture and ProductDistribution and is used to compute the standard deviation
I had a look at your commit, which is a more general way to solve the issue than my solution. I don't know/remember how to integrate your commit in this PR. Maybe it is better to restart from your work?

sofianehaddad:
@regislebrun ok. In my UC, there is no RandomMixture but only ProductDistribution !
btw, you can cherry-pick my commit or restart it (as it is a small fix)

regislebrun:
@sofianehaddad I cherry picked your commit, then added some other developments and squashed everything at the end. Concerning the UC, I preferred to use the exact same example as the ticket, but I will use yours as it is more focused on the actual problem.

regislebrun:
Do we wait for Denis to be back from holidays? Michael is expecting these developments for his work on PDMP...

jschueller:
Sofiane had a close look, ok for me

PR #807, opened 2018-07-16
Explicitely use CMAKE_CROSSCOMPILING_EMULATOR

jschueller:
So we do not rely on Linux' binfmt_misc feature which in turn allows to
get rid of wine in the host enviroment when cross-compiling in a docker
container.


jschueller:
Related: https://gitlab.kitware.com/cmake/cmake/merge_requests/2217

jschueller:
found another way: use a shell script

PR #808, opened 2018-07-16
Explicitely use CMAKE_CROSSCOMPILING_EMULATOR

jschueller:
So we do not rely on Linux' binfmt_misc feature which in turn allows to
get rid of wine in the host environment when cross-compiling in a docker
container.


regislebrun:
+1 I am not really able to comment on this PR...

PR #809, opened 2018-07-17
[CircleCI] Go back to using swig master

jschueller:


jschueller:
Fixed in swig master: https://github.com/swig/swig/pull/1279

PR #810, opened 2018-07-24
Define Drawable line width as scalar

jschueller:


regislebrun:
Did you check if R is happy with a non integer line width?

jschueller:
Only tested with matplotlib, I just had a look at R, it seems so:
http://www.melissaclarkson.com/resources/R_guides/documents/line_styles_Ver2.pdf

PR #811, opened 2018-07-24
Chaospy distribution

regislebrun:
This PR allows to wrap chaospy univariate distributions (which is a lot of distributions) into OT distributions. The multivariate distributions are not supported as they implement CDF computation only for independent components, and do it with error (try a multivariate Normal distribution with a diagonal covariance matrix).


jschueller:
you can add chaospy to circle.yml so it's ci-tested: `pip3 install matplotlib ...`

regislebrun:
@jschueller I found several references to scipy in the sources, do I have to add chaospy in the same files too? I don't know what to do in in .travis.yml, install-openturns-dep.bat, debian/changelog and debian/control.

jschueller:
No need: .travis uses debian packages and chaospy is not in debian; same reason for distro/debian; install-openturns-dep.bat is no longer used

jschueller:
you broke something ?

regislebrun:
Sofiane told me that the parallel between scipy and chaospy distributions was not exact: the computeQuantile() method was missing (the one with the *tail* argument). Even if it is deduced from computeScalarQuantile() it looked natural to add it the way you did for scipy distributions. And in the way I broke something, because what the inv() method of chaopy returns can be silently converted into a float, but cannot be used to build a list of float.

sofianehaddad:
@regislebrun my msg was about `computeScalarQuantile`
Btw both `computeQuantile` and `computeScalarQuantile` require the `tail`argument (see https://github.com/openturns/openturns/blob/master/python/src/PythonDistribution.cxx#L306 and https://github.com/openturns/openturns/blob/master/python/src/PythonDistribution.cxx#L411)

regislebrun:
@sofianehaddad It appears that computeScalarQuantile() has a flag in C++ and not as a Python method! It is quite strange, so I modified the Python version to match the C++ version.

PR #812, opened 2018-07-27
Added DistFunc::pSpearmanCorrelation

regislebrun:
This method allows to compute the probability to observe a sample Spearman correlation coefficient less or equal to a given value for a bidimensional sample of given size with independent components. It is usefull eg to test for null Spearman correlation as in HypothesisTest. BTW, I changed the implementation and the documentation of HypothesisTest::Spearman(), HypothesisTest::PartialSpearman() and HypothesisTest::FullSpearman() to reflect their actual behavior: to test for the *absence* of rank correlation.
The implementation is based on the exact distribution given [here](http://www.luke-g.com/math/spearman/index.html)
for sample size not greater than 26 (which is larger than any implementation in R), otherwise it is based on an Edgeworth expansion identical to the one in Algotihm AS 89 [here](https://www.jstor.org/stable/2347111)


lib/src/Uncertainty/Distribution/DistFunc.cxx
```diff
@@ -1006,6 +1006,47 @@ Indices DistFunc::rPoisson(const Scalar lambda,
   return result;
 }
 
+/********************************************************************************************************************************/
+/* Spearman rho distribution: let (X_i) and (Y_i) be two independent samples of size n. This function computes the probability that the sample Spearman correlation between the samples is less or equal (tail = false) or is greater (tail = ture) than the given value.
```
jschueller: DistFunc.cxx:1011:1: warning: "/*" within comment [-Wcomment]

jschueller:
Some suggestion about testing: the branch `if (size <= 26)` is not tested, also HypothesisTest::Spearman may be moved to a separate test so that its run whether or not R-rot is present.

regislebrun:
Good idea. I messed up my local git repository (no more master...) so first I fix it, then back to code.

jschueller:
Cool, I'll probably issue PR with the independence test improvement from #784 based on this work.

jschueller:
If you want we can merge the first commit so I can propose the independent copula identification enhancement in chaos, if you can move the testresult modification to another PR before the end of the week.

regislebrun:
@jschueller Ok. Not 100% sure to have the time to propose a new PR by the end of the week but I will do my best.

jschueller:
I meant the TestResult PR can wait, just drop the last commit here and I think it could be merged.

jschueller:
It's not really a header, but rather a big chunk of code which contains a big array.

PR #816, opened 2018-08-01
Fix Sample.__str__ ellipsis condition

jschueller:
size_ must be used, not data.getSize(), which is the global number of
elements, also chose to use '>=' instead of '>'.


regislebrun:
+1

PR #818, opened 2018-08-01
Fix BurrFactory error checking

jschueller:
sumLogXC cannot be 0 else k is inf,
but now we also check that sumLogXC is not inf (=> k=0, and sumRatio=-nan)


regislebrun:
+1

PR #820, opened 2018-08-02
Fix HistogramFactory::computeSilvermanBandwidth

jschueller:
Using the 2nd rule (useQuantile=False) always returns QauntileEpsilon


jschueller:
@regislebrun is it good for you ?

regislebrun:
+1

PR #823, opened 2018-08-09
Graph: apply margins within log bounds

jschueller:
In log-scale, check that the margin would not exceed the range

This is needed in the spectral estimation example from doc branch.

### DIFF
--- a/lib/src/Base/Graph/GraphImplementation.cxx
+++ b/lib/src/Base/Graph/GraphImplementation.cxx
@@ -629,10 +629,9 @@ void GraphImplementation::computeBoundingBox() const
     return;
   }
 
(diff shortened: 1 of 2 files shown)


regislebrun:
IMO log scales should apply only for positive values, zero is not allowed (an exception should be thrown if some value is <= 0). In this case, taking a margin is dividing the min value by a given factor and multiplying the max value by the same factor. The usual margin should be interpreted as a power of ten.

jschueller:
Good idea.

sofianehaddad:
seems ok. Test ? Even a picture associated to this PR ?

jschueller:
There is an example in the doc branch.

PR #824, opened 2018-08-09
Fix composed distributions description warnings

jschueller:
The warning from (https://github.com/openturns/openturns/pull/794/commits/4a0012a109d2b33552f86abc1484df967e32cccf) is really common with ComposedDistribution.


regislebrun:
+1

PR #829, opened 2018-09-06
Parametrize statistical tests by first kind risk

jschueller:
... instead of 1-risk

follow-up of #812

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -39,6 +39,7 @@
  * Deprecated UserDefinedCovarianceModel constructor based on a Collection<CovarianceMatrix>
  * Deprecated FieldFunction,FieldToPointFunction::getSpatialDimension
  * Removed VertexFunction class
+ * Deprecated LinearModelRSquared, LinearModelAdjustedRSquared
 
 === Python module ===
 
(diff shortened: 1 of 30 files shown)


jschueller:
I noticed that in the R2 tests (LinearModelRSquared, LinearModelAdjustedRSquared) the threshold is still the r2, not the risk level.

regislebrun:
@jschueller What is the rational behind this choice for R2 tests?

jschueller:
Hum... the definition of the P-value must be changed.

sofianehaddad:
What about fitting tests ?

regislebrun:
@jschueller I rewrote the documentation of TestResult in a commit related to PR#812 with the definition of the p-value&co, but I had to drop it in order to speed-up the integration process of the PR. From my point of view, the current TestResult class is a little bit dumb as it does not include the value of the test statistics nor the associated threshold, so you don't have any room to store the value of the test statistics...

jschueller:
@sofianehaddad, changed missing FittingTest too

jschueller:
Should I drop the 2nd commit ?

jschueller:
I dropped the last commit, the linear model test result is unchanged.

sofianehaddad:
IMO we should remove the stat test associated to LinearModel as we are just comparing `R2` to a specific threshold

Otherwise seems ok for me, +1

jschueller:
I can deprecate it.

regislebrun:
+1

PR #830, opened 2018-09-06
Improve resource map

regislebrun:
Now, there are separated maps for String, Scalar, UnsignedInteger and Bool in order to avoid costly conversions.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -42,11 +42,13 @@
  * Deprecated FieldFunction,FieldToPointFunction::getSpatialDimension
  * Removed VertexFunction class
  * Deprecated LinearModelRSquared, LinearModelAdjustedRSquared
+ * Added ResourceMap::GetType to access the type of a given key.
 
 === Python module ===
 
 === Miscellaneous ===
  * CMake >= 2.8.4 is required to build OpenTURNS
+ * The format of openturns.conf has changed to make a distiction between types of keys
 
 === Bug fixes ===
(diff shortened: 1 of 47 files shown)


jschueller:
Did you push the parallel-threads fix ?
```
ResourceMap.cxx:343:51: error: invalid conversion from 'char*' to 'OT::UnsignedInteger {aka long unsigned int}' [-fpermissive]
   setAsUnsignedInteger( "parallel-threads", getenv("NUMBER_OF_PROCESSORS") );
```

jschueller:
One drawback is that the user has to use the right api to set a value.
Consider SquaredExponential-DefaultTheta, in http://openturns.github.io/openturns/latest/user_manual/_generated/openturns.ResourceMap.html, look at ```SquaredExponential-DefaultTheta``` it is displayed as ```1``` but the user doesnt know if it's a float or an int, and whether he should use setAsUnsignedInteger or setAsScalar.
Can we infer the type of a key with the new API ? At least there should be a "Type" column in that documentation.

jschueller:
Another idea would be keeping the single string map but to cache converted values.

regislebrun:
I am not very fond of the cache idea. I added a type accessor to ResourceMap and enhanced the documentation, IMO it is enough to make the user happy.

sofianehaddad:
Needs a rebase

regislebrun:
rebase done, everything looks alright, no?

jschueller:
Ok for me.

PR #833, opened 2018-09-10
Added random algebra SVD computation to KarhunenLoeveSVDAlgorithm

regislebrun:
The implementation proposes two methods. The first one is based on https://arxiv.org/abs/0909.4061. It GREATLY improve the speed of the computation for large cases but the accuracy depends on the problem dimension. The second one is based on Nathan Halko, Per-Gunnar Martisson, Yoel Shkolnisky and Mark Tygert, "An algorithm for the principal component analysis of large data sets", arXiv:1007.5510v2, 19 Mar 2011.


sofianehaddad:
why setting only random svd for KL purposes ? I mean we could get other applications thus IMO the development should be done outside KL. What do you think ?

jschueller:
I thought of this too. Anyways at least it lacks some doc and refs to the methods.

regislebrun:
I updated the documentation and the comments in the code. Is it ok now?

lib/src/Uncertainty/Process/KarhunenLoeveSVDAlgorithm.cxx
```diff
@@ -179,11 +180,128 @@ void KarhunenLoeveSVDAlgorithm::run()
...
+  {
+    LOGINFO("Use random SVD");
+    // Maximum number of singular values to compute
+    MatrixImplementation & A = designMatrix;
+    const UnsignedInteger m = A.getNbRows();
+    const UnsignedInteger n = A.getNbColumns();
+    const UnsignedInteger rank = std::min(std::min(ResourceMap::GetAsUnsignedInteger("KarhunenLoeveSVDAlgorithm-RandomSVDMaximumRank"), A.getNbColumns()), A.getNbRows());
+    LOGINFO(OSS() << "Maximum rank=" << rank);
+    LOGINFO(OSS() << "A=" << A.getNbRows() << "x" << A.getNbColumns());
+    if (ResourceMap::Get("KarhunenLoeveSVDAlgorithm-RandomSVDVariant") == "redsvd")
```
jschueller: Halko2010

PR #837, opened 2018-09-14
Fix error handling in operator(Sample)

jschueller:
We have to check for exceptions after the evaluation of each point.
Closes http://trac.openturns.org/ticket/965


regislebrun:
+1

mbaudin47:
Funny but correct unit test : it fails if the code succeeds !

PR #838, opened 2018-09-17
Deep-copy MemoizeEvaluation cache

jschueller:
Currently the same instance of cache is shared when the underlying
evaluation is also a MemoizeEvaluation. This causes values to be
duplicated in the cache.

Either we could disable the cache in the top-level MemoizeEvaluation or
deep copy the cache. Here I choose to have a copy for the user to
retrieve easily the values.


jschueller:
cc @adumasphi

PR #840, opened 2018-09-17
Do not allow to add twice the pair to the cache

jschueller:
Else the keys_ attribute contains duplicates.
Does not hurt the insertion performance, see bench from #606.

dae33cf is important here, while the other changes are cosmetic.

I did not see in #838 that this should be fixed in the Cache class.


regislebrun:
+1

jschueller:
Also fixed getInputCache/getOutputCache order consistency

PR #841, opened 2018-09-17
 Doc: Fix invalid modindex link

jschueller:


regislebrun:
Not clear for me what issue is addressed here, but +1 if you find it useful ;-)!

jschueller:
There is a dead link in contents.rst, we just remove it to avoid the 404, nothing funky.
cc @aurelieladier

PR #845, opened 2018-09-21
Improve sobol indices experiment

regislebrun:
Now, the constructor based on a Distribution can build other experiments than a MonteCarloExperiment, namely a LHSExperiment or a LowDiscrepancyExperiment. It is controlled by the key 'SobolIndicesExperiment' in ResourceMap, and the construction takes into account both the limitation of SobolSequence in terms of dimension and the specific needs of Sobol indices computation in terms of shuffling and scrambling.
    
The SobolIndicesAlgorithm has also been fixed in order to allow the use of the asymptotic distribution only when the underlying sampling method generates IID samples, otherwise the formulas are no more valid.


jschueller:
Ok for me once you squash these commits and fix the doc.

PR #848, opened 2018-09-22
Improved Distribution::computeQuantile()

regislebrun:
Now, the quantile function is continuous. It fixes ticket #915.


regislebrun:
@jschueller Maybe the underlying code was not so wrong ;-)

jschueller:
We need some test

regislebrun:
added tests. If the output is empty, it is ok (at least for all the univariate distributions with a factory).

sofianehaddad:
Thanks, interesting test :-)

PR #852, opened 2018-09-24
 Add a generic Sobol' plot function 

jschueller:
Add SobolIndicesAlgorithm::DrawSobolIndices method to plot external Sobol indices.
We use it to fix chaos examples as mentioned in http://trac.openturns.org/ticket/958.


lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
```diff
@@ -594,54 +594,73 @@ Sample SobolIndicesAlgorithmImplementation::computeIndices(const Sample & ,
   throw new NotYetImplementedException(HERE);
 }
 
-/** Method that draw (plot) the sensitivity graph */
-Graph SobolIndicesAlgorithmImplementation::draw() const
+
+Graph SobolIndicesAlgorithmImplementation::DrawSobolIndices(const Description & inputDescription,
+                                                            const Point & firstOrderIndices,
+                                                            const Point & totalOrderIndices)
 {
```
sofianehaddad: we can also add a title argument

lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
```diff
@@ -594,54 +594,73 @@ Sample SobolIndicesAlgorithmImplementation::computeIndices(const Sample & ,
...
+  Graph graph(DrawSobolIndices(inputDesign_.getDescription(), getAggregatedFirstOrderIndices(), getAggregatedTotalOrderIndices()));
+  if (outputDesign_.getDimension() > 1)
+    graph.setTitle(OSS() << "Aggregated Sobol' indices - " << getClassName());
+  else
+    graph.setTitle(OSS() << "Sobol' indices - " << getClassName());
+  const UnsignedInteger dimension = mergedFirstOrderIndices_.getDimension();
+
+  // Draw confidence intervals
   if (bootstrapSize_ > 0 && confidenceLevel_ > 0.0)
   {
```
sofianehaddad: We might check if we use asymptotic estimation ? in that case, no need to have bootstrapSize_ > 0

lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
jschueller: It's easily set through the graph.

lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
jschueller: Indeed, fixed

PR #854, opened 2018-09-25
Fix student factory

regislebrun:
The previous code was completely wrong. Now, the factory is able to estimate multivariate Student distributions, with the restriction of nu>2 due to the use of first and second moments for the location and scale parameters. The correlation is estimated in a robust way using the sample Kendall tau matrix, and the number of degrees of freedom nu is obtained by profiled log-likelihood maximization.

I also improved the elliptical distributions: now the log-PDF is computed directly on a logarithmic scale, improving the accuracy of log-likelihood computations.


jschueller:
The test slightly fail on mingw though.

jschueller:
The test slightly fail on mingw though, please fix the numerical precision.

regislebrun:
@jschueller done

jschueller:
The python student test fails, seems setnumericalprecision(4) would do it like the cxx test.

jschueller:
The test fails on appveyor, small numeric diff.

regislebrun:
@jschueller I hope this time it will be ok. Otherwise, I will have to remove all the significant figures from the result ;-)!

jschueller:
it's ok now on appveyor, thanks

PR #855, opened 2018-09-26
Improved the performance of MarginalTransformationEvaluation

regislebrun:
The improvement of ResourceMap (direct access to a Scalar instead of a systematic conversion from a String) lead to an improvement of x1.94 wrt OT 1.11 on the benchmark (Uniform, Triangular)->(Normal, Normal). In this PR, the introduction of a specific attribute to switch to tail evaluation gives an improvement of x2.48 wrt OT 1.11.


jschueller:
I think you can just use "static", do you have the script of the benchmark for me to test ?

PR #858, opened 2018-09-28
Deprecated RandomVector specific ctors

jschueller:
As we did for Function, we disambiguate the RandomVector API:

- RandomVector(Point) => ConstantRandomVector
- RandomVector(Distribution, RandomVector) => ConditionalRandomVector
- RandomVector(Function, RandomVector) => CompositeRandomVector
- RandomVector(FunctionalChaosResult) => FunctionalChaosRandomVector

Only RandomVector(Distribution) is kept (UsualRandomVector), what do you
think about that ? Should we deprecate it too ? Find a better name than UsualRandomVector ?

In the future this will help simplifying the swig module structure.


regislebrun:
+1

PR #859, opened 2018-09-28
Fix 1-d NormalCopula::getParametersCollection

jschueller:
It's better to use DistributionImplementation::getParametersCollection
as it relies on getParameter which returns an empty point in NormalCopula.

Now NormalCopula(1).getParametersCollection() returns a collection with
an empty point instead of an empty collection and fixes http://trac.openturns.org/ticket/970


regislebrun:
+1

PR #861, opened 2018-09-28
Typo in openturns.conf

jschueller:
On a side note I noticed the types are inconsistent with the xml tags:
```
static const char * XMLTag_value                  = "value_string";
static const char * XMLTag_value_scalar           = "value_scalar";
static const char * XMLTag_value_unsigned_integer = "value_unsigned_integer";
static const char * XMLTag_value_bool             = "value_bool";
```
vs
```
if (it != mapString_.end()) return "string";
if (it != mapScalar_.end()) return "float";
if (it != mapUnsignedInteger_.end()) return "unsigned int";
if (it != mapBool_.end()) return "bool";
```

I propose to use Python types instead: "str", "float", "int", "bool", it will show more nicely in the ResourceMap doc, what do you think ?

### DIFF
--- a/lib/src/Base/Common/ResourceMap.cxx
+++ b/lib/src/Base/Common/ResourceMap.cxx
@@ -33,9 +33,9 @@ BEGIN_NAMESPACE_OPENTURNS
 static const char * ConfigurationFileName = "openturns.conf";
 #if defined OPENTURNS_HAVE_LIBXML2
 static const char * RootElementName = "openturns-configuration";
-static const char * XMLTag_value                  = "value_string";
-static const char * XMLTag_value_scalar           = "value_scalar";
-static const char * XMLTag_value_unsigned_integer = "value_unsigned_integer";
+static const char * XMLTag_value_str              = "value_str";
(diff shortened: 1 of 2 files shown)


lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -33,9 +33,9 @@ BEGIN_NAMESPACE_OPENTURNS
 static const char * ConfigurationFileName = "openturns.conf";
 #if defined OPENTURNS_HAVE_LIBXML2
 static const char * RootElementName = "openturns-configuration";
-static const char * XMLTag_value                  = "value_string";
-static const char * XMLTag_value_scalar           = "value_scalar";
-static const char * XMLTag_value_unsigned_integer = "value_unsigned_integer";
+static const char * XMLTag_value                  = "value_str";
```
regislebrun: It would be better to name the variable XMLTag_value_str here, and so on.

regislebrun: +1

jschueller: ok!

PR #863, opened 2018-10-01
Wrong type in MarginalTransformationEvaluation

jschueller:
introduced by #855


regislebrun:
@jschueller Oupss... sorry!

+1

PR #868, opened 2018-10-08
Fix crash in Python conversions

jschueller:
Fixes empty array conversions crashes with GLIBCXX_ASSERTIONS when accessing &std::vector[0]


regislebrun:
+1

PR #870, opened 2018-10-14
Various enhancements

regislebrun:
I will keep this PR open as long as possible before the next release in order to integrate several enhancements listed on the trac site. For now, it contains:
  *  An enhanced interface for KernelSmoothing in order to fix [Ticket #660](http://trac.openturns.org/ticket/660)
  * A replacement of the basic binning by a linear binning in KernelSmoothing
  * The possibility to specify the bounds values and the side where a boundary correction is needed. It is a partial answer to Ticket #408](http://trac.openturns.org/ticket/408).
  *  A minor enhancement of Interval to take into account infinite bounds
  * The ability to partially load a Study if some objects cannot be reloaded, in order to reload files coming from previous versions
  * Added a simplified TruncatedDistribution constructor. It allows to define more concisely and in a more natural way a TruncatedDistribution for Bayesian modelling.
  * Fixed a bug in setParameter()/getParameter() where the order in which the bounds were serialized were different.
  *  Fixed a bug in KernelMixture: the PDF/CDF approximations were not recomputed when changing the parameters.
  * Improved Multinomial documentation. Now, the link with the classical definition (eg Wikipedia) is made explicit. The illustration has also been updated.
  *  Improved Distribution::drawPDF(). Now it produces something meaningful for 2D discrete distributions.
  * Improved RandomMixture. Now the construction of a RandomMixture is much more efficient in the analytical case, in particular for the 1D case.
  * Improved UserDefined. Now, it uses the alias method for sampling, as described in Devroye86, Chapter 3. It has an optimal space and time complexity of resp. O(n) and O(1), instead of O(nlog(n)) and O(log(n)) with the previous implementation. For a distribution with a support of size 10^4 the speed-up is of ~300%, and much higher for higher dimensional cases.


sofianehaddad:
Ok for me
IMO 4af8e9c ffbc395 and a256a44 should be squashed into one commit
Same remark for 2756916  and b5d8013

jschueller:
I agree.

regislebrun:
done.

python/test/t_DistFunc_discrete.py
```diff
@@ -0,0 +1,20 @@
...
+TESTPREAMBLE()
+RandomGenerator.SetSeed(0)
+
+try:
+    # Discrete related functions
+    # rDiscrete
+    p = [0.1, 0.2, 0.3, 0.4]
+    print("rDiscrete(", p, ")=%.6g" %
+                      DistFunc.rDiscrete(p))
+    print("rDiscrete(", p, ", 100)=",
```
jschueller: The print of p is formatted differently on the autobuilder and that test fails, could you wrap it in a Point ?
http://autobuilder.openturns.org/openturns/pull/870/squeeze-x86_64-ot-linux/LastTest.log

regislebrun: @jschueller Done. I used an Indices instead of a Point as the returned value was an UnsignedInteger.

jschueller: It's "p" that's printed wrong, not the result.

regislebrun: @jschueller Ouch! Fixed.

PR #876, opened 2018-10-17
Require CMake 2.8.8

jschueller:
The new dynamic lookup macro uses string(MD5 ...) signature which is not
available before 2.8.7.
I figured we can bump cmake a bit (december 2011) and get rid of a few
more workarounds.


regislebrun:
+1

PR #877, opened 2018-10-22
Remove citations from DI::getIsoPT docstring

jschueller:
Turns out we cannot use citations cleanly in inherited docsrtings:
https://github.com/sphinx-doc/sphinx/issues/5528

We can simply remove these as the docstring already points to the
theoric section where citations are.


regislebrun:
+1

PR #880, opened 2018-10-23
Fix sobol evaluation in special case

jschueller:
When the dim is 2 and we want to compute second order indices the code
throwed as samples size did not match in setDesign.
Now we always extend the design when in dim=2 in setDesign.


jschueller:
cc @adumasphi

regislebrun:
+1

PR #881, opened 2018-10-25
Viewer: Force external Axes properties

jschueller:


python/src/viewer.py
```diff
@@ -223,10 +207,24 @@ def __init__(self,
...
+        self._ax[0].set_ylim([bb.getLowerBound()[1], bb.getUpperBound()[1]])
+
+        # set scale
+        if (graph.getLogScale() == ot.GraphImplementation.LOGX) or (graph.getLogScale() == ot.GraphImplementation.LOGXY):
+            self._ax[0].set_xscale('log')
+        if (graph.getLogScale() == ot.GraphImplementation.LOGY) or (graph.getLogScale() == ot.GraphImplementation.LOGXY):
+            self._ax[0].set_yscale('log')
+
+        # set title
+        self._ax[0].set_title(self.ToUnicode(graph.getTitle()))
```
sofianehaddad: For me, title is a property for figure !

`
self._fig.suptitle(self.ToUnicode(graph.getTitle()))
`

jschueller: Ok. Note that the title is a bit higher now, not just on top of the graph frame.

regislebrun:
+1

PR #882, opened 2018-10-26
Add ParametricPointToFieldFunction

jschueller:


regislebrun:
@jschueller Do you plan to add other parametric functions, eg. ParametricPointToPointConnection? Is this class a shortcut for a composition between a ParametricFunction and a PointToFieldFunction or is it more?
If not, it is ok for me.

jschueller:
Yes, this is needed because you cannot compose a ParametricFunction and a PointToFieldFunction with PointToFieldConnection as you need a ref to the function; this could be possible with a PointToFieldConnection+linearfunction, but this solution is complicated.

PR #883, opened 2018-10-26
Bugfix in EmpiricalBernsteinCopula

sofianehaddad:
When building the EmpiricalBernsteinCopula, if provided sample is not a  "copula sample", we
enforce sample size to be multiple of binNumber, thus is becomes a copula.

This fixes http://trac.openturns.org/ticket/975

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -96,6 +96,7 @@
  * #968 (Empty legend make OT crash)
  * #970 (Composing gaussian copulas can crash the chaos)
  * #972 (FittingTest::ChiSquared slow and buggy)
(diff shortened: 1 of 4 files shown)


regislebrun:
+1

jschueller:
Could you add a test ?

sofianehaddad:
yes but not before the end of the week

jschueller:
You mean wednesday ? :)

sofianehaddad:
Sorry I mean Friday

jschueller:
No problem, the release is not going to happen before next week.

sofianehaddad:
@jschueller done

regislebrun:
+1

PR #887, opened 2018-10-30
Various fixes

jschueller:


jschueller:
cc @bchaingne for the nlopt commits

bchaigne:
ok for the constraint error computation

bchaigne:
I agree with the change on using the relative output tolerance but then the residualErrorHistory is no more consistent: it computes the absolute value. The doc tells it should contain the norm of the Lagrangian gradient anyway.

bchaigne:
ok for the doc

jschueller:
@bchaigne, see d55b55c

bchaigne:
Ok, now I see that none of the algorithms actually implemented the `residualError` properly as the doc tells (i.e. gradient of the Lagrangian). I think the doc of `OptimizationResult::getResidualError()` should be updated too.

sofianehaddad:
@jschueller agreed with @bchaigne 

If `residualError` does not refers to the gradient of Lagragian, documentation should be updated.

jschueller:
@sofianehaddad ok

python/test/t_SobolIndicesAlgorithm_std.expout
```diff
@@ -361,3 +361,35 @@ Aggregated total order indices interval =  [0.481412, 0.505571]
  [ 0.00667425 0          ]]
 [[ 0          0.00667425 ]
  [ 0.00667425 0          ]]
+First order indices =  [0.906196,0.0868706,-0.00929482]
+Total order indices =  [0.911593,0.0977023,-1.62617e-11]
```
regislebrun: @jschueller Will this tiny value be a problem on different platforms?

python/test/t_SobolIndicesAlgorithm_std.expout
```diff
@@ -361,3 +361,35 @@ Aggregated total order indices interval =  [0.481412, 0.505571]
...
+Aggregated first order indices interval =  [0.881264, 0.931129]
+[0.0670776, 0.106664]
+[-0.0291558, 0.0105662]
+Aggregated total order indices interval =  [0.892125, 0.931061]
+[0.0895866, 0.105818]
+[-1.62617e-11, -1.62616e-11]
```
regislebrun: @jschueller Same remark here

python/test/t_SobolIndicesAlgorithm_std.expout
```diff
@@ -361,3 +361,35 @@ Aggregated total order indices interval =  [0.481412, 0.505571]
...
+First order indices =  [0.86802,0.0947589,0]
+Total order indices =  [0.878764,0.100685,9.93901e-15]
+Aggregated first order indices interval =  [0.837938, 0.898101]
+[0.0860047, 0.103513]
+[0, 0]
+Aggregated total order indices interval =  [0.859507, 0.898021]
+[0.0926155, 0.108754]
+[9.93901e-15, 9.93901e-15]
+First order indices =  [0.900039,0.092229,-0.000294721]
+Total order indices =  [0.898092,0.098666,-1.83997e-15]
```
regislebrun: @jschueller and here

python/test/t_SobolIndicesAlgorithm_std.expout
```diff
@@ -361,3 +361,35 @@ Aggregated total order indices interval =  [0.481412, 0.505571]
...
+Aggregated first order indices interval =  [0.897154, 0.902924]
+[0.073066, 0.111392]
+[-0.0197411, 0.0191517]
+Aggregated total order indices interval =  [0.879021, 0.917163]
+[0.0958158, 0.101516]
+[-1.83997e-15, -1.83997e-15]
```
regislebrun: @jschueller and here

python/test/t_SobolIndicesAlgorithm_std.expout
jschueller: no, it passes everywhere

jschueller:
@regislebrun, @sofianehaddad I added several more commits to use the Bernstein through the distributionfactory api.

jschueller:
@regislebrun interesting that you mention Normal with non pd correlation, I was just investigating a case where NormalCopulaFactory fails, and I think we already banned regularization a long time ago. What would be the implications of allowing spd correlations ?

jschueller:
@regislebrun could some kind of regularization be used to turn the matrix into an SPD one ?

python/test/t_SobolIndicesAlgorithm_std.expout
jschueller: Oh, you were right, it fails on i386

regislebrun:
@jschueller It is tempting to multiply all the off-diagonal entries by 1-dimension * SpecFunc::Precision but with no guarantee that it will give something stable enough... If you have something to do it looks like a natural choice, making the matrix more diagonal dominant.

jschueller:
@regislebrun I added 2 more commits to add BernsteinCopulaFactory to GetMultivariateFactories, fix the default ctor of EmpirialBernsteinCopula

regislebrun:
@jschueller for me its ok.

PR #888, opened 2018-11-06
Integration of fmath

regislebrun:
- Added SpecFunc::Exp(): This method allows to use the *fast* implementation of the exponential function provided by https://github.com/herumi/fmath/blob/master/fmath.hpp. The resulting speedup of EmpiricalBernsteinCopula::computePDF() is significant (x2.3).

- Replaced std::exp() by SpecFunc::Exp(): All the computations of exponential of real numbers have been replaced in order to benefit from the new high performance implementation of the exponential function.


sofianehaddad:
Could you please provide here a small bench ?

.circleci/run_docker_mingw.sh
```diff
@@ -15,7 +15,7 @@ CXXFLAGS="-D_hypot=hypot -D_GLIBCXX_ASSERTIONS" ${ARCH}-w64-mingw32-cmake \
   -DPYTHON_LIBRARY=${MINGW_PREFIX}/lib/libpython${PYMAJMIN}.dll.a \
   -DPYTHON_EXECUTABLE=/usr/bin/${ARCH}-w64-mingw32-python${PYMAJMIN}-bin \
   -DPYTHON_SITE_PACKAGES=Lib/site-packages \
-  -DUSE_TBB=OFF \
+  -DUSE_TBB=OFF -DCMAKE_CXX_FLAGS="-march=native" \
```
jschueller: I 'd rather not to have to use "march=native". What happens is that the ci uses mingw in it's 32 bits variant (64 is available too though). It works because the host processor actually is x86_64, so it has sse2. But the client machine could run an i686 processor and this will crash with an illegal instruction. 

A better way to do that is to explicitely test if sse2 is available:
```
try_compile (OPENTURNS_HAVE_SSE2
  ${CMAKE_CURRENT_BINARY_DIR}
  ${CMAKE_CURRENT_SOURCE_DIR}/cmake/sse2.cxx)
```

with cmake/sse2.cxx:
```
#include <emmintrin.h>

int main()
{
  __m128d v = _mm_setzero_pd();
  return 0;
}
```

Then we only use fmath only if OPENTURNS_HAVE_SSE2.

jschueller: ok

jschueller:
What version of gcc do you use ?

jschueller:
Maybe it's more useful to try to just use -ffast-math:
std::exp / -O2 : 0m9,231s
std::exp / -O2 -ffast-math : 0m2,020s
fmath::expd / -O2: 0m1,629s
fmath::expd / -O2 -ffast-math: 0m2,029s


```
#include <iostream>

#include "fmath.hpp"

int main() {
  double sum = 0;
    for (double x = 0; x <= 1000.0; x += 1e-6) {
        sum += fmath::expd(x);
//         sum += std::exp(x);
    }
}
```

regislebrun:
@jschueller I don't agree with your proposal. The option -ffast-math relaxes a lot of the controls made on floating point arithmetic, such as the associativity of elementary operations. Several parts of OT have been written with an exact IEEE conformance in background and this option would break these algorithms as there is no more control on how the rounding errors accumulate. BTW, the fact that this option makes fmath::expd() run slower is a sign that something is going wrong.

I use the following flags to compile OT:
"-Wall -Wextra -O3 -mfpmath=sse -mtune=native -march=native -msse -msse2 -msse3 -msse4.1 -msse4.2 -mavx -maes -msse4 -mmmx"
but such options are useless on your benchmark as I just checked.

So a very localized optimization, even if written in SSE2 assembly language, looks better to me than a global relaxation of the IEEE norm for a numerical library.

jschueller:
Yes, I see some tests failing with fastmath enabled.

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -503,6 +509,42 @@ Scalar SpecFunc::DiLog(const Scalar x)
...
+{
+  const UnsignedInteger size = x.getSize();
+#ifdef OPENTURNS_HAVE_FMATH
+  Point result(size);
+  for (UnsignedInteger i = 0; i < size; ++i)
+    result[i] = fmath::expd(x[i]);
+  return result;
+  /*  if (size == 0) return Point(0);
+  std::vector< double > result(x.begin(), x.end());
+  if (size % 2 == 0)
```
jschueller: maybe n should be a multiple of 4 instead of 2 when avx is used

regislebrun:
@jschueller There is this comment line 472 in fmath.hpp:
`remark : -ffast-math option of gcc may generate bad code for fmath::expd`

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
regislebrun: @jschueller In fmath.hpp, lines 536-537, the documentation of fmath::expd_v reads:
```
	px : pointer to array of double
	n : size of array(assume multiple of 2 or 4)
```

jschueller:
see https://github.com/openturns/openturns/pull/888/commits/36202c91a09a761ac2c5c967dc72d5f7e12e7f89

jschueller:
nice, can you squash commits 1-3-4, plus commits 2-5 ?
also can you try setting ARCH=x86_64 in run_docker_mingw.sh to make sure this passes on mingw ?

jschueller:
SpecFunc fails withtout output for msvc 14, could be crash, maybe the avx2 variant is broken

jschueller:
Ok it works on mingw too.
Only the glm test output changes a bit on both Linux & mingw.
But the SpecFunc test crashes on msvc, weird, I tested it separately and it's ok: https://ci.appveyor.com/project/jschueller/fmath/build/job/u292ko1qk5oj1eu7.
I rerun it.

jschueller:
Maybe just disable it for mscv ?
```
if (OPENTURNS_HAVE_SSE2 AND NOT MSVC) 
    set (OPENTURNS_HAVE_FMATH TRUE)
```

jschueller:
so only the glm test fails now due to a precision issue

jschueller:
Great, unfortunately the glm test is still failling:
```
-covariance (full optim)=class=AbsoluteExponential scale=class=Point name=Unnamed dimension=1 values=[0.133] amplitude=class=Point name=Unnamed dimension=1 values=[0.191]
-trend (full optim)=[[-0.103,1.01]]
+covariance (full optim)=class=AbsoluteExponential scale=class=Point name=Unnamed dimension=1 values=[0.01] amplitude=class=Point name=Unnamed dimension=1 values=[0.191]
+trend (full optim)=[[-0.111,1.01]]
```

jschueller:
the cmake/sse2.cxx file should be removed (sse2.c is used)

regislebrun:
I don't understand why some tests are still failing under mingw as I deactivated the use of fast_math on this target. Quite annoying to spend so much tie on such a bug!

jschueller:
The test also fails on Linux; maybe fmath does not yield consistent results between avx and sse.
Try to run that test with fmath disabled on your box.

PR #892, opened 2018-11-16
Remove rot dependency in HypothesisTest

sofianehaddad:
In this PR, we implement : 
 - The `Pearson` correlation test
- The `ChiSquared` independence test

The objective here is to drop the `rot` dependency
It also fixes http://trac.openturns.org/ticket/977


jschueller:
Still in progress ?

sofianehaddad:
@jschueller the last part (FullRegression) will be handled in a new PR

sofianehaddad:
@jschueller indeed

PR #894, opened 2018-11-19
Optim: Track progress wrt eval number

jschueller:
In SQP, Abdorackwitz the number of iterations was used.
It was missing from TNC, Cobyla, NLopt.
For multistart the budget is the total of local searches.


jschueller:
cc @aurelieladier

regislebrun:
+1

PR #891, opened 2018-11-13
Updating correlation analysis

sofianehaddad:
In this PR, we add missing theory documentation on `Standard Rank Regression Correlation`. We also fix some details in documentation (bad/missing links, SRC expression...)

Finally PCC/PRCC do not rely on anymore `LinearModelFactory`, thus on `rot`

PR #895, opened 2018-11-19
SRC: Fix estimator using Y variance

jschueller:
We have to use the variance of Y sample in the denominator to take the
noise epsilon into account instead (before #891 it was also inacurrate)
IE SRC indices do not sum to 1 if there's no linear relationship.

http://openturns.github.io/openturns/latest/theory/reliability_sensitivity/ranking_src.html
http://openturns.github.io/openturns/master/theory/reliability_sensitivity/ranking_src.html

cc @adumasphi


adumasphi:
To me, the problem is that with the current formulation where the sum of all SRC = 1, we cannot see if the linear model hypothesis is validated or not. If we really use Var(Y), thus the sum of all SRC may be not equal to 1 and then indicates that the linear model does not fit the data so the SRC are not relevant.

sofianehaddad:
@adumasphi I agreed about the assessment of linearity.  
The issue is that when user gets the coefficients, he might be interested in comparing to other importances factors/sensitivity indices.
If he gets a linear chaos expansion, then first order indices sum to 1, even if linearity is not validated. In such situation, ranking issued from the two methods is the same, but sensitivity values differ.

By the way, I think we should take into account noise to get the "true" definition of SRC indices

sofianehaddad:
@regislebrun good idea! A ResourceMap key...

regislebrun:
... or an optional argument to the method.

sofianehaddad:
@jschueller agreed for the RM or method's argument?

jschueller:
With what default then ?

sofianehaddad:
IMO default should take into account noise

sofianehaddad:
@jschueller could you merge the PR ? I will integrate "scale" and "signed" purposed later (ticket 980)

jschueller:
yup

PR #896, opened 2018-11-23
Viewer: handle barplot fillstyle

jschueller:
xref http://trac.openturns.org/ticket/979
I dont know how to modify the hatch linestyle though.

barplot


regislebrun:
+1

PR #897, opened 2018-11-23
Add Opt++ solvers

jschueller:
Adds a new dependency to Opt++: https://software.sandia.gov/opt++/
Optpp is the user interface. The OptppImplementation tree hides the
implementation of all solvers (we want to encapsulate opt++ headers).

Some notes:
- Opt++ does not compile out of the box for osx, see https://github.com/openturns/optpp/commit/3e9f57afb6f251a6ecd657b89cc9db1f342dae67
- Opt++ does not link properly its dependencies (visible on platforms where undefined symbols are not allowed): https://github.com/openturns/optpp/commit/94e8224f0ebe732cfc6785dd10d26e39a2156fe1
- There is no way to retrieve the error message
- Only a return code is provided, but not consistently across all
algorithms
- So what I could do is only check for progress using the output points
variance
- By default opt++ can exit on error unless UseExceptions
is defined instead of SimulateExceptions in newmat11/include.h:20
I think this may happen with OptBCEllipsoid where it simply print stuff to the console and exit
- The documentation can differ from the code on supported constraint
types, added a summary of capabilities of each solver (did the same for NLopt in #899)
- OptGSS requires to be compiled separately because
of a declaration clash
- I did not find a clean way to interrupt the optimization, falling back to throwing
- under certain condition OptBaNewton might hang (kappa too bug in root solver for line search)

... so Opt++ is not so good as a c++ library, I hope it's better at maths :)


lib/src/Base/Optim/OPTppImplementation.cxx
```diff
@@ -0,0 +1,796 @@
...
+
+  result.setEvaluationNumber(size);
+  result.setOptimalPoint(optimizer);
+  result.setOptimalValue(Point(1, optimalValue));
+  result.setLagrangeMultipliers(computeLagrangeMultipliers(optimizer));
+  setResult(result);
+}
+
+
+/** Minimum step accessor */
```
regislebrun: no double stars comment in .cxx

lib/src/Base/Optim/OPTppImplementation.cxx
```diff
@@ -0,0 +1,796 @@
...
+{
+  minimumStep_ = minimumStep;
+}
+
+Scalar OPTppImplementation::getMinimumStep() const
+{
+  return minimumStep_;
+}
+
+/** Maximum step accessor */
```
regislebrun: no double stars comment in .cxx

lib/src/Base/Optim/OPTppImplementation.cxx
```diff
@@ -0,0 +1,796 @@
...
+{
+  maximumStep_ = maximumStep;
+}
+
+Scalar OPTppImplementation::getMaximumStep() const
+{
+  return maximumStep_;
+}
+
+/** Maximum line search tolerance accessor */
```
regislebrun: no double stars comment in .cxx

lib/src/Base/Optim/OPTppImplementation.cxx
```diff
@@ -0,0 +1,796 @@
...
+{
+  lineSearchTolerance_ = lineSearchTolerance;
+}
+
+Scalar OPTppImplementation::getLineSearchTolerance() const
+{
+  return lineSearchTolerance_;
+}
+
+/** Maximum line search iteration number accessor */
```
regislebrun: no double stars comment in .cxx

lib/src/Base/Optim/OPTppImplementation.cxx
```diff
@@ -0,0 +1,796 @@
...
+{
+  maximumLineSearchIterationNumber_ = maximumLineSearchIterationNumber;
+}
+
+UnsignedInteger OPTppImplementation::getMaximumLineSearchIterationNumber() const
+{
+  return maximumLineSearchIterationNumber_;
+}
+
+/** Trust region size accessor */
```
regislebrun: no double stars comment in .cxx

lib/src/Base/Optim/OPTppImplementation.cxx
```diff
@@ -0,0 +1,796 @@
...
+Description OPTppImplementation::AlgorithmNames_;
+
+
+void OPTppImplementation::InitializeProblem(int ndim, NEWMAT::ColumnVector& x)
+{
+  if (!P_current_opt_)
+    throw InternalException(HERE) << "P_current_opt_ is null";
+  const UnsignedInteger dimension = P_current_opt_->getProblem().getDimension();
+  const Point startingPoint(P_current_opt_->getStartingPoint());
+  for (UnsignedInteger i = 0; i < dimension; ++ i)
```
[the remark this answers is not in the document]
jschueller: no, their ColumnVector type does not implement iterators

jschueller:
@regislebrun dont worry the minimization flag is weel handled: we pass f/-f to the solver each time it requests an evaluation or a gradient and retrieve the optimal value with reversing too.

regislebrun:
@jschueller and what about the Hessian?

jschueller:
Indeed it seems that one was not taking sign into account, fixed.

regislebrun:
+1

PR #898, opened 2018-11-26
NLopt: fix constraint error computation

bchaigne:
Negative values should be checked for constraint violation since
positive values means feasibility.

### DIFF
--- a/lib/src/Base/Optim/NLopt.cxx
+++ b/lib/src/Base/Optim/NLopt.cxx
@@ -364,7 +364,7 @@ void NLopt::run()
       Point h(getProblem().getInequalityConstraint()(inP));
       for (UnsignedInteger k = 0; k < getProblem().getInequalityConstraint().getOutputDimension(); ++ k)
       {
-        h[k] = std::max(h[k], 0.0);// convention h(x)>=0 <=> admissibility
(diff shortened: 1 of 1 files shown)


lib/src/Base/Optim/NLopt.cxx
```diff
@@ -364,7 +364,7 @@ void NLopt::run()
       Point h(getProblem().getInequalityConstraint()(inP));
       for (UnsignedInteger k = 0; k < getProblem().getInequalityConstraint().getOutputDimension(); ++ k)
       {
-        h[k] = std::max(h[k], 0.0);// convention h(x)>=0 <=> admissibility
+        h[k] = std::min(h[k], 0.0);// convention h(x)>=0 <=> admissibility
```
regislebrun: If we expect a positive error, it should be:
```
h[k] = -std::min(h[k], 0.0);// convention h(x)>=0 <=> admissibility
```
or
```
h[k] = std::max(-h[k], 0.0);// convention h(x)>=0 <=> admissibility
```
no?

jschueller: we have to only take into account the violated constraints (h<0)
then we compute the inf norm of h

jschueller:
included in #899

PR #900, opened 2018-12-04
Some improvements

sofianehaddad:
* Add a specific `DrawCorrelationCoefficients` method to SobolIndicesAlgorithm. Maybe we should consider move this static method (to CorrelationAnalysis class ?)
* Minor fixes : remove deprecated methods, force using Sample(i, j) instead of Sample[i,j]
* Update documentation : using new draw method, fix level for statistical tests


regislebrun:
+1 Not very fond of all these hard-coded constants & attributes in DrawCorrelationCoefficients, but it does the job and it avoids to pollute ResourceMap with too many keys. Maybe the color should be user-defined, as it is a real pain to change it afterward.

PR #909, opened 2019-01-10
Python: Enable 0+Point operation

jschueller:
it enables to write sum(Sample), while still disallowing float+Point operator


sofianehaddad:
@jschueller why do you need this specific radd?

jschueller:
it allows to write sum(Sample) instead of sum(array(Sample)), as sum starts with 0

regislebrun:
Why do you want to write sum(Sample)? It corresponds to a flat view of a Sample, which is not what we decided when we restricted the asPoint() method to work only for 1D sample. I would prefer to write sum(Sample.asPoint()) rather than to have a radd limited to 0!

regislebrun:
or let's decide that one can add a scalar and a sample. But before to go this way we should discuss about the implication all together.

jschueller:
No, it's not the flat sum but the componentwise sum.

regislebrun:
Ok I was refering to your comment `sum(Sample) instead of sum(array(Sample))`, so Sample is a Point ;-)? Could you describe the use-case more precisely, as:
  * `sum(point)` works well if `point` is a `Point`
  * `sum(array(sample))` returns a scalar if `sample` is a `Sample` of dimension 2

Do you expect `sum(sample)` to provide the same result as `sample.computeMean()*sample.getSize()`? If yes, +1 for this PR

jschueller:
No, array(Sample) has the same dim as Sample, so sum(array(sample))=Point with the same dim too (point-wise sum)

So yes basically sum(Sample) returns sample.computeMean()*sample.getSize()

regislebrun:
Ok my bad. With the current master, the following code:
```
import openturns as ot
import numpy as np
sample = ot.Normal(2).getSample(5)
print(sum(np.array(sample)))
```
produces
```
[-1.55578827  2.51975251]
```
while
```
from openturns import *
from numpy import *
sample = Normal(2).getSample(5)
print(sum(array(sample)))
```
produces
```
0.9639642366685628
```
The danger of global imports...

So +1

jschueller:
Oh.... I did not see that sum could come from numpy too,sorry.

This PR is all about the built-in sum operator.

sofianehaddad:
ok for me

PR #999, opened 2019-01-28
Viewer: Dont use matplotlib 2-d input

jschueller:
This changes calls to matplolibs to use 1-d arrays (Point) instead of 2-d arrays (Sample), eg:
```
x = data.getMarginal(0).asPoint()
Axes.pie(x)
```

This works around an issue in matplotlib with latest numpy: https://github.com/matplotlib/matplotlib/issues/13286, cc @pcaclin

### DIFF
--- a/python/src/viewer.py
+++ b/python/src/viewer.py
@@ -299,9 +299,9 @@ class View(object):
 
             # retrieve data
             data = drawable.getData()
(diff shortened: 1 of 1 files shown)


sofianehaddad:
ok for me
+1

regislebrun:
+1

PR #1012, opened 2019-02-12
Ka/python function

KamalAbboud:
Implementation of the multiprocessing mechanism for func_sample argument 
in PythonFunction class.


jschueller:
Hi, you need to rebase against current master for the tests to pass.

PR #1015, opened 2019-02-14
WelchFactory: Boost build(Field)

jschueller:
perf

The use of std::copy to copy around contiguous data from a timeseries yields
a serious x127 speedup (317s->2.5s) in a new testcase and becomes comparable to scipy.

The bottleneck is the FFT decomposition now as expected.


regislebrun:
+1 well done!

PR #1020, opened 2019-02-22
Fix TruncatedDistribution::setBounds

jschueller:
It was not updating (c|p)df(Lower|Upper)Bound_ attributes


jschueller:
cc @aurelieladier

regislebrun:
+1

PR #1026, opened 2019-02-27
Python: Add Distribution unary minus

### LINKED ISSUE #1024: The unary operator is undefined for Normal distribution
The unary operator - is undefined for Normal distribution.

```
import openturns as ot
x = ot.Normal()
y = -x
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-45aa6b0fd8d1> in <module>()
      1 import openturns as ot
      2 x = ot.Normal()
----> 3 y = -x
TypeError: bad operand type for unary -: 'Normal'
```
This sounds weird, since the distribution of y is ... x ! In the general case, this should be ot.Normal(-mu,sigma), isn'it ?

jschueller:
Fixes #1024


regislebrun:
+1

PR #1022, opened 2019-02-24
Fix BIC doc

mbaudin47:
Fixes the description of the KS test so that it matches the implementation.

PR #1032, opened 2019-03-01
Clarification of the Kolmogorov-Smirnov doc.

mbaudin47:
Fix #1022. 
The goal of the update is to clarify the two cases: the parameters are either known or estimated.

### DIFF
--- a/python/doc/theory/data_analysis/kolmogorov_test.rst
+++ b/python/doc/theory/data_analysis/kolmogorov_test.rst
@@ -51,17 +51,21 @@ In general, the threshold p-value is chosen to be 0.05:
     p_{ref} = 0.05
 
 Based on the p-value, 
+
 - if  :math:`p<p_{ref}`, we reject the candidate distribution,
+
 - otherwise, the candidate distribution is not rejected.
 
 Two situations may occur in practice.
+
 - the parameters of the distribution under test are known,
+
 - the parameters of the distribution under test are estimated from data. 
 
 If the parameters of the distribution under test are known, 
 algorithms are available to directly compute 
 the distribution of :math:`\widehat{D}_N` both for *N* large 
-(asymptotic distribution) or for *N* finite. 
+(asymptotic distribution) or for *N* small (exact distribution). 
 This is because the distribution of :math:`\widehat{D}_N` does 
 not depend on the candidate distribution. 
 
--- a/python/doc/theory/data_analysis/kolmogorov_test.rst
+++ b/python/doc/theory/data_analysis/kolmogorov_test.rst
@@ -55,16 +55,18 @@ Based on the p-value,
 - otherwise, the candidate distribution is not rejected.
 
 Two situations may occur in practice.
+- the parameters of the distribution under test are known,
+- the parameters of the distribution under test are estimated from data. 
 
-- The parameters of the distribution under test are known. 
-In this case, algorithms are available to directly compute 
+If the parameters of the distribution under test are known, 
+algorithms are available to directly compute 
 the distribution of :math:`\widehat{D}_N` both for *N* large 
 (asymptotic distribution) or for *N* finite. 
 This is because the distribution of :math:`\widehat{D}_N` does 
 not depend on the candidate distribution. 
-- The parameters of the distribution under test are estimated 
-from data. 
-In this case, the :math:`\widehat{D}_N` is generally smaller, because 
+
+If the parameters of the distribution under test are estimated 
+from data, the :math:`\widehat{D}_N` statistic is generally smaller, because 
 the parameters of the distribution have been computed from the 
 data. 
 In general, the distribution of :math:`\widehat{D}_N` is not known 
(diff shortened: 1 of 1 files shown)


jschueller:
The commit title should reflect the actual modification, and you can still reference the bug number in the rest of the commit message.

jschueller:
Still fails at kolmogorov.rst:60

mbaudin47:
This is true: thanks for watching. I try a fix ; let's see what happens.

python/doc/theory/data_analysis/kolmogorov_test.rst
```diff
@@ -33,28 +33,45 @@ With a sample :math:`\left\{ x_1,x_2,\ldots,x_N \right\}`, the distance is estim
...
+The candidate distribution will not be rejected if and only if 
+:math:`p` is larger than a given threshold probability. 
+In general, the threshold p-value is chosen to be 0.05:
+
+.. math::
+
+    p_{ref} = 0.05
+
+Based on the p-value, 
+- if  :math:`p<p_{ref}`, we reject the candidate distribution,
```
sofianehaddad: I think you should skip a line

python/doc/theory/data_analysis/kolmogorov_test.rst
```diff
@@ -33,28 +33,45 @@ With a sample :math:`\left\{ x_1,x_2,\ldots,x_N \right\}`, the distance is estim
...
+- otherwise, the candidate distribution is not rejected.
+
+Two situations may occur in practice.
+- the parameters of the distribution under test are known,
```
sofianehaddad: same here (sky lines)

python/doc/theory/data_analysis/kolmogorov_test.rst
```diff
@@ -33,28 +33,45 @@ With a sample :math:`\left\{ x_1,x_2,\ldots,x_N \right\}`, the distance is estim
...
+- the parameters of the distribution under test are estimated from data. 
+
+If the parameters of the distribution under test are known, 
+algorithms are available to directly compute 
+the distribution of :math:`\widehat{D}_N` both for *N* large 
+(asymptotic distribution) or for *N* finite.
```
sofianehaddad: For N small, we use the exact computation

python/doc/theory/data_analysis/kolmogorov_test.rst
```diff
@@ -33,28 +33,45 @@ With a sample :math:`\left\{ x_1,x_2,\ldots,x_N \right\}`, the distance is estim
...
+the distribution of :math:`\widehat{D}_N` both for *N* large 
+(asymptotic distribution) or for *N* finite. 
+This is because the distribution of :math:`\widehat{D}_N` does 
+not depend on the candidate distribution. 
+
+If the parameters of the distribution under test are estimated 
+from data, the :math:`\widehat{D}_N` statistic is generally smaller, because 
+the parameters of the distribution have been computed from the 
+data. 
+In general, the distribution of :math:`\widehat{D}_N` is not known
```
sofianehaddad: I would say that the distribution of D_n is lightly differ comparing to D distribution and we should involve Bootstrap methods to better estimate the pvalue

python/doc/theory/data_analysis/kolmogorov_test.rst
mbaudin47: Ok.

python/doc/theory/data_analysis/kolmogorov_test.rst
mbaudin47: OK

python/doc/theory/data_analysis/kolmogorov_test.rst
mbaudin47: I am not sure I clearly understand this specific comment. I suppose that you suggest to describe the algorithm which presumably uses bootstrap, but this is not true because the algorithm does not involve bootstrap. Please look at:

https://github.com/openturns/openturns/blob/master/lib/src/Uncertainty/StatTests/FittingTest.cxx#L272

The algorithm is the following.
* Estimate the parameters of the distribution A based on the sample S. 
* Compute the KS distance D between A and S.
* Get the size N of the sample.
* For i in sampleSize (e.g. sampleSize = 1000)
  * Generate the sample T from distribution A and size N.
  * Estimate the parameters of the distribution B based on the sample T.
  * Compute and store the KS distance between T and B.
* Compute the p-value of the KS distance D within the KS distance sample.

There is no bootstrap here, in the sense that we generate samples from a distribution, not from the sample itself. I seems to me that this algorithm does not fit in the current documentation, but this is just a choice.

sofianehaddad: es, excepting mistakes in the written lines. Btw, global idea is to highlight the difference between distribution's parameters in case these are to be estiamted or are known

jschueller:
see #1079

PR #1037, opened 2019-03-06
 t_coupling_tools: /bin/kill may not exist 

jschueller:
t_coupling_tools may fail on conda-forge docker images where /bin/kill is not available


regislebrun:
+1

PR #1039, opened 2019-03-07
getMarginal-related enhancements

jschueller:
* Now ComposedCopula::getMarginal handles the copulas blocs to be visited
in any order:

```
# test ComposedCopula.getMarginal in reverse
copula = ComposedCopula([IndependentCopula(2), NormalCopula(2)])
print(copula.getMarginal([3, 2, 1, 0]))

ComposedCopula(NormalCopula(R = [[ 1 0 ]
 [ 0 1 ]]), IndependentCopula(dimension = 2))
```

* Also we Implement MarginalDist;::getParametersCollection to handle the non-contiguous case when used in iso-probabilistic transformations:

Else the parameters of the isoprobabilistic transformations involving
MarginalDistribution (from a ComposedCopula with non-contiguous blocs) cannot be retrieved:
```
import openturns as ot
copula = ot.ComposedCopula([ot.IndependentCopula(2), ot.IndependentCopula(2)])
distribution = ot.ComposedDistribution([ot.Normal()]*4, copula)
print(ot.DistributionTransformation(ot.IndependentCopula(4), distribution.getMarginal([3,1,2,0])))
```

### DIFF
--- a/lib/src/Uncertainty/Distribution/ComposedCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/ComposedCopula.cxx
@@ -501,36 +501,51 @@ Distribution ComposedCopula::getMarginal(const Indices & indices) const
   CopulaCollection marginalCopulas;
   const UnsignedInteger indicesSize = indices.getSize();
   const UnsignedInteger size = copulaCollection_.getSize();
-  // For each copula, see if there is something to extract
-  UnsignedInteger currentPosition = 0;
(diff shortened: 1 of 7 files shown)


jschueller:
@sofianehaddad @regislebrun can you have a look a this one ? I'd like to include that fix in the 1.12 maintenance branch for our colleagues at EDF.

regislebrun:
+1 good job. It remains to speed-up the fully interleaved marginals...

PR #1044, opened 2019-03-13
Serialization fixes

jschueller:
Memoize was missing in the catalog.
Also the cache hit count and the enabled flag are now persistent across
serializations.

Also fixes a bunch of serialization issues.

cc @adumasphi


regislebrun:
+1

Issue #1045, opened 2019-03-14
LowDiscrepancyExperiment/ComposedCopula correlation across blocks ?


jschueller:
This example with LowDiscrepancyExperimentseems seems to yield a perfect monotone relashionship between variables from disjoint copula blocs:

```
import openturns as ot
E = ot.Normal()
F = ot.Beta()
L= ot.LogNormal()
I = ot.Uniform()
distribution = ot.ComposedDistribution([E,F,L,I], ot.ComposedCopula([ot.ClaytonCopula(), ot.IndependentCopula(2)]))
experiment = ot.LowDiscrepancyExperiment(ot.SobolSequence(4), distribution, 1000)
#experiment = ot.MonteCarloExperiment(distribution, 1000)
sample = experiment.generate()
print(sample.computeSpearmanCorrelation())


# E/L = 1
[[  1           0.683142    1          -0.00320276 ]
 [  0.683142    1           0.683142    0.663224   ]
 [  1           0.683142    1          -0.00320276 ]
 [ -0.00320276  0.663224   -0.00320276  1          ]]
```

Any thoughts on this @regislebrun ?

regislebrun:
There is a bug in both ComposedCopula::getIsoProbabilisticTransformation() and ComposedCopula::getInverseIsoProbabilisticTransformation() methods. The `shift` variable is not incremented in the loop. I have a PR for that.

PR #1046, opened 2019-03-14
Fixed ComposedCopula

regislebrun:
A small bug in iso-probabilistic transformations of ComposedCopula made them work in a Monte Carlo setting but not in a quasi-Monte Carlo setting.
It fixes the bug #1045


jschueller:
Thanks! You forgot to update the expout.

regislebrun:
Oups, done.

PR #1047, opened 2019-03-15
Fix iso-transformation parameter gradient

jschueller:
I fixed the parameter gradient of the iso-transformation with the current state of marginal/copula parameters in transformation, as it failed on this example that uses the rosenblatt transformation:
```
import openturns as ot
model = ot.SymbolicFunction(['x0', 'x1', 'x2', 'x3'], ['-(6+x0^2-x1+x2+3*x3)'])
dim = model.getInputDimension()
marginals = [ot.Normal(5.0, 3.0) for i in range(dim)]
distribution = ot.ComposedDistribution(marginals, ot.ComposedCopula([ot.ClaytonCopula(), ot.NormalCopula()]))
#distribution = ot.Normal([5]*dim, [3]*dim, ot.CorrelationMatrix(dim))
#distribution = ot.ComposedDistribution(marginals, ot.IndependentCopula(dim))
distribution.setDescription(['marginal'+str(i) for i in range(dim)])
vect = ot.RandomVector(distribution)
output = ot.CompositeRandomVector(model, vect)
event = ot.Event(output, ot.Greater(), 0.0)
solver = ot.Cobyla()
solver.setMaximumEvaluationNumber(200)
solver.setMaximumAbsoluteError(1.0e-10)
solver.setMaximumRelativeError(1.0e-10)
solver.setMaximumResidualError(1.0e-10)
solver.setMaximumConstraintError(1.0e-10)
algo = ot.FORM(solver, event, distribution.getMean())
algo.run()
result = algo.getResult()
hasoferReliabilityIndexSensitivity = result.getHasoferReliabilityIndexSensitivity()
print(hasoferReliabilityIndexSensitivity)
```

### DIFF
--- a/lib/src/Base/Func/AggregatedEvaluation.cxx
+++ b/lib/src/Base/Func/AggregatedEvaluation.cxx
@@ -209,7 +209,7 @@ UnsignedInteger AggregatedEvaluation::getOutputDimension() const
 
 
 /* Gradient according to the marginal parameters */
-Matrix AggregatedEvaluation::parameterGradient(const Point & ) const
+Matrix AggregatedEvaluation::parameterGradient(const Point & inP) const
 {
   Matrix result(getParameter().getDimension(), getOutputDimension());
   const UnsignedInteger size = functionsCollection_.getSize();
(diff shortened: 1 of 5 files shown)


jschueller:
@sofianehaddad @regislebrun hello, could someone review this for wednesday ? I want to merge this real quick for the 1.12 maintenance branch. Sorry for merging too quickly lately, it should be the last bugfix we need for the gui.

regislebrun:
+1 at one time we should find the motivation to *actually* fix the bug ;-)!

Issue #1062, opened 2019-03-28
Unable to save least squares


RRiva:
Dear all,

I'm trying to save LinearLeastSquares and QuadraticLeastSquares objects, but when it comes to loading them, the code returns an error.

Best regards,
Riccardo

```
import openturns as ot
import pickle

# Copied from the guide.
formulas = ['cos(x1 + x2)', '(x2 + 1) * exp(x1 - 2 * x2)']
myFunc = ot.SymbolicFunction(['x1', 'x2'], formulas)
data  = [[0.5,0.5], [-0.5,-0.5], [-0.5,0.5], [0.5,-0.5]]
data += [[0.25,0.25], [-0.25,-0.25], [-0.25,0.25], [0.25,-0.25]]
myLeastSquares = ot.LinearLeastSquares(data, myFunc)
#myLeastSquares = ot.QuadraticLeastSquares(data, myFunc)
myLeastSquares.run()

# Save.
fid = open('lin.pkl', 'wb')
pickle.dump(myLeastSquares, fid)
fid.close()

# Load.
fid = open('lin.pkl', 'rb')
myLeastSquares_2 = pickle.load(fid)
fid.close()
```

jschueller:
Hello, this will be fixed in next version, in the mean time you can serialize the result function: `myLeastSquares.getResponseSurface()`

RRiva:
Thanks a lot! 😊

PR #1063, opened 2019-03-28
Misc enhancements

jschueller:
- really filter the thisown attribute in sphinx
- fixes for #1062, #1064
- fix for cmake python libs detection
- cleanups
- update cotire/exprtk
- drop validation files install


regislebrun:
+1

PR #1069, opened 2019-04-06
Fix two LHS documentation issues

mbaudin47:
* Replaces myPlane by myExperiment in LowDiscrepancyExperiment and LHSExperiment
* Since OptimalLHSExperiment cannot be used directly (this is a base class), added the appropriate "See also"


python/doc/pyplots/LHSExperiment.py
```diff
@@ -6,9 +6,9 @@
 # Generate sample with the given plane
 distribution = ot.ComposedDistribution([ot.Uniform(0, 1)] * 2)
 size = 10
-myPlane = ot.LHSExperiment(distribution, size)
+myExperiment = ot.LHSExperiment(distribution, size)
```
jschueller: just use 'experiment' instead or 'myExperiment'

jschueller:
see #1079

PR #1076, opened 2019-04-10
Improving statistical tests

sofianehaddad:
Now, the `TestResult` class contains a `statistic` attribut. 
As all statistical tests are implemented in `C++`, we can now fill this attribut.

Of course we get a specific accessor, which makes possible a full comparison of statistical tests implemented in the library and outside the platform.

We also fix some tests `LinearModelTest` : we implement all tests only in constructors with `LinearModelResult` (x, y, result,...). As it is generic, the degrees of freedom was wrong (size-dim-1): the underlying model could be different from a simple linear expansion and thus number of parameters is not necessary `dimension+1`


sofianehaddad:
@jschueller @adumasphi I did not update `LinearModelDurbinWatson`, `LinearModelBreuschPagan`, `LinearModelHarrisonMcCabe`. It assumes that `LinearModelResult` correponds to a basic linear regression. Isn't it? should I update these tests and consider a more "general" underlying model ?

lib/src/Uncertainty/StatTests/LinearModelTest.cxx
```diff
@@ -75,9 +75,12 @@ TestResult LinearModelTest::LinearModelFisher(const Sample & firstSample,
   // H0 : Beta_i = 0
   // H1 : Beta_i < 0 or Beta_i > 0
   // The statistics follows a Fisher distribution
-  const Scalar statistic = sumSquaredResiduals / dimension / (sumSquaredExplained / (df));
+  const Scalar numerator = sumSquaredExplained / (size - df - 1);
+  const Scalar denumerator = sumSquaredResiduals / df;
```
regislebrun: denominator

lib/src/Uncertainty/StatTests/LinearModelTest.cxx
```diff
@@ -75,9 +75,12 @@ TestResult LinearModelTest::LinearModelFisher(const Sample & firstSample,
   // H0 : Beta_i = 0
   // H1 : Beta_i < 0 or Beta_i > 0
   // The statistics follows a Fisher distribution
-  const Scalar statistic = sumSquaredResiduals / dimension / (sumSquaredExplained / (df));
+  const Scalar numerator = sumSquaredExplained / (size - df - 1);
```
regislebrun: Maybe a check for size - df - 1 > 0?

ChangeLog
```diff
@@ -26,6 +26,7 @@
  * Deprecated OptimizationProblem(levelFunction, levelValue)
  * Deprecated (Linear|Quadratic)(LeastSquares|Taylor)::getResponseSurface
  * FittingTest::BestModelBIC returns the Distribution and the BIC value
+ * Add statistic attribut and accessor to the TestResult class
```
regislebrun: statistic*s*! It must be changed everywhere, sorry...

python/src/TestResult.i
```diff
@@ -26,11 +26,13 @@ namespace OT {
     PyObject * item_1 = PySequence_Fast_GET_ITEM( newPyObj.get(), 1 ); check<_PyBool_>( item_1 );
     PyObject * item_2 = PySequence_Fast_GET_ITEM( newPyObj.get(), 2 ); check<_PyFloat_>( item_2 );
     PyObject * item_3 = PySequence_Fast_GET_ITEM( newPyObj.get(), 3 ); check<_PyFloat_>( item_3 );
+    PyObject * item_4 = PySequence_Fast_GET_ITEM( newPyObj.get(), 4 ); check<_PyFloat_>( item_3 );
```
adumasphi: Should it be item_4 ?

python/test/t_LinearModelTest_std.expout
```diff
@@ -1,7 +1,7 @@
...
+LinearModelFisher pvalue=0.00
+LinearModelResidualMean pvalue=1.00
+Durbin Watson =  class=TestResult name=Unnamed type=DurbinWatson binaryQualityMeasure=false p-value threshold=0.05 p-value=0.000219316 statistic=3.69566 description=[H0: auto.cor=0]
```
adumasphi: It might be nice to have a test for Durbin Watson with the cases H0:auto.cor > 0 and H0:auto.cor < 0.

lib/src/Uncertainty/StatTests/LinearModelTest.cxx
sofianehaddad: It is done in LinearModelResult

python/src/TestResult.i
sofianehaddad: indeed, thanks

sofianehaddad:
> > @jschueller @adumasphi I did not update `LinearModelDurbinWatson`, `LinearModelBreuschPagan`, `LinearModelHarrisonMcCabe`. It assumes that `LinearModelResult` correponds to a basic linear regression. Isn't it? should I update these tests and consider a more "general" underlying model ?
> 
> Minor changes is fine for as well.

@adumasphi  `LinearModelResult` assumes we solve a least squares problem :

$$ Y = \psi(X) \alpha $$

such as $y \approx \sum_k \alpha_k \psi_k(X)$. If $\psi(X)=(1,x_1,...,x_d)$ we get the usual linear regression. In the current implementation of the statistical tests, we can set any `psi` basis but with a limitation : its size should be dimension+1. Thus the question is the meaning of these tests with a basis that differ from the canonical one ? If we can have any basis we should update the checks! Otherwise IMHO we should remove constructors with a `basis`

ChangeLog
sofianehaddad: indeed...

lib/src/Uncertainty/StatTests/LinearModelTest.cxx
sofianehaddad: fixed, thanks

ChangeLog
jschueller: statistics_ ? I dont see it used at the plural anywhere, https://en.wikipedia.org/wiki/Test_statistic

jschueller:
pyinstallcheck_docstring_metamodel fails because of the rounding

ChangeLog
regislebrun: ok my bad, go for statistic_

regislebrun:
+1

jschueller:
Can you make sure to pass the last failing test for tomorrow ?

sofianehaddad:
@jschueller should be ok now

jschueller:
cool, could you trigger a rebuild either using a commit or through circleci to check that it succeeds on mingw now that the ci issue is fixed ?

jschueller:
nevermind

sofianehaddad:
@jschueller strange that we could not restart a build in circle ?

jschueller:
yes, I can restart mine, but not yours :(

sofianehaddad:
hmmm I can not restart mine too...

PR #1082, opened 2019-04-12
Add Ceres solver

jschueller:
Add Ceres optimization solvers http://ceres-solver.org/
    
Ceres provides trust-region and line-search methods to solve least-squares problems.
    
Line-search methods do not support bounds and can also be used on general problems.
    
No inequality/equality constraint support is available.
    
The implementation was a bliss, only one minor issue reported:
https://github.com/ceres-solver/ceres-solver/issues/470
    
As numerous parameters are used to tweak the different algorithms,
we rely only on the ResourceMap to optionally set them; ie no
attributes/accessors are used.


jschueller:
Oh, and I gave up on ROL, that's why I'm switched to Ceres. DLib is on the way by my colleague.

lib/src/Base/Optim/Ceres.cxx
```diff
@@ -0,0 +1,550 @@
...
+    const Point outP(problem.getResidualFunction()(inP));
+    std::copy(outP.begin(), outP.end(), residuals);
+    algorithm_.evaluationInputHistory_.add(inP);
+    algorithm_.evaluationOutputHistory_.add(Point(1, 0.5 * outP.normSquare()));
+
+    // gradient
+    if (jacobians)
+    {
+      const Matrix gradient(problem.getResidualFunction().gradient(inP));
+      for (UnsignedInteger i = 0; i < n; ++ i)
```
regislebrun: No way to use std::copy here? Or memcopy?

lib/src/Base/Optim/Ceres.cxx
```diff
@@ -0,0 +1,550 @@
...
+    const Point outP(problem.getObjective()(inP));
+    *cost = problem.isMinimization() ? outP[0] : -outP[0];
+    algorithm_.evaluationInputHistory_.add(inP);
+    algorithm_.evaluationOutputHistory_.add(outP);
+
+    // gradient
+    if (jacobian)
+    {
+      const Matrix gradient(problem.getObjective().gradient(inP));
+      for (UnsignedInteger i = 0; i < n; ++ i)
```
regislebrun: std::copy here?

PR #1089, opened 2019-04-18
Add calibration examples

mbaudin47:
This is a gross draft, designed to let the reviews possibles. Please do not review too deeply.


mbaudin47:
This is ready for reviewing.

jschueller:
Each example should preferably be self-contained ; we cannot reuse data nor scripts.

mbaudin47:
Ok: i understand. So i guess that we may discuss how to include (part of) the graphical features within the library. Another option would be to make so that creating the required graphics is easy, which is not the present situation. 

Here is an overview:
1. the prediction and observations versus input plot,
1. the predictions/observations validation plot,
1. the distribution of the parameter theta,
1. the distribution of the residuals.

I will rework these examples so that the smallest possible number of graphics functions is used.

mbaudin47:
In order to solve the graphics issue, i created a dedicated "PostCalibrationResult" class, as a draft of what could become a C++ graphics class. This class has 4 methods:
* drawThetaDistribution
* drawObservationsVsPredictions
* drawResiduals
* drawModelVsObservations (should perhaps be renamed drawObservationsVsInputs)

The class is based only on OT objects (no matplotlib). If the design seems correct, i suggest to convert this into a plain C++ code, so that the NB rely on the C++ class.

jschueller:
I suggest CalibrationAnalysis as the class name, so it's closer to what is done for the linear model.

regislebrun:
The graphics are nice when working on models with scalar input and output, but they don't extend easily to higher dimension. From my point of view, it should be put into a teaching module rather than included into the library, excepted if most of the studies involve such models. BTW, drawThetaDistribution() should be renamed drawParameterDistributions() or drawParameterPriorVsPosterior(). Once again, it will be difficult to extend the method to parameters of dimension greater than 2, and even in the 2D case it will be hard to read...

mbaudin47:
@regislebrun I understand your concerns. I already prepared two different situations
* flooding and Chaboche has a input sample with size 100 and dimension 1 and same for the output with 3 parameters to calibrate
* logistic has a input sample with size 1 and dimension 22 and same for the output with 2 parameters to calibrate
So these 3 examples alread contain some variability in the settings.
However, I am currently preparing a test-case where the input sample will be with size 100 and dimension 2 and the output is a sample with size 100 and dimension 3. Moreover, the three outputs will be of different units. I will then see what happens.

mbaudin47:
The following NB:

https://github.com/openturns/openturns/blob/2cfdacfbc01e190cfcb5d07fd9be9c46e816d548/python/doc/examples/calibration/calibration_deflection_tube.ipynb

shows a generalized calibration with multi-dimensional output, multidimensional input and multi-dimensional parameter. A different (but not more complicated) would appear if we wanted to calibrate two different time series considered a two realizations of a multi-D vector. In this case, the current graphics would have to be "transposed", so that the two times series are  represented on the same graph. This is not an issue anyway and might be done when appropriate.

However, my main concern is that I cannot see how to develop this based on only basic OT graphics. The core issue is the lack of a "subplot" system, which forces the current implementation to be based on Matplotlib. The required component would be a sort of "Pairs" graph (but with correct axes instead of the current implementation which does not show the X and Y axes values). This is way beyond the examples I thought I would create.

I can see three solutions.
* Redo the examples, by removing the graphics parts. This would be a pity, because it would make the whole component much more complicated to grab for users, making the calibration process rather abstract.
* Making a separate Python module which would provide the required graphics features. 
* Implementing the needed "subplot" within OT. This is ultimate. It would, furthermore, allow to modify the Pairs graph. I don't think I can do this by myself.

mbaudin47:
I am waiting for opinions on the option I should select: what solution would you choose?

mbaudin47:
This is good to review:
* I removed the use of .csv files : the data is generated, then used within the NB.
* I use only code within the NB : there is no class outside the NB anymore.

PR #1091, opened 2019-04-21
Various enhancements

regislebrun:
The main contributions are:
+ the ability to provide a global covariance error in BLUE and ThreeDVAR, to cope with situations where the error has a varying covariance or where there is a correlation between the errors
+ the visual test of fit between a distribution and a sample set
+ the ability to draw the survival function for distributions
+ the ability to specify if the discretization has to be done on a linear grid or a logarithmic grid for the various distribution graphs

PR #1093, opened 2019-04-23
Improved ParametericFunction

regislebrun:
Now parameterGradient() is based on the underlying function gradient() instead of finite differences.


jschueller:
oh, and pyinstallcheck_docstring_bayesian seems to fail now

regislebrun:
Done (after 1h of torture with my physiotherapist...). Sorry for the duplicated work...

mbaudin47:
Thank you very much ! After all, torture seems efficient...

jschueller:
Needs a rebase after #1091

PR #1104, opened 2019-04-26
Improve cloning python wrappers

sofianehaddad:
Fix openturns#1099 

 * Fix `PythonDistribution::setParameter` : need to compute the new range
 * Fix  `PythonDistribution::clone` : copy constructor PythonDistribution(other) should duplicate the internal attribut
 * Apply fix to `PythonEvaluation`, `PythonExperiment`, `PythonFieldToFunction`, `PythonGradient`, `PythonHessian`, `PythonRandomVectorImplementation` classes


regislebrun:
+1

Issue #1112, opened 2019-05-09
Several bugs in Multinomial


regislebrun:
The computePDF() method sometimes return wrong values, as well as getSupport() and computeProbability(). These problems are shown by the following script:
```
import openturns as ot

def computeProbability(multinomial, interval):
    support = multinomial.getSupport(interval)
    if support.getSize() == 0:
        return 0.0
    return sum(multinomial.computePDF(support).asPoint())

cases = [[5, [0.25]*3, [0, 1, 2], [1, 2, 3]], [5, [0.25]*4, [0, 1, 2, 0], [1, 2, 3, 5]], [4, [0.5]*2, [1, 2], [2, 3]], [10, [0.2]*5, [1.0]*5, [3.0]*5], [100, [0.25]*4, [24]*4, [28]*4]]
for case in cases:
    print("="*50)
    N = case[0]
    p = case[1]
    a = case[2]
    b = case[3]
    interval = ot.Interval(a, b)
    d_OT = ot.Multinomial(N, p)
    print("p (support)=", computeProbability(d_OT, interval))
    print("p (OT)     =", d_OT.computeProbability(interval))
    print("cdf (support)=", computeProbability(d_OT, ot.Interval([0]*len(b), b)))
    print("cdf (OT)     =", d_OT.computeCDF(b))
```
The values of the CDF are correct, showing that there is are bugs in computePDF() and getSupport(), and the values of p are clearly incorrect.

regislebrun:
The fix is almost there...

PR #1113, opened 2019-05-09
Fix issue #1112 (Several bugs in Multinomial)

regislebrun:


regislebrun:
Any remaining issue here?

Issue #1111, opened 2019-05-09
Function::draw() does not take the scale parameter properly into account


regislebrun:
When specifying a logarithmic scale on X or Y, one get either exceptions (functions from R to R) or wrong results (functions from R^2 to R), as shown by the following script:
```
import openturns as ot

f = ot.SymbolicFunction("x", "x^3")
g = f.draw(1e-3, 1e3, 100, 0)
ot.Show(g)
try:
    g = f.draw(1e-3, 1e3, 100, 1)
    ot.Show(g)
except:
    print("bug scale=1")
try:
    g = f.draw(1e-3, 1e3, 100, 2)
    ot.Show(g)
except:
    print("bug scale=2")
try:
    g = f.draw(1e-3, 1e3, 100, 3)
    ot.Show(g)
except:
    print("bug scale=3")

f = ot.SymbolicFunction(["x", "y"], ["x^2*y^2"])
g = f.draw([1e-1]*2, [1e1]*2, [100]*2, 0)
ot.Show(g)
g = f.draw([1e-1]*2, [1e1]*2, [100]*2, 1)
ot.Show(g)
g = f.draw([1e-1]*2, [1e1]*2, [100]*2, 2)
ot.Show(g)
g = f.draw([1e-1]*2, [1e1]*2, [100]*2, 3)
ot.Show(g)
```

regislebrun:
The fix is almost there...

PR #1114, opened 2019-05-09
Fixed issue #1111 (Function::draw() does not take the scale parameter properly into account)

regislebrun:


jschueller:
needs a rebase

PR #1123, opened 2019-05-14
Update GeneralizedPareto_doc.i.in

adutfoy:


regislebrun:
+1

jschueller:
This definition is correct, why do you want to remove it ? Some comments would help.

PR #1128, opened 2019-05-16
Small fixes in tests

sofianehaddad:


jschueller:
-1 for the first commit, it should be done before branching the release
will be done for the 1.14rc

sofianehaddad:
Ok I will do it later

regislebrun:
+1

PR #1136, opened 2019-05-18
Fix ev3 exponent var node derivation

jschueller:
Fixes the Joe copula formula derivation: "1-((0.5*(1-t))^(-1.0/0.5)+(0.5*t)^(-1.0/0.5))^(-0.5)"


jschueller:
cc @KieranDelamotte

regislebrun:
+1

Issue #1139, opened 2019-05-19
CopulaImplementation should derive from DistributionImplementation


mbaudin47:
This is required for the MarshallOlkin copula:

https://github.com/openturns/openturns/compare/master...mbaudin47:newMarshallOlkinCopula#diff-eca7910e8a05945828d49f6006be96cf

However, this change breaks some unit tests, including Frank copula:

    [ 42%] Building CXX object lib/test/CMakeFiles/t_FrankCopula_std.dir/t_FrankCopula_std.cxx.o
    /home/osboxes/GIT/openturns/lib/test/t_FrankCopula_std.cxx: In function ‘int main(int, char**)’:
    /home/osboxes/GIT/openturns/lib/test/t_FrankCopula_std.cxx:77:65: 
    error:     ‘OT::ContinuousDistribution’ is not a base of ‘OT::FrankCopula’
               << " ddf (FD)=" << copula.ContinuousDistribution::computeDDF(point)
                                                                 ^~~~~~~~~~

jschueller:
just remove the "ContinuousDistribution::"

regislebrun:
Yes I have done the job, but the PR is not yet ready. In the mean time, you can replace ContinuousDistribution by DistributionImplementation in the C++ tests.

Issue #1129, opened 2019-05-16
Segmentation fault with RandomMixture


sofianehaddad:
Hi,
The following scripts make python crash :
```
import openturns as ot
dist = ot.RandomMixture([ot.Uniform()]*200)
dist.computeCDF([0])
```

Issue #1119, opened 2019-05-10
The parameter description is not taken into account in CalibrationResult


regislebrun:
The prior and posterior parameter distributions should get their description from the parameter description of the model when available, which is not the case. The resulting graphs are much less readable without this information. The model seen as a function of the parameter should also be part of the result to allow for a graphical assessment of the calibration.

Issue #1080, opened 2019-04-12
Brent, Secant, Bissection documentation


adutfoy:
Hi

The classes Brent, Secant, Bissection in the API documentation don't give any examples... it should be added.

Anne

josephmure:
This issue was fixed by #1144 and should be closed @adutfoy @jschueller.

sofianehaddad:
Already fixed in #1144

PR #1144, opened 2019-05-23
Various fixes

regislebrun:
This PR proposes various fixes/improvements:
  * StandardDistributionPolynomialFactory is now able to identify the Binomial case
  * Fixed a segfault in HyperbolicAnisotropicEnumerateFunction::inverse() when the evaluation cache is empty. It fixes issue #1129.
  * Replaced remaining uses of double indexing of Sample in a few classes
  * Added an illustration in solver documentation (Brent, Bisection, Secant), it fixes issue #1080 
  * Made CopulaImplementation derives from DistributionImplementation. It fixes issue #1139.
  * Fixed a circular call in RandomMixture. It fixes issue #1129.
  * Improved the simplification mechanism in TruncatedDistribution, fixing an infinite waiting time in RandomMixture
  * Fixed CalibrationResult by providing additional missing information. It fixes issue #1119.
  * Fixed GEV parameter gradient (see issue #1058)


lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
```diff
@@ -81,9 +82,10 @@ GaussianNonLinearCalibration::GaussianNonLinearCalibration(const Function & mode
     algorithm_ = OptimizationAlgorithm::Build(leastSquaresNames[0]);
   else
     algorithm_ = MultiStart(TNC(), LowDiscrepancyExperiment(SobolSequence(), Normal(candidate, CovarianceMatrix(candidate.getDimension())), ResourceMap::GetAsUnsignedInteger("NonLinearLeastSquaresCalibration-MultiStartSize")).generate());
+  parameterPrior_.setDescription(model.getParameterDescription());
 }
 
-namespace ThreeDVARFunctions
+namespace GaussianNonLinearFunctions
```
mbaudin47: Excellent: I also saw this point, but have not published the PR yet.

lib/src/Uncertainty/Algorithm/OrthogonalBasis/StandardDistributionPolynomialFactory.cxx
```diff
@@ -196,6 +196,12 @@ void StandardDistributionPolynomialFactory::checkSpecificFamily()
     referenceFamily = KrawtchoukFactory(static_cast<UnsignedInteger>(parameter[0]), parameter[1]);
     hasClassMatch = true;
   }
+  if (measureType == "Bernoulli")
```
mbaudin47: The "switch" C++ statement would be more appropriate, isn't it ?

regislebrun: Doesn't work with complex types, only simple types (int, char & co)

lib/test/t_FisherSnedecor_std.cxx
```diff
@@ -79,7 +79,7 @@ int main(int, char *[])
     Scalar eps = 1e-5;
     Point DDF = distribution.computeDDF( point );
     fullprint << "ddf     =" << DDF << std::endl;
-    fullprint << "ddf (FD)=" << distribution.ContinuousDistribution::computeDDF(point) << std::endl;
+    fullprint << "ddf (FD)=" << distribution.DistributionImplementation::computeDDF(point) << std::endl;
```
jschueller: It doesn't like this is even needed: the test succeeds with `distribution.computeDDF`

lib/src/Uncertainty/Distribution/RandomMixture.cxx
```diff
@@ -2293,32 +2293,17 @@ Scalar RandomMixture::computeProbability(const Interval & interval) const
...
-    for (UnsignedInteger i = 0; i < gridX.getSize(); ++i)
-      if (interval.contains(gridX[i])) cdf += samplePDF(i, 0) * elementaryVolume;
-    return cdf;
-#endif
+    if (isContinuous()) return computeProbabilityContinuous(interval);
+    // Generic implementation for discrete distributions
+    if (isDiscrete())   return computeProbabilityDiscrete(interval);
+    // Generic implementation for general distributions
+    return computeProbabilityGeneral(interval);
   }
```
jschueller: Does it fix #1129 ? if so the changelog must be updated.

jschueller:
Please add tests for the random mixture fix and enumerate function fixes.

I'm not sure about the LinearFunction enhancement, how about my suggestion to add a Null/ZeroHessian class instead ?

The rest is good to go.

lib/test/t_FisherSnedecor_std.cxx
regislebrun: It was an attempt to embed a validation of the analytical formula (distribution.computeDDF()) and a finite differences approximation (distribution.DistributionImplementation::computeDDF()). I can remove it if you think it is useless.

lib/src/Uncertainty/Distribution/RandomMixture.cxx
regislebrun: Yes it fixes it. I modify the changelog asap.

regislebrun:
Adding a new class looked a little bit overkill to me for the purpose. I wanted to mimic what we do in several places (e.g. Normal with independent components): the trivial case is treated in a specific way without the need of a new class (such as an IndependentMultivariateNormal). But I can add the new class, its tests and its documentation later: we simply drop this commit from the PR.

jschueller:
ok

lib/test/t_FisherSnedecor_std.cxx
jschueller: if you can, use sed as it shows up in about 50 tests

regislebrun:
The failure on VS10 is a timeout:
```
99% tests passed, 1 tests failed out of 467
Total Test time (real) = 287.47 sec
The following tests FAILED:
	213 - pyinstallcheck_BoxCoxFactory_glm (Timeout)
Errors while running CTest
```
Anything else wrong?

PR #1157, opened 2019-06-03
Add Point, Matrix unary minus

### LINKED ISSUE #1153: Unitary minus for points and matrices
```
import openturns as ot
A = ot.Matrix(3,3)
B = -A
```
produces
```
TypeError: bad operand type for unary -: 'Matrix'
```

jschueller:
Closes #1153


mbaudin47:
Thanks ! 
Note that the AppVeyor bug is a "Build execution time has reached the maximum allowed time for your plan (60 minutes).".

regislebrun:
+1

jschueller:
What do you mean any scalar ? element-wise minus/add ?

sofianehaddad:
Yes, element-wise addition/substraction
We should implement **all** or **no one**! We can not provide "partial" operations like unary minus only...

jschueller:
I do not agree : unary minus is an extension of scalar(-1)*vector, you'd want to add vector+-scalar, this is another subject, and we should consider the implications of this, I think there was a trac ticket about some possible downsides.

sofianehaddad:
I missed the fact that Product is already implemented. Thus ok for this PR
Concerning +/- operations (http://trac.openturns.org/ticket/858) we will see later

PR #1165, opened 2019-06-08
Update gaussian linear calibration

mbaudin47:
This updates the implementation of gaussian linear calibration, in order to use a more stable code based on Cholesky decomposition combined with linear least squares. 
Fixes #1165


lib/src/Uncertainty/Bayesian/GaussianLinearCalibration.cxx
```diff
@@ -113,33 +113,54 @@ GaussianLinearCalibration::GaussianLinearCalibration(const Sample & modelObserva
 /* Performs the actual computation. Must be overloaded by the actual calibration algorithm */
 void GaussianLinearCalibration::run()
 {
+  // Compute the difference of output observations and output predictions
   const Point deltaY(modelObservations_.getImplementation()->getData() - outputObservations_.getImplementation()->getData());
+  // Compute covariance matrix B of the parameter and its Cholesky decomposition
   CovarianceMatrix B(getParameterCovariance());
-  const IdentityMatrix IB(B.getDimension());
-  const CovarianceMatrix invB(B.solveLinearSystem(IB).getImplementation());
+  const TriangularMatrix LB(B.computeCholesky());
```
regislebrun: You should not do that. Get the Cholesky factors or its inverse from the prior distribution, see GaussianNonLinearCalibration::run(). The computation has already been done when creating the prior distribution.

lib/src/Uncertainty/Bayesian/GaussianLinearCalibration.cxx
```diff
@@ -113,33 +113,54 @@ GaussianLinearCalibration::GaussianLinearCalibration(const Sample & modelObserva
...
 	{
-	  const UnsignedInteger size = outputObservations_.getSize();
 	  for (UnsignedInteger i = 0; i < size; ++i)
 	    for (UnsignedInteger j = 0; j < dimension; ++j)
 	      for (UnsignedInteger k = 0; k < dimension; ++k)
 		R(i * dimension + j, i * dimension + k) = errorCovariance_(j, k);
 	}
     }
-  const IdentityMatrix IR(R.getDimension());
-  const CovarianceMatrix invR(R.solveLinearSystem(IR).getImplementation());
```
regislebrun: Same thing here: first, create the error distribution, then extract the relevant matrices.

lib/src/Uncertainty/Bayesian/GaussianLinearCalibration.cxx
```diff
@@ -113,33 +113,54 @@ GaussianLinearCalibration::GaussianLinearCalibration(const Sample & modelObserva
...
-  const Point thetaStar(getCandidate() - K * deltaY);
-  const SquareMatrix L((IB - K * M).getImplementation());
-  const CovarianceMatrix covarianceThetaStar((K * R * K.transpose() + L * B * L.transpose()).getImplementation());
+  // Compute the Cholesky decomposition of R
+  const TriangularMatrix LR(R.computeCholesky());
+  // Compute inv(LB), the first part of the extended design matrix
+  const IdentityMatrix ILB(LB.getDimension());
+  const Matrix invLB(LB.solveLinearSystem(ILB).getImplementation());
+  // Compute inv(LR)*J, the second part of the extended design matrix
+  const CovarianceMatrix invLRJ(LR.solveLinearSystem(gradientObservations_).getImplementation());
```
regislebrun: Here and above, the inversion of the Cholesky factors have been done in the corresponding Normal distributions.

mbaudin47:
This is not ready for a review: the numerical results are different.

mbaudin47:
@regislebrun : thanks for the excellent advices. The current implementation takes your suggestions into account.

The missing part is the output dimension, which is supposed to be multidimensionnal in the general case, and in the unit tests. The current implementation only manages one single output, and this is why the computations fails. I work on this topic to fix the code.

sofianehaddad:
@jschueller your PR contains also some commits from master. You should rebase your branch and drop useless commits

mbaudin47:
After extensive testing, I updated the reference values of the unit test for the linear gaussian calibration and here is why. The details of the numerical values are in the NB:

https://github.com/mbaudin47/pycrash/blob/master/calibration/calibration_unittest_LinearGaussian.ipynb

and this message references the cells in the NB.

* I found that the unit test was poorly chosen in the sense that the condition number of the Jacobian matrix is *very* large and roughly equal to 3.677e5 (cell [65]).

The model here has two outputs - the ill conditioning comes from the first output "a + b * exp(c * x)". The parameter a is especially difficult to estimate, as shown by a large confidence interval (or, equivalently, by a large covariance) ; the 95% marginal interval are (cell [16]):

a: [4.45374, 6.92998]
b: [0.076765, 0.0896614]
c: [0.991617, 0.992986]

* Moreover, the previous implementation was based on the inverse of the Kalman matrix, which condition number is close to 8.494e17, i.e. the worst possible value (cell [65]). This is generated by the matrices inverted in the naive Kalman implementation.

* The robust new implementation uses a matrix which condition number is 4.363e5 (cell [67]). This explains why the new implementation has a greater accuracy.

* I also compared the corresponding cost function value. The old implementation leads to the parameter value [0.998045,0.0933438,0.991264], which is associated to a cost equal to 67.78 (cell [50]). The new implementation leads to the parameter value [5.69186,0.0832132,0.992301] , which is associated to a cost equal to 32.05 (cell [52]). This implies that the better estimate of the MAP also reduces the cost function value.

* This however, was not very reassuring: it would be more comfortable to get exactly the same results with both old and new implementation on some test case. This is the reason why i created another test case, linear, for which the Jacobian matrix is not so ill-conditionned.

https://github.com/mbaudin47/pycrash/blob/master/calibration/calibration_linear_bidim_LinearGaussian.ipynb

For this test case, both the old and the new implementation produce the parameter [11.0496,6.77479,-9.11727]. This is because the Kalman matrix has here a condition number equal to 1.428, which is close to being perfect. Moreover, to be *really* sure, i compared the outputs with respect to the nonlinear algorithm GaussianNonLinearCalibration, which produces the same estimate (but not the same covariance because the gradient is estimated at the optimum, instead of the candidate). 

All in all, I trust the numerical outputs of the updated code more than the old one.

mbaudin47:
The unit test fails on circleci because of "ModuleNotFoundError: No module named 'numpy'". This is because I used numpy to check the value of the covariance matrix. I did that because the assert_almost_equal function does not allow for OT's own covariance matrices. How may I fix the issue ?

jschueller:
It should not be too hard to overload assert_almost_equal for matrix, see lib/include/openturns/OTtestcode.hxx, you can basically copy-paste the code for Sample.

mbaudin47:
Given the conflicting files, I created the PR #1175 which is much cleaner and simpler. This is why I plan to close the PR #1165 if this can be accepted.

PR #1174, opened 2019-06-17
Pareto and estimation features

jschueller:
* Add GPD location parameter u
* Add Pareto (specific case of GPD)
* Add ParetoFactory (mle/lsq/moments methods)
* Add LogNormal least-squares estimator
* Add generic LeastSquaresDistributionFactory, works well for Pareto
* Now MethodOfMomentFactory uses the LeastSquaresProblem API
* New OptimizationAlgorithm apis to list available solvers for a given problem


jschueller:
@sofianehaddad I'd like to merge this one, can you have a look ?

lib/src/Uncertainty/Distribution/GeneralizedPareto.cxx
```diff
@@ -368,12 +376,32 @@ Scalar GeneralizedPareto::getXi() const
...
 }
 
+/* U accessor */
+void GeneralizedPareto::setU(const Scalar u)
+{
+  if (u != u_)
+  {
+    u_ = u;
+    isAlreadyComputedMean_ = false;
+    isAlreadyComputedCovariance_ = false;
```
sofianehaddad: hmm covariance is invariant with translating data. Is there any reason why you set it to `false` here?

jschueller: I guess it was just a copy-paste

lib/src/Uncertainty/Distribution/LeastSquaresFactory.cxx
```diff
@@ -0,0 +1,414 @@
...
+    UnsignedInteger unknownParameterSize = unknownParameterIndices_.getSize();
+    for (UnsignedInteger j = 0; j < unknownParameterSize; ++ j)
+    {
+      effectiveParameter[unknownParameterIndices_[j]] = parameter[j];
+    }
+    // set known values
+    UnsignedInteger knownParametersSize = knownParameterIndices_.getSize();
+    for (UnsignedInteger j = 0; j < knownParametersSize; ++ j)
+    {
+      effectiveParameter[knownParameterIndices_[j]] = knownParameterValues_[j];
```
sofianehaddad: Is it necessary ?
IMHO we can do it in ctor (and once) as know parameters (with values) are not supposed to change

lib/src/Uncertainty/Distribution/LeastSquaresFactory.cxx
```diff
@@ -0,0 +1,414 @@
...
+    }
+    try
+    {
+      distribution.setParameter(effectiveParameter);
+    }
+    catch (Exception &)
+    {
+      return Point(getOutputDimension(), SpecFunc::LogMinScalar);
```
sofianehaddad: why 'Min'?
If we want to minimize residuals, maybe we should set value to 'Max'

lib/src/Uncertainty/Distribution/LeastSquaresFactory.cxx
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    Point result(getOutputDimension());
+
+    // compute residual of the CDF distribution vs sample
+    const Sample cdfSample = distribution.computeCDF(sample_);
```
sofianehaddad: please use () instead of =

lib/src/Uncertainty/Distribution/LeastSquaresFactory.cxx
```diff
@@ -0,0 +1,414 @@
...
+    UnsignedInteger unknownParameterSize = unknownParameterIndices_.getSize();
+    for (UnsignedInteger j = 0; j < unknownParameterSize; ++ j)
+    {
+      effectiveParameter[unknownParameterIndices_[j]] = parameter[j];
+    }
+    // set known values
+    UnsignedInteger knownParametersSize = knownParameterIndices_.getSize();
+    for (UnsignedInteger j = 0; j < knownParametersSize; ++ j)
+    {
+      effectiveParameter[knownParameterIndices_[j]] = knownParameterValues_[j];
```
sofianehaddad: same remark as for `Evaluation`

lib/src/Uncertainty/Distribution/LeastSquaresFactory.cxx
```diff
@@ -0,0 +1,414 @@
...
+    }
+    distribution.setParameter(effectiveParameter);
+    // Matrix result
+    MatrixImplementation result(parameter.getSize(), getOutputDimension());
+    // Evaluate the gradient
+    const Sample logPdfGradientSample(distribution.computeLogPDFGradient(sample_).getMarginal(unknownParameterIndices_));
+    const Point logPdfGradient(logPdfGradientSample.computeMean());
+
+    return result;
```
sofianehaddad: something goes wrong here.
`result` is a `Matrix` with `0` as did not fill it

lib/src/Uncertainty/Distribution/Pareto.cxx
```diff
@@ -0,0 +1,382 @@
...
+  return result;
+}
+
+
+/* Get the PDF of the distribution */
+Scalar Pareto::computePDF(const Point & point) const
+{
+  if (point.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
+  const Scalar x = point[0] - gamma_;
+  if (x <= 0.0) return 0.0;
```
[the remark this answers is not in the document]
jschueller: yes, it's, x>=gamma+beta

lib/src/Uncertainty/Distribution/LogNormalFactory.cxx
```diff
@@ -317,4 +322,121 @@ LogNormal LogNormalFactory::buildAsLogNormal() const
...
+  }
+
+private:
+  Sample sample_;
+  Sample dataOut_;
+};
+
+
+LogNormal LogNormalFactory::buildMethodOfLeastSquares(const Sample & sample) const
+{
```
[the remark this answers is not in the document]
jschueller: In this case it is analytical, so this is more efficient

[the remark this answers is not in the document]
sofianehaddad: Maybe but in `Pareto`, even if it is analytical, the code relies on the `LeastSquaresDistributionFactory`.  So why this difference ?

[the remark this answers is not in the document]
jschueller: I guess it was because of the change of variables: here I wanted to write the change of variables as in the formula, in pareto no need to change variables, so the LeastSquaresDistributionFactory is directly usable.

lib/src/Uncertainty/Distribution/LogNormalFactory.cxx
```diff
@@ -317,4 +322,121 @@ LogNormal LogNormalFactory::buildAsLogNormal() const
...
+    LinearLeastSquares leastSquares(dataIn, dataOut_);
+    leastSquares.run();
+    const Scalar a0 = leastSquares.getConstant()[0];
+    const Scalar a1 = leastSquares.getLinear()(0, 0);
+    const Scalar sigmaLog = 1.0 / a1;
+    const Scalar muLog = -a0 * sigmaLog;
+    Point result(size);
+    for (UnsignedInteger i = 0; i < size; ++ i)
+    {
+      result[i] = dataOut_(i, 0) - (dataIn(i, 0) / sigmaLog - muLog/sigmaLog);
```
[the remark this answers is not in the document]
jschueller: Why not for the global residual (float, sum of residuals), but here we need the residual on each output, so I'd be more reluctant to add it in the public interface.

lib/src/Uncertainty/Distribution/ParetoFactory.cxx
```diff
@@ -0,0 +1,120 @@
...
+
+Pareto ParetoFactory::buildMethodOfLikelihoodMaximization(const Sample & sample) const
+{
+  const MaximumLikelihoodFactory factory(buildMethodOfMoments(sample));
+  return buildAsPareto(factory.build(sample).getParameter());
+}
+
+Pareto ParetoFactory::buildMethodOfLeastSquares(const Sample & sample) const
+{
+  const LeastSquaresFactory factory(buildMethodOfMoments(sample));
```
sofianehaddad: IMHO we should have similar procedure as for `LogNormalFactory:: buildMethodOfLeastSquares`

For a fixed `gamma`, we derive `alpha` and `beta` coefficients. Residual is difference between empirical `log-survival function` and one issued from distribution (linear regression)

lib/src/Uncertainty/Distribution/ParetoFactory.cxx
```diff
@@ -0,0 +1,120 @@
...
+{
+  const MaximumLikelihoodFactory factory(buildMethodOfMoments(sample));
+  return buildAsPareto(factory.build(sample).getParameter());
+}
+
+Pareto ParetoFactory::buildMethodOfLeastSquares(const Sample & sample) const
+{
+  const LeastSquaresDistributionFactory factory(buildMethodOfMoments(sample));
+  return buildAsPareto(factory.build(sample).getParameter());
+}
```
sofianehaddad: IMHO we should have similar procedure as for `LogNormalFactory:: buildMethodOfLeastSquares`

For a fixed `gamma`, we derive `alpha` and `beta` coefficients. Residual is difference between empirical `log-survival function` and one issued from distribution (linear regression)

jschueller: why not, maybe later on

sofianehaddad: ok, otherwise seems ok for me

jschueller: cool thanks

PR #1181, opened 2019-06-28
Fejer1 creation

M2CI-ONERA:
Development of Fejer1 during OpenTURNS formation days


jschueller:
thanks!

jschueller:
moved to #1548, thanks!

PR #1182, opened 2019-06-29
Fixed Dirac covariance

adutfoy:
The covariance was computed as the identity matrix instead of the null matrix. The Dirac distribution is now better integrated with distribution algebra.


jschueller:
pyinstallcheck_docstring_dist_bundle3 is failing

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -4899,6 +4899,11 @@ Distribution DistributionImplementation::sqr() const
       Point parameters(getParameter());
       return new ChiSquare(parameters[0]);
     }
+  if (getClassName() == "Normal")
+    {
+      Point parameters(getParameter());
```
sofianehaddad: const

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -305,9 +294,14 @@ Distribution DistributionImplementation::operator * (const Scalar value) const
   if (dimension_ != 1) throw NotYetImplementedException(HERE) << "In DistributionImplementation::operator * (const Scalar value) const: can multiply by a constant 1D distributions only.";
   if (value == 0.0) return new Dirac(Point(1, 0.0));
   if (value == 1.0) return clone();
+  if (getClassName() == "Dirac") return new Dirac(getRealization()[0] * value);
   const Collection< Distribution > coll(1, *this);
   const Point weight(1, value);
-  return new RandomMixture(coll, weight);
```
sofianehaddad: we might check `value` here and return `*this` in case it equals `1`

lib/src/Uncertainty/Model/DistributionImplementation.cxx
regislebrun: done, thx

lib/src/Uncertainty/Model/DistributionImplementation.cxx
regislebrun: I can't find this line in the code...

jschueller: its done just below that, note that its simpler to just check `value` instead of building the mixture:
```
if (value == 1.0)
  return *this;
```

regislebrun: Please note that the special cases Distribution*0 and Distribution*1 are checked at the very beginning of the method (return clone() instead of *this, can be changed if needed). The *key point* is that even if value is not equal to 0 or 1, it can be simplified in the RandomMixture eg by being integrated into the mean/standard deviation of a Normal: 2*Normal(1,1)=Normal(2,2), which is done in the RandomMixture and is detected because there is a unique atom with a unit weight after the simplification process!

jschueller: ok, thanks

PR #1183, opened 2019-07-05
Various enhancements

jschueller:
- Segfault in BoundingVolumeHierarchy with a single simplex
- NLopt tweaks (drop nocedal, all global algos require finite bounds)
- Python: Add `Domain.__contains__`: to use the syntax "if x in domain:"
- Defer alias initialization in UserDefined : it is better to not use the randomgenerator in the initialization of objects (they may be called in xml serialization..., or instanciating new intermediate object may break tests)
- In Hypergeometric the alias method attributes were not used
- Fix ANCOVA for sparse decomposition
- Add UserDefined::setParameter
- Allow KernelSmoothing to return a Dirac when size=1 (avoids throwing in non linear calibration when bootstrapsize==1)
- Add ResourceMap::Reload()


lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -716,7 +717,7 @@ void ResourceMap::loadDefaultConfiguration()
   // KarhunenLoeveQuadratureAlgorithm parameters //
   setAsScalar( "KarhunenLoeveQuadratureAlgorithm-RegularizationFactor", 0.0);
 
-  // KarhunenLoeveSVDAlgorithm parameters //
+  // KarhunenLoeveSVDAlgorloadConfigurationFile()ithm parameters //
```
regislebrun: Typo here

lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -961,7 +962,7 @@ void ResourceMap::loadDefaultConfiguration()
   setAsScalar( "TruncatedDistribution-DefaultThresholdRealization", 0.5 );
 
   // TruncatedNormalFactory parameters //
-  setAsScalar( "TruncatedNormalFactory-SigmaLowerBound", 1.0e-4 );
+  setAsScalar( "TmapBool_runcatedNormalFactory-SigmaLowerBound", 1.0e-4 );
```
regislebrun: and here

lib/src/Base/Common/ResourceMap.cxx
jschueller: woaw... I dont know what happened, looks like I commited a bit too fast, thanks

PR #1189, opened 2019-07-19
Improved Mixture::getRealization()

regislebrun:
Now it is based on DistFunc::rDiscrete() instead of UserDefined::getRealization()

With master, a small benchmark gives:
```
Uniform weights, speed=4.47e+06 rng/s
Non-uniform weights, speed=4.25e+06 rng/s
```
and with this PR:
```
Uniform weights, speed=8.63e+06 rng/s
Non-uniform weights, speed=7.70e+06 rng/s
```
ie the performance is almost doubled. The benchmark code is:
```
import openturns as ot
from time import time

n = 1000
size = 10000000
p = ot.RandomGenerator.Generate(n)

ot.RandomGenerator.SetSeed(0)
mixture = ot.Mixture([ot.Normal()]*n)
t0 = time()
res = mixture.getSample(size)
t1 = time()
print("Uniform weights, speed=", size / (t1 - t0), "rng/s")

ot.RandomGenerator.SetSeed(0)
mixture = ot.Mixture([ot.Normal()]*n, p)
t0 = time()
res = mixture.getSample(size)
t1 = time()
print("Non-uniform weights, speed=", size / (t1 - t0), "rng/s")
```


jschueller:
I thinks its better not to initialize the alias in the ctor because it uses the RNG, see:
https://github.com/openturns/openturns/pull/1183

regislebrun:
Fixed the call to RandomGenerator in the constructor

lib/src/Uncertainty/Distribution/Mixture.cxx
```diff
@@ -839,7 +843,10 @@ void Mixture::save(Advocate & adv) const
 {
   DistributionImplementation::save(adv);
   adv.saveAttribute( "distributionCollection_", distributionCollection_ );
-  adv.saveAttribute( "weightsDistribution_", weightsDistribution_ );
+  adv.saveAttribute( "base_", base_ );
+  adv.saveAttribute( "alias_", alias_ );
```
jschueller: I wouldnt save base_/alias_ attributes, they can be regenerated.

PR #1191, opened 2019-07-25
[RFC] Improved KernelMixture::computePDF() performances

regislebrun:
There are two main bottlenecks:
  * The first one is due to the creation and destruction of Counter at each call to getImplementation() for interface objects.
  * The second one is due to the creation and destruction of Point during the call to kernel_.computePDF()

A more generic answer should be provided (hence the RFC), but this one already makes a speedup ranging from 9x to 18x!
 
The benchmark I use is:
```
from openturns import *

ResourceMap.SetAsUnsignedInteger( "Distribution-MinimumVolumeLevelSetSamplingSize", 20000 )

ref = Mixture([Normal(2), Normal([1.0]*2, [0.5]*2, IdentityMatrix(2))])
alpha = 0.03
level_ref, beta_ref = ref.computeMinimumVolumeLevelSetWithThreshold(1.0 - alpha)
size = 10000
data = ref.getSample(size)
from time import time
for kernel in [Uniform(), Epanechnikov(), Triangular(), Normal()]:
    print("kernel=", kernel)
    t0 = time()
    reconstructed = KernelSmoothing(kernel).build(data)
    level_rec, beta_rec = reconstructed.computeMinimumVolumeLevelSetWithThreshold(1.0 - alpha)
    t1 = time()
    print("t=%.2f" % (t1 - t0), "s")
```
With master, the results are:
```
kernel= Uniform(a = -1, b = 1)
t=14.79 s
kernel= Epanechnikov()
t=15.07 s
kernel= Triangular(a = -1, m = 0, b = 1)
t=14.83 s
kernel= Normal(mu = 0, sigma = 1)
t=21.89 s
```
and with this PR:
```
kernel= Uniform(a = -1, b = 1)
t=0.80 s
kernel= Epanechnikov()
t=0.82 s
kernel= Triangular(a = -1, m = 0, b = 1)
t=0.90 s
kernel= Normal(mu = 0, sigma = 1)
t=2.23 s
```
I also provide two flame graphs to illustrate the bottleneck
perf_master
perf_PR


lib/src/Uncertainty/Distribution/KernelMixture.cxx
```diff
@@ -41,7 +41,7 @@ static const Factory<KernelMixture> Factory_KernelMixture;
 /* Default constructor */
 KernelMixture::KernelMixture()
   : ContinuousDistribution()
-  , kernel_()
+  , p_kernel_()
```
jschueller: I guess this should not be left uninitialized, maybe `p_kernel_(Distribution().getImplementation())`

sofianehaddad: indeed

sofianehaddad:
Good idea, +1

jschueller:
Did you benchmark separately the impact of interface ojects and Point creation ?

By relying on std::shared_ptr alone I'm already shaving 8-10% on your benchmark.

So if the impact of the interface is the same, it would be more interesting to not do it specifically here but globally.

FYI, my patch in Pointer.hxx:
```
diff --git a/lib/src/Base/Common/openturns/Pointer.hxx b/lib/src/Base/Common/openturns/Pointer.hxx
index ef35c7aae..b66db88c6 100644
--- a/lib/src/Base/Common/openturns/Pointer.hxx
+++ b/lib/src/Base/Common/openturns/Pointer.hxx
@@ -23,6 +23,7 @@
 
 #include "openturns/OTprivate.hxx"
 #include "openturns/SharedPointer.hxx"
+#include <memory>
 
 BEGIN_NAMESPACE_OPENTURNS
 
@@ -65,7 +66,7 @@ public:
 #ifdef SWIG
   typedef SharedPointer<T> pointer_type;
 #else
-  typedef IMPL<T> pointer_type;
+  typedef std::shared_ptr<T> pointer_type;
 #endif
 
   /**
@@ -136,7 +137,9 @@ public:
   template <class Base>
   Pointer & assign(const Pointer<Base> & ref)
   {
-    ptr_.assign(ref.ptr_);
+//     ptr_.assign(ref.ptr_);
+    std::cout << "Pointer::assign"<<std::endl;
+//     ptr_ = std::make_shared(ref);
     return *this;
   }
```

regislebrun:
@jschueller Hi I am back to business. I did some additional benchmarks. With this branch and the old shared pointers, I get:
```
kernel= Uniform(a = -1, b = 1)
t=0.81 s
kernel= Epanechnikov()
t=0.83 s
kernel= Triangular(a = -1, m = 0, b = 1)
t=0.93 s
kernel= Normal(mu = 0, sigma = 1)
t=2.26 s
```
With the current master I get:
```
kernel= Uniform(a = -1, b = 1)
t=12.38 s
kernel= Epanechnikov()
t=13.43 s
kernel= Triangular(a = -1, m = 0, b = 1)
t=14.41 s
kernel= Normal(mu = 0, sigma = 1)
t=17.97 s
```
With both I get:
```
kernel= Uniform(a = -1, b = 1)
t=0.83 s
kernel= Epanechnikov()
t=0.82 s
kernel= Triangular(a = -1, m = 0, b = 1)
t=0.94 s
kernel= Normal(mu = 0, sigma = 1)
t=2.23 s
```
So the main improvement is mainly due to the avoidance of the Counter manipulation, and this PR leads to a significant improvement over master.

jschueller:
ok, I expected a bit more magic from the cxx11 changes thought : |

lib/src/Uncertainty/Distribution/KernelMixtureTmp.cxx
```diff
@@ -0,0 +1,928 @@
+//                                               -*- C++ -*-
```
jschueller: I assume this source file was added accidentally, please remove it

regislebrun: Done

regislebrun:
I made additional speed improvements. Consider the following benchmark:
```
import openturns as ot
from time import time

size = 1000
d1 = ot.Normal()
d2 = ot.Normal()
p = d1 * d2
s = p.getSample(size)
t0 = time()
res = p.computePDF(s)
t1 = time()
print("Speed=", size / (t1 - t0), "pdf/s")
```
With master I get:
```
Speed= 5336.597115980623 pdf/s
```
By replacing Distribution by Pointer<DistributionImplementation> I get:
```
Speed= 6086.90253595881 pdf/s
```
Then by implementing the PDF kernel as a Function instead of a bound method I get:
```
 Speed= 9663.771662133502 pdf/s
```
There are still a lot of Point created/destroyed. I extended the GaussKronrod class to work with UniVariateFunction models and implemented the PDF kernel as a UniVariateFunction, and then I got:
```
Speed= 118513.26085353314 pdf/s
```
so a final speed-up of 22x!
Replacing Normal() by Beta(), master gives:
```
Speed= 10082.341762180316 pdf/s
```
and the various improvements already described give:
```
Speed= 37257.71140530983 pdf/s
```
so a speed-up of 3.7, not bad but not so impressive. It is due to the fact that there is no specific implementation of Scalar computePDF(const Scalar scalar), so the algorithm calls Scalar computePDF(const Point & point), which creates a lot of temporary objects. I implemented the method and got:
```
Speed= 89404.7967078228 pdf/s
```
now the speed-up is nearly 9x.

IMHO, the following points should be addressed:
- every distribution with 1D version (either pure 1D such as Gamma, or with an 1D version such as Dirichlet) should have specific implementations of computePDF(), computeLogPDF(), computeCDF(), computeComplementaryCDF() based on scalar arguments
- the integration algorithms and the 1D equation solvers should work on UniVariateFunction objects
- Point should *not* be persistent (maybe a new PersistentPoint class?) to avoid costly Counter creation

jschueller:
yes, the idfactory definetely slows things down for us on such cases, and here again I was hoping my changes in AtomitInc using c++11's std::atomic would help a bit : |

jschueller:
in any case if you want this to be merged you should first rebase to solve the optpp error

regislebrun:
Good to go?

jschueller:
yes

PR #1193, opened 2019-07-31
Chaospy-support for multivariate distributions

jonathf:
Extend Chaospy distribution support to include stochastically independent multivariate distributions.


jonathf:
@jschueller, yes, thank you. I will add tests before submitting the PR for review.

Side note, the doctest example is commented out. Can I assume this is because doctests are preferred to be run without depending on chaospy being installed? If so, is there a reason the formulation isn't on the more canonical format:

    >>> import chaospy as cp  # doctest: +SKIP

jschueller:
Yes, we could use that syntax, thanks for pointing that out

jonathf:
@regislebrun,

I forgot to thank you for adding support for chaospy to OpenTURNS. So thank you. It is appreciated. Following up (a year later), I will try expand on your example to support multivariate distributions.

jschueller:
you will need to expand the range argument and check if cdf is implemented:
```
def __init__(self, dist):
        super(ChaospyDistribution, self).__init__(len(dist))
        if len(dist)>1:
            raise Exception("Multivariate chaospy don't implement CDF computation")
        self._dist = dist
        bounds = dist.range()
        self.__range = Interval(bounds[0], bounds[1])
```

python/src/Distribution.i
```diff
@@ -156,21 +156,27 @@ class ChaospyDistribution(PythonDistribution):
...
+        from chaospy import Iid, J, get_dependencies
+        independent = len(dist) == 1
+        independent |= isinstance(dist, Iid)
+        independent |= isinstance(dist, J) and not get_dependencies(*dist)
+        if not independent:
+            raise Exception(
+                "Dependent chaospy distributions doesn't implement CDF computation")
         self._dist = dist
         bounds = dist.range()
         self.__range = Interval(bounds[0], bounds[1])
```
jschueller: you will need to extend the range attribute

jonathf: I am not following.

`dist.range()` returns a numpy array of shape `(2, len(dist))`.
Looks like `Interval` takes lists as input, but accepts the numpy array of shape `(1,)` in the current implementation. I imagine that is also the case for shape `(len(dist),)`, which is what will be the case if the code is unmodified.

Do you mind explaining what about the range attribute that needs to be extended?

jschueller: ok, I guess it works then

python/test/t_Distribution_chaospy.py
```diff
@@ -37,3 +35,6 @@
     print('kurtosis=', distribution.getKurtosis())
     print('kurtosis(sampling)=', sample.computeKurtosis())
     print('range=', distribution.getRange())
+    print('scalar quantile=%.6g' % distribution.computeScalarQuantile(cdf))
```
jschueller: ChaospyDistribution.computeScalarQuantile should raise if dim>1

python/src/Distribution.i
```diff
@@ -156,21 +156,27 @@ class ChaospyDistribution(PythonDistribution):
...
     --------
     >>> import openturns as ot
-    >>> # import chaospy as cp
-    >>> # chaospy_dist = st.Triangular(1.0, 2.0, 3.0)
-    >>> # distribution = ot.Distribution(ot.ChaospyDistribution(chaospy_dist))
-    >>> # distribution.getRealization()
+    >>> import chaospy as cp  # doctest: +SKIP
+    >>> chaospy_dist = cp.J(cp.Triangular(1.0, 2.0, 3.0), cp.F(4.0, 5.0))  # doctest: SKIP
+    >>> distribution = ot.Distribution(ot.ChaospyDistribution(chaospy_dist))  # doctest: SKIP
+    >>> distribution.getRealization()  # doctest: SKIP
```
jschueller: isnt it "+SKIP" ? (with a +)

jonathf: Yes it is. Thanks.

jschueller:
thanks!

PR #1206, opened 2019-08-31
New HistogramFactory build

### LINKED ISSUE #1205: Histogram from its classes
The current API does not provide a way to create an histogram from its classes. 
Indeed, according to:

https://github.com/openturns/openturns/blob/a9346d73f4f0b3353ce50e2c0726eaab9b37696b/lib/src/Uncertainty/Distribution/HistogramFactory.cxx#L49

the current API does provide the following constructors: 
```
ot.HistogramFactory().build(sample)
ot.HistogramFactory().build(sample,binNumber)
ot.HistogramFactory().build(sample,bandwidth)
```
Notice that this constructors are undocumented:

http://openturns.github.io/openturns/master/user_manual/_generated/openturns.HistogramFactory.html

There is no way to create an histogram from the classes with the following constructor:
```
ot.HistogramFactory().build(sample,first,width)
```
where `first` is the left value of the first class and `width` is a `Point` which contains the widths of the classes. This is bizarre, since the first and width parameters can be obtained from an existing histogram. Hence, we cannot create a new histogram given the data that are actually used to define it!
Here is a use-case for the new constructor. It allows to first create an histogram, get its (first,width) parameters, performs selections in the data, then compute the histogram with the new data using the old classes: this way, the classes of the two histograms perfectly match. This is graphically nice.

mbaudin47:
This fixes #1205.
The current implementation is naive, in the sense that it does not make use of a sample sort. Hence, its performance is poor with a large sample or with a large number of bins.

For a reason that I do not understand, the test fails to execute:
```
$ ./t_HistogramFactory_std 
./t_HistogramFactory_std: symbol lookup error: 
./t_HistogramFactory_std: undefined symbol: 
_ZNK2OT16HistogramFactory5buildERKNS_6SampleEdRKNS_5PointE
```
I cannot find why, since the build succeeds: how is that possible that a symbol is not found?


lib/src/Uncertainty/Distribution/HistogramFactory.cxx
```diff
@@ -133,6 +140,63 @@ Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
   return result;
 }
 
+Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
+    const Scalar first,
+    const Point width) const
```
regislebrun: Width should be passed by reference as it is a complex type (ie an object)

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
```diff
@@ -133,6 +140,63 @@ Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
...
+{
+  const UnsignedInteger size = sample.getSize();
+  if (size == 0) throw InvalidArgumentException(HERE) << "Error: cannot build an Histogram based on an empty sample.";
+  if (sample.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: can build an Histogram only if dimension equals 1, here dimension=" << sample.getDimension();
+  const UnsignedInteger binNumber = width.getSize();
+  if (binNumber == 0) throw InvalidArgumentException(HERE) << "Error: expected a positive number of bin, got 0.";
+  // Construct the histogram
+  Point heights(binNumber, 0.0);
```
regislebrun: No need to initialize the Point to zero as it is done by default.

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
```diff
@@ -133,6 +140,63 @@ Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
...
+  if (!SpecFunc::IsNormal(minvalue) || !SpecFunc::IsNormal(maxvalue)) throw InvalidArgumentException(HERE) << "Error: cannot build an Histogram distribution if data contains NaN or Inf";
+  if (maxvalue == minvalue)
+  {
+    const Scalar epsilon = ResourceMap::GetAsScalar("Distribution-DefaultCDFEpsilon");
+    const Scalar delta = std::max(std::abs(minvalue), 10.0) * epsilon;
+    Histogram result(minvalue - 0.5 * delta, Point(1, delta), Point(1, 1.0));
+    result.setDescription(sample.getDescription());
+    return result;
+  }
+  if (minvalue<first) throw InvalidArgumentException(HERE) << "Error: the minimum of the sample is smaller than first";
```
regislebrun: add spaces around <

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
```diff
@@ -133,6 +140,63 @@ Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
...
+  // Compute the right bound of the histogram = first + the sum of the widths
+  Scalar rightBound = first;
+  for(UnsignedInteger j = 0; j < binNumber; ++j)
+  {
+    rightBound += width[j];
```
regislebrun: Maybe you should check that width[j] is > 0, or at least >= 0?

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
```diff
@@ -133,6 +140,63 @@ Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
...
+  }
+  if (maxvalue>rightBound) throw InvalidArgumentException(HERE) << "Error: the maximum of the sample is greater than the right boundary of the histogram";
```
regislebrun: Spaces around >

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
```diff
@@ -133,6 +140,63 @@ Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
...
+  // Count the number of points in each class with a naive loop
+  Scalar left = first;
+  Scalar right;
+  for(UnsignedInteger j = 0; j < binNumber; ++j) // Loop over the classes
+  {
+    right = left + width[j];
+    for(UnsignedInteger i = 0; i < size; ++i)    // Loop over the values
+    {
+      if ((sample(i,0) >= left) and (sample(i,0)<right))
```
regislebrun: no 'and', use &&

regislebrun: With this test, if a point is exactly equal to maxvalue it is not taken into account

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
```diff
@@ -133,6 +140,63 @@ Histogram HistogramFactory::buildAsHistogram(const Sample & sample,
...
+      }
+    left = right;
+  }
+  // Compute the area
+  Scalar area;
+  for(UnsignedInteger j = 0; j < binNumber; ++j) // Loop over the classes
+  {
+    area += width[j] * heights[j];
+  }
+  const Scalar inverseArea = 1.0 / area;
```
regislebrun: What is the purpose of introducing inverseArea?

lib/src/Uncertainty/Distribution/openturns/HistogramFactory.hxx
```diff
@@ -50,12 +50,18 @@ public:
                      const Scalar bandwidth) const;
   Distribution build(const Sample & sample,
                      const UnsignedInteger binNumber) const;
+  Distribution build(const Sample & sample,
+                     const Scalar first,
+                     const Point width) const;
```
regislebrun: Here width should be passed by reference to avoid costly copy

lib/src/Uncertainty/Distribution/openturns/HistogramFactory.hxx
```diff
@@ -50,12 +50,18 @@ public:
...
   Distribution build() const;
   Histogram buildAsHistogram(const Sample & sample) const;
   Histogram buildAsHistogram(const Sample & sample,
                              const Scalar bandwidth) const;
   Histogram buildAsHistogram(const Sample & sample,
                              const UnsignedInteger binNumber) const;
+  Histogram buildAsHistogram(const Sample & sample,
+                             const Scalar first,
+                             const Point width) const;
```
regislebrun: Here width should be passed by reference as it avoids costly copy, and it is implemented with a reference, hence the undefined symbol you get.

mbaudin47:
Thank you very much for the review: I'll fix this soon.

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
mbaudin47: OK.

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
mbaudin47: OK

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
mbaudin47: That's right: OK.

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
mbaudin47: Ok for the "and". 
For the sample(i,0)==maxvalue case, I understand your point, but what you wrote is probably not what you meant. In the current implementation, there is an issue if the maxvalue is exactly equal to rightBound. In order to manage this case, I change the condition "if (maxvalue > rightBound)" to "(maxvalue >= rightBound)". This fixes the issue.

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
mbaudin47: Ok. And I also replace ">" with ">=", as mentioned in the above conversation.

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
mbaudin47: I do not understand what you mean: the inverseArea variable is then used in the creation of the Histogram, so that the area of the histogram is equal to 1. Am I missing something here ?

lib/src/Uncertainty/Distribution/openturns/HistogramFactory.hxx
mbaudin47: You are right: OK.

lib/src/Uncertainty/Distribution/openturns/HistogramFactory.hxx
mbaudin47: Thanks for the information: I missed that "&" also provided a performance advantage at this point.

lib/src/Uncertainty/Distribution/HistogramFactory.cxx
mbaudin47: OK

mbaudin47:
Below is a full-length motivating example for this development.
I consider the flooding example. 
There are four inputs Q, Ks, Zv, Zm and one output H. 
I want to see if the input flowrate Q has an impact on getting high values of the output H. 
In order to see this, I compare the unconditional distribution of Q with conditional distribution of Q given that H is greater than a given threshold. 
I choose the 90%-quantile of H as the threshold.
If the two distributions are close to each other, then the variable cannot an any statistical impact.
```
#!/usr/bin/env python
# coding: utf-8

import openturns as ot

# Create the marginal distributions of the parameters
dist_Q = ot.TruncatedDistribution(ot.Gumbel(558., 1013.), 0, ot.TruncatedDistribution.LOWER)
dist_Ks = ot.TruncatedDistribution(ot.Normal(30.0, 7.5), 0, ot.TruncatedDistribution.LOWER)
dist_Zv = ot.Uniform(49.0, 51.0)
dist_Zm = ot.Uniform(54.0, 56.0)
marginals = [dist_Q, dist_Ks, dist_Zv, dist_Zm]

# Create the joint probability distribution
distribution = ot.ComposedDistribution(marginals)
distribution.setDescription(['Q', 'Ks', 'Zv', 'Zm'])

# create the model
model = ot.SymbolicFunction(['Q', 'Ks', 'Zv', 'Zm'],
                            ['(Q/(Ks*300.*sqrt((Zm-Zv)/5000)))^(3.0/5.0)'])

# create a sample
size = 500
inputSample = distribution.getSample(size)

outputSample = model(inputSample)

sample = ot.Sample(size,5)
sample[:,0:4] = inputSample
sample[:,4] = outputSample
sample[0:5,:]

sampleQ = sample[:,0]

numberOfBins = 10
histogram = ot.HistogramFactory().buildAsHistogram(sampleQ,numberOfBins)
graph = histogram.drawPDF()
graph.setLegends(["Q"])

import numpy as np

def computeConditionnedSample(sample, alpha = 0.9, criteriaComponent = None, selectedComponent = 0):
    '''
    Return values from the selectedComponent-th component of the sample. 
    Selects the values according to the alpha-level quantile of the criteriaComponent-th component of the sample.
    '''
    dim = sample.getDimension()
    if criteriaComponent is None:
        criteriaComponent = dim - 1
    sortedSample = sample.sortAccordingToAComponent(criteriaComponent)   
    quantiles = sortedSample.computeQuantilePerComponent(alpha)
    quantileValue = quantiles[criteriaComponent]
    sortedSampleCriteria = sortedSample[:,criteriaComponent]
    indices = np.where(np.array(sortedSampleCriteria.asPoint())>quantileValue)[0]
    conditionnedSortedSample = sortedSample[int(indices[0]):,selectedComponent]
    return conditionnedSortedSample

# Extract the sub-sample of the input flowrates Q which leads to large values of the output H.

alpha = 0.9
criteriaComponent = 4
selectedComponent = 0
conditionnedCampleQ = computeConditionnedSample(sample,alpha,criteriaComponent,selectedComponent)

# We could as well use:
# ```
# conditionnedHistogram = ot.HistogramFactory().buildAsHistogram(conditionnedCampleQ)
# ```
# but this creates an histogram with new classes, corresponding to conditionnedCampleQ.
# We want to use exactly the same classes as the full sample, so that the two histograms match.

first = histogram.getFirst()
width = histogram.getWidth()
conditionnedHistogram = ot.HistogramFactory().buildAsHistogram(conditionnedCampleQ,first,width)

graphConditionnalQ = conditionnedHistogram.drawPDF()
graphConditionnalQ.setColors(["blue"])
graphConditionnalQ.setLegends(["Q|H>H_%s" % (alpha)])

graph.add(graphConditionnalQ)
```

conditionnal-histogram

We see that the two distribution are not close to each other: Q has an impact on the event "H>H_90%".

mbaudin47:
Good for review.

mbaudin47:
@regislebrun : If I understand correctly our discussion, do you request an algorithm to estimate the bandwidth from the data in order to approve this PR, or this can be accepted in the current state ? I agree that, in general, the bandwidth should be adapted from the data, but the current implementation provides the basic building block when the classes are set by the user. I have already created an algorithm for the estimation of the bandwidth based on cross-validation, but it did not seem to be bulletproof at the time where I performed the experiments: I would not put this specific algorithm in the lib without further improvements.

PR #1215, opened 2019-09-06
Add MarginalEvaluation/Gradient/Hessian

jschueller:
Speeds up marginal evaluation in Function.getMarginal (replaces the Composed with Symbolic or Linear)


jschueller:
cc @aurelieladier

PR #1220, opened 2019-09-15
Group sobol indice

mbaudin47:
Creates a new getSobolGroupedTotalIndex method for FunctionalChaosSobolIndices. 
This method computes the total Sobol' indice of a group of variables. 
The algorithm is quite naive in the sense that it uses the following property of a group of variable u:

Su+Tc=1

where Su is the first order Sobol' indice of the group u and Tc is the total order Sobol' indice of the group c, where c is the complementary group of u. 
Hence, 

Tu=1-Sd

where d is the complementary group of u. 
This is why the tricky part is to compute the complementary group ; then we use the getSobolGroupedIndex method (which already exists). 

The naive part is that the current implementation is not accurate when Sd is close to 1, which leads to Tu close to zero. In this case, catastrophic cancellation will take place. An accurate implementation would be based on the sum of the squared coefficients of the appropriate multi-indices (given that these squared coefficients are all positive, cancellation cannot happen here). My humble opinion is that this accuracy limitation will never been seen in practice, given that the polynomial chaos is generally just an approximation of the actual function.

By the way, I also reorganized the metamodeling example section and added examples in the FunctionalChaosSobolIndices "Examples" doc section.


mbaudin47:
The build_linux fails because of an example :
```
Failed example:
    chaosSI.getSobolIndex(0)
Expected:
    0.3400236040839249
Got:
    0.34002360408392485
```
How may I fix this ?

jschueller:
use ellipsis:
```
>>> chaosSI.getSobolIndex(0)
0.340236...
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -350,6 +350,50 @@ Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const UnsignedInteger v
...
 }
 
+/* Sobol grouped total index accessor */
+Scalar FunctionalChaosSobolIndices::getSobolGroupedTotalIndex(const Indices & variableIndices,
+    const UnsignedInteger marginalIndex) const
+{
+  // Compute the complementary indice
+  const UnsignedInteger inputDimension = functionalChaosResult_.getDistribution().getDimension();
+  const UnsignedInteger groupSize = variableIndices.getSize();
+  const UnsignedInteger complementaryGroupSize = inputDimension - groupSize;
```
jschueller: you need to check that the size of variableIndices doesnt exceed dimension else this will overflow

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -350,6 +350,50 @@ Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const UnsignedInteger v
...
+  Indices complementaryVariableIndices(complementaryGroupSize);
+  UnsignedInteger i = 0;
+  for (UnsignedInteger j = 0; j < inputDimension; ++j)
+  {
+    if (allVariableIndices[j]>=0)
+    {
+        complementaryVariableIndices[i] = UnsignedInteger(allVariableIndices[j]);
+        i=i+1;
+    }
+  }
```
jschueller: The code that computes `complementaryVariableIndices ` is too complicated, use:
```
Indices complementaryVariableIndices;
for (UnsignedInteger j = 0; j < inputDimension; ++ j)
  if (!variableIndices.contains(j))
    complementaryVariableIndices.add(j);
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -350,6 +350,50 @@ Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const UnsignedInteger v
...
+  Scalar complementaryFirstIndex = getSobolGroupedIndex(complementaryVariableIndices, marginalIndex);
+  Scalar groupTotalIndex = 1.0-complementaryFirstIndex;
+  return groupTotalIndex;
+}
+/* Sobol total index accessor */
+Scalar FunctionalChaosSobolIndices::getSobolGroupedTotalIndex(const UnsignedInteger variableIndex,
+    const UnsignedInteger marginalIndex) const
+{
+  Indices index(1);
+  index[0] = variableIndex;
```
jschueller: you can use the ctor: `Indices(1, variableIndex) `

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: You are right: this is fixed. 
As far as I understand, the current policy is to check the input arguments at the level where it is appropriate: am i correct ?

jschueller: yes

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: Excellent: this is much simpler indeed. But the "complement" - that i did not know - method does exactly the required job: i'd rather use it.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: Correct. Actually, the same statements are here and there in the FunctionalChaosSobolIndices.cxx: I replace them all with your suggestion.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -233,8 +233,7 @@ Scalar FunctionalChaosSobolIndices::getSobolIndex(const Indices & variableIndice
 Scalar FunctionalChaosSobolIndices::getSobolIndex(const UnsignedInteger variableIndex,
     const UnsignedInteger marginalIndex) const
 {
-  Indices index(1);
-  index[0] = variableIndex;
+  Indices index(1, variableIndex);
```
mbaudin47: Add `const` in front of Indices

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -290,12 +289,11 @@ Scalar FunctionalChaosSobolIndices::getSobolTotalIndex(const Indices & variableI
 Scalar FunctionalChaosSobolIndices::getSobolTotalIndex(const UnsignedInteger variableIndex,
     const UnsignedInteger marginalIndex) const
 {
-  Indices index(1);
-  index[0] = variableIndex;
+  Indices index(1, variableIndex);
```
mbaudin47: Add `const` in front of Indices

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -341,15 +339,35 @@ Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const Indices & variabl
...
 }
 
-/* Sobol index accessor */
+/* Sobol (first order) index accessor */
 Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const UnsignedInteger variableIndex,
     const UnsignedInteger marginalIndex) const
 {
-  Indices index(1);
-  index[0] = variableIndex;
+  Indices index(1, variableIndex);
```
mbaudin47: Add `const` in front of Indices

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -341,15 +339,35 @@ Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const Indices & variabl
...
+  // Compute total index from complementary first index
+  Scalar complementaryFirstIndex = getSobolGroupedIndex(complementaryVariableIndices, marginalIndex);
+  Scalar groupTotalIndex = 1.0-complementaryFirstIndex;
+  return groupTotalIndex;
+}
+/* Sobol total index accessor */
+Scalar FunctionalChaosSobolIndices::getSobolGroupedTotalIndex(const UnsignedInteger variableIndex,
+    const UnsignedInteger marginalIndex) const
+{
+  Indices index(1, variableIndex);
```
mbaudin47: Add `const` in front of Indices

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -341,15 +339,35 @@ Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const Indices & variabl
...
+/* Sobol grouped total index accessor */
+Scalar FunctionalChaosSobolIndices::getSobolGroupedTotalIndex(const Indices & variableIndices,
+    const UnsignedInteger marginalIndex) const
+{
+  // Compute the complementary indice
+  const UnsignedInteger inputDimension = functionalChaosResult_.getDistribution().getDimension();
+  if (!variableIndices.check(inputDimension)) throw InvalidArgumentException(HERE) << "The variable indices of a Sobol indice must be in the range [0, dim-1] and must be different.";
+  Indices complementaryVariableIndices = variableIndices.complement(inputDimension);
+  // Compute total index from complementary first index
+  Scalar complementaryFirstIndex = getSobolGroupedIndex(complementaryVariableIndices, marginalIndex);
```
mbaudin47: Add `const` in front of `Scalar`

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
```diff
@@ -341,15 +339,35 @@ Scalar FunctionalChaosSobolIndices::getSobolGroupedIndex(const Indices & variabl
...
 }
 
+/* Sobol grouped total index accessor */
+Scalar FunctionalChaosSobolIndices::getSobolGroupedTotalIndex(const Indices & variableIndices,
+    const UnsignedInteger marginalIndex) const
+{
+  // Compute the complementary indice
+  const UnsignedInteger inputDimension = functionalChaosResult_.getDistribution().getDimension();
+  if (!variableIndices.check(inputDimension)) throw InvalidArgumentException(HERE) << "The variable indices of a Sobol indice must be in the range [0, dim-1] and must be different.";
+  Indices complementaryVariableIndices = variableIndices.complement(inputDimension);
```
mbaudin47: Add `Const`

mbaudin47: ok

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: ok

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: ok

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: ok

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: ok

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosSobolIndices.cxx
mbaudin47: ok

sofianehaddad:
This feature is not mentioned in `ChangeLog`

jschueller:
i'll add it

PR #1248, opened 2019-09-17
Fix ProductCovarianceModel parameters description

sofianehaddad:
In order to perform analytical evaluation of amplitude, parameter description should be `amplitude_0`, not amplitude

This fixes openturns#1247


regislebrun:
+1

PR #1274, opened 2019-09-28
Created new graphics examples

mbaudin47:
Created a new Examples section with graphics examples:
* basics: create graphics Curve and Cloud, add, configure line style, line width, get valid options, create a grid of graphics etc...
* create a filled area
* create a contour plot for the log-likelihood function


regislebrun:
+1

These examples are really nice. I have only one concern regarding the log-likelihood contours. It looks more like a matplotlib example than an OpenTURNS example. You could have used the following code:
```
def logLikelihood(X):
    mu = X[0]
    sigma = X[1]
    a = sample.getMin()[0]
    b = sample.getMax()[0]
    distribution = ot.TruncatedNormal(mu,sigma,a,b)
    samplesize = sample.getSize()
    return -distribution.computeLogPDF(sample).computeMean() * samplesize

# Draw the logLikelihood function
graph = ot.PythonFunction(2, 1, logLikelihood).draw([-3.0, 0.1], [5.0, 7.0], [50]*2)
# Fine-tuning
# By default each level is a dedicated contour in order to have one color per contour, but they all share the same grid and data
# Define the user levels
levels = [718.57, 718.6, 718.7, 718.8, 719.,720., 750.]
# Take the first contour as the only one with multiple levels
contours = graph.getDrawable(0)
# Here we ask for many iso-values in the same data so the color will be the same for all curves
contours.setLevels(levels)
# Inline the level values
contours.setDrawLabels(True)
# Substitute the collection of drawables by a collection reduced to this unique contour
graph.setDrawables([contours])
# Remove the legend
graph.setLegendPosition("")
# Set the titles
graph.setTitle("Log-Likelihood")
graph.setXTitle(r"$\mu$")
graph.setYTitle(r"$\sigma$")
graph
```
This way, the function is *much more* efficient as it uses the vectorized evaluation of the log-PDF over a sample (no significant speed-up for size=11, already a x4 speed-up for size=100 and x30 for size=1000), the overall code has the same size (18 lines without comments) but exposes much more OT features.
Do you agree with this proposal?

mbaudin47:
The speedup is a good point. Graphically speaking, the basic Contour has many colors, but the customized Contour only has one. Is it possible to configure the graphics so as to get different colors?

mbaudin47:
In the new version of the example, I added your excellent advice ; however, I still feel that the Matplotlib way is interesting. Why not to keep both of them?

regislebrun:
Because there are many other places to learn matplotlib ;-) For the colors, you can use the following code:
```
# Fine-tuning
# By default each level is a dedicated contour in order to have one color per contour, but they all share the same grid and data
# Define the user levels
levels = [718.57, 718.6, 718.7, 718.8, 719.,720., 750.]
# Take the first contour as the only one with multiple levels
contour = graph.getDrawable(0)
# Build a range of colors
ot.ResourceMap.SetAsUnsignedInteger('Drawable-DefaultPalettePhase', len(levels))
palette = ot.Drawable.BuildDefaultPalette(len(levels))
drawables = list()
for i in range(len(levels)):
    # Here we ask for many iso-values in the same data so the color will be the same for all curves
    contour.setLevels([levels[i]])
    # Inline the level values
    contour.setDrawLabels(True)
    # Substitute the collection of drawables by a collection reduced to this unique contour. We have to copy the drawable because a Python list stores only pointers
    drawables.append(ot.Drawable(contour))
# Replace the drawables
graph.setDrawables(drawables)
# Remove the legend
graph.setLegendPosition("")
# Add colors
graph.setColors(palette)
# Set the titles
graph.setTitle("Log-Likelihood")
graph.setXTitle(r"$\mu$")
graph.setYTitle(r"$\sigma$")
ot.Show(graph)
```

mbaudin47:
> Because there are many other places to learn matplotlib ;-)

Ah! This is true! Thanks for the code. I just updated the example according to your suggestion. The result is very nice. I also fixed a bug: the values in the labels were duplicated so that only the last one was correct. This is because the "setLevels" method did not update the labels when the number of levels does not change. This is wrong: the number of levels may not change, but the content can; hence, the new code updates the labels everytime. 

The resulting example will satisfy you, i hope!

regislebrun:
+1

PR #1276, opened 2019-10-03
Various enhancements

jschueller:


regislebrun:
+1

jschueller:
I thought it'd be best to not have to build with full dependencies to test docstrings, maybe i'll reconsider with what is most difficult to install

jschueller:
@regislebrun @sofianehaddad I ended up pushing much more commits here, maybe you'll want to have a new look at it, I also fixed the Bonmin detection

python/src/HMatrix.i
```diff
@@ -40,6 +41,7 @@ public:
 
   virtual void compute(OT::UnsignedInteger i, OT::UnsignedInteger j, OT::Matrix* localValues) const
   {
+    OT::InterpreterLock lock;
```
sofianehaddad: Unfortunately this does not fix the current issue we have with parallel hmat
I have to see later

sofianehaddad:
Except the `hmat` case, it is ok for me

python/src/HMatrix.i
jschueller: Ok, I'll leave it for later then

.travis.yml
```diff
@@ -113,7 +113,7 @@ script:
               -DUSE_COTIRE=ON
               -DCOTIRE_MAXIMUM_NUMBER_OF_UNITY_INCLUDES="-j4"
               . &&
-        make install -j4;
+        make install VERBOSE=1 -j4;
```
sofianehaddad: probably c++11 : default "standard" is usually libc++ instead of libstc++

jschueller: yeah, that's what I wanted to know.
But I dont know why it complains if the default is c++11.

lib/src/Base/Type/HermitianMatrix.cxx
```diff
@@ -63,6 +63,17 @@ HermitianMatrix::HermitianMatrix(const HermitianMatrix & h)
...
+HermitianMatrix & HermitianMatrix::operator=(const HermitianMatrix & rhs)
+{
+  if (this != &rhs)
+  {
+    SquareComplexMatrix::operator=(rhs);
+    hasBeenHermitianized_ = rhs.hasBeenHermitianized_;
+  }
+  return *this;
+}
+
```
regislebrun: Why do you have to define the copy constructor explicitly? What is wrong with the one automatically built by the compiler?

lib/src/Base/Type/SymmetricMatrix.cxx
```diff
@@ -51,13 +51,24 @@ SymmetricMatrix::SymmetricMatrix(const MatrixImplementation & i)
...
+SymmetricMatrix & SymmetricMatrix::operator=(const SymmetricMatrix & rhs)
+{
+  if (this != &rhs)
+  {
+    SquareMatrix::operator=(rhs);
+    hasBeenSymmetrized_ = rhs.hasBeenSymmetrized_;
+  }
+  return *this;
+}
+
```
regislebrun: Same question here?

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
```diff
@@ -225,13 +230,16 @@ void SubsetSampling::run()
     coefficientOfVariationPerStep_.add(sqrt(coefficientOfVariationSquare));
 
     // stop if the number of subset steps is too high, else results are not numerically defined anymore
-    if (std::abs(pow(probabilityEstimate, 2.)) < SpecFunc::MinScalar)
+    if (probabilityEstimate < minimumProbability_)
       throw NotDefinedException(HERE) << "Probability estimate too low: " << probabilityEstimate;
```
regislebrun: too low or too small?

lib/src/Base/Type/HermitianMatrix.cxx
jschueller: Its deprecated in newer c++ standard to have only one of copy constructor or copy assignement operators, it's either both or neither

lib/src/Base/Optim/Bonmin.cxx
```diff
@@ -28,8 +28,8 @@
 #include "openturns/PersistentObjectFactory.hxx"
 #ifdef OPENTURNS_HAVE_BONMIN
   #include "openturns/BonminProblem.hxx"
-  #include "BonBonminSetup.hpp"
-  #include "BonCbc.hpp"
+  #include <BonBonminSetup.hpp>
```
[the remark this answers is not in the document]
jschueller: ^^

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
jschueller: too small yes

Issue #1275, opened 2019-09-28
The TruncatedNormalFactory sometimes fail


mbaudin47:
The following script creates a TruncatedNormalFactory from a sample:
```
import openturns as ot
ot.RandomGenerator.SetSeed(0)
a = -1
b=2.5
mu = 2.
sigma = 3.
distribution = ot.TruncatedNormal(mu,sigma,a,b)
distFactory = ot.TruncatedNormalFactory()    
sample = distribution.getSample(11)
newDistribution = distFactory.build(sample)
print("Built-in Fitted=",newDistribution)
```
This prints
```
Built-in Fitted= TruncatedNormal(mu = -16060.1, sigma = 561.146, a = -0.696535, b = 2.33702)
```
The bounds are correctly estimated, given that the sample min is -0.5701 and the sample max is 2.211. The mu and sigma parameters are completely wrong, with a very negative mu and a large sigma. 
In this particular case, the PDF cannot even be drawn:
```
>>> newDistribution.drawPDF()
Traceback (most recent call last):
  File "<ipython-input-67-bf36ab6df2ab>", line 1, in <module>
    newDistribution.drawPDF()
  File "/my/path/openturns/build/install/lib/python3.7/site-packages/openturns/model_copula.py", line 6265, in drawPDF
    return _model_copula.Distribution_drawPDF(self, *args)
TypeError: InvalidArgumentException : Error: cannot draw a PDF with xMax <= xMin, here xmin=-11500.3 and xmax=-11500.3
```
In this case the optimizer (TNC) uses a large number of function evaluations and only stops when the difference between two consecutive likelihood function values is close to zero.

The current implementation uses a clever scaling of the data and disables the estimation of the bounds of the TruncatedNormal. This is why I tested the basic MaximumLikelihoodFactory class:
```
MLfactory = ot.MaximumLikelihoodFactory(ot.TruncatedNormal())
newDistribution = MLfactory.build(sample)
print("ML Fitted      =",newDistribution)
```
this prints:
```
ML Fitted      = TruncatedNormal(mu = 0.00949447, sigma = 1.00589, a = -0.934979, b = 0.934979)
```

This cannot be good: the TruncatedNormal max is 0.9350 while the sample max is 2.211. 

I assume that estimating the parameters of a TruncatedNormal is instrinsically difficult. Using the method of moments would be an option, but this is not so easy given that this requires to solve nonlinear equations (there are references on this topic). 

[...]

PR #1277, opened 2019-10-07
Improve Kolmogorov-Smirnov test

### LINKED ISSUE #1061: FittingTest.Kolmogorov is slow with distribution factory
Within the OT GUI, we compute the P-Value of the KS-test in the case where the parameters are estimated from data.
The problem is that the list contains 20 distributions, which leads to a long computation. 

Denote by p1 the p-value evaluated assuming that the parameters are known. The p-value p1 is fast to computed. Denote p2 the p-value assuming that the parameters are estimated. We always have p2<p1. Therefore, an idea is that, if p1 is lower than, say, 0.001, then we do not have to compute p2. 

Given that p2 is to be computed, we may want to use the smallest number of simulations required to achieve a given precision. For this purpose, we may use the ProbabilitySimulationAlgorithm. 

These two different ideas are in the script in attachment.
* Computing only p1 when possible is implemented is called the "low-cost algorithm" in the script. 
* Computing only p1 when possible, and p2 with an Event algorithm is called the "event-based algorithm" in the script.

[test-kolmogorov-smirnov-performance.txt](https://github.com/openturns/openturns/files/3014191/test-kolmogorov-smirnov-performance.txt)

The first algorithm uses never more costly KS tests than required: it allows to save some simulations. But there are still several distribution factories which are associated with p2-values in the range [0.001,1]. For these factories, the second algorithm costs much. 

The last point is disappointing: I do not see how we could use a reduction variance method. This is because we have no hand on the underlying random variable.

This is not a bug report: this is an enhancement.

mbaudin47:
The goal of this proposal is to increase the speed of the Kolmogorov-Smirnov test when the parameters of the distribution are estimated from a sample. Although the ultimate goal is to fix #1061, it only partially solves it and more work would be required. 

Since the FisherSnedecor was shown to have the least performance, this proposal creates a new FisherSnedecorFactory::buildMethodOfMoments method. 
* This method is used as a starting point for the likelihood maximization. This is, in general, better than using the parametersLowerBound in the current implementation. As a result, when the buildMethodOfMoments fails, the default "build" method fails: I consider that this is a reasonable behavior, but this might be discussed. 
* In the KS test experiments that I did, this increases the performance because the sample made the buildMethodOfMoments method to fail. This speeds up the benchmark by rejecting the FisherSnedecor distribution.

Aside from this, the PR improves the doc of the LogNormalFactory. 

This PR is related to #1275 and #1061.

### DIFF
--- a/lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
+++ b/lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
@@ -52,8 +52,8 @@ FisherSnedecor FisherSnedecorFactory::buildMethodOfMoments(const Sample & sample
   if (mu <= 1.0) throw InvalidArgumentException(HERE) << "Error: cannot estimate a FisherSnedecor distribution based on a sample with sample mean less than 1 using the method of moments.";
   const Scalar sigma2 = sample.computeCovariance()(0,0);
   if (sigma2 == 0.0) throw InvalidArgumentException(HERE) << "Error: cannot estimate a FisherSnedecor distribution based on a constant sample using the method of moments.";
-  const Scalar d2 = 2*mu/(mu-1);
-  const Scalar d1 = 2*std::pow(d2,2)*(d2-2)/(std::pow(d2-2,2)*(d2-4)*sigma2-2*std::pow(d2,2));
+  const Scalar d2 = 2 * mu / (mu - 1);
+  const Scalar d1 = 2 * std::pow(d2,2.0) * (d2 - 2)/(std::pow(d2 - 2.0,2.0) * (d2 - 4) * sigma2 - 2 * std::pow(d2,2.0));
   FisherSnedecor result(d1, d2);
   result.setDescription(sample.getDescription());
   return result;
@@ -82,7 +82,7 @@ DistributionFactoryResult FisherSnedecorFactory::buildEstimator(const Sample & s
 FisherSnedecor FisherSnedecorFactory::buildAsFisherSnedecor(const Sample & sample) const
 {
   // Use method of moments as starting point
-  FisherSnedecor estimatedMomentsDistribution = buildMethodOfMoments(sample);
+  const FisherSnedecor estimatedMomentsDistribution = buildMethodOfMoments(sample);
   Point parametersFromMoments = estimatedMomentsDistribution.getParameter();
 
   const UnsignedInteger dimension = build().getParameterDimension();
--- a/lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
+++ b/lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
@@ -45,6 +45,20 @@ FisherSnedecorFactory * FisherSnedecorFactory::clone() const
 
 /* Here is the interface that all derived class must implement */
 
+/* Algorithm associated with the method of moments */
+FisherSnedecor FisherSnedecorFactory::buildMethodOfMoments(const Sample & sample) const
+{
+  const Scalar mu = sample.computeMean()[0];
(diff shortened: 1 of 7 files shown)


lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
```diff
@@ -45,6 +45,20 @@ FisherSnedecorFactory * FisherSnedecorFactory::clone() const
...
 
+/* Algorithm associated with the method of moments */
+FisherSnedecor FisherSnedecorFactory::buildMethodOfMoments(const Sample & sample) const
+{
+  const Scalar mu = sample.computeMean()[0];
+  if (mu <= 1.0) throw InvalidArgumentException(HERE) << "Error: cannot estimate a FisherSnedecor distribution based on a sample with sample mean less than 1 using the method of moments.";
+  const Scalar sigma2 = sample.computeCovariance()(0,0);
+  if (sigma2 == 0.0) throw InvalidArgumentException(HERE) << "Error: cannot estimate a FisherSnedecor distribution based on a constant sample using the method of moments.";
+  const Scalar d2 = 2*mu/(mu-1);
+  const Scalar d1 = 2*std::pow(d2,2)*(d2-2)/(std::pow(d2-2,2)*(d2-4)*sigma2-2*std::pow(d2,2));
```
regislebrun: Please ad some spaces around operators in your formula. Some compilers will complain if you take an integral power of a foating point value, it is more robust to change constants from their integral writing into their floating point writing: `std::pow(d2-2,2)` becomes `std::pow(d2 - 2.0, 2.0)`

lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
```diff
@@ -67,6 +81,10 @@ DistributionFactoryResult FisherSnedecorFactory::buildEstimator(const Sample & s
 
 FisherSnedecor FisherSnedecorFactory::buildAsFisherSnedecor(const Sample & sample) const
 {
+  // Use method of moments as starting point
+  FisherSnedecor estimatedMomentsDistribution = buildMethodOfMoments(sample);
```
regislebrun: const FisherSnedecor

lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
mbaudin47: Ok, it makes sense, given that this amounts to x^y=exp(y*log(x)) when x is a floating point. I fixed this.

lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
mbaudin47: Ok: I must focus on this point. This is fixed.

lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx
```diff
@@ -76,7 +94,7 @@ FisherSnedecor FisherSnedecorFactory::buildAsFisherSnedecor(const Sample & sampl
 
   // override starting point
   OptimizationAlgorithm solver(factory.getOptimizationAlgorithm());
-  solver.setStartingPoint(parametersLowerBound);
+  solver.setStartingPoint(parametersFromMoments);
```
jschueller: parametersLowerBound is unused

mbaudin47: No: parametersLowerBound is used later, to set the lower bounds of the optimization problem in the MaximumLikelihoodFactory:

https://github.com/openturns/openturns/blob/60dbd347039aba890dd3b68dcf833f02780ebdbf/lib/src/Uncertainty/Distribution/FisherSnedecorFactory.cxx#L101

PR #1280, opened 2019-10-11
[RFC] Add ResourceMapBase class

pcaclin:
Add ResourceMapBase class for local options management
    
Define a new class ResourceMapBase for storage of (key, value) pairs, and providing related accessors. Values can be of type String, UnsignedInteger, Scalar or Bool.

ResourceMapBase can be used to set local options for other OT objects (e.g. OptimizationAlgorithmImplementations).

ResourceMap is re-defined as a class embedding a ResourceMapBase attribute, and providing services to ensure singleton-like behavior and compatibility with previous implementation.


sofianehaddad:
@jschueller what is the status of this PR ?

jschueller:
I dont think its relevant anymore, we wanted to clean things up but maybe we ended adding more bloat.

PR #1291, opened 2019-10-15
Improved kriging examples

mbaudin47:
* Added example to manage an arbitrary basis in kriging: based on LegendreFactory
* Adding new example to estimate hyperparameters with a customized optimization algorithm
* Replaced KrigingRandomVector by ConditionedGaussianProcess in the simulate example
* Fixed confusions in the GLM example


mbaudin47:
> Ok for me.
> But these throw an error with my sphinx version ; can you make sure to insert a blank line after each math section (after each closing $$) ?

Thanks for review the content: I just fixed the issues. How is that possible that your sphinx version fails, while the continuous integration succeeds? Did you use a specific option which has additional checks? That would be great to have this on the C.I., because the current checks seems not to be fail-safe.

PR #1292, opened 2019-10-15
Improved MarshallOlkinCopula

mbaudin47:
Added the case alpha=beta=0.
Formatted the code as required in review.

### DIFF
--- a/lib/src/Uncertainty/Distribution/MarshallOlkinCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/MarshallOlkinCopula.cxx
@@ -106,12 +106,12 @@ Scalar MarshallOlkinCopula::computeCDF(const Point & point) const
   const Scalar vbeta = std::pow(v, beta_);
   if (ualpha>vbeta)
   {
-    const Scalar p = std::pow(u, 1-alpha_) * v;
+    const Scalar p = std::pow(u, 1 - alpha_) * v;
     return p;
   }
   else
   {
-    const Scalar p = u * std::pow(v, 1-beta_);
+    const Scalar p = u * std::pow(v, 1 - beta_);
     return p;
   }
 }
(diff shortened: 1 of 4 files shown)


lib/src/Uncertainty/Distribution/MarshallOlkinCopula.cxx
```diff
@@ -142,7 +142,6 @@ Bool MarshallOlkinCopula::hasIndependentCopula() const
 void MarshallOlkinCopula::setAlpha(const Scalar alpha)
 {
   if ((alpha < 0.0) || (alpha > 1.0)) throw InvalidArgumentException(HERE) << "Alpha MUST be in [0, 1], here alpha=" << alpha;
```
sofianehaddad: if (!(alpha >= 0.0) && !(alpha <= 1.0))

We should use the "common" way for checking arguments

mbaudin47: Ok. This way is indeed used in:
https://github.com/openturns/openturns/blob/cdc0df551b77e9f4dcfafc85b2d4206a90e0fad3/lib/src/Uncertainty/Distribution/WeibullMax.cxx#L344
but not in:
https://github.com/openturns/openturns/blob/cdc0df551b77e9f4dcfafc85b2d4206a90e0fad3/lib/src/Uncertainty/Distribution/AliMikhailHaqCopula.cxx#L374

I will fix this the way you suggest.

lib/src/Uncertainty/Distribution/MarshallOlkinCopula.cxx
```diff
@@ -106,12 +106,12 @@ Scalar MarshallOlkinCopula::computeCDF(const Point & point) const
...
+    const Scalar p = std::pow(u, 1 - alpha_) * v;
     return p;
   }
   else
   {
-    const Scalar p = u * std::pow(v, 1-beta_);
+    const Scalar p = u * std::pow(v, 1 - beta_);
     return p;
   }
 }
```
regislebrun: Here you also have to deal with the special cases, as you are not allowed to take the null power of something. See MinCopula and IndependentCopula for the relevant codes.

mbaudin47: I did not think of this point. 
So I took http://www.cplusplus.com/reference/cmath/pow/ as the reference and analyzed the various cases.

*"If the base is finite negative and the exponent is finite but not an integer value, it causes a domain error."*
* if (ualpha>vbeta) and u<0: impossible
* otherwise if v<0: impossible

*"If both base and exponent are zero, it may also cause a domain error on certain implementations."*
* If (ualpha>vbeta) and u==0 and alpha_==1.
* otherwise if v==0 and beta_==1.

Both cases can happen. I first added the corresponding unit tests and did not observe any crash. This does not imply that the code is failsafe, given that the reference mentions "on certain implementations". This is why I added a specific treatment. 

*"If base is zero and exponent is negative, it may cause a domain error or a pole error (or none, depending on the library implementation)."*
* If (ualpha>vbeta) and u==0 and alpha_>1: impossible
* otherwise if v==0 and beta_>1: impossible

*"The function may also cause a range error if the result is too great or too small to be represented by a value of the return type."*
* I do not see how this could be possible, given that both the base and the exponent are in [0,1].

I think that all cases are now correctly implemented.

PR #1293, opened 2019-10-17
TBB: Disable OpenBLAS threading

jschueller:
This disables openblas threading when TBB is enabled.
I need confirmation that this fixes the openblas threading conflict.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ option (USE_COTIRE                   "Use cotire for unity builds"              
 option (USE_IPOPT                    "Use Ipopt for MINLP problems"                                          ON)
 option (USE_BONMIN                   "Use Bonmin for MINLP problems"                                         ON)
 option (USE_SPECTRA                  "Use Spectra for eigenvalues computation"                               ON)
+option (USE_OPENMP                   "Use OpenMP to disable threading"                                       ON)
+option (USE_OPENBLAS                 "Use OpenBLAS to disable threading"                                     ON)
 
 option (BUILD_PYTHON                 "Build the python module for the library"                               ON)
 option (BUILD_SHARED_LIBS            "Build shared libraries"                                                ON)
@@ -354,6 +356,15 @@ if (USE_IPOPT)
   endif ()
 endif ()
 
+if (USE_OPENMP)
+  find_package (OpenMP)
+  if (OPENMP_FOUND)
+    set (OPENTURNS_HAVE_OPENMP TRUE)
+    set (CMAKE_CXX_FLAGS "${OpenMP_CXX_FLAGS} ${CMAKE_CXX_FLAGS}")
+    set (CMAKE_SHARED_LINKER_FLAGS "${OpenMP_CXX_FLAGS} ${CMAKE_SHARED_LINKER_FLAGS}")
+  endif ()
+endif ()
+
 if (MSVC)
   # Disable some warnings
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4197 /wd4244 /wd4251 /wd4267 /wd4996")
@@ -420,6 +431,17 @@ if (OPENTURNS_HAVE_DLFCN_H)
     list (APPEND OPENTURNS_PRIVATE_LIBRARIES ${LIBDL_LIBRARIES})
   endif ()
   check_function_exists (dladdr OPENTURNS_HAVE_DLADDR)
+  set (CMAKE_REQUIRED_LIBRARIES)
+endif ()
+
+if (USE_OPENBLAS)
+  set (CMAKE_REQUIRED_LIBRARIES ${LAPACK_LIBRARIES})
+  check_function_exists("openblas_set_num_threads" OPENTURNS_HAVE_OPENBLAS_SET_NUM_THREADS)
+  check_function_exists("goto_get_num_procs" OPENTURNS_HAVE_GOTO_GET_NUM_PROCS)
+  if (OPENTURNS_HAVE_OPENBLAS_SET_NUM_THREADS AND OPENTURNS_HAVE_GOTO_GET_NUM_PROCS)
+    set (OPENTURNS_HAVE_OPENBLAS TRUE)
+  endif ()
+  set (CMAKE_REQUIRED_LIBRARIES)
 endif ()
 
 # Some useful macros to ease CMakeLists.txt file writing
--- a/lib/src/Base/Common/TBB.cxx
+++ b/lib/src/Base/Common/TBB.cxx
@@ -29,9 +29,23 @@
 
 #include "openturns/TBB.hxx"
 #include "openturns/ResourceMap.hxx"
+#include "openturns/Exception.hxx"
 
 #ifdef OPENTURNS_HAVE_TBB
 #include <tbb/task_scheduler_init.h>
+#endif
+
+#ifdef OPENTURNS_HAVE_OPENMP
+#include <omp.h>
+#endif
+
+#ifdef OPENTURNS_HAVE_OPENBLAS
+extern "C" {
+// This function is private
+int goto_get_num_procs(void);
+// This one is public but redeclare it in case regular cblas headers are used
+void openblas_set_num_threads(int num_threads);
+}
 #endif
 
 BEGIN_NAMESPACE_OPENTURNS
@@ -39,6 +53,7 @@ BEGIN_NAMESPACE_OPENTURNS
 static pthread_mutex_t TBB_InstanceMutex_;
 static TBB * TBB_P_instance_ = 0;
 static const TBB_init initializer_TBB;
+static UnsignedInteger TBB_NumberOfThreads_ = 1;
 
 #ifdef OPENTURNS_HAVE_TBB
 tbb::task_scheduler_init * TBB_P_scheduler_ = 0;
@@ -56,12 +71,18 @@ Bool TBB::IsAvailable()
 
 void TBB::SetNumberOfThreads(const UnsignedInteger numberOfThreads)
 {
+  if (!numberOfThreads)
+    throw InvalidArgumentException(HERE) << "Number of threads must be positive";
 #ifdef OPENTURNS_HAVE_TBB
   delete TBB_P_scheduler_;
   TBB_P_scheduler_ = new tbb::task_scheduler_init(numberOfThreads);
-#else
-  (void)numberOfThreads;
 #endif
+  TBB_NumberOfThreads_ = numberOfThreads;
+}
+
+UnsignedInteger TBB::GetNumberOfThreads()
+{
+  return TBB_NumberOfThreads_;
 }
 
 void TBB::Enable()
@@ -90,6 +111,7 @@ TBB_init::TBB_init()
 #else
     pthread_mutex_init( &TBB_InstanceMutex_, NULL );
 #endif
+
     TBB_P_instance_ = new TBB;
   }
   TBB::Enable();
@@ -107,5 +129,39 @@ TBB_init::~TBB_init()
 #endif /* OPENTURNS_HAVE_TBB */
 }
 
+
+TBBContext::TBBContext()
+: ompNumThreads_(0)
+, openblasNumThreads_(0)
+{
+  if (TBB::GetNumberOfThreads() > 1)
+  {
+    // disable threading
+#ifdef OPENTURNS_HAVE_OPENMP
+    ompNumThreads_ = omp_get_max_threads();
+    omp_set_num_threads(1);
+#endif
+#ifdef OPENTURNS_HAVE_OPENBLAS
+    openblasNumThreads_ = goto_get_num_procs();
+    openblas_set_num_threads(1);
+#endif
+  }
+  (void) ompNumThreads_;
+  (void) openblasNumThreads_;
+}
+
+TBBContext::~TBBContext()
+{
+  if (TBB::GetNumberOfThreads() > 1)
+  {
+    // restore threading
+#ifdef OPENTURNS_HAVE_OPENMP
+  omp_set_num_threads(ompNumThreads_);
+#endif
+#ifdef OPENTURNS_HAVE_OPENBLAS
+  openblas_set_num_threads(openblasNumThreads_);
+#endif
+  }
+}
 
 END_NAMESPACE_OPENTURNS
--- a/lib/include/OTconfig.h.in
+++ b/lib/include/OTconfig.h.in
@@ -94,6 +94,12 @@
 /* Define to 1 if you have the CMinpack library. */
 #cmakedefine OPENTURNS_HAVE_CMINPACK
 
+/* Define to 1 if you have the OpenMP specification. */
+#cmakedefine OPENTURNS_HAVE_OPENMP
+
+/* Define to 1 if you have the openblas library. */
+#cmakedefine OPENTURNS_HAVE_OPENBLAS
+
 /* Define to 1 if you have the <stdlib.h> header file. */
 #cmakedefine OPENTURNS_HAVE_STDLIB_H
 
--- a/lib/src/Base/Common/openturns/TBB.hxx
+++ b/lib/src/Base/Common/openturns/TBB.hxx
@@ -106,6 +106,17 @@ void parallel_sort( ITERATOR first, ITERATOR last )
 
 BEGIN_NAMESPACE_OPENTURNS
 
+// disables blas threading inside TBB parallel regions
+class TBBContext
+{
+public:
+  TBBContext();
+  ~TBBContext();
+private:
+  int ompNumThreads_;
+  int openblasNumThreads_;
+};
+
 class OT_API TBB
 {
 public:
@@ -131,6 +142,7 @@ public:
   static inline
   void ParallelFor( UnsignedInteger from, UnsignedInteger to, const BODY & body, std::size_t gs = 1 )
   {
+    TBBContext context;
     tbb::parallel_for( tbb::blocked_range<UnsignedInteger>( from, to, gs ), body );
   }
 
@@ -138,6 +150,7 @@ public:
   static inline
   void ParallelReduce( UnsignedInteger from, UnsignedInteger to, BODY & body, std::size_t gs = 1)
   {
+    TBBContext context;
     tbb::parallel_reduce( tbb::blocked_range<UnsignedInteger>( from, to, gs ), body );
   }
 
@@ -145,16 +158,22 @@ public:
   static inline
   void ParallelSort( ITERATOR first, ITERATOR last )
   {
+    TBBContext context;
     tbb::parallel_sort( first, last );
   }
 
+  /* Whether TBB is available */
   static Bool IsAvailable();
+
+  /* Enable/disable */
   static void Enable();
   static void Disable();
 
-private:
+  /* Accessor to the maximum number of threads */
   static void SetNumberOfThreads(const UnsignedInteger numberOfThreads);
+  static UnsignedInteger GetNumberOfThreads();
 
+private:
   friend struct TBB_init;
 
 }; /* end class TBB */
--- a/python/src/TBB.i
+++ b/python/src/TBB.i
@@ -6,6 +6,8 @@
 
 %include TBB_doc.i
 
+%ignore OT::TBBContext;
+
 %ignore OT::TBB::Split;
 %ignore OT::TBB_init;
 
--- a/python/src/TBB_doc.i.in
+++ b/python/src/TBB_doc.i.in
@@ -1,5 +1,8 @@
 %feature("docstring") OT::TBB
-"Threading parameters."
+"Threading parameters.
+
+By default the number of threads is the value associated to the `parallel-threads` key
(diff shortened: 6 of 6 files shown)


lib/src/Base/Common/TBB.cxx
```diff
@@ -85,6 +88,12 @@ TBB_init::TBB_init()
 #else
     pthread_mutex_init( &TBB_InstanceMutex_, NULL );
 #endif
+
+#if defined(OPENTURNS_HAVE_TBB) && defined(_OPENMP)
+    // we dont want threading at multiple levels (eg openblas)
+    omp_set_num_threads(1);
```
sofianehaddad: how to reactivate `openmp` ?

jschueller: I did not add a method to do that

regislebrun: So the number of openmp threads is fixed at compile time, no way for the user to benefit from blas parallelism even if he does not use the TBB/OpenMP combo. Maybe an entry in ResourceMap could be beneficial here...

regislebrun: I am currently compiling your branch to perform some tests...

jschueller: A resourcemap entry would be unusable here as if we want it to be the default, it has to happen at the import.
But, this was just of proof of concept: maybe having this as the default behavior is not acceptable.
I propose to add a method, TBB::DisableOpenMPThreads() to to that,
and eventually TBB::EnableOpenMPThreads()

sofianehaddad: @jschueller yes and apply these methods inside `computeParallelPDF(sample)` for example

regislebrun:
@sofianehaddad It may be a good idea to have a look [here](https://github.com/jschueller/openturns/pull/5) to see how badly TBB and OpenMP interact. IMO OpenMP should be deactivated even on Windows. Most of the time is spent discretizing the covariance model as shown by perf/flamegraph, so OpenMP is of no advantage on any platform for kriging, and it induces many bad interactions in many other places in OT.

So +1 for this PR for me

sofianehaddad:
-1 for me
Let us make things clear. Problem you are mentioning happens when **locally** we use both `TBB` and `OpenMP`. For example, when using `Distribution.computeParallelPDF(Sample)` 
In general, `TBB` and `OpenMP` are used differently without conflicts.

In the PR you suppose to simply **disable** `OpenMP` which is of "nonsense". We need to master the few classes/methods where we might have conflicts instead.

With this PR, suppose you are performing `KrigingAlgorithm`. `TBB` is enabled, `OpenMP` disabled. Every iteration of the likelihood function requires a Cholesky factorisation, which will be done sequentially. It is clearly a regression. 
IMHO we need to think about how to handle the few cases...

lib/src/Base/Common/TBB.cxx
regislebrun: In my opinion OpenMP is used only in low-level classes, even in process-related algorithms (including kriging) the blocking factor is the discretization of the covariance, where OpenMP is of no help. For large problems, there is much more to gain using hmat than lapack, so let's keep the SSE2 part of OpenBLAS and forget about its OpenMP prallelism!

jschueller:
@sofianehaddad kriging does not use tbb in the learning step
but I seee your point : I dont remember if we had a testcase for this

regislebrun:
@sofianehaddad Do you have a single example where disabling OpenMP hurts performance on kriging? It is absolutely not what I observe on any test case! The factorization step is mostly not noticeable compared to the discretization step, which is done at each step of the learning phase, and if you are working on really large examples where it becomes to be noticeable then switch to hmat as it will gives you a HUGE improvement.
In *theory* having parallel blas should improve kriging, but *in practice* it is a parallel discretization which provides speed-up.

jschueller:
I pushed a slightly better version: now omp is disabled only during a tbb parallel region

regislebrun:
@jschueller @sofianehaddad Do you have a set of benchmarks to support the fact that *it is a slightler version* or *it is clearly a regression*? Nothing is clear in this landscape! See my last experiment regarding Gu, in all cases activating OpenMP is the bad option, either with TBB activated or not. I have a benchmark to support my point, so please provide one to support yours.

regislebrun:
Or give an option to disable OpenMP locally, globaly or to enable it everytime ;-)!

regislebrun:
Here is an updated benchmark:
timings

I use the first 100,200,...,2500 points in the database and perform 5 consecutive creation&run of KrigingAlgorithm, with TBB enabled and with or without OpenMP, using either LAPACK or HMAT. As we can see, on my computer the combe TBB+OpenMP starts to become competitive wrt TBB alone only at the very end of the experiment, when HMAT starts to become competitive too. So from my side the extra heat is not worth the speedup ;-)!

I also observed that from one run to the other, the HMAT timing can change a lot. For example for size=2500 the five consecutive runs give:
```
  size= 2500
  rep= 0
  t= 112.67556071281433
  rep= 1
  t= 91.74262022972107
  rep= 2
  t= 299.84150862693787
  rep= 3
  t= 108.26588368415833
  rep= 4
  t= 112.25342988967896
```
It is probably because AcaRandom... is somewhat random! But a better understanding of this behaviour is needed.

jschueller:
Beware, the convergence history and the number of calls to the likelihood function depends on the values attained which can differ when TBB is enabled, or when TNC is used (I had to use another algo, for example neldermead from nlopt and disabled TBB to measure correctly the impact of #1414).
Also, hmat internally uses rand() in several places, but I could not figure if it has anything to do with acarandom, and did not have to explicitely call srand yet.
I have not verified, but maybe the blas calls are also non-deterministic with openblas when allowing openmp threads.

regislebrun:
Here are the results I get with your branch:
timings
No actual change here... I may be doing something wrong with my benchmark, I will have a closer look at your recommendations. Nevertheless, the script looks quite natural and gives poor results.

jschueller:
You're right, the openblas_set_num_threads seems to have has no effect here, only omp_set_num_threads is effective, I'm going back to my previous commit.

sofianehaddad:
Seems strange
@jschueller @regislebrun with elementary checks (meaning before going through the `KrigingAlgorithm`), we might evaluate the performance of some operations.
Consider the following script : 
```
from __future__ import print_function
import os
import sys

from time import time
import openturns as ot
import numpy as np

ot.Log.Show(ot.Log.ALL)

def assemble_covariance(mesh, rho, covariance_model):
    # Some statistics
    vertices = mesh.getVertices()
    spatial_range = mesh.getVertices().computeRange()
    covariance_model.setScale(rho * spatial_range)
    # Number of vectors over which to check the factorization
    outputDimension = covariance_model.getOutputDimension()
    full_dimension = vertices.getSize() * outputDimension
    T_assemblage = -1.0
    failed = False
    t0 = time()
    mat = covariance_model.discretize(vertices)
    T_assemblage = time() - t0
    return failed, T_assemblage, mat

def factor_covariance(mat):
    # Some statistics
    failed = False
    try:
        t0 = time()
        L = mat.computeCholesky()
        T_factorisation = time() - t0
    except :
        failed = True
    return failed, T_factorisation

def build_mesh(ID, input_dimension):
    """
    Build regular mesh 
    """

    mesher = ot.IntervalMesher([ID]*input_dimension)
    lowerbound = [-1.0] * input_dimension
    upperBound = [1] * input_dimension
    interval = ot.Interval(lowerbound, upperBound)
    mesh= mesher.build(interval)
    return mesh

ID = 21
rho = 0.5
input_dimension = 3
covariance_model = ot.SquaredExponential([1.0] * input_dimension)

mesh = build_mesh(ID, input_dimension)

print("Mesh : ID=", ID, "vertices=", mesh.getVerticesNumber(), "simplices=", mesh.getSimplicesNumber())

print("*" * 50)
print("#" * 50)
print("=" * 50)
WARNING = '\033[91m'
OKGREEN = '\033[92m'
ENDC = '\033[0m'
msg = "ID=" + str(ID) + " rho=" + str(rho) + " covariance_model=" + str(covariance_model)
print(msg)
try:
    failed, T_assemblage, mat = assemble_covariance(mesh, rho, covariance_model)
    msg2 = "failed? " + str(failed) + " T_assemblage=" + format(T_assemblage, '7.2f')
    print(OKGREEN + msg2 + ENDC)

    failed, T_factorisation = factor_covariance(mat)
    msg2 = "failed? " + str(failed) + " T_facto=" + format(T_factorisation, '7.2f') + "s"
    print(OKGREEN + msg2 + ENDC)

except:
    failed = True
if failed:
    msg2 = "failed? " + str(failed)
    print(WARNING + msg2 + ENDC)
```

We can notice that using `OPENBLAS_NUM_THREADS=1` before launching the script (using current `master`) increase drastically CPU time for factorisation. 
I guess we might have similar results with this branch when relying on `openblas` configuration. 
I will check with 3c745cabc9abe97f9e7f37f1746e2e2fa0ea5f09

jschueller:
in fact both apis control different things: http://scv.bu.edu/examples/python/examples/parallel/multithread/ so we might want to have both, like hmat does

sofianehaddad:
@jschueller yes agreed

jschueller:
@sofianehaddad Here are some results with your script
```
# master:
# failed? False T_assemblage=   4.98
# failed? False T_facto=   6.66s

# master+OPENBLAS_NUM_THREADS=1:
# failed? False T_assemblage=   5.22
# failed? False T_facto=   6.93s

# tbb_omp:
# failed? False T_assemblage=   4.94
# failed? False T_facto=   6.51s
```

sofianehaddad:
@jschueller all in all it's strange!

With my 8 cores, I have :
```
#master:
failed? False T_assemblage=   5.82
failed? False T_facto=   8.56s

#master OPENBLAS_NUM_THREADS=1:
failed? False T_assemblage=   5.82
failed? False T_facto=  40.73s

#tbb_omp (https://github.com/sofianehaddad/openturns/tree/tbb_omp)
failed? False T_assemblage=   5.83
failed? False T_facto=   8.61s

#tbb_omp ( 2a89dca96767b70836b534b6329316c5db94bcec )
failed? False T_assemblage=   5.80
failed? False T_facto=   8.61s
```
Big difference here! If I add `ot.TBB.Disable()`, `T_assemblage` value moves from `5.8s` to `10`.

jschueller:
I get a much slower time with OMP_NUM_THREADS=1, but not with OPENBLAS_NUM_THREADS=1.

sofianehaddad:
0.3.5
I will update it today

regislebrun:
Comme les modèles stationnaires n'ont pas de version parallèle de discretize(), j'ai un peu modifié le benchmark de Sofiane pour utiliser FractionalBrownianMotionModel (en 1D donc) et mieux voir ce qui se passe avec les TBBs.

Avec TBB:
```
master:
ID=10001 rho=0.5 covariance_model=FractionalBrownianMotionModel(scale=[0.5], amplitude=[1], exponent=[0.7])
failed? False T_assemblage=   6.60
failed? False T_facto=   4.94s

master, OMP_NUM_THREADS=1:
ID=10001 rho=0.5 covariance_model=FractionalBrownianMotionModel(scale=[0.5], amplitude=[1], exponent=[0.7])
failed? False T_assemblage=   6.46
failed? False T_facto=   8.36s

tbb_omp:
ID=10001 rho=0.5 covariance_model=FractionalBrownianMotionModel(scale=[0.5], amplitude=[1], exponent=[0.7])
failed? False T_assemblage=   4.17
failed? False T_facto=   4.61s
```
Sans TBB:
```
master:
ID=10001 rho=0.5 covariance_model=FractionalBrownianMotionModel(scale=[0.5], amplitude=[1], exponent=[0.7])
failed? False T_assemblage=  16.50
failed? False T_facto=   4.88s

master, OMP_NUM_THREADS=1:
ID=10001 rho=0.5 covariance_model=FractionalBrownianMotionModel(scale=[0.5], amplitude=[1], exponent=[0.7])
failed? False T_assemblage=  16.10
failed? False T_facto=   8.60s

tbb_omp:
ID=10001 rho=0.5 covariance_model=FractionalBrownianMotionModel(scale=[0.5], amplitude=[1], exponent=[0.7])
failed? False T_assemblage=  15.78
failed? False T_facto=   4.78s
```
Je vois bien le gain d'un facteur 4 des TBBs pour l'assemblage, et celui d'un facteur 2 pour la factorisation. Sur ma machine la parallélisation TBB semble plus efficace que celle via omp.

regislebrun:
J'ai un autre bench, plus proche de ce que je fais avec le LIP6: échantillonner et évaluer la PDF d'un modèle multivarié. J'ai pris le cas d'une loi de Student, dont l'évaluation de la PDF fait appel aux blas, et dont l'évaluation sur un échantillon est parallélisée via TBBs:
```
from openturns import *
from time import time

#TBB.Disable()
size = 1000000
dim = 100
R = CorrelationMatrix(dim)
for i in range(dim-1):
    R[i, i+1]=1.0/dim
    
d = Student(5.2, [0.0]*dim, [1.0]*dim, R)
t0 = time()
sample = d.getSample(size)
t1 = time()
dt = t1 - t0
print("sampling speed=", size / dt, "rng/s, t=", dt, "s")
t0 = time()
pdf = d.computePDF(sample)
t1 = time()
dt = t1 - t0
print("pdf speed=", size / dt, "eval/s, t=", dt, "s")
```
Avec TBBs:
```
master:
sampling speed= 72721.5429359592 rng/s, t= 13.751083374023438 s
pdf speed= 229129.36550898076 eval/s, t= 4.364346742630005 s

master, OMP_NUM_THREADS=1:
sampling speed= 228007.12891933596 rng/s, t= 4.3858277797698975 s
pdf speed= 1329881.5146867696 eval/s, t= 0.7519466876983643 s

tbb_omp:
sampling speed= 73990.15303720353 rng/s, t= 13.51531195640564 s
pdf speed= 231154.29891584697 eval/s, t= 4.326114654541016 s
```
Là c'est clair, net et précis: l'utilisation d'OMP pourrit complètement la performance, et le correctif de la branche ne sauve pas l'affaire.
Sans TBB:
```
master:
sampling speed= 72438.25539666906 rng/s, t= 13.804860353469849 s
pdf speed= 143447.82677245606 eval/s, t= 6.971175670623779 s

master, OMP_NUM_THREADS=1:
sampling speed= 226957.73518873978 rng/s, t= 4.40610671043396 s
pdf speed= 436496.5872166555 eval/s, t= 2.290968656539917 s

tbb_omp:
sampling speed= 100569.58686806972 rng/s, t= 9.943363904953003 s
pdf speed= 150757.18544586207 eval/s, t= 6.633183002471924 s
```
Voilà pourquoi je trouve que les TBBs font un job utile et pas omp.

A un moment ce serait bien de pouvoir suivre l'évolution des performances régulièrement, mais je ne vois pas trop comment faire ça bien...

jschueller:
@regislebrun tu utilise quelle version de ma branche ? la dernière où je désactive openmp+openblas ou l'ancienne version ?

regislebrun:
@jschueller j'utilise la dernière version. Sofiane m'a fait remarquer dans un autre fil de discussion que la détection de certaines fonctions d'OpenBLAS ne se passait pas bien:
[CMakeError.log](https://github.com/openturns/openturns/files/4387745/CMakeError.log)

J'utilise la toute dernière version d'OpenBLAS, fraîchement compilée. Sofiane est en train d'investiguer le problème.

jschueller:
oh, oui, il faut débugger ça

jschueller:
Je crois que ce qui ne vas pas c'est que j'utilise BLAS_LIBRARIES au lieu de LAPACK_LIBRARIES dans le CMakeLists.

jschueller:
Je viens de faire la modification.

CMakeLists.txt
```diff
@@ -420,6 +431,17 @@ if (OPENTURNS_HAVE_DLFCN_H)
     list (APPEND OPENTURNS_PRIVATE_LIBRARIES ${LIBDL_LIBRARIES})
   endif ()
   check_function_exists (dladdr OPENTURNS_HAVE_DLADDR)
+  set (CMAKE_REQUIRED_LIBRARIES)
+endif ()
+
+if (USE_OPENBLAS)
+  set (CMAKE_REQUIRED_LIBRARIES ${LAPACK_LIBRARIES})
```
sofianehaddad: great

sofianehaddad:
@jschueller le problème était là. Je lui ai indiqué comment régler le souci
Maintenant avec cette PR ça doit être réglé!

regislebrun:
Ok cool je teste ça.

jschueller:
Ah il y a un autre souci: j'utilisais omp_get_num_threads au lieu de omp_get_max_threads, du coup après une première région tbb ça revenait à désactiver openmp complètement dans les régions non tbb, je corrige, et il faut refaire des benchs :[

jschueller:
il y a encore un souci: on désactive les threads openmp/openblas même quand tbb n'est pas activé ou même disponible, décidément

jschueller:
J'ai lancé le benchmark de Régis pour différentes valeurs de OMP_NUM_THREADS sans changer le nombre thread tbb ni OPENBLAS_NUM_THREADS, sur ma machine on voit bien le conflit tbb/omp lorsque le nombre de threads openmp augmente:


```
master:
TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 1
OPENBLAS_NUM_THREADS= 1
sampling speed= 117890.58051140679 rng/s, t= 8.482441902160645 s
pdf speed= 1792179.5023537208 eval/s, t= 0.5579798221588135 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 2
OPENBLAS_NUM_THREADS= 1
sampling speed= 87703.63322247176 rng/s, t= 11.402036190032959 s
pdf speed= 9846.616723867317 eval/s, t= 101.55772566795349 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 4
OPENBLAS_NUM_THREADS= 1
sampling speed= 73007.84025440953 rng/s, t= 13.697159051895142 s
pdf speed= 3666.1186303969 eval/s, t= 272.76804184913635 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 8
OPENBLAS_NUM_THREADS= 1
sampling speed= 67145.41689062859 rng/s, t= 14.893049240112305 s
pdf speed= 1202.826899592036 eval/s, t= 831.3748223781586 s

...

tbb_omp:
TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 1
OPENBLAS_NUM_THREADS= 1
sampling speed= 104753.63986895306 rng/s, t= 9.546207666397095 s
pdf speed= 1982100.950337651 eval/s, t= 0.5045151710510254 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 2
OPENBLAS_NUM_THREADS= 1
sampling speed= 86072.31154900606 rng/s, t= 11.618138074874878 s
pdf speed= 2056992.3106235412 eval/s, t= 0.4861466884613037 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 4
OPENBLAS_NUM_THREADS= 1
sampling speed= 77344.68569233846 rng/s, t= 12.929136514663696 s
pdf speed= 1979687.256627979 eval/s, t= 0.5051302909851074 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 8
OPENBLAS_NUM_THREADS= 1
sampling speed= 67477.71337596423 rng/s, t= 14.819707870483398 s
pdf speed= 2041867.4187048567 eval/s, t= 0.4897477626800537 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 16
OPENBLAS_NUM_THREADS= 1
sampling speed= 57747.20815101279 rng/s, t= 17.31685447692871 s
pdf speed= 2040317.9240661748 eval/s, t= 0.49011969566345215 s

TBB_NUM_THREADS= 40
OMP_NUM_THREADS= 32
OPENBLAS_NUM_THREADS= 1
sampling speed= 59253.17227659405 rng/s, t= 16.876733541488647 s
pdf speed= 2033899.7516242377 eval/s, t= 0.491666316986084 s
```

La différence entre master et tbb_omp sur le cas nominal OMP_NUM_THREADS= 1 est je pense expliquée par les appels aux fonctions omp_get_max_threads/omp_set_num_threads, goto_get_num_procs/openblas_set_num_threads car en enlevant omp_get_max_threads et goto_get_num_procs je réduis déjà l'écart de quasiment moitié (de 500ms environ), mais du coup ce n'est plus changeable dynamiquement, donc je laisse comme ça pour l'instant.

jschueller:
Ah et j'ai ajouté des accesseurs aux nombres de threads TBB:
```
/* Accessor to the maximum number of threads */
  static void SetNumberOfThreads(const UnsignedInteger numberOfThreads);
  static UnsignedInteger GetNumberOfThreads();
```

J'avais aussi ajouté des clés pour les nombres de threads omp/openblas que l'on autorise à l'intérieur des régions parallèles TBB, mais rien ne vaut jamais mieux que 1, même en divisant le nombre de threads tbb.

regislebrun:
Pour moi c'est ok, il y a assez de moyens de régler les choses pour qu'on s'en sorte avec cette PR.

Issue #1214, opened 2019-09-06
There is no example with IntegrationStrategy on a database


mbaudin47:
The current advanced chaos example:

http://openturns.github.io/openturns/master/examples/meta_modeling/functional_chaos_advanced_ctors.html

does not show how to create a chaos from integration on a database. 

Moreover, it does not show that other integration strategies such as Monte-Carlo, LHS or Low Discrepancy are available.

I have an example at hand, that I wish to share. I would assign this ticket to myself if I could.

josephmure:
Did #1296 fix this?

sofianehaddad:
See https://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_chaos_cantilever_beam_integration.html and https://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_advanced_ctors.html

mbaudin47:
> Did #1296 fix this?

No. I created an example using `GaussProductExperiment`, which is the most classical way of performing chaos with integration. However, it does not show how to create a chaos from integration and a Monte-Carlo or LHS or QMC design. Notice that this is possible, since the input argument of `IntegrationStrategy` is an `Experiment`. This is a very interesting property of the OpenTURNS library, which is very difficult to implement in other contexts with full generality. Hence, it highly deserves more than one example. For example, in addition to request mentioned in the title of this issue, we could show the convergence (or lack of convergence) on specific functions when the sample size increases : Ishigami would be a nice example, but g(x)=|x| or the g-Sobol' function would show something that is not shown in the library yet. Please feel free to pick the relevant part of the code at otbenchmark:

https://github.com/mbaudin47/otbenchmark/blob/master/otbenchmark/GSobolSensitivity.py

josephmure:
I think I will let you handle this ;)

mbaudin47:
The example I have is at:

https://github.com/mbaudin47/otsupgalilee-eleve/blob/master/5-Chaos/Exercice-chaos-cantilever-beam-Integration.ipynb

mbaudin47:
This issue will be closed when PR #2081 is merged, because the required example is provided in python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cantilever_beam_integration.py.

PR #1294, opened 2019-10-17
Improved and create new chaos examples

mbaudin47:
This PR improves the polynomial chaos and S.A. examples. 

The order of the examples in the "Metamodeling" section is changed so that simpler examples come first. 

* chaos_ishigami.ipynb: this example should come first, given that it is classical, easy and simple.
* chaos_cantilever_beam_integration.ipynb: *new*. It shows the integration method, which had no example before. **Please check** that the polynomial degree selected in `FixedStrategy` is consistent with respect to the parameters of the `GaussProductExperiment` class. 
* functional_chaos.ipynb: fixed. The example was wrong in the sense that the metamodel validation was so poor that the fact that the predictivity was close to zero could not be seen. To simplify the task, I changed the model a bit so that high polynomial degrees are not required anymore. This is still a difficult example, nonlinear with high interactions. 
* functional_chaos_advanced_ctors.ipynb: beautified by printing the comments in the text.
* functional_chaos_exploitation.ipynb
* functional_chaos_database.ipynb
* chaos_ishigami_grouped_indices.ipynb
* chaos_build_distribution.ipynb: *new* - the role of this function was not clear enough from the existing doc.
* functional_chaos_graphs.ipynb: simplified
* chaos_draw_validation.ipynb: separated from functional_chaos_graphs.ipynb and updated to use the dedicated MetamodelValidation class.

Also clarified the "Reliability, Sensitivity" section.
* Separated Reliability from Sensitivity examples in the table of contents.
* Clarified the example which shows the S.A. by sampling for the Ishigami test case in sensitivity_sobol.ipynb. Clarified how to estimate first and total indices (without second order indices) and including second order indices. Showed how to change the estimator.


mbaudin47:
This only partially solves #1214: this is because the current example in chaos_cantilever_beam_integration.ipynb does not show how to use MonteCarlo, LHS or low discrepancy experiments in the integration method.

python/doc/examples/meta_modeling/functional_chaos.ipynb
```diff
@@ -11,11 +11,21 @@
...
    "metadata": {},
    "source": [
-    "In this basic example we are going to create a global approximation of a model response using functional chaos.\n",
+    "In this example we are going to create a global approximation of a model response using functional chaos.\n",
     "\n",
-    "Here\n",
+    "Let $h$ be the function defined by:\n",
+    "$$\n",
+    "g(\\mathbf{x}) = \\left[\\cos(x_1 + x_2), (x_2 + 1) e^{x_1}\\right]\n",
+    "$$\n",
```
jschueller: In this PR too, can you make sure you add a blank line after each closing $$ ?

mbaudin47: Oups. Fixed.

jschueller:
see #1296

PR #1295, opened 2019-10-18
Improved HMatrix

regislebrun:
Now the regularization is done based on the largest eigenvalue of the matrix. Thanks to Romain for finding this trick, which allows to skip the many regularization loops done in the current implementation. The default values of some ResourceMap keys have also been updated to more meaningful values.

A quick benchmark of the discretization and factorization of an exponential kernel (scale = 0.1xsize of the mesh) against several industrial meshes show the following results (lapack+openblas vs hmat-oss on a 4 cores laptop with 64GB ram):
```
2268 vertices: lapack 0.26s error 2.9e-16 vs hmat 1.44s error 1.4e-3
9644 vertices: lapack 6.25s error 5.2e-16 vs hmat 29s error 1.5e-3
18292 vertices: lapack 30s error 7.2e-16 vs hmat 86s error 2.9e-3
134331 vertices: lapack out of memory vs hmat 4300s error ??? (no lapack reference)
```
with master, we get no answer for the two last cases using hmat: the old regularization procedure was unable to stabilize the matrix!
Using the (non-free) parallel version of hmat, the hmat factorization outperforms the lapack factorization by 25% for 18292 vertices. While lapack needs more than 64GB for 134331 vertices, hmat needs only 400MB!

There are still room for improvements:
  - accelerate the assembly procedure, which always take a significant part of the process
  - implement a new compression method better suited for covariance kernels (ongoing work by Romain)
  - ...


lib/src/Base/Stat/HMatrixImplementation.cxx
```diff
@@ -307,6 +307,34 @@ void HMatrixImplementation::assemble(const HMatrixTensorRealAssemblyFunction& f,
...
+      currentEigenVector = nextEigenVector / nextEigenValue;
+      gemv('N', 1.0, currentEigenVector, 0.0, nextEigenVector);
+      nextEigenValue = nextEigenVector.norm();
+      precision = std::abs(nextEigenValue - currentEigenValue);
+      found = precision <= epsilon * nextEigenValue;
+      LOGDEBUG(OSS() << "(" << iteration << ") precison=" << precision << ", relative precision=" << precision / nextEigenValue << ", found=" << found);
+      if (found) break;
+      currentEigenValue = nextEigenValue;
+    }
+  if (!found) LOGWARN(OSS() << "Did not reached the target relative precision=" << epsilon << ", got relative precision=" << precision / nextEigenValue);
```
jschueller: Did not reach

lib/src/Base/Stat/HMatrixImplementation.cxx
```diff
@@ -320,19 +348,40 @@ void HMatrixImplementation::factorize(const String& method)
...
-    context.progress = NULL;
-    static_cast<hmat_interface_t*>(hmatInterface_)->factorize_generic(static_cast<hmat_matrix_t*>(hmat_), &context);
-  }
-  catch (std::exception& ex)
-  {
-    // hmat::LapackException is not yet exported
-    throw InternalException(HERE) << ex.what();
-  }
+  // Compute a reasonable regularization factor
+  Scalar lambda = 2.0 * computeApproximateLargestEigenValue() * ResourceMap::GetAsScalar("HMatrix-AssemblyEpsilon");
```
sofianehaddad: Care we should have the **same** value that is contained in https://github.com/openturns/openturns/pull/1295/files#diff-227275030f245b905d3e4ae25d6b9054R815

lib/src/Base/Stat/HMatrixImplementation.cxx
```diff
@@ -320,19 +348,40 @@ void HMatrixImplementation::factorize(const String& method)
...
+	  LOGDEBUG("Factorization ok");
+	}
+      catch (std::exception& ex)
+	{
+	  // hmat::LapackException is not yet exported
+	  msg = ex.what();
+	  // Double the current regularization factor by adding it another time
+	  addIdentity(lambda);
+	  // And double its value for next loop
+	  lambda += lambda;
```
sofianehaddad: ok for 2 but I think this value should be in a way compared to CovModel(0) ?
@poncetr ?

regislebrun: I don't think so as the initial value for lambda is the eigenvalue of the covariance matrix, ie it is already in the correct scale.

lib/src/Base/Stat/HMatrixImplementation.cxx
regislebrun: I must have missed something here... I don't see the point!

lib/src/Base/Stat/HMatrixImplementation.cxx
sofianehaddad: If we consider a `CorrelationModel` instead of `CovarianceModel` (`amplitude` set to 1),
we might easily have maximal eigen value around 1000 if we consider medium/large sized matrices.
Thus adding `1` to the diagonal for regularising the matrix is statistically "incorrect" as we are changing the covariance structure.

lib/src/Base/Stat/HMatrixImplementation.cxx
sofianehaddad: If `HMatrixParameters` contains an `AssemblyEpsilon` different, it might cause troubles.

regislebrun: Ok I see your point: I completely overlooked this epsilon_. In fact, it should be removed as the new development is here exactly to provide automatically the best possible value for it, taking into account both the characteristics of the matrix (its largest eigenvalue) and the compression error (ie the level of accuracy one can reasonably expect from the compression)

lib/src/Base/Stat/HMatrixImplementation.cxx
regislebrun: I still don't understand where you see a 1 being added to the diagonal. Tell me where I am wrong:
  + I want to factorize a matrix obtained through the discretization of a given covariance model, regardless of its scaling (ie covariance or correlation)
  + To this end I first compress it using an h-matrix representation. It introduces a numerical noise which can break the definite positiveness of the matrix
  + To fix this problem I look for the optimal scaling of the regularization: something which allows me to shift all the eigenvalues in the positive side. It is a proportion of the largest eigenvalue, modulated by the amount of numerical noise generated by the compression

There is no scaling done between the assembly and the factorization of the matrix. Even if you scaled it before or if you scale it after the factorization, if the factorization succeeded it is ok for the other computations.

sofianehaddad: Agreed. IMO you should also control the regularization : `2 * eps * lamba_max` is ok but with the loop we should assert that final `translating` parameter is less or equal than `0.5 * amplitude`

We should control positiveness and "behavioural" aspects

regislebrun: There is definitely additional work to do to ensure that the global problem is well-posed, but IMO:
  + it is well beyond the scope of this PR (eg: is the mesh adapted to the sampling of the covariance model? In particular, do we have clusters of vertices much denser than the scale of the model?)
  + this PR already improves a lot wrt the current implementation, which is essentially useless for the user -> this point alone should be enough to adopt this PR!

BTW, a control based on the eigenvalue takes into account both the amplitude of the model (linked to the output) and the geometry of the sampling points (linked to the scale, ie the input), so the final check should incorporate both aspects. I will add something like this. At the end of the regularization loop, which should end after1 or 2 iterations (not taking into account round-off errors), either the factorization is done, which is a certificate for positiveness, or an exception is thrown, allowing the user to take an action. Could you elaborate a little bit more on the behavioural aspects? I don't see the mathematics beyond this sentence ;-)!

poncetr: I confirm the sclaing of the regularization, as explained by Regis.

From our tests we have observed a linear relationship between the prescribed assembly error of the HMatrix (set by the key ```HMatrix-AssemblyEpsilon``` of the ```ResourceMap```). The regression coefficient is in practice close to one.

If a much larger coefficient is required to ensure the positivity of the approximation of the covariance matrix, then it is likely that the HMatrix library
failed to satisfy the prescribed accuracy, which may especially happen with the "ACA partial" compression method (set by the key ```HMatrix-CompressionMethod``` of the '''ResourceMap```). In this case, it is indeed necessary to abort the computation and suggest to the user to use a more reliable compression algorithm.

FYI, a randomized ACA algorithm is being tested to prevent these possible problems, with the same complexity as the "ACA partial" algorithm.

jschueller:
Should this PR make it into 1.14rc1 ? I see there's a conflict.

sofianehaddad:
@jschueller yes for 1.14rc1

lib/src/Base/Stat/HMatrixImplementation.cxx
```diff
@@ -307,6 +307,34 @@ void HMatrixImplementation::assemble(const HMatrixTensorRealAssemblyFunction& f,
...
 
+/* Compute an approximation of the largest eigenValue (in magnitude) using power iterations */
+Scalar HMatrixImplementation::computeApproximateLargestEigenValue(const Scalar epsilon)
+{
+  const UnsignedInteger dimension = getNbRows();
+  Point currentEigenVector(dimension, 1.0);
+  Point nextEigenVector(dimension);
+  gemv('N', 1.0, currentEigenVector, 0.0, nextEigenVector);
+  Scalar nextEigenValue = nextEigenVector.norm();
+  Scalar currentEigenValue = nextEigenValue / std::sqrt(dimension);
```
jschueller: this line fails on appveyor, it should be:
std::sqrt(1.0 * dimension)

regislebrun: Ok done.

PR #1304, opened 2019-10-29
Various enhancements

jschueller:


regislebrun:
+1 I fixed the segfault in a local branch the same way, and I switched from Function to UniVariateFunction (with additional evolutions of Solver) to gain a 3x speed in the transformation. I will update my code after the merge of this PR

PR #1313, opened 2019-11-10
Fixed distribution arithmetic

regislebrun:
The weight was not taken into account if a simplification had occured in a sum or a difference.


regislebrun:
@jschueller This PR should be backported to the 1.14 version ASAP!

jschueller:
do we need to add a testcase ?

sofianehaddad:
@jschueller @regislebrun I think we should update https://github.com/openturns/openturns/blob/master/python/test/t_Distribution_arithmetic.py and add a specific testcase

regislebrun:
done

PR #1315, opened 2019-11-12
GLM: Use LogMaxScalar for logdet hmat variant

jschueller:
This way its consistent with the lapack variant

cc @josephmure


regislebrun:
+1

PR #1316, opened 2019-11-13
Hypergeometric: Fix computeScalarQtle when dirac

jschueller:
There are several combinations of parameters {n,k,m} where the distribution
is equivalent to a dirac (k=0, m=0).
In these configurations computeScalarQuantile can exit early to avoid
freezes.


regislebrun:
+1 Good catch!

PR #1332, opened 2019-11-20
Various enhancements

jschueller:


regislebrun:
+1

lib/src/Uncertainty/Distribution/WeibullMinMuSigma.cxx
```diff
@@ -104,7 +104,7 @@ Matrix WeibullMinMuSigma::gradient() const
   const Scalar dbetadgamma = ((operator()(newParameters + pt) - operator()(newParameters - pt)) / (2 * epsilon))[0];
   const Scalar dalphadgamma = ((operator()(newParameters + pt) - operator()(newParameters - pt)) / (2 * epsilon))[1];
 
-  SquareMatrix nativeParametersGradient(IdentityMatrix(3));
+  Matrix nativeParametersGradient(IdentityMatrix(3));
```
sofianehaddad: Matrix nativeParametersGradient(3, 2); ??

jschueller: no, last colum is [0,0,1]

sofianehaddad: oh yes indeed I missed it.
So what do we need to explicitly set `Matrix` instead of `SquareMatrix`?

jschueller: It was related to the return type being Matrix, but I dont remember the warning name.

PR #1366, opened 2020-02-01
HMat: Default to ACARandom compression method

### LINKED ISSUE #1365: hmat compression method selection
- HMatrixImplementation::setKey does not handle aca random
- HMatrixImplementation::setKey does not take the same strings as HMatParameters
- I could be a good idea to reuse the same strings as the hmat c++ enum names instead of introducing new ones: 'ACA full' -> 'AcaFull', 'ACA random' -> 'AcaRandom' ...
- Could we default to the aca random compression method in the ResourceMap instead of aca full ?
- I'd prefer the ResourceMap key to use a string instead of an int, it's more explicit

Tell me what do you think, and I'll make the corresponding changes.
cc @poncetr

jschueller:
Closes #1365


regislebrun:
+1

PR #1367, opened 2020-02-03
Speed up KL P1 SVD decomposition

jschueller:
- The use the iterative SVD decomposition method (from https://github.com/yixuan/spectra, inspired by arpack) to speed up KL P1 decomposition : now we can specify if we want only N modes instead of the full decomposition, like LAPACK does.
- The sparse storage (new SparseMatrix class implementing CSC storage) of the gram matrix helps reduce the memory usage.


regislebrun:
+1 good job! 10^6 vertices is no more a problem with HMAT+sparse

PR #1387, opened 2020-02-26
Fixed Distribution arithmetic

regislebrun:
The simplification mechanism was wrong when the RandomMixture generated by the distribution arithmetic was reduced to a unique atom and a non-zero constant: the constant was not taken into account. As a result, eg Poisson(5.0)+1 was aqual to Poisson(5.0).


sofianehaddad:
Could you please add `Poisson(5.0)+1` to `t_Disstribution_arithmetic` ?
Otherwise it is ok

regislebrun:
done

PR #1398, opened 2020-03-06
Fixed UserDefined

regislebrun:
The univariate user defined distribution was reporting a non-independent copula. It was thus impossible to build the associated Gauss quadrature rule using GaussProductExperiment.


jschueller:
could you rebase your PRs ?

PR #1414, opened 2020-03-14
HMat: Save state during factorization

### LINKED ISSUE #1404: Bug in KrigingAlgorithm+hmat-oss
The following script:
```
import openturns as ot
from time import time

ot.Log.Show(ot.Log.INFO)
ot.ResourceMap.SetAsString( "KrigingAlgorithm-LinearAlgebra", "HMAT")
#ot.ResourceMap.SetAsString("HMatrix-CompressionMethod", "Svd")
#ot.ResourceMap.SetAsString("HMatrix-CompressionMethod", "AcaFull")
#ot.ResourceMap.SetAsString("HMatrix-CompressionMethod", "AcaPartial")
#ot.ResourceMap.SetAsString("HMatrix-CompressionMethod", "AcaPlus")
ot.ResourceMap.SetAsString("HMatrix-CompressionMethod", "AcaRandom")
learningInputData  = ot.Sample.ImportFromCSVFile("InputBug.csv")
learningOutputData = ot.Sample.ImportFromCSVFile("OutputBug.csv")
basis = ot.ConstantBasisFactory(learningInputData.getDimension()).build()
algo = ot.KrigingAlgorithm(learningInputData, learningOutputData, ot.AbsoluteExponential(), basis)
t0 = time()
algo.run()
t1 = time()
print("t=", t1 - t0, "s")
```
produces the following error:
```
[hmat] assert failure isLower at /home/regis/hmat-oss-1.6.1/src/h_matrix.cpp:1931 void hmat::HMatrix<T>::lltDecomposition(hmat_progress_t*) [with T = double; hmat_progress_t = hmat_progress_struct]
/home/regis/local/hmat/lib/libhmat-oss.so(+0x8d798)[0x7f841d979798]
/home/regis/local/hmat/lib/libhmat-oss.so(+0x8d84f)[0x7f841d97984f]
/home/regis/local/hmat/lib/libhmat-oss.so(+0xadc49)[0x7f841d999c49]
/home/regis/local/hmat/lib/libhmat-oss.so(_ZN4hmat15RecursionMatrixIdNS_7HMatrixIdEEE25recursiveLltDecompositionEP20hmat_progress_struct+0x99)[0x7f841d9998c9]
/home/regis/local/hmat/lib/libhmat-oss.so(_ZN4hmat7HMatrixIdE16lltDecompositionEP20hmat_progress_struct+0x5a)[0x7f841d999bea]
/home/regis/local/hmat/lib/libhmat-oss.so(_ZN4hmat13HMatInterfaceIdE9factorizeE20hmat_factorization_tP20hmat_progress_struct+0x47)[0x7f841d9b0567]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZN2OT21HMatrixImplementation9factorizeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x302)[0x7f841fc2cc42]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT27GeneralLinearModelAlgorithm33computeHMatLogDeterminantCholeskyEv+0x20f)[0x7f84200f43ef]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT27GeneralLinearModelAlgorithm27computeReducedLogLikelihoodERKNS_5PointE+0x1ef)[0x7f84200f6c6f]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT27GeneralLinearModelAlgorithm30ReducedLogLikelihoodEvaluationclERKNS_5PointE+0x10)[0x7f84200fd220]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT24EvaluationImplementationclERKNS_6SampleE+0x29a)[0x7f841f9b621a]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT10EvaluationclERKNS_6SampleE+0x11)[0x7f841f9bb2a1]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT17MemoizeEvaluationclERKNS_6SampleE+0x96b)[0x7f841fa9cbeb]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT10EvaluationclERKNS_6SampleE+0x11)[0x7f841f9bb2a1]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT35NonCenteredFiniteDifferenceGradient8gradientERKNS_5PointE+0xd1)[0x7f841f768bb1]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT8Gradient8gradientERKNS_5PointE+0xe)[0x7f841f9bdaee]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT22FunctionImplementation8gradientERKNS_5PointE+0x40)[0x7f841f9ab580]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZNK2OT8Function8gradientERKNS_5PointE+0x11)[0x7f841f9a78c1]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZN2OT3TNC27ComputeObjectiveAndGradientEPdS1_S1_Pv+0x3c5)[0x7f841fbff095]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(tnc+0x213)[0x7f841fbf74b3]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZN2OT3TNC3runEv+0x826)[0x7f841fc00e26]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZN2OT27GeneralLinearModelAlgorithm28maximizeReducedLogLikelihoodEv+0x125)[0x7f84200fb4e5]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZN2OT27GeneralLinearModelAlgorithm3runEv+0x59)[0x7f84200fbe39]
/home/regis/openturns-adm/build/install/lib/libOT.so.0(_ZN2OT16KrigingAlgorithm3runEv+0x4c)[0x7f84200e3f1c]
/home/regis/openturns-adm/build/install/lib64/python3.7/site-packages/openturns/_metamodel.so(+0xe8873)[0x7f84059d8873]
/lib64/libpython3.7m.so.1.0(_PyMethodDef_RawFastCallKeywords+0x2f9)[0x7f842199da29]
/lib64/libpython3.7m.so.1.0(_PyCFunction_FastCallKeywords+0x25)[0x7f842199dae5]
/lib64/libpython3.7m.so.1.0(_PyEval_EvalFrameDefault+0x7170)[0x7f842193a830]
/lib64/libpython3.7m.so.1.0(+0x685f3)[0x7f84219335f3]
/lib64/libpython3.7m.so.1.0(_PyEval_EvalFrameDefault+0x6f0d)[0x7f842193a5cd]
/lib64/libpython3.7m.so.1.0(_PyEval_EvalCodeWithName+0x8fe)[0x7f84219e6b3e]
/lib64/libpython3.7m.so.1.0(PyEval_EvalCodeEx+0x3e)[0x7f84219e6dae]
```
I get the same error with AcaPartial and AcaPlus. I am using the current master (1c261ba38b0f30b4a67da5eb5fe60f9a737a3619) with the latest version of hamat-oss (1.6.1). It looks like an hmat-oss bug but not sure as I can't figure out how to reproduce it outside of OT.
Here are the data files:
[InputBug.zip](https://github.com/openturns/openturns/files/4315109/InputBug.zip)
[OutputBug.zip](https://github.com/openturns/openturns/files/4315111/OutputBug.zip)

jschueller:
We have to create a copy as the factorization can leave the matrix in a broken state and should not be reused
Introduces a 30% performance penalty on Regis testcase.

Closes #1404

### DIFF
--- a/lib/src/Base/Stat/HMatrixImplementation.cxx
+++ b/lib/src/Base/Stat/HMatrixImplementation.cxx
@@ -215,6 +215,32 @@ HMatrixImplementation::HMatrixImplementation(const HMatrixImplementation& other)
 #endif
 }
 
+/* Copy assignment operator */
+HMatrixImplementation & HMatrixImplementation::operator=(const HMatrixImplementation & other)
+{
+#ifdef OPENTURNS_HAVE_HMAT
+  if (this != &other)
+  {
+    // destroy current
+    if (hmatInterface_ != NULL && hmat_ != NULL)
+      static_cast<hmat_interface_t*>(hmatInterface_)->destroy(static_cast<hmat_matrix_t*>(hmat_));
+
+    if (other.hmatClusterTree_.get())
+    {
+      hmat_cluster_tree_t* ptr_other_ct = static_cast<hmat_cluster_tree_t*>(other.hmatClusterTree_.get()->get());
+      hmatClusterTree_ = new HMatrixClusterTree(hmat_copy_cluster_tree(ptr_other_ct), other.hmatClusterTree_.get()->getSize());
+      hmat_cluster_tree_t* ptr_ct = static_cast<hmat_cluster_tree_t*>(hmatClusterTree_.get()->get());
+
+      hmat_interface_t* ptr_interface = static_cast<hmat_interface_t*>(other.hmatInterface_);
+      hmat_matrix_t* hmat_copy = ptr_interface->copy(static_cast<hmat_matrix_t*>(other.hmat_));
+      ptr_interface->set_cluster_trees(hmat_copy, ptr_ct, ptr_ct);
+      hmat_ = hmat_copy;
+    }
+  }
+#endif
+  return *this;
+}
+
 /* Virtual constructor */
 HMatrixImplementation * HMatrixImplementation::clone() const
 {
@@ -350,6 +376,11 @@ void HMatrixImplementation::factorize(const String& method)
 
   // Compute a reasonable regularization factor
   Scalar lambda = 2.0 * computeApproximateLargestEigenValue() * ResourceMap::GetAsScalar("HMatrix-AssemblyEpsilon");
+
+  // create a backup copy as the factorization can leave the matrix in a broken state and should not be reused
+  hmat_matrix_t* hmatBackup = static_cast<hmat_matrix_t*>(hmat_);
+  hmat_ = static_cast<hmat_interface_t*>(hmatInterface_)->copy(static_cast<hmat_matrix_t*>(hmatBackup));
+
   // Do regularization
   addIdentity(lambda);
   Bool done = false;
@@ -365,6 +396,10 @@ void HMatrixImplementation::factorize(const String& method)
       context.factorization = fact_method;
       context.progress = NULL;
       static_cast<hmat_interface_t*>(hmatInterface_)->factorize_generic(static_cast<hmat_matrix_t*>(hmat_), &context);
+
+      // ditch the original instance
+      static_cast<hmat_interface_t*>(hmatInterface_)->destroy(static_cast<hmat_matrix_t*>(hmatBackup));
+
       done = true;
       LOGDEBUG("Factorization ok");
     }
@@ -372,6 +407,11 @@ void HMatrixImplementation::factorize(const String& method)
     {
       // hmat::LapackException is not yet exported
       msg = ex.what();
+
+      // ditch the copy and restart from the original instance
+      static_cast<hmat_interface_t*>(hmatInterface_)->destroy(static_cast<hmat_matrix_t*>(hmat_));
+      hmat_ = static_cast<hmat_interface_t*>(hmatInterface_)->copy(static_cast<hmat_matrix_t*>(hmatBackup));
+
       // Double the current regularization factor by adding it another time
       addIdentity(lambda);
       // And double its value for next loop
--- a/lib/src/Base/Stat/openturns/HMatrixImplementation.hxx
+++ b/lib/src/Base/Stat/openturns/HMatrixImplementation.hxx
@@ -121,6 +121,9 @@ public:
   /** Virtual copy constructor */
   virtual HMatrixImplementation * clone() const;
 
+  /** Copy assignment operator */
+  HMatrixImplementation& operator=(const HMatrixImplementation& rhs);
+
   // Destructor
   virtual ~HMatrixImplementation();
 
--- a/python/src/HMatrixImplementation.i
+++ b/python/src/HMatrixImplementation.i
@@ -4,6 +4,8 @@
 #include "openturns/HMatrixImplementation.hxx"
 %}
 
+%ignore OT::HMatrixImplementation::operator=(const HMatrixImplementation &);
+
 %include HMatrixImplementation_doc.i
 
 %include openturns/HMatrixImplementation.hxx


jschueller:
cc @jeromerobert

jeromerobert:
> Introduces a 30% performance penalty on Regis testcase.

That's odd. 30% compared to what ?

jschueller:
compared to not saving the matrix data for the regularization loop (I increased the initial regulation value to get only 1 iteration to allow comparison with the unfixed code)

regislebrun:
In my view there are two strategies for the iterations:
+ to always make a copy of the compressed matrix before to try to factorize it, and try as many regularization steps as needed to get something good. Each factorization try is 30% more expansive than the current code, including the first try.
+ to try a factorization with no copy and to reassemble and recompress the matrix from scratch if it fails. Each try but the first is then 100% more expansive.

So if the cost of a factorization is C, for N tries, the first strategy costs 1.3xCxE[N], the second strategy costs C+1.3x(E[N]-1), where E[N] is the expected number of tries to get a success.

According to @poncetr , N>1 is quite rare if the regularization starts with a factor of 2\lambda_max\epsilon_assembly, it is why I limited the number of tries to Nmax=2 by defaults (see "HMatrix-FactorizationIterations" in ResourceMap).

So what are the options to improve the current PR?
- to take 4\lambda_max\epsilon_assembly, perform a unique try and throw an error if it fails
- to keep 2\lambda_max\epsilon_assembly at the initial step and make the copy of the compressed matrix more efficient
- to implement the second strategy, with no copy but a rebuilt from scratch if the first try fails
- ...

Anyway, a penalty of 30% is not such a big deal when any other method (ie LAPACK, AcaFull, or Svd) is more than 10x slower even for moderately large cases.

jeromerobert:
`copy` is almost instant compared to factorization and assembly, so I'm not sure to understand where does this 30% comes from. Here are the timing I get on a part of the execution with the PR:
```
INF - Reduced log-likelihood=-4589.86
HMatrix assembly: 1792ms
HMatrix copy/create: 4ms
HMatrix copy=: 4ms
HMatrix factorization: 1905ms
HMatrix copy=: 4ms
HMatrix factorization: 1905ms
HMatrix copy=: 5ms
HMatrix factorization: 1960ms
INF - Reduced log-likelihood=-4648.96
HMatrix assembly: 1793ms
HMatrix copy/create: 4ms
HMatrix copy=: 5ms
HMatrix factorization: 1936ms
INF - Reduced log-likelihood=-4588.16
HMatrix assembly: 1801ms
HMatrix copy/create: 4ms
HMatrix copy=: 6ms
HMatrix factorization: 1941ms
INF - Reduced log-likelihood=-4585.57
HMatrix assembly: 1795ms
HMatrix copy/create: 5ms
HMatrix copy=: 4ms
HMatrix factorization: 1936ms
HMatrix copy=: 4ms
HMatrix factorization: 1936ms
HMatrix copy=: 5ms
HMatrix factorization: 1954ms
```

jeromerobert:
By the way, is this expected to have 2 `copy` after the assembly ? (one in the = operator and one in the `HMatrixImplementation` constructor)

poncetr:
I agree with N=1 in general. Yet :
- this in only true on the cases I have looked at. Yet there exist (pathological) cases where this won't be true.
- for robustness, recopy is better
- @jeromerobert should be right on the the time repartition : recopy should be 
Thus, a recopy for each regularization seems to be the way to go for me.

regislebrun:
Ok 30% is the cost of the second regularization loop. No surprise here, and the price is not too high to pay for a robust and automatic regularization. If the user knows that its case is pathological he can always increase the initial regularization factor.
It may be a good idea to see if we can avoid one of the two copies found by Jerome.

Otherwise, +1

jschueller:
Forget my benchmark, I did not see the test is not reproducible as hmat uses rand(), also TBB usage may introduce randomness.

jschueller:
Actually the cost is much lower, I get less that 10% ; the benchmark was not deterministic with TBB and TNC enabled.

regislebrun:
Using this branch and taking the 2100 first points in the database, during the optimization loop of GLMAlgo I get:
```
DBG - Factorization, regularization loop 0, regularization factor=0.015181
DBG - Must increase the regularization to 0.0303619 because In lltDecomposition, diagonal index 16 has an invalid value -6.35911
DBG - Factorization, regularization loop 1, regularization factor=0.0303619
DBG - Must increase the regularization to 0.0607238 because In lltDecomposition, diagonal index 16 has an invalid value -6.35911
DBG - Factorization, regularization loop 2, regularization factor=0.0607238
DBG - Must increase the regularization to 0.121448 because In lltDecomposition, diagonal index 16 has an invalid value -5.44898
DBG - Factorization, regularization loop 3, regularization factor=0.121448
DBG - Must increase the regularization to 0.242895 because In lltDecomposition, diagonal index 16 has an invalid value -4.32041
DBG - Factorization, regularization loop 4, regularization factor=0.242895
DBG - Must increase the regularization to 0.48579 because In lltDecomposition, diagonal index 16 has an invalid value -3.08551
DBG - Factorization, regularization loop 5, regularization factor=0.48579
DBG - Must increase the regularization to 0.971581 because In lltDecomposition, diagonal index 16 has an invalid value -1.86069
DBG - Factorization, regularization loop 6, regularization factor=0.971581
DBG - Must increase the regularization to 1.94316 because In lltDecomposition, diagonal index 16 has an invalid value -0.694915
DBG - Factorization, regularization loop 7, regularization factor=1.94316
DBG - Must increase the regularization to 3.88632 because In lltDecomposition, diagonal index 22 has an invalid value -505.752
DBG - Factorization, regularization loop 8, regularization factor=3.88632
DBG - Must increase the regularization to 7.77265 because In lltDecomposition, diagonal index 22 has an invalid value -112.553
DBG - Factorization, regularization loop 9, regularization factor=7.77265
DBG - Must increase the regularization to 15.5453 because In lltDecomposition, diagonal index 22 has an invalid value -47.8497
  File "bug.py", line 14, in <module>
    algo.run()
  File "/home/regis/openturns-julien-ref/build_hmat_factorize/install/lib64/python3.7/site-packages/openturns/metamodel.py", line 5517, in run
    return _metamodel.KrigingAlgorithm_run(self)
RuntimeError: InternalException : In lltDecomposition, diagonal index 22 has an invalid value -47.8497
```
For this data set, most of the optimization steps need only one regularization, but some of them need up to 14 steps (ie more than the default 10). It must be something strange in the data...

jschueller:
@regislebrun I was working on changes to copy the matrix only once as @jeromerobert noticed.
I cannot reproduce the issue with the first 2100 points with the latest code.

jschueller:
Also I suspect my copy constructor was not correct.
@jeromerobert is the hmat_interface_t* instance tied to the hmat_matrix_t* instance ? What about hmat_cluster_tree_t* ?
```
 /* Copy assignment operator */
HMatrixImplementation & HMatrixImplementation::operator=(const HMatrixImplementation & other)
{
#ifdef OPENTURNS_HAVE_HMAT
  if (this != &other)
  {
    // destroy current
    if (hmatInterface_ != NULL && hmat_ != NULL)
      static_cast<hmat_interface_t*>(hmatInterface_)->destroy(static_cast<hmat_matrix_t*>(hmat_));

    if (other.hmatClusterTree_.get())
    {
      hmat_cluster_tree_t* ptr_other_ct = static_cast<hmat_cluster_tree_t*>(other.hmatClusterTree_.get()->get());
      hmatClusterTree_ = new HMatrixClusterTree(hmat_copy_cluster_tree(ptr_other_ct), other.hmatClusterTree_.get()->getSize());
      hmat_cluster_tree_t* ptr_ct = static_cast<hmat_cluster_tree_t*>(hmatClusterTree_.get()->get());

      hmat_interface_t* ptr_interface = static_cast<hmat_interface_t*>(other.hmatInterface_);
      hmat_matrix_t* hmat_copy = ptr_interface->copy(static_cast<hmat_matrix_t*>(other.hmat_));
      ptr_interface->set_cluster_trees(hmat_copy, ptr_ct, ptr_ct);
      hmat_ = hmat_copy;
    }
  }
#endif
  return *this;
}
```

jschueller:
@regislebrun also HMatrix-FactorizationIterations is not 2 but 10 currently

jeromerobert:
A `hmat_matrix_t` must always be created/updated/destroyed with the same `hmat_interface_t`. There is basically one `hmat_interface_t` by scalar type (typically 4). It would be strange to create more than 4 `hmat_interface_t` but it should work. 2 different `hmat_interface_t` of the same scalar type could be used with a `hmat_matrix_t` if their scalar type are compatible.

A `hmat_cluster_tree_t` can be shared between multiple `hmat_matrix_t`. A `hmat_cluster_tree_t` must be kept alive until it has no more `hmat_matrix_t` pointing to it else you'll a segfault or worst. It's possible to bind a `hmat_cluster_tree_t` life cycle to a `hmat_matrix_t` using the own_cluster_trees function. In that case destroying a `hmat_matrix_t` will also destroyed its owned cluster trees.

jschueller:
@jeromerobert thanks

jschueller:
moved to #1402

PR #1420, opened 2020-03-23
Improved gu

regislebrun:
This PR proposes the following changes:
- Improved GLMAlgo: Now the algebra is done using better suited operators. The parameter gradient is also scaled in-place.
- I added a way to check if the active parameters are exactly the scale parameters and I have illustrated its use by providing a much more efficient implementation of parameterGradient() in AbsoluteExponential


jschueller:
thanks, I included your algebra commit in my branch.
The parameterGradient improvements could be done in a separate branch to be included before.

mbaudin47:
Is this PR still required? Is'nt the relevant part already merged in the master?

josephmure:
Yes, this PR is still required, nothing has been merged. I was supposed to do someting, I think, but I've never gotten around to it and now I no longer remember what it was. Do you remember, @regislebrun @jschueller ?

Anyway, I think this work needs to wait until the great GLM/Kriging overhaul that was tabled by the cotech.

jschueller:
I dont remember a task assigned to you, but I know we proposed to include this in a module in the mean time.

PR #1429, opened 2020-03-31
Improve kriging

regislebrun:
This is an attempt to speed-up kriging metamodel evaluation. On the following benchmark I get a 2x speedup:
```
import openturns as ot
from time import time

ot.TBB.Disable()
size = 2000
study = ot.Study("Bench_kriging.xml.gz")
study.load()
metaModel = ot.Function()
study.fillObject("metaModel", metaModel)
print("meta-model=", metaModel)
inSample = ot.Normal(metaModel.getInputDimension()).getSample(size)
t0 = time()
res = metaModel(inSample)
t1 = time()
print("Speed=", size / (t1 - t0), "evals/s")
```
I get the following results:
```
master:
Speed= 370.9020406313455 evals/s
this branch:
Speed= 754.3130215312524 evals/s
```
The compressed XML file is attached. It is based on a sample of size 7819 of a function from R^5 to R.
[Bench_kriging.xml.gz](https://github.com/openturns/openturns/files/4411571/Bench_kriging.xml.gz)
There are further possible improvements, mainly to avoid the creation of temporary Point during the evaluation of the covariance model between each point of the learning database and the point given as a input of the meta-model. I don't see how to do this without exposing an iterator on the internal storage of the sample (data_.begin()). Any advice is welcome!


lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
```diff
@@ -141,14 +141,58 @@ struct KrigingEvaluationPointFunctor
...
 }; // struct KrigingEvaluationPointFunctor
 
+struct KrigingEvaluationPointFunctor1D
+{
+  const Point & input_;
+  const KrigingEvaluation & evaluation_;
+  Scalar accumulator_;
+
+  KrigingEvaluationPointFunctor1D(const Point & input,
+                                  const KrigingEvaluation & evaluation)
```
jschueller, line 151: I dont see where that ctor is needed

sofianehaddad: In https://github.com/openturns/openturns/pull/1429/files#diff-41b14220e03d778dfa73ce9f1e261d7eR186

sofianehaddad:
> There are further possible improvements, mainly to avoid the creation of temporary Point during the evaluation of the covariance model between each point of the learning database and the point given as a input of the meta-model. I don't see how to do this without exposing an iterator on the internal storage of the sample (data_.begin()). Any advice is welcome!

@regislebrun please have a look at https://github.com/sofianehaddad/openturns/commits/improved_kriging

Still some improvements & checks are needed

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
jschueller: I meant the copy ctor(other, TBB::Split)

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
```diff
@@ -141,14 +141,58 @@ struct KrigingEvaluationPointFunctor
...
+    : input_(input)
+    , evaluation_(evaluation)
+    , accumulator_(evaluation.getOutputDimension())
+  {}
+
+  KrigingEvaluationPointFunctor1D(const KrigingEvaluationPointFunctor1D & other,
+                                  TBB::Split)
```
jschueller, line 158: here

jschueller:
Note that the openmp equivalent is a one liner with the same speedup:
```
  const UnsignedInteger trainingSize = inputSample_.getSize();
  Matrix R(dimension, trainingSize * dimension);
#pragma omp parallel for schedule(dynamic) firstprivate(R)
  {
    for (UnsignedInteger j = 0; j < trainingSize; ++j)
    {
      const CovarianceMatrix localCovariance(covarianceModel_(inS[i], inputSample_[j]));
      for (UnsignedInteger columnIndex = 0; columnIndex < dimension; ++ columnIndex)
        for (UnsignedInteger rowIndex = 0; rowIndex < dimension; ++ rowIndex)
          R(rowIndex, columnIndex + j * dimension) = localCovariance(rowIndex, columnIndex);
    }
    result[i] = R * gamma_.getImplementation()->getData();
  }
```

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
sofianehaddad: I see. It is useless but it is a copy of `KrigingEvaluationPointFunctor`, which has the same actors ;-)

regislebrun: At this point it looks quite good. With master we have:
perf_master

We see that 2/3 of the time spent in KrigingEvaluation::operator() is spent in CovarianceModelImplementation::operator() and 1/3 is spent in the creation/destruction of temporary Point. On the benchmark I get 370 evals/s

By replacing the call to CovarianceModelImplementation::operator() by a call to CovarianceModelImplementation::computeAsScalar(), we avoid a costly multiplication by a 1x1 matrix and also the creation of some Point to transfer arguments and values:
perf_regis

Now, half of the time is spent computing exp(), which is much better. On the benchmark I get 761 evals/s. But now the time spent in the creation/destruction of Point, even if lower than before, represents half of the total time.

Thanks to Sofiane's improvement, using iterators instead of duplicating the data, we completely remove this part:
perf_sofiane

Now, approximately 90% of the time is spent in computeAsScalar(), in which 2/3 of the time is spent computing exp(). On the benchmark I get 1570 evals/s. The final speed-up is 4.24x, not so bad for the very common case of scalar output!

regislebrun:
Ready to go? Thks to Sofiane for its very efficient additional input!

jschueller:
the conditionedguassian process fails on mingw with different values, I wonder what happens since tbb is not even used

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
```diff
@@ -201,15 +247,56 @@ struct KrigingEvaluationSampleFunctor
...
+  inline void operator()( const TBB::BlockedRange<UnsignedInteger> & r ) const
+  {
+    const UnsignedInteger inputDimension = input_.getDimension();
+    const UnsignedInteger start = r.begin();
+    const UnsignedInteger size = r.end() - start;
+    Scalar value = 0.0;
+    for (UnsignedInteger i = 0; i != size; ++i)
+    {
+      for (UnsignedInteger j = 0; j < trainingSize_; ++j)
+        value += evaluation_.covarianceModel_.getImplementation()->computeAsScalar(input_.getImplementation()->data_begin() + (start + i) * inputDimension, evaluation_.inputSample_.getImplementation()->data_begin() + (j * inputDimension) ) * evaluation_.gamma_(j, 0);
```
sofianehaddad, line 277: Sorry @regislebrun I introduced an issue 
Could you fix it please ?
```
-    Scalar value = 0.0;
     for (UnsignedInteger i = 0; i != size; ++i)
     {
       for (UnsignedInteger j = 0; j < trainingSize_; ++j)
-        value += evaluation_.covarianceModel_.getImplementation()->computeAsScalar(input_.getImplementation()->data_begin() + (start + i) * inputDimension, evaluation_.inputSample_.getImplementation()
-      output_(start + i, 0) = value;
+        output_(start + i, 0) += evaluation_.covarianceModel_.getImplementation()->computeAsScalar(input_.getImplementation()->data_begin() + (start + i) * inputDimension, evaluation_.inputSample_.get
```

jschueller:
It's the first commit that fail the tests when TBB is disabled at compile time, nothing to do with mingw.

sofianehaddad:
@jschueller probably due to https://github.com/openturns/openturns/pull/1429/commits/65469f88b3104aba8d91039e76e03cf7ba0ae815#diff-41b14220e03d778dfa73ce9f1e261d7eR154 as the `accumulator` should be initialised to 0

But it was fixed

jschueller:
Ah, it was hidden in your commit :/

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
regislebrun: done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
```diff
@@ -201,15 +247,56 @@ struct KrigingEvaluationSampleFunctor
...
+      output_(start + i, 0) = value;
+    }
+  } // operator()
```
jschueller, line 280: this loop would be simpler to iterate on the global indices (begin, end), instead of (0, size) as we never need the local index : we always use start + i !

python/test/t_EfficientGlobalOptimization_std.py
```diff
@@ -149,7 +149,7 @@ def ackley(X):
 # print(result.getOutputSample())
 
 # local refinement
-algo2 = ot.NLopt(problem, 'LD_LBFGS')
+algo2 = ot.TNC(problem)
```
jschueller, line 152: then it must be reenabled in python/test/CMakeLists.txt unconditionally to nlopt:
```
if (NLopt_FOUND)
  ot_pyinstallcheck_test (NLopt_std)
  ot_pyinstallcheck_test (EfficientGlobalOptimization_std)
endif ()
```

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
regislebrun: correct. I am wondering if it would be better to increment a counter by inputDimension instead of doing a multiplication at each step.

jschueller: I'd say dont bother: the improvement should be too tiny compared to the floating point operations that cost much more. Also the compiler should already optimize this kind of stuff.

python/test/t_EfficientGlobalOptimization_std.py
regislebrun: Ok. I saw I was using a quite old version of NLopt (2.4.2). Maybe the new version is better? Do you have a feedback on it? I also planed to directly interface lbfgsb as the NLopt interface is (or was) pure crap. But it introduces another dependence, and I only found a Fortran implementation of the algorithm (and don't want to reintroduce a Fortran dependence). Do you have an idea on how to benefit from this algorithm with the same level of quality as in scipy?

jschueller: nlopt still uses the same code from luksan originally written in fortran and translated to c whereas scipy uses the nocedal implementation, in fortran.

There are several c ports around though, like
https://github.com/stephenbeckr/L-BFGS-B-C

jschueller: If you want I can have a go at it.

regislebrun: Oooppppsss I made a confusion between lbfgsb and slsqp! The problem Benoit Chaigne found when debuggin ANATOLE was that NLopt does'nt pass the stoping criteria as is to the solver, but instead do strange things leading to convergence troubles.  I don't know if lbfgsb is impacted too, but the problem seems to be at an upper level than the crude algorithms.

jschueller: this probably needs to be reported to nlopt

regislebrun: Benoit did it in May 2018. I didn't check if it has been taken into account. But it is a side story wrt kriging.

Is this PR ok?

jschueller: yes

PR #1433, opened 2020-04-06
Fix histogram

### LINKED ISSUE #1423: The computeSilvermanBandwidth of the HistogramFactory has no help
* The `computeSilvermanBandwidth` method has no entry in the table of contents : its arguments are undocumented. 

* The `computeSilvermanBandwidth` method name is not appropriate. According to Scott ("Multivariate density estimation", Scott, 1992), the optimal AMISE bandwidth was published in 1979 by Scott and 1981 by Diaconis and Freedman. Hence, I suggest to rename this method with the simpler `computeBandwidth` name.

* When the second argument `useQuantile` is True (by default), the `computeSilvermanBandwidth` method returns the robust bandwith where the standard deviation is estimated based on the quantiles. In this case, 

$$
h = \frac{IQ}{2\Phi^{-1}(0.75)} \times \left(\frac{24 \sqrt{\pi}}{n}\right)^{\frac{1}{3}}
$$

where $\Phi^{-1}$ is the quantile function of the gaussian standard distribution and $IQ$ is the inter-quartiles range:

$$
IQ = Q_3 - Q_1
$$

where 
$$
Q_3 = q(0.75), \qquad Q_1 = q(0.25).
$$

* When `useQuantile` is false, the equation is the one already present in the doc. 

* The default bandwidth should be documented as being optimal for the gaussian distribution. This is because the printed bandwidth corresponds to the minimum of the AMISE when the roughness is supposed to be the one of the gaussian distribution.

mbaudin47:
* Fixes #1423 by fixing the help page.
* Renaming the method HistogramFactory::computeSilvermanBandwidth into HistogramFactory::computeBandwidth. This breaks the compatibility with OT 1.13, but I think that this is for the best. If the renaming scheme cannot be accepted, all I have to do is to drop the commit 3324d37965b610235af187c67104b5520117f5b3

The Circle CI checks are a success, with the exception of the "Publish documentation" step which fails because of a login/password issue.


jschueller:
The 1st commit message should look like this:
```
HistogramFactory: Fix doc

Closes #1423 
```
ie dont mention the issue # in the commit title, but in the body where "Closes" allows to automatically close it upon merge

jschueller:
The ci check probably because you enable circleci on your fork.

PR #1439, opened 2020-04-08
Update GaussLegendre_doc.i.in

adutfoy:
The method integrateWithNodes was not documented. The method integrate had a wrong documentation.


jschueller:
pulled in #1441

PR #1440, opened 2020-04-08
Update IntegrationStrategy_doc.i.in

adutfoy:
Changed LeastSquaresStrategy into IntegrationStrategy in the documentation.


jschueller:
pulled in #1441

PR #1452, opened 2020-04-14
Mesh: Fix numerical issue in checkPointInSimplex

### LINKED ISSUE #1445: symbolic distribution arithmetics
this is a proposition to enhance cases where we want to compute the distribution of a*X/(X-b):
```
X = ot.Normal(100.0, 10.0)

Y1 = 100*dist_X/(dist_X-200)
Y1.setDescription(['Y1'])

f= ot.SymbolicFunction(['x'],['100*x/(x-200)'])
Y2 = ot.CompositeDistribution(f, dist_X)
Y2.setDescription(['Y2'])

graph = Y1.drawPDF()
curve2 = Y2.drawPDF().getDrawable(0)
curve2.setColor('blue')
graph.add(curve2)
graph
```
dist

In the first case we compute 1/(X-b) which is badly conditioned then the distribution of the product,
but in the second case we provide the whole function in one go, and it does what we want.

The point is we have to see the whole expression and notice X is several times in the expression before actually computing the distribution.

A solution could be creating a new distribution that aggregates the leaf distributions and the operator tree and decide whether to create a composite etc.

### LINKED ISSUE #1449: P1LagrangeInterpolation sometimes fails
Hi,
I am trying to interpolate the values of a Field defined on a Mesh m1 at the vertices of another Mesh m2, using P1LagrangeInterpolation. I can create the interpolation object but when I use it I get an exception (InvalidArgumentException with no description...). Either the two meshes are incompatible and it should be checked in the constructor, or there is a more subtile bug. The strange thing is that it occurs with no apparent regularity (eg when mesh2 has a simplex fully outside of mesh1), as the following example shows:
```
import openturns as ot

N = 5
m1 = ot.IntervalMesher([1]*2).build(ot.Interval([0.0]*2, [1.0]*2))
ns1 = m1.getSimplicesNumber()
m2 = ot.IntervalMesher([N]*2).build(ot.Interval([-1.0]*2, [2.0]*2))
ns2 = m2.getSimplicesNumber()

ot.ResourceMap.SetAsUnsignedInteger("Mesh-LargeSize", 5000)
g = m1.draw()
g.add(m2.draw())
c = ["red"]*(1+ns1) + ["blue"]*(1+ns2)
print(len(c))
g.setColors(c)
g.setTitle("Interpolation from mesh 1 to mesh 2, N=" + str(N))
view = otv.View(g, square_axes=True)
view.save("Interpolation_N_" + str(N) + ".png")
view.close()


interpolation = ot.P1LagrangeInterpolation(m1, m2, 1)
nv1 = m1.getVerticesNumber()
values = ot.Sample(nv1, [1.0])
res = interpolation(values)
print("res=", res)
```
When I use this script with `N=4` it works like a charm, corresponding to the following meshes:
Interpolation_N_4

but when I use it with `N=5` I get the exception:
Interpolation_N_5

jschueller:
Closes #1449

### DIFF
--- a/lib/src/Base/Func/P1LagrangeInterpolation.cxx
+++ b/lib/src/Base/Func/P1LagrangeInterpolation.cxx
@@ -171,7 +171,7 @@ void P1LagrangeInterpolation::computeProjection()
     {
       if (!inputMesh_.checkPointInSimplexWithCoordinates(outputVertices[i], simplexIndices[i], coordinates))
       {
-        throw InvalidArgumentException(HERE);
+        throw InvalidArgumentException(HERE) << "P1LagrangeInterpolation: point #" << i << " is not inside simplex";
       }
       IndicesCollection::const_iterator cit = simplices.cbegin_at(simplexIndices[i]);
       // Points are sorted to avoid memory gaps during matrix-matrix multiplication
--- a/lib/src/Base/Geom/Mesh.cxx
+++ b/lib/src/Base/Geom/Mesh.cxx
@@ -192,12 +192,13 @@ Bool Mesh::checkPointInSimplexWithCoordinates(const Point & point,
     Point & coordinates) const
 {
   if (index >= getSimplicesNumber()) return false;
+  const Scalar epsilon = ResourceMap::GetAsScalar("Mesh-VertexEpsilon");
   if (dimension_ == 1)
   {
     const Scalar x = point[0];
     const Scalar x0 = vertices_(simplices_(index, 0), 0);
     const Scalar x1 = vertices_(simplices_(index, 1), 0);
-    if ((x - x0) * (x - x1) > SpecFunc::ScalarEpsilon)
+    if ((x - x0) * (x - x1) > epsilon)
       return false;
     coordinates = Point(2);
     if (x0 == x1)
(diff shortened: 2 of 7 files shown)


lib/src/Base/Geom/Mesh.cxx
```diff
@@ -223,16 +223,16 @@ Bool Mesh::checkPointInSimplexWithCoordinates(const Point & point,
...
     {
       return false;
     }
+    const Scalar x = point[0] - x0;
+    const Scalar y = point[1] - y0;
     coordinates = Point(3);
     coordinates[1] = (x02 * y - y02 * x) / det;
     coordinates[2] = (x * y01 - y * x01) / det;
-    coordinates[0] = 1.0 - coordinates[1] - coordinates[2];
+    coordinates[0] = 0.5 + (0.5 - coordinates[1] - coordinates[2]);
```
regislebrun, line 235: or maybe `(0.5 - coordinates[1]) + (0.5 - coordinates[2])`, but at the end I thin we will have to introduce an epsilon to check if the given point is inside of a slightly larger simplex

lib/src/Base/Geom/Mesh.cxx
```diff
@@ -223,16 +223,16 @@ Bool Mesh::checkPointInSimplexWithCoordinates(const Point & point,
...
     return coordinates[0] >= 0.0 && coordinates[0] <= 1.0 && coordinates[1] >= 0.0 && coordinates[1] <= 1.0 && coordinates[2] >= 0.0 && coordinates[2] <= 1.0;
```
regislebrun, line 236: With an epsilon, the test would become:
```
return coordinates[0] >= -epsilon && coordinates[0] <= 1.0+epsilon && coordinates[1] >= -epsilon && coordinates[1] <= 1.0+epsilon && coordinates[2] >= -epsilon && coordinates[2] <= 1.0+epsilon;
```
and here, as the coordinates are within [0,1], a default choice of epsilon=1e-14 should be ok. In fact, it depends on the condition number of the linear function mapping the global coordinates into the barycentric coordinates.

regislebrun:
I tested your branch on the 3D test-case, and it works! Thanks a lot!

+1

lib/src/Base/Func/P1LagrangeInterpolation.cxx
```diff
@@ -171,7 +171,7 @@ void P1LagrangeInterpolation::computeProjection()
     {
       if (!inputMesh_.checkPointInSimplexWithCoordinates(outputVertices[i], simplexIndices[i], coordinates))
       {
-        throw InvalidArgumentException(HERE);
+        throw InvalidArgumentException(HERE) << "Point #" << i << " is not inside simplex";
```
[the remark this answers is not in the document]
sofianehaddad: otherwise it is ok +1

PR #1464, opened 2020-04-22
Various fixes

regislebrun:
This PR provides several small improvements:
+ Now the nested MemoizeEvaluation objects are flattened: if you encapsulate a MemoizeFunction into another MemoizeFunction, the history and caches are merged at the upper level (it was already done) but the underlying evaluation is taken from the encapsulated MemoizeFunction, otherwise both cache&history are augmented at each call, resulting into a waste of resources
+ Now the user can choose to store or to forget the collection of results corresponding to each optimization problem solved during the exploration, eg to reduce the memory footprint of the algorithm.
+ Now it is possible to select between MonteCarlo or QuasiMonteCarlo to compute entropy for high dimension models. While QuasiMonteCarlo may be more accurate for a given sample size, it was taking too long to generate a sample for distributions with no dedicated implementation of the Rosenblatt transformation (eg the EmpiricalBernsetinCopula class).
+ Now the copula property is not enforced during the learning phase, leading to a smooth objective function.
+ The computeLogPDF method is no more based on BLAS. It allows to reduce the memory footprint of the class while preserving the performance thanks to TBBs when computed over a sample.
+ The compression level can be specified at creation time for studies, allowing for a huge saving of disk space for large objects.


lib/src/Base/Optim/openturns/MultiStart.hxx
```diff
@@ -46,7 +47,8 @@ public:
 
   /** Constructor with parameters */
   MultiStart(const OptimizationAlgorithm & solver,
-             const Sample & startingPoints);
+             const Sample & startingPoints,
+             const Bool keepResults = ResourceMap::GetAsBool("MultiStart-KeepResults"));
```
jschueller, line 51: I dont like such flags being added to ctors because it prevents extending the api.
I'd prefer if we could just provide an accessor for it.

lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -737,6 +737,9 @@ void ResourceMap::loadDefaultConfiguration()
   addAsScalar("AbdoRackwitz-DefaultTau", 0.5);
   addAsUnsignedInteger("AbdoRackwitz-DefaultMaximumEvaluationNumber", 100000);
 
+  // MultiStart parameters //
+  addAsBool("MultiStart-KeepResults", true);
```
sofianehaddad, line 741: why not having in mind a `storeResult` with `UnsignedInteger` ? We might set `0` if we don't want to store any result or limit the sizing (for example, default value = `1000`)

regislebrun: I don't see the point of keeping only a part of the results. The use-case behind the storage of the results was to partition the starting points according to their corresponding optimal point, and when it is not needed I propose to deactivate it. With your proposal, it becomes less straightforward to keep everything: you have to set the bound to the size of the starting points sample, which cannot be set in the ResourceMap. Do you have a use-case where only a part of these results would be meaningful?

sofianehaddad: no use case for the moment

regislebrun:
Is this PR planed to be merged into master too?

jschueller:
sure, it's merged now

Issue #1465, opened 2020-04-27
The Sample constructor based on a list and an integer should not exist


mbaudin47:
We can create a 1D sample with:
```
sample = ot.Sample([2.0, 2.0, 1.0, 1.0, 2.0, 3.0, 1.0, 2.0, 2.0, 1.0], 1)
```
This should not be possible, since a list cannot be a structure for a sample: a 2D object should be used as input, instead of a list. Hence, the previous example should be:
```
sample = ot.Sample([[2.0], [2.0], [1.0], [1.0], [2.0], [3.0], [1.0], [2.0], [2.0], [1.0]], 1)
```
However, this example requires a significant amount of square brackets, which may explain why this constructor is available.

mbaudin47:
Before dropping this constructor, a solution should be found in order to get a simple method to create such a sample.
Below is an attempt based on numpy:
```
>>> import openturns as ot
>>> import numpy as np
>>> data = np.array([2.0, 2.0, 1.0, 1.0, 2.0, 3.0, 1.0, 2.0, 2.0, 1.0])
>>> data = data.reshape(10, 1)
>>> sample = ot.Sample(data)
>>> sample.getSize()
10
>>> sample.getDimension()
1
```

sofianehaddad:
I think we have all elements to generate a sample properly.
Notice that the C++ API implements a `setData` method. Thus we might add a static method to the `Sample` class (`BuildFromSequence`) for example. 
```
>>> import openturns as ot
>>> import numpy as np
>>> data = np.array([2.0, 2.0, 1.0, 1.0, 2.0, 3.0, 1.0, 2.0, 2.0, 1.0])
>>> sample = ot.Sample.BuildFromSequence(data, 10, 1)
>>> sample.getSize()
10
>>> sample.getDimension()
1
```
What do you think?

jschueller:
Agreed

mbaudin47:
True : `BuildFromSequence` is the very correct way of handling this. However, it is not documented, which is a bug: #1589

mbaudin47:
The `BuildFromSequence` actually does not exist: I suggest to implement it because the commit 9613095d12bbcc2a3d6b2fce8e14fcb482739fc3 is a regression which makes the library more complicated to use.

mbaudin47:
@sofianehaddad : The release is coming, which introduces a change in the API. The following code works with 1.15:

```
n = 20
x = ot.Sample(range(n), 1)
```
but fails in the current master. 
[...]

PR #1467, opened 2020-04-27
Modernize code

jschueller:
Modernize code:
- Require C++11: (std::vector.data(), inherited constructors, std::atomic, std::shared_ptr), drop Atomic, SharedPointer etc, cleanup Log, also override would be useful
- Require cmake >=3.2 and drop all the quirks
- Require swig >=3.x and drop all the quirks
- Require matplotlib =>1.3.1 and drop all the quirks

Drop Sample(sequence, int) ctor (#1465)


jschueller:
no, cmake just doesnt mention it missing anymore, the idea is to send the message that muparser is not so important these days

Issue #1455, opened 2020-04-17
GeneralLinearModelAlgorithm::setCovarianceModel could lead to unexpected behavior of parameter optimization in KrigingAlgorithm


tbittar:
Suppose that we construct a metamodel for a function and then add points sequentially. We compute a new metamodel at every iteration but reestimate the hyperparameters every k>1 iterations. We use the previous hyperparameters as starting points for this hyperparameter optimization. The problem is that the hyperparameters might not be updated.

Consider the following example:
```python

import openturns as ot

# We wish to create a metamodel for f
f = ot.SymbolicFunction(['x'],  ['sin(x)'])

ot.RandomGenerator.SetSeed(0)
inputSample = ot.Uniform(0, 10).getSample(10)
outputSample = f(sampleX)

basis = ot.ConstantBasisFactory(1).build()
covarianceModel = ot.SquaredExponential([1.0])

# First metamodel computation without parameter optimization
algo = ot.KrigingAlgorithm(sampleX, sampleY, covarianceModel, basis)
algo.setOptimizeParameters(False)
algo.run()

result = algo.getResult()
print(result.getCovarianceModel())  # Return SquaredExponential(scale=[1], amplitude=[1]), ok

# Now compute the metamodel using the covariance model that is output by the previous KrigingAlgorithm, this time we aim at optimizing the hyperparameters
algo2 = ot.KrigingAlgorithm(sampleX, sampleY, result.getCovarianceModel(), basis)
algo2.setOptimizeParameters(True)  # We expect to optimize the covariance parameters
algo2.run()

result2 = algo2.getResult()
print(result2.getCovarianceModel()) # Return SquaredExponential(scale=[1], amplitude=[1]), the covariance parameters are not optimized

# Origin of the problem
print(covarianceModel.getActiveParameter())  # Return [0, 1]
print(result.getCovarianceModel().getActiveParameter())  # Return []

# This is how we can force the optimization of parameters
covarianceModel2 = result.getCovarianceModel() 
covarianceModel2.setActiveParameter([0,1])  # We need to manually set the active parameters
algo3 = ot.KrigingAlgorithm(sampleX, sampleY, covarianceModel2, basis)
algo3.setOptimizeParameters(True)
algo3.run()
result3 = algo3.getResult()
print(result3.getCovarianceModel())  # Return SquaredExponential(scale=[0.743496], amplitude=[5.19971]), parameters have been optimzed
```

[...]

PR #1472, opened 2020-05-04
misc fixes

jschueller:
- #1455
- xmlstmgr::hasAttribute to look in object members


lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -92,8 +93,9 @@ class ExpectedImprovementEvaluation : public EvaluationImplementation
     const Scalar sk2 = metaModelResult_.getConditionalMarginalVariance(x);
     const Scalar sk = sqrt(sk2);
     if (!SpecFunc::IsNormal(sk)) return Point(1, -SpecFunc::MaxScalar);
-    const Scalar ratio = fmMk / sk;
-    Scalar ei = fmMk * DistFunc::pNormal(ratio) + sk * DistFunc::pNormal(ratio);
+    const Point ratio (1, fmMk / sk);
+    const Normal gaussian(1);
+    Scalar ei = fmMk * gaussian.computeCDF(ratio) + sk * gaussian.computePDF(ratio);
```
sofianehaddad, line 98: This is due to 6cde2f5220271f0095a016a21fe81cec2b85d220
My bad ☹️
We need to avoid creating Normal(1). I will implement `DistFunc::dNormal`

jschueller: we need a bench, I guess in 1-d it's not so bad
else we could declare it as static so the ctor is called once

PR #1483, opened 2020-05-14
Fixed the doc of KernelSmoothing

mbaudin47:
Clarify the default bandwidth rule.
Make the link between the mixed rule and the corresponding ResourceMap key.


jschueller:
I'm afraid I cant merge this because it doesnt build

jschueller:
moved to #1509

PR #1486, opened 2020-05-15
Various bugfixes

### LINKED ISSUE #1124: DistributionFactory::buildAsXXX methods not documented
Various `DistributionFactory::buildAsXXX` methods (for example, `WeibullFactory::buildAsWeibull`, `NormalFactory::buildAsNormal` ) are not documented (maybe all?)

### LINKED ISSUE #1443: Not all distributions have a getRoughness() method
The roughness of a distribution is the integral of the square of the PDF. It is a significant part of the non parametric estimation theory and practice. For example, it allows to compare the kernels in terms of performance of estimating the density, because this expression is involved in the AMISE criterion. The `getRoughness` method of a distribution computes it. However, it is not available for a distribution defined from `PythonDistribution`. 

```
import openturns as ot
class Quartic(ot.PythonDistribution):
    def __init__(self):
        super(Quartic, self).__init__(1)
        self.c = 15.0 / 16

    def computeCDF(self, x):
        u = x[0]
        if u <= -1:
            p = 0.0
        elif u >= 1:
            p = 1.0
        else:
            p = 0.5 + 15.0 / 16.0 * u \
                - 5.0 / 8.0 * pow(u, 3) \
                + 3.0 / 16.0 * pow(u, 5)
        return p

    def computePDF(self, x):
        u = x[0]
        if u < -1 or u > 1:
            y = 0.0
        else:
            y = self.c * (1 - u**2)**2
        return y

    def getRange(self):
        return ot.Interval(-1.0, 1.0)

distribution = ot.Distribution(Quartic.Quartic())
R = distribution.getRoughness()
```
produces:
```
RuntimeError: NotYetImplementedException : In 
DistributionImplementation::getRoughness() const
```

It is unfortunate that we cannot compare the Quartic kernel to the other kernels. In this case, we have to use the PythonDistribution, because it is not available in the library.

### LINKED ISSUE #1485: The Normal().getRoughness() method is wrong.
The `getRoughness` method of the `Normal` distribution returns a correct result only in 1 dimension. For dimension greater than 1, the result is wrong.

```
dimension = 2
distribution = ot.Normal(dimension)
R = distribution.getRoughness()
```
produces 0.28209479177387814. But the correct result is 0.0795775, with an absolute error lower than 3.e-7. 

This is because the current formula is:
```
wrongR = ot.Point([1.0 / (2.0 * np.sqrt(np.pi))])
```
but the correct formula for a standard normal multivariate gaussian is:
```
exactRforStandardNormal = ot.Point([1.0 / ((2.0 * np.sqrt(np.pi)) ** dimension)])
```
The faulty code is here:

https://github.com/openturns/openturns/blob/38d0aad0f1d0dfcc6db661dba681846630418d7e/lib/src/Uncertainty/Distribution/Normal.cxx#L712

The formula I suggest is only correct for a diagonal covariance (equal to 1!). Generalizing the formula would require further investigation. A way to validate the code is to use the Python prototype provided in #1443, because it works whatever the distribution. A quick "fix" would be to generate a NotImplementedException for non unity dimension.

sofianehaddad:
Various fixes : 
 - Fix  #1124
 - Fix #1443
 - Fix #1485

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,7 @@
  * Fixed TBB performance when used together with OpenBLAS
 
 === Bug fixes ===
+ * #1124 (DistributionFactory::buildAsXXX methods not documented)
  * #1213 (The legend of the graphics in MetaModelValidation is wrong)
  * #1222 (There is no kriging example based on HMAT)
  * #1331 (FittingTest_BestModelBIC sometimes fail)
--- a/ChangeLog
+++ b/ChangeLog
@@ -56,6 +56,7 @@
  * #1432 (Expected improvement-based EfficientGlobalOptimization stopping criterion could be improved)
  * #1437 (OrderStatisticsMarginalChecker bound message)
  * #1438 (Normal distribution: computeComplementaryCDF)
+ * #1443 (Not all distributions have a getRoughness() method)
  * #1448 (Memory consumption leads to crash)
  * #1449 (P1LagrangeInterpolation sometimes fails)
  * #1456 (truncation of distribution)
--- a/ChangeLog
+++ b/ChangeLog
@@ -62,6 +62,7 @@
  * #1456 (truncation of distribution)
  * #1461 (ComparisonOperator().getImplementation().getClassName() segfault)
  * #1471 (The graphics of KarhunenLoeveQuadratureAlgorithm has no axes)
+ * #1485 (The Normal().getRoughness() method is wrong)
 
 == 1.14 release (2019-11-13) == #release-1.14
 
--- a/lib/src/Uncertainty/Distribution/Normal.cxx
+++ b/lib/src/Uncertainty/Distribution/Normal.cxx
@@ -709,7 +709,22 @@ Distribution Normal::getStandardRepresentative() const
 Scalar Normal::getRoughness() const
 {
   // 0.2820947917738781434740398 = 1 / (2 * sqrt(Pi))
-  return 0.2820947917738781434740398 / getSigma()[0];
+  if (dimension_ == 1)
+    return 0.2820947917738781434740398 / getSigma()[0];
+
+  Scalar roughness = 1.0;
+  if (hasIndependentCopula())
+  {
+    for (UnsignedInteger d = 0; d < dimension_; ++d)
+      roughness *= 0.2820947917738781434740398 / getSigma()[d];
+    return roughness;
+  }
+  else
+  {
+    for (UnsignedInteger d = 0; d < dimension_; ++d)
+      roughness *= 0.2820947917738781434740398 / cholesky_(d, d);
+  }
+  return roughness;
 }
 
 /* Get the kurtosis of the distribution */
--- a/lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
+++ b/lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
@@ -1687,6 +1687,74 @@ protected:
     const DistributionImplementation * p_distribution_;
   };  // class EntropyKernel
 
+  // Class used to wrap the computePDF()**2 method for integration purpose
+  class PDFSquaredWrapper : public EvaluationImplementation
+  {
+  public:
+    PDFSquaredWrapper(const DistributionImplementation *p_distribution)
+        : EvaluationImplementation(), p_distribution_(p_distribution)
+    {
+      // Nothing to do
+    }
+
+    PDFSquaredWrapper *clone() const
+    {
+      return new PDFSquaredWrapper(*this);
+    }
+
+    Point operator()(const Point &point) const
+    {
+      const Scalar pdf = p_distribution_->computePDF(point);
+      return Point(1, pdf * pdf);
+    }
+
+    Sample operator()(const Sample &sample) const
+    {
+      // Keep the parallelism of computePDF
+      Sample pdfSquared(p_distribution_->computePDF(sample));
+      for (UnsignedInteger i = 0; i < pdfSquared.getSize(); ++i)
+        pdfSquared(i, 0) = pdfSquared(i, 0) * pdfSquared(i, 0);
+      return pdfSquared;
+    }
+
+    UnsignedInteger getInputDimension() const
+    {
+      return p_distribution_->getDimension();
+    }
+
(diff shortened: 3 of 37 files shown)


python/test/t_Normal_std.py
```diff
@@ -19,6 +19,10 @@ def cleanPoint(inPoint):
             inPoint[i] = 0.0
     return inPoint
 
+# First dimension 1 with default ctor
```
mbaudin47, line 22: Please add three unit tests to cover the features:
* a multivariate normal with independent copula
* a multivariate normal with non diagonal covariance matrix
* a distribution with singularities

lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
```diff
@@ -1687,6 +1687,74 @@ protected:
     const DistributionImplementation * p_distribution_;
   };  // class EntropyKernel
 
+  // Class used to wrap the computePDF()**2  method for interpolation purpose
```
mbaudin47, line 1690: "interpolation" should be replaced with "integration"

lib/src/Uncertainty/Distribution/Normal.cxx
```diff
@@ -709,7 +709,22 @@ Distribution Normal::getStandardRepresentative() const
...
+  if (hasIndependentCopula())
+  {
+    for (UnsignedInteger d = 0; d < dimension_; ++d)
+      roughness *= 0.2820947917738781434740398 / getSigma()[d];
+    return roughness;
+  }
+  else
+  {
+    for (UnsignedInteger d = 0; d < dimension_; ++d)
+      roughness *= 0.2820947917738781434740398 / cholesky_(d, d);
```
mbaudin47, line 725: This sounds much less obvious for me, because I do not see clearly why the off-the-diagonal terms do not impact the result. Did you find a theoretical reference or did a personal calculation?

python/test/t_Normal_std.expout
```diff
@@ -1,5 +1,6 @@
-Standard distribution =  Normal(mu = 0, sigma = 1)
-Get roughness =  0.28209
+Standard distribution =  Normal(mu = 0, sigma = 1) roughness =  0.28209
+2D distribution =  class=Normal name=Normal dimension=2 mean=class=Point name=Unnamed dimension=2 values=[0,0] sigma=class=Point name=Unnamed dimension=2 values=[1,1] correlationMatrix=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0,0,1] roughness =  0.07958
+2D with correlation distribution =  class=Normal name=Normal dimension=2 mean=class=Point name=Unnamed dimension=2 values=[0,0] sigma=class=Point name=Unnamed dimension=2 values=[1,2] correlationMatrix=class=CorrelationMatrix dimension=2 implementation=class=MatrixImplementation name=Unnamed rows=2 columns=2 values=[1,0.3,0.3,1] roughness =  0.04171
```
mbaudin47, line 3: Where do these reference values come from?

lib/src/Uncertainty/Distribution/Normal.cxx
```diff
@@ -709,7 +709,22 @@ Distribution Normal::getStandardRepresentative() const
...
   // 0.2820947917738781434740398 = 1 / (2 * sqrt(Pi))
-  return 0.2820947917738781434740398 / getSigma()[0];
+  if (dimension_ == 1)
+    return 0.2820947917738781434740398 / getSigma()[0];
+
+  Scalar roughness = 1.0;
+  if (hasIndependentCopula())
+  {
+    for (UnsignedInteger d = 0; d < dimension_; ++d)
+      roughness *= 0.2820947917738781434740398 / getSigma()[d];
```
mbaudin47, line 719: This is clear.

lib/src/Uncertainty/Distribution/Normal.cxx
sofianehaddad: @mbaudin47 you just need to write what is the `roughness`, ie \int density^2(x) dx

sofianehaddad: calcul_roughness

lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
sofianehaddad: Done

mbaudin47:
The calculation is neat: +1 for you!
The `pyinstallcheck_Normal_std` unit test fails. From the detail, I guess that the .expout just needs to be refreshed.

python/test/t_Normal_std.py
sofianehaddad: @mbaudin47 No. 
```
/* Get the PDF singularities inside of the range - 1D only */
Point DistributionImplementation::getSingularities() const
{
  if (dimension_ != 1) throw NotDefinedException(HERE) << "Error: cannot ask for PDF singularities for multivariate distributions.";
  return Point(0);
}
```

sofianehaddad:
> The calculation is neat: +1 for you!
> The `pyinstallcheck_Normal_std` unit test fails. From the detail, I guess that the .expout just needs to be refreshed.

I dropped the changes on the test. Now I had a more specific test for `roughness` with no more `distributions`

python/test/t_Normal_std.expout
sofianehaddad: For standard distribution, litterature. 
I implemented an estimate using simulation (similar to your code)

ChangeLog
```diff
@@ -15,7 +15,9 @@
 === Miscellaneous ===
 
 === Bug fixes ===
-
+ * #1124 (DistributionFactory::buildAsXXX methods not documented)
+ * #1443 (Not all distributions have a getRoughness() method)
+ * #1485 (The Normal().getRoughness() method is wrong)
```
jschueller, line 20: this one should go to the 1.15 ; can you move it to target the 1.15 branch, else you can edit the commit and i'll cherry-pick to my misc fixes pr that targets 1.15

sofianehaddad: Done

PR #1487, opened 2020-05-15
Add new examples

mbaudin47:
* KernelSmoothing: added an example for sensitivity to bandwidth
* Metamodel: adding a new example for model selection
* Polynomial chaos: explained the transformation
* Polynomial chaos: adding an example to shows the sensitivity to the degree of the polynomial


mbaudin47:
This is ready for review.

jschueller:
can you rebase on master to check that it still succeeds ?

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "\n",
+    "$$\n",
+    "f_1(x) = w_1 f_A(x) + w_2 f_B(x)\n",
+    "$$\n",
+    "\n",
+    "for any $x\\in\\mathbb{R}$ where $f_A$ is the density of the gaussian distribution \n",
+    "$\\mathcal{N}(0,1)$, $f_B$ is the density of the gaussian distribution \n",
+    "$\\mathcal{N}(3/2,(1/3)^2)$ and the weights are $w_1 = \\frac{3}{4}$ and $w_2 = \\frac{1}{4}$.\n",
+    "\n",
+    "This is a mixture of two gaussians: 1/4th of the observations have the $\\mathcal{N}(0,1)$ \n",
```
josephmure, line 21:
```suggestion
    "for any $x\\in\\mathbb{R}$ where $f_A$ is the density of the Normal distribution \n",
    "$\\mathcal{N}(0,1)$, $f_B$ is the density of the Normal distribution \n",
    "$\\mathcal{N}(3/2,(1/3)^2)$ and the weights are $w_1 = \\frac{3}{4}$ and $w_2 = \\frac{1}{4}$.\n",
    "\n",
    "This is a mixture of two Normal distributions: 1/4th of the observations have the $\\mathcal{N}(0,1)$ \n",
```
@adutfoy advocated for more consistency in our vocabulary: _Normal_ distributions and _Gaussian_ processes

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "distribution and 3/4th of the observations have the $\\mathcal{N}(3/2,(1/3)^2)$ distribution. \n",
+    "This example is considered in (Wand, Jones, 1994), page 14, Figure 2.3.\n",
+    "\n",
+    "We consider a sample generated from independent realizations of $f_1$ and \n",
+    "want to approximate the distribution from kernel smoothing. \n",
+    "More precisely, we want to observe the sensitivity of the resulting density on the \n",
```
josephmure, line 27:
```suggestion
    "More precisely, we want to observe the sensitivity of the resulting density to the \n",
```

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "## References\n",
+    "\n",
+    "* \"Kernel Smoothing\", M.P.Wand, M.C.Jones. Chapman and Hall / CRC (1994).\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Simulate the mixture by merging two samples\n",
```
josephmure, line 39:
```suggestion
    "## Generate the mixture by merging two samples\n",
```
@adutfoy Is that a situation where "generate" should be used instead of "simulate"?

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "\n",
+    "In this section, we show that a mixture of two gaussian distributions is just the merged sample of two independent gaussian distributions. In order to generate a sample with size $n$, we generate a first sample with size $\\lfloor w_1 n\\rfloor$ from the first gaussian distribution $f_A$ and a second sample with size $\\lfloor w_2 n\\rfloor$ from the second gaussian distribution $f_B$. Finally, we merge the two sub-samples to create the final sample."
```
josephmure, line 41:
```suggestion
    "In this section, we show that a mixture of two Normal distributions is nothing more than the merged sample of two independent Normal distributions. In order to generate a sample with size $n$, we sample $\\lfloor w_1 n\\rfloor$ points from the first Normal distribution $f_A$ and $\\lfloor w_2 n\\rfloor$ points from the second Normal distribution $f_B$. Then we merge the two samples."
```
I think this formulation is a little easier.

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "import openturns.viewer as otv\n",
+    "import pylab as pl\n",
+    "import numpy as np"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We consider a quite large sample size $n=1000$."
```
josephmure, line 60:
```suggestion
    "We choose a rather large sample size: $n=1000$."
```

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+   "outputs": [],
+   "source": [
+    "n = 1000"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Then we define the two gaussians and their parameters. "
```
josephmure, line 76:
```suggestion
    "Then we define the two Normal distributions and their parameters. "
```

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "w2 = 1.0 - w1\n",
+    "distribution1 = ot.Normal(0.0, 1.0)\n",
+    "distribution2 = ot.Normal(1.5, 1.0 / 3.0)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We generate two independent sub-samples from the two gaussian distributions."
```
josephmure, line 95:
```suggestion
    "We generate two independent sub-samples from the two Normal distributions."
```

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "sample = ot.Sample(sample1)\n",
+    "sample.add(sample2)\n",
+    "sample.getSize()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In order to see the result, we build a kernel smoothing approximation on the sample and see the result. In order to keep it simple, we consider the default bandwidth selection rule."
```
josephmure, line 141:
```suggestion
    "In order to see the result, we build a kernel smoothing approximation on the sample. In order to keep it simple, let us use the default bandwidth selection rule."
```

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+   ],
+   "source": [
+    "fit.drawPDF()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We see that there are two modes in the distributions. However, these two modes are somewhat unclearly distinguished. The first suggestion would be to increase the sample size. However, it might be interesting to see if the bandwidth selection rule can be better chosen: this is the purpose of the next section."
```
josephmure, line 179:
```suggestion
    "We see that the distribution of the merged sample has two modes. However, these modes are not clearly distinct. To distinguish them, we could increase the sample size. However, it might be interesting to see if the bandwidth selection rule can be better chosen: this is the purpose of the next section."
```
I think keeping it simple is better.

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+   ],
+   "source": [
+    "factory.getBandwidth()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We see that the default bandwidth produced a bandwidth close to 0.17."
```
josephmure, line 253:
```suggestion
    "We see that the default bandwidth is close to 0.17."
```

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+      ]
+     },
+     "execution_count": 12,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
```
josephmure, line 273: Could you switch the colors and use a dotted curve for the mixture? This would make it consistent with the later graphs.

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "graph.setLegends([\"Mixture\", \"Kernel smoothing\"])\n",
+    "graph.setLegendPosition(\"topleft\")\n",
+    "graph"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We see that the result is similar to the previous one. However, it seems that the mixture is not so well approximated."
```
josephmure, line 287:
```suggestion
    "We see that the result of the kernel smoothing approximation of the mixture is similar to the previous one and could be improved."
```
There is some ambiguity in the original formulation because it is not clear we are comparing the current kernel smoothing to the previous one.

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+   "metadata": {},
+   "source": [
+    "## Sensitivity to the bandwidth"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this section, we observe the sensitivity of the kernel smoothing to the bandwidth. We consider the three following bandwidths: the small bandwidth 0.05, the large bandwidth 0.54 and 0.18 which is in-between. For each bandwidth, we use the second optional argument of the `build` method in order to configure the bandwidth to a specific value. "
```
josephmure, line 301:
```suggestion
    "In this section, we observe the sensitivity of the kernel smoothing to the bandwidth. We consider the three following bandwidths: the small bandwidth 0.05, the large bandwidth 0.54 and 0.18 which is in-between. For each bandwidth, we use the second optional argument of the `build` method in order to select a specific bandwidth value. "
```

python/doc/examples/data_analysis/smoothing_mixture.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "    graph.setXTitle(\"X\")\n",
+    "    view = otv.View(graph, figure=fig, axes=[ax])\n",
+    "    pl.ylim(top = 0.5)  # Common y-range"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We see that when the bandwidth is too low, then the resulting kernel smoothing has many more modes than the distribution it is supposed to approximate. When the bandwidth is too large, then the approximated distribution is too smooth and has only one mode instead of the expected two modes which are in the mixture distribution. When the bandwidth is equal to 0.18, then the two modes are correctly represented."
```
josephmure, line 348:
```suggestion
    "We see that when the bandwidth is too small, the resulting kernel smoothing has many more modes than the distribution it is supposed to approximate. When the bandwidth is too large, the approximated distribution is too smooth and has only one mode instead of the expected two modes which are in the mixture distribution. When the bandwidth is equal to 0.18, the two modes are correctly represented."
```

josephmure:
Once again, I will need some time to review! There are several entirely new notebooks, so this is a large addition to the documentation.

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Sensitivity of the polynomial chaos to the degree\n",
```
josephmure, line 7:
```suggestion
    "# Polynomial chaos is sensitive to the degree\n",
```
I think this conforms to @adutfoy's prescription for example titles.

josephmure, line 7:
```suggestion
    "# Polynomial chaos: sensitivity to the degree\n",
```
This conforms to *my* prescription for example titles ;)

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+   "outputs": [],
+   "source": [
+    "maxDegree = 4"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To make real tests, we suggest to use the following parameter value."
```
josephmure, line 48:
```suggestion
    "For real tests, we suggest using the following parameter value:"
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+   "outputs": [],
+   "source": [
+    "# maxDegree = 7 "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We define the parameters of the cantilever beam problem."
```
josephmure, line 64:
```suggestion
    "Let us define the parameters of the cantilever beam problem."
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "    chaosalgo.run()\n",
+    "    result = chaosalgo.getResult()\n",
+    "    return result"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The following function computes  the sparsity rate of the polynomial chaos. To do this, we compute number of coefficients in the decomposition assuming a linear enumerate rule and a fixed truncation. The sparsity rate is the complement of the ratio between the number of coefficients selected from LARS and the total number of coefficients in the full polynomial basis."
```
josephmure, line 160:
```suggestion
    "The following function computes  the sparsity rate of the polynomial chaos. To do this, we compute the number of coefficients in the decomposition assuming a linear enumeration rule and a fixed truncation. The sparsity rate is the complement of the ratio between the number of coefficients selected from LARS and the total number of coefficients in the full polynomial basis."
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "    # Compute rate\n",
+    "    sparsityRate = 1.0 - nbcoeffs / P\n",
+    "    return sparsityRate"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The following functions compute and plots the Q2 predictivity coefficients within the validation plot."
```
josephmure, line 186:
```suggestion
    "The following functions compute and plot the Q2 predictivity coefficients within the validation plot."
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+   "outputs": [],
+   "source": [
+    "n_valid = 1000  # size of the test design"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The seed is selected to get *interesting* results. "
```
josephmure, line 260: I think this deserves an explanation.

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "magicSeed = 43  # 127 is funny too\n",
```
josephmure, line 269: "funny" in what way? Is the user supposed to execute the notebook with 127 to see what changes?

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "    ax = fig.add_subplot(1, maxDegree, totalDegree)\n",
+    "    _ = ot.viewer.View(graph, figure=fig, axes=[ax])\n",
+    "    pl.suptitle(\"Metamodel validation\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We see that, when the degree of the polynomial increases, then the Q2 coefficient decreases. We also see that the sparsity rate increases which shows that, while the basis size grows rapidly with the degree, the algorithm selects a smaller fraction of this basis. This shows that the algorithm performs its taks of selecting relevant coefficients. However, this selection does not seem to be sufficient to mitigate the large number of coefficients. \n",
```
josephmure, line 323:
```suggestion
    "We see that when the degree of the polynomial increases, the Q2 coefficient decreases. We also see that the sparsity rate increases: while the basis size grows rapidly with the degree, the algorithm selects a smaller fraction of this basis. This shows that the algorithm performs its task of selecting relevant coefficients. However, this selection does not seem to be sufficient to mitigate the large number of coefficients. \n",
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "\n",
+    "Of cours, this example has been selected to make a gradually decreasing predictivity. We are going to see that this situation is actually easy to reproduce. "
```
josephmure, line 325:
```suggestion
    "Of course, this example is designed to make a predictivity decrease gradually. We are going to see that this situation is actually easy to reproduce. "
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "    pl.xlabel(\"Degree\")\n",
+    "    pl.ylabel(\"Q2 (%)\")\n",
+    "    return"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Each experiment is repeated a number of times. "
```
josephmure, line 398:
```suggestion
    "Each experiment is repeated several times. "
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "N = 20  # size of the train design\n",
+    "Q2sample = computeSampleQ2(N, n_valid, numberAttempts, maxDegree)\n",
+    "plotQ2Boxplots(Q2sample, N)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We see that when the size of the design of experiments is as small as 20, it is more appropriate to use a very low degree polynomial equal to 1. Considering the polynomial degree 4 is risky. "
```
josephmure, line 448:
```suggestion
    "We see that when the size of the design of experiments is as small as 20, it is more appropriate to use a very low degree polynomial. Here 1 performs best and 4 is risky. "
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "N = 30  # size of the train design\n",
+    "Q2sample = computeSampleQ2(N, n_valid, numberAttempts, maxDegree)\n",
+    "plotQ2Boxplots(Q2sample, N)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We a sample size equal to 30, we can generally use a polynomial degree equal to 2."
```
josephmure, line 489:
```suggestion
    "With a 30-point design set, a polynomial degree of 2 is usually advisable."
```

python/doc/examples/meta_modeling/chaos_beam_sensitivity_degree.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "When the sample size increases, the Q2 computation becomes less sensitive to the polynomial degree."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Conclusion\n",
+    "\n",
+    "We observe that on the cantilever beam example, to use a polynomial total degree equal to 4, we need a sample size at least equal to 50 to get a satisfactory and reproducible Q2. When the degree is equal to 4, if we use a small sample size, then, depending on the particular sample, the predictivity coefficient can be very low (i.e less than 0.5). When the sample size as small as 20, then a polynomial degree equal to 1 is safer in this case, but the limited sample size may have an impact on other statistics that we may derive from a metamodel that was calculated on such a small training sample.\n",
```
josephmure, line 539:
```suggestion
    "We observe that on the cantilever beam example, to use a polynomial total degree equal to 4, we need a sample size at least equal to 50 to get a satisfactory and reproducible Q2. When the degree is equal to 4, if the sample is small, then depending on the particular sample, the predictivity coefficient can be very low (i.e. less than 0.5). With a sample size as small as 20, a polynomial degree of 1 is safer. However the limited sample size may have an impact on other statistics that could be derived from a metamodel calculated on such a small training sample.\n",
```

python/doc/examples/meta_modeling/chaos_distribution_transformation.ipynb
```diff
@@ -0,0 +1,300 @@
...
+   "source": [
+    "sampleZ = Z.getSample(5)\n",
+    "sampleZ"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the third step, we create the transform T which maps X into Z."
```
josephmure, line 176:
```suggestion
    "In the third step, we create the transform T which maps X to Z."
```

python/doc/examples/meta_modeling/chaos_distribution_transformation.ipynb
```diff
@@ -0,0 +1,300 @@
...
+    "Tinverse(sampleZ)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Conclusion\n",
+    "\n",
+    "Most of the times, as a user of the library, we do not use the `DistributionTransformation` class directly, as this is done automatically by the polynomial chaos classes. However, it helps to understand how this transform is done, because this clarifies why the coefficients of the chaos expansion cannot be related to the input random variable $X$: the expansion is based on the standard variables $\\xi$. Hence, the absolute values of the corresponding coefficient has no straightforward interpretation even though squaring them allows to see the the part of variance associated with each coefficient. "
```
josephmure, line 275:
```suggestion
    "The `DistributionTransformation` class is rarely used directly because the polynomial chaos classes perform the transformation automatically. However, understanding how this transform is done clarifies why the coefficients of the chaos expansion cannot be related to the input random variable $X$: the expansion is based on the standard variables $\\xi$. Hence, the absolute values of the corresponding coefficients have no straightforward interpretation, even though squaring them yields the part of the global variance associated with each coefficient. "
```

python/doc/examples/meta_modeling/overfitting_model_selection.ipynb
```diff
@@ -0,0 +1,860 @@
...
...TrSkepUpznNiUjrAStWwMGDRiMyNZsxA7Zvh3XrIEMi50IoWtQoRNC+vVER5Qk0mva0pyhFGcSgFAZsc+uW9HwL4Y6eiTEr2iBG05wllOaQzJbm5E4tHMLFZq/h6SWT6pCISdOESC0mMYmHPKQf/Ywe3A0bYMECoyc8Nfr9dxg6FFatguzZk7Zv3bqQJw8sXfrETccwhpOcZC5zzfs00o4933YrNSiEeLLRgYq2HcIJC03Dv+RiKCP4wqMr50b9AKk8ncFZndInqbDwBPmWzLA6FGcRa9I0pVRCk6Zp4DullAZm2srAPkLmZng8d742cI/r65GmB11e7YJniCd1/q5Dxo8/poy/P179+xNsdXB2Et/rpsLCKNulC3+2acNfly7BpUtJPm6OBg0oNnAge/PlM/LF47Er1y4mPz+Z6Qem83PYz/FukxwlTp3i35w5uRwcbPr/S2l8C2EhPz+IxIO2AX8SEfIU3z3zIYGeX1I5chHQyurwRDzW/NiXNl45yeDzutWhOIxJk6ZV1lpfsjXOtyiljmutf4i7kczN8HjufG3gPtf3PM9T5YUq1HihBuVpBfny4dOsGV4HDsCzz1odnunifd369oWXXqL4uHEUT27d7ipV+GfWalY3u8K0a80pWBACA433ToC97GUyk9nIRsr9r1yKruEREyfiXb48JXx9Tf9/KWknQlispZ8HU86t5Z3IBpw570mu5V9A//5w/brVoYk4LnGJ/PM3k6l1F1MmgXAVWuvqWusX43l8A1yOmnvhcZOmaa0v2b5eAdYA5R0VvxCO9gIvsJKV+OHHIQ5BzZqcb9EC3nknddQA37ULliyB2bNTdK8M+krRPmQYna6OROmIWBXBznCGBjRgLnMph8kNb3C/UoNCiNiiZko7xjHw8YGGDWHwYKvDEnFMDR3Le6sUmVp0sDoUZ5KYSdMyKaWyRH0P1EQmRxNu7nVeZypTqUtdLnKRPxs2hP/9zyi5584VUMLCoGNHmDwZcuVK0aECAmDtg5rcJBvvsQqA0FAYEBBBbWozmMG8wztmRP0oKTUohHvLQAa60pXxjDcWBAbC6tWwXyqyOYt/+ZerX3+JqlARnn7a6nCcSWImTfMGdimlfgP2Ahu01pssiVYIB2pKU3rQgzrU4U6auzB1Kvz5J4wZY3Vo9jN+vJFa8957KT6UUflL8Rk9ac+c6OUXQ
```
josephmure, line 775: Suggestion rejected by Github
    *We see that the RMSE on the train set continuously decreases, reaching zero when the polynomial degree is such that the number of coefficients is equal to the train dataset sample size. In this extreme situation, the least squares solution is equivalent to solving a linear system of equations: this leads to null residuals.*

jschueller:
I'm closing this one, its likely going to be lost after #1574 unless @KieranDelamotte wants to pick it up

mbaudin47:
@jschueller : It seems to me that the PR should have been merged long before @KieranDelamotte even started his own work, since the PR is OK since July.

jschueller:
i did not merge it because of joseph remarks

mbaudin47:
This is true: Joseph often has more remarks than I have time to include.
However, I cannot include them now, because the PR is closed:
 
image

Please reopen it, so that I can include Joseph's suggestion. Even if you do not want to merge, including Joseph comments will make Kieran's work much easier.

jschueller:
ok

mbaudin47:
@josephmure : please approve this PR, or push a new one, because there is no button "approve all Joseph's suggestion in a single batch commit". I fear that the result is just that these PRs are going to /dev/null, which was not your initial intent, I guess.

josephmure (approved):
I am not sure what you want me to do, but I can approve this PR.

josephmure:
I've just created PR #1604 which is the same as this one but with all "apply suggestion" commits squashed, so I think this PR should be closed.

PR #1490, opened 2020-05-19
Calibration: Cannot calibrate a function with unobserved inputs.

### LINKED ISSUE #1327: Cannot calibrate a function with unobserved inputs
In this example, we calibrate the flooding model. We assume that there are true, but unknown, values of the parameters Q, Ks and DeltaZ. We also assume that we are given a sample of noisy observed river heights H, but no observed input. In other words, we assume that there is no input observation. 

```
# ## Generate the observations
import numpy as np
import openturns as ot
# We define the model $g$ which has 3 inputs and one output H.
# 
def functionFlooding(X) :
    L = 5.0e3
    B = 300.0
    Q, K_s, DeltaZ = X
    alpha = DeltaZ/L
    if alpha < 0.0 or K_s <= 0.0:
        H = np.inf
    else:
        H = (Q/(K_s*B*np.sqrt(alpha)))**(3.0/5.0)
    return [H]
g = ot.PythonFunction(3, 1, functionFlooding) 
g = ot.MemoizeFunction(g)
g.setOutputDescription(["H (m)"])
# Set the parameters to be calibrated.
Q = ot.Dirac(1013.0)
K_s = ot.Dirac(30.0)
DeltaZ = ot.Dirac(5.0)
# Create the joint input distribution.
inputRandomVector = ot.ComposedDistribution([Q, K_s, DeltaZ])
# Create a Monte-Carlo sample of the output H.
# This is a sample which contains a constant value.
nbobs = 100
inputSample = inputRandomVector.getSample(nbobs)
outputH = g(inputSample)
# Observe the distribution of the output H.
print(outputH[0:5])
# Generate the observation noise and add it to the output of the model.
sigmaObservationNoiseH = 0.1 # (m)
noiseH = ot.Normal(0.,sigmaObservationNoiseH)
sampleNoiseH = noiseH.getSample(nbobs)
Hobs = outputH + sampleNoiseH
# ## Setting the calibration parameters
# Define the value of the reference values of the $\theta$ parameter. 
QInitial = 1000.
KsInitial = 20.
DeltaZInitial = 2.
thetaPrior = ot.Point([QInitial,KsInitial,DeltaZInitial])
# The following statement create the calibrated function from the model.
calibratedIndices = [0,1,2]
mycf = ot.ParametricFunction(g, calibratedIndices, thetaPrior)
# Therefore, the calibrated function has no input and this is why we set an empty list as input argument.
mycf([])
# ## Calibration with linear least squares
# Since we have no input observation, we set the empty list `[]` as input.
algo = ot.LinearLeastSquaresCalibration(mycf, [], Hobs, thetaPrior, "SVD")
```

The previous script generates the following output:
```
TypeError: InvalidArgumentException : Error: expected a model of input dimension=1, got input dimension=0
```

The equations make clear the fact that the dependency to X is optional: only the dependency to theta is mandatory. 

The line which generates the exception is:

https://github.com/openturns/openturns/blob/e28c9eae3b145d3602f800cfacd1bc8d942219c8/lib/src/Uncertainty/Bayesian/LinearLeastSquaresCalibration.cxx#L57

This is consistent: the code requires that the dimension of the input sample matches the input dimension of the model. We are, however, in a situation were an empty input sample is provided.

I tried to create an empty sample with dimension 0:
```
inputObservations = ot.Sample([],0)
```
This kills the library!

So there are two bugs:
* the `Sample` class fails;
* the case where there is no observed input sample is not taken into account in the calibration classes.

My guess is that a new constructor should be provided, where there is no input sample:
```
LinearLeastSquaresCalibration(model, outputObservations, candidate, methodName)
```

mbaudin47:
Cannot calibrate a function with unobserved inputs. Fixes #1327


mbaudin47:
This is ready for review. 
Including it into v1.15 would be appreciated.

mbaudin47:
@regislebrun : does it match your expectations?

regislebrun:
As a general comment: good job!

IMO the underlying numerical method should not appear in the constructor. The constructor should expose the data of the problem at hand, not the way to solve it (which is already mostly defined by the choice of the class). An accessor and a meaningful default in ResourceMap is much more in the spirit of the other classes. BTW, I am not very fond of strings as descriptors, as it has only a weak semantic and is quite error-prone (for the user: uppercase/lowercase, mispeling, for the developer: a switch to be maintained and extended each time a new method is added)

mbaudin47:
You are talking about this:
https://github.com/openturns/openturns/blob/a60832bce370880580d320191fae503e52256ade/lib/src/Uncertainty/Bayesian/openturns/LinearLeastSquaresCalibration.hxx#L62
and that:
https://github.com/openturns/openturns/blob/a60832bce370880580d320191fae503e52256ade/lib/src/Uncertainty/Bayesian/openturns/GaussianLinearCalibration.hxx#L73
right?
I agree that this reveals a poor design which is quite uncommon in the library. In both cases, the argument `MethodName` is used by the linear least squares method:
https://github.com/openturns/openturns/blob/a60832bce370880580d320191fae503e52256ade/lib/src/Uncertainty/Bayesian/LinearLeastSquaresCalibration.cxx#L119
and:
https://github.com/openturns/openturns/blob/a60832bce370880580d320191fae503e52256ade/lib/src/Uncertainty/Bayesian/GaussianLinearCalibration.cxx#L179
so I guess that the design issue is the LeastSquaresMethod::Build() static method. 
I will dig into this and try to suggest alternatives.

mbaudin47:
If you agree, I would rather leave this PR as is, because it solves the problem it was meant to solve. I reported the issue that you mention in a separate issue: #1511.

regislebrun:
So +1 and we need to fix #1511 as soon as possible

jschueller:
I dont think adding more constructors is the way to go ; its more explicit to use the same constructor with a sample of dim 0 `ot.Sample(n, 0)` rather than omitting it.
But, the tests you are adding are useful, so I propose to keep them, but calling the old constructor with a sample of dim 0.

jschueller:
I added your tests/examples in #1565

mbaudin47:
I understand that using a dimension-0 `Sample` is a workaround that temporarily solves the issue. 

This does not however correspond to the user's requirements (actually, 3 users). When we do not have any observed inputs, we should not have to specify one empty argument. The actual issue here is that we cannot use non-positional input arguments in C++, so that we must set it, even if it is empty.

If the argument is to leave only the minimum possible number of constructors in order to minimize development / maintenance work, there are many constructors in the library that could be suppressed by the same point of view. Here is just a few of them:
* `LinearLeastSquaresCalibration(model, inputObservations, outputObservations, candidate, methodName)` could be suppressed, because the user may define a `LinearFunction` and use `LinearLeastSquaresCalibration(modelObservations, gradientObservations, outputObservations, candidate, methodName)`.
* Same for `GaussianLinearCalibration`.
This amounts to unnecessarily increase the user's code. 

@regislebrun : may you share your point of view?

regislebrun:
100% to suppress most of the constructors ;-)! More seriously, even if many of the questions can be answered looking at the maths or the software architecture, one have also to take into account non-scientific aspects:
- the tradition in the community
- the trade-off between the time gain by the users and the pain for the maintainers
- as a subset, is it something that ease my life or the life of the typical user?
What I dislike the most is to adopt a field-specific jargon in a math library (FunctionalChaos is an example of that) as it may divert the user from the relevant bibliography. In second place comes the ready-to-use classes hiding most of the necessary complexity. Then come  the classes with constructors here only to help one user among dozens. It is nice for him but the burden introduced is not worth it. Are we in this situation? Is it often the case that the model has no input, only parameters to calibrate? If yes IMO we should add the requested constructor, otherwise we shouldn't. I can't decide here (even if the change looks quite harmless)

mbaudin47:
Ok, I get the point. So let us leave the PR as is: we may reopen it if necessary if other users feel the need for this particular constructor.

mbaudin47:
> I added your tests/examples in #1565

The notebook https://github.com/mbaudin47/openturns/blob/a60832bce370880580d320191fae503e52256ade/python/doc/examples/calibration/calibration_withoutobservedinputs.ipynb was not included in the examples, isn'it?

PR #1493, opened 2020-05-19
Update estimate_probability_form.ipynb

adutfoy:
Changed the parameters of the use case.

### DIFF
--- a/python/doc/examples/reliability_sensitivity/estimate_probability_form.ipynb
+++ b/python/doc/examples/reliability_sensitivity/estimate_probability_form.ipynb
@@ -46,7 +46,7 @@
     "Failure occurs when the beam deviation is too large:\n",
     "\n",
     "$$\n",
-    "d \\ge 30 (cm)\n",
+    "d \\ge 0.22 (m)\n",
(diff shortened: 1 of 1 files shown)


mbaudin47:
It seems to fail because of indentation:
```
sphinx.errors.SphinxWarning: /home/devel/project/build/python/src/sphinx_build/examples/reliability_sensitivity/estimate_probability_form.ipynb:7:Unexpected indentation.
```

mbaudin47:
It would be interesting for the reader to have an information on the application. Here is a suggestion.

_We are interested in the the vertical deviation of a diving board created by a children diver. We consider a children which weight generates a force approximately equal to 300N (i.e. almost 30 kg). Because of the uncertainty in the weight of the person, we consider that the force is a random variable. The length of the diving board is between 2.5 m and 2.6 m. The Young modulus corresponds to the fiberglass material, a material often used for diving boards. Uncertainties in the production of the material are taken into account in the Young modulus and the section modulus of the board._

Does this text fit your intent?

jschueller:
moved to #1509

PR #1498, opened 2020-05-22
EGO: Fixed criteria. 

### LINKED ISSUE #1495: Wrong formula for Expected Improvement evaluation ?
In line 96 of ```EfficientGlobalOptimization.cxx```, the Expected Improvement is computed as follows: 
``` Scalar ei = fmMk * DistFunc::pNormal(ratio) + sk * DistFunc::pNormal(ratio);``` where pNormal is the CDF of the normal distribution. 

I think that we should use the PDF of the normal distribution for the term after ```sk```. Is it a bug or did I miss something ?

mbaudin47:
Fixes #1495

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -80,6 +80,7 @@
  * #1456 (truncation of distribution)
  * #1461 (ComparisonOperator().getImplementation().getClassName() segfault)
  * #1471 (The graphics of KarhunenLoeveQuadratureAlgorithm has no axes)
+ * #1495 (Wrong formula for Expected Improvement evaluation)
 
 == 1.14 release (2019-11-13) == #release-1.14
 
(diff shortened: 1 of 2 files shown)


mbaudin47:
This however fixes only part of the underlying design issue. The issue was caused by a poor design in which the objective function cannot be accessed from outside the class. Therefore, any flaw in the implementation remains hidden until an indirect cause highlights the problem. In order to be testable, this part of the code must be evaluable outside. Hence, what we need is a getExpectedImprovementFunction() method which returns the function as the `getObjectiveFunction` method does for `GeneralLinearModelAlgorithm`. This way, we can 1) create a unit test for this method which checks the value and compares it with a reference value, 2) plot it in the examples to show how it looks like and compare with reference graphics.

jschueller:
actually you can retrieve the criterion through getOptimizationAlgorimth().getProblem().getObjectiveFunction() once it has run.

jschueller:
cherry-picked into to #1472

PR #1500, opened 2020-05-25
Update some examples

sofianehaddad:
This is a copy of #1478 in order to integrate it to the 1.15

Some highlights : 
 - using new API
 - Pay attention to docstrings regarding parametrization of distribution
 - Fix some examples due to new parametrization

 Probably more examples need to be reviewed


jschueller:
I already cherry picked them

PR #1503, opened 2020-05-25
Added DistFunc::dNormal

sofianehaddad:
Purpose here is to add a new static method for the evaluation of Gaussian pdf.


sofianehaddad:
@jschueller do you think we can integrate it to 1.15 ? In that way, we may fix EGO issue using this new feature

jschueller:
not critical, it will just be slightly faster

mbaudin47:
This is related to the performance of the implementation of the EGO criteria, isn'it:

https://github.com/openturns/openturns/pull/1498/files#diff-7b5d8ecc63c1ee329b91a6274215e305

Are there other use-cases of the dNormal function in the code?

regislebrun:
I can see (at least) the following places where SpecFunc::dNormal() could be used:
- SmoothedUniform
- KernelSmoothing
- LogNormal
- StudentFunction
- NormalCopula
- DistFunc (other places than in dNormal() ;-))
- Normal
- RandomMixture
- TruncatedNormal
- NatafIndependentCopulaGradient
- InverseNatafIndependentCopulaGradient
- NatafIndependentCopulaHessian
- InverseNatafIndependentCopulaHessian
and probably other places I missed...

sofianehaddad:
@regislebrun  I replaced when needed/identified. 
Probably we can do more/best but needs time...

jschueller:
@regislebrun good to go ?

regislebrun:
+1
I listed several places where this new function could be used, not to have the job done right now but to motivate its introduction in the library!

Issue #1515, opened 2020-06-04
KernelSmoothing build failure


josephmure:
The KernelSmoothing build() method sometimes fails inexplicably.
 ```
 import openturns as ot
 import numpy as np
 
 
 #a1 is a 254 x 1 matrix with 5 non-zero elements equal to 0.25
 #a1 is a 255 x 1 matrix with 5 non-zero elements equal to 0.25
 a1 = np.append(np.repeat(0.0,249), np.repeat(0.25,5)).reshape(-1, 1)
 a2 = np.append(np.repeat(0.0,250), np.repeat(0.25,5)).reshape(-1, 1)
 
 #a1 et a2 are turned into Samples s1 and s2
 s1 = ot.Sample(a1)
 s2 = ot.Sample(a2)
 
 #Kernel smoothing succeeds with s1, but fails with s2
 ks = ot.KernelSmoothing()
 k1 = ks.build(s1)
 k2 = ks.build(s2)
 ```
 The Distribution k1 is constructed without error, but the last line of this script produces the following error:
 ```
 RuntimeError: InternalException : Error: Brent method requires that the function takes different signs at the endpoints of the given starting interval, here infPoint=0, supPoint=0, value=0, f(infPoint) - value=-nan and f(supPoint) - value=-nan
 ```
 I realize that kernel smoothing should not be attempted on 2-valued Samples anyway, but I would like to understand the error.
 It looks like the Brent algorithm does not realize there is more than one value in the Sample when building k2, although it does when building k1.

jschueller:
the problem is we dont check that the variance is > 0

sofianehaddad:
@jschueller agreed that we should add some checks. However the variance here is not 0
The issue is related to the evaluation of the bandwidth using the `MixedPlugin` method

```
bw1 = ks.computeMixedBandwidth(s1) # ok
bw2 = ks.computeMixedBandwidth(s2) # fails and provide the previous msg
```

jschueller:
yes, falling back to silverman is also a solution, i'll do that

sofianehaddad:
@jschueller I see your point concerning the `sd` equals to 0. Indeed the bw is evaluated only through a small sample and not the full one. Small being here `250`
If we increase this size to 255 for example (using `ot.ResourceMap.SetAsUnsignedInteger("KernelSmoothing-SmallSize", 255)`), the problem vanishes...

Probably we also should consider how to select the small sample, check `sd` value... In other words makes the estimate more robust

[...]

PR #1517, opened 2020-06-08
KernelSmoothing: fixed solve the equation rule

mbaudin47:
* Fix the calculation of psi: the N-1 expression is wrong, because the KS estimator is an average, not a variance.
* Clarified the doc of the KernelSmoothing-CutOffPlugin parameter
* KernelSmoothing: fixed failure when phiGammaH<0. This fixes the exception which is sometimes raised when the cutoff parameter is set too low (see the unit test for reference). This bug was *not* #1515 initially, but the discussion raised this issue afterwards. The pow function can be used, but with a negated argument, which produces a negative output, as expected.

Work in progress.

### DIFF
--- a/lib/src/Uncertainty/Distribution/KernelSmoothing.cxx
+++ b/lib/src/Uncertainty/Distribution/KernelSmoothing.cxx
@@ -146,7 +146,7 @@ struct PluginConstraint
         if (std::abs(x) < cutOffPlugin) phi += 2.0 * hermitePolynomial_(x) * std::exp(-0.5 * x * x);
       }
     }
-    const Scalar res = phi / ((N_ * (N_ - 1.0)) * std::pow(h, order_ + 1.0) * std::sqrt(2.0 * M_PI));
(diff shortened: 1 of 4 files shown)


mbaudin47:
This is a partial fix, because my code does not seem to build.

regislebrun:
I already started to work on the bug yesterday and decided to add two methods to SpecFunc to compute the integral power and integral root of scalars, taking into account the parity of the exponent to set the sign correctly. There are several places where std::pow should be replaced to fix the bug, not only where you made the change.

Concerning the replacement of N*(N-1) by N*N, you are right, there is an inconsistency between equation (21) and equations (29)&(30) in the reference. Well done.

jschueller:
see #1588

mbaudin47:
Closing this PR is invalid, since the update in t_KernelSmoothing_std.py was not integrated in #1588.

jschueller:
I missed that part, including it my release branch

mbaudin47:
Great: thanks!

mbaudin47:
Including the unit test was done at 6e5c0ca471c3b62c0fa17008ff83a0b1777b5e2a. 
This is why I delete the branch.

PR #1519, opened 2020-06-10
Fix KernelSmoothing::computeMixedBandwidth

regislebrun:
This PR provides the following improvements:
+ the SpecFunc::IPow() and SpecFunc::IRoot() methods to compute the nth power or nth root of a scalar of arbitrary sign (n being a signed integer)
+ a better selection of a subsample to compute the mixed bandwidth in KernelSmoothing, based on a quasi-random selection of the sub-sample
+ the correct extraction of nth root for negative intermediate values
+ a better default choice for the precision at which the plugin constraint is solved (the comment says "loosely solved" in the C++ code, but the default value was 0 for the absolute precision)
Feel free to cherry-pick things in this PR to improve PR #1509 or PR #1517.

### DIFF
--- a/lib/src/Base/Func/SpecFunc/SpecFunc.cxx
+++ b/lib/src/Base/Func/SpecFunc/SpecFunc.cxx
@@ -1135,6 +1135,33 @@ UnsignedInteger SpecFunc::NextPowerOfTwo(const UnsignedInteger n)
   return powerOfTwo;
 }
 
+// Integer power
+Scalar SpecFunc::IPow(const Scalar x, const SignedInteger n)
+{
+  if (n == 0) return 1.0;
+  if (x == 0) return 0.0;
+  if (x < 0.0)
+    {
+      if (n % 2) return -std::pow(-x, 1.0 * n);
+      return std::pow(-x, 1.0 * n);
+    }
+  return std::pow(x, 1.0 * n);
+}
+  
+// Integer root
+Scalar SpecFunc::IRoot(const Scalar x, const SignedInteger n)
+{
+  if (n == 0) throw InvalidArgumentException(HERE) << "Cannot take the zeroth root of anything!";
+  if (x == 0) return 0.0;
+  if (x < 0.0)
+    {
+      if (n % 2 == 0) throw InvalidArgumentException(HERE) << "Cannot take an even root of a negative number";
+      return -std::pow(-x, 1.0 / n);
+    }
+  return std::pow(x, 1.0 / n);
+}
+  
+
 // Compute the number of bits sets to 1 in n
 // Best known algorithm for 64 bits n and fast multiply
 UnsignedInteger SpecFunc::BitCount(const Unsigned64BitsInteger n)
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,9 +30,9 @@
  * #1474 (optimization defaults)
  * #1507 (Leak in Collection typemaps)
  * #1510 (ot.Ceres('LEVENBERG_MARQUARDT') and ot.Ceres('DOGLEG') do not handle bound constraints)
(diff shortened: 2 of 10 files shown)


jschueller:
I just tested it: it fixes the two testcases from #1515, now I have to figure out how to integrate it with my changes and michael's

regislebrun:
@jschueller we can do it the other way: I merge everything into my branch (the raw way: copy-paste of code sections) if it is simpler for you. BTW, I have not updated the tests and the documentation as I expected my code to be cherry-picked into another branch, but I can do it if I am in charge of the merge.

jschueller:
That works too, just pick my changes for the test & changelog, and also add a test for michael's testcase (see #1515)

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -1135,6 +1135,29 @@ UnsignedInteger SpecFunc::NextPowerOfTwo(const UnsignedInteger n)
   return powerOfTwo;
 }
 
+// Integer power
+Scalar SpecFunc::IPow(const Scalar x, const SignedInteger n)
+{
+  if (n == 0) return 1.0;
+  if (x == 0) return 0.0;
+  if (x < 0.0) return -std::pow(-x, 1.0 * n);
```
sofianehaddad, line 1143: what should be the output of `IPow(-5.0, 2)`? I think you should check `n%2`, isn't it?

jschueller:
@sofianehaddad the missing includes were not raising any error because the ci uses unity builds, I just cherry-picked the commit into master.

regislebrun:
I agree with @sofianehaddad, the missing headers in `InverseNataf*` are a pain in the ass as it prevents master to compile the incremental way (no unity). I don't know who is responsible for that (and I don't care) but master should be fixed ASAP.

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
regislebrun: Right! I will fix it ASAP. Concerning the documentation, my first intent was to provide a fix for the issue in the C++ code, to be cherry-picked and completed (doc+test) by someone else as there were already two PRs on this issue. But I propose to be the one in charge of solving #1515, including code from the other PRs and adding test&doc to this PR.

sofianehaddad: Concerning the doc it is more general. As the class is public, we should provide some docstrings for **all existing methods/attributes**

I can start soon with *attributes* and some methods

regislebrun:
@sofianehaddad Ok. How do you want to proceed? You create a branch where we all contribute, or you take the action alone? For each method in SpecFunc/DistFunc we need:
- the documentation including an example
- a C++ test and its Python counterpart
Anything else?

regislebrun:
For now I am struggling with NLopt giving different results on my computer than the reference results, even in the bounded case...

sofianehaddad:
@regislebrun I am only focused on doc.
I think examples should be very small here (it is the SpecFunc)
I can start with a branch and cherry-pick other branches if someone wants to contribute.
I miss the C++/python tests. If the idea is following the current "template", thus let's forget.

sofianehaddad:
@regislebrun tracking `NLOpt` differences is not trivial. I recently experienced it with `EGO` and having similar troubles when performing the "same" algos (with OT & scipy)...

jschueller:
see #1588

mbaudin47:
Your PR does not include the update in the Py unit test provided at : https://github.com/openturns/openturns/pull/1517/files#diff-84c6ac15f293197160f452e68a574dd1

jschueller:
I added in my release branch

PR #1269, opened 2019-09-25
Various fixes

regislebrun:
This PR provides the following fixes/improvements:
  * More efficient computeConditionalPDF() for NormalCopula
  * Resolution of [Bug with Normal.computeMinimumLevelSet method](https://github.com/openturns/openturns/issues/1185)
  * Resolution of [computeProbability clamped by Domain-SmallVolume](https://github.com/openturns/openturns/issues/1190)
  * Resolution of [Sample::sort & Sample::sortAccordingToAComponent only return new Samples](https://github.com/openturns/openturns/issues/1202)

PR #1527, opened 2020-06-19
Update GaussianLinearCalibration_doc.i.in

mbaudin47:
Clarifies the description of the Normal error distribution.


jschueller:
moved to #1269

PR #1530, opened 2020-06-22
Drop normalization in Kriging

sofianehaddad:
This pull-request drops the input transformation inside KrigingAlgorithm. 
There is  : 
 - No boolean to set to True/False for data preprocessing 
 - No more internal transformation handled

User should make **its own mapping** outside the kriging algorithm. We should add later some documentation on how to efficiently do such a mapping.

Coming soon : same cleaning for the *GeneralLinearModelAlgorithm*


lib/include/openturns/OTtestcode.hxx
```diff
@@ -331,6 +332,23 @@ inline void assert_almost_equal(const Sample & a, const Sample & b, Scalar rtol
...
+  const UnsignedInteger columns = a.getNbColumns();
+
+  for (UnsignedInteger j = 0; j < columns; ++j)
+  {
+    for (UnsignedInteger i = 0; i < rows; ++i)
+    {
+      assert_almost_equal(a(i, j), b(i, j), rtol, atol);
+    }
+  }
+}
```
jschueller, line 351: great :)

lib/include/openturns/OTtestcode.hxx
```diff
@@ -331,6 +332,23 @@ inline void assert_almost_equal(const Sample & a, const Sample & b, Scalar rtol
   }
 }
 
+inline void assert_almost_equal(const Matrix &a, const Matrix &b, Scalar rtol = 1.0e-5, Scalar atol = 1.0e-8)
```
jschueller, line 335: great :)

josephmure: If I understand correctly, this means we will be able to test matrix equality. This will not be available in regular code, though, will it? Only in unit tests?

lib/test/t_KrigingAlgorithm_std.cxx
```diff
@@ -62,14 +62,12 @@ int main(int, char *[])
       Sample Y2(model(X2));
 
       Basis basis(ConstantBasisFactory(dimension).build());
-      SquaredExponential covarianceModel(Point(1, 1e-05), Point(1, 4.11749));
+      SquaredExponential covarianceModel(Point(1, 1e-02), Point(1, 4.50736));
```
josephmure, line 65: Why is this change necessary?

python/doc/examples/meta_modeling/kriging_beam_arbitrary_trend.ipynb
```diff
@@ -124,6 +124,22 @@
     "Y_train = model(X_train)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We notice here that input sample contains huge values. Thus we start by transforming data :"
```
josephmure, line 131: Good thinking! Examples need to show how to transform the data if the KrigingAlgorithm no longer does it.

python/test/t_MetaModelValidation_std.py
```diff
@@ -6,99 +6,93 @@
...
-    print("Kriging scoring")
-    print("Q2 = ", round(metaModelValidationKG.computePredictivityFactor(), 3))
-    ot.PlatformInfo.SetNumericalPrecision(2)
-    print("Residual sample = ", repr(
-        metaModelValidationKG.getResidualSample()))
-
-except:
-    import sys
-    print("t_MetaModelValidation_std.py", sys.exc_info()[0], sys.exc_info()[1])
+ot.PlatformInfo.SetNumericalPrecision(3)
```
josephmure, line 9: Why remove the try-catch?

lib/include/openturns/OTtestcode.hxx
sofianehaddad: Yes. Now me can check matrix equality. It is only in unit tests.
Do you have a specific use case where you need it in the code?

lib/test/t_KrigingAlgorithm_std.cxx
sofianehaddad: Because we dropped the normalization. In practice I will say it is useless as we perform an optimization algorithm. However final results might slightly differ considering the algorithm, the architecture and so. Thus we start from the optimal point.
The point here was obtained using MultiStart and strong convergence criteria (outside this specific test)

python/doc/examples/meta_modeling/kriging_beam_arbitrary_trend.ipynb
sofianehaddad: Indeed. Probably more examples are required. I am trying to find the best way for

python/test/t_MetaModelValidation_std.py
sofianehaddad: To mimic other unit tests. As we have an assert, it is useless to use try/catch

lib/include/openturns/OTtestcode.hxx
josephmure: It may be convenient sometimes but I am not sure it is worth putting in the code. In Python, users can always convert matrices to numpy arrays for comparison purposes.

lib/test/t_KrigingAlgorithm_std.cxx
josephmure: Ah, of course.

python/test/t_MetaModelValidation_std.py
josephmure: Thanks Sofiane!

lib/include/openturns/OTtestcode.hxx
sofianehaddad: Sure. IMHO it is fully convenient *only* for tests (C++/Python).

sofianehaddad: Sure. IMHO it is fully convenient *only* for tests (C++/Python).

jschueller:
@sofianehaddad do you want to modify GLM here too, or should I merge it as is ?

sofianehaddad:
@jschueller you can merge it. New PR for GLM

python/doc/examples/meta_modeling/kriging_beam_arbitrary_trend.ipynb
mbaudin47: I do not get the point of getting the transformation without using it in the example. Should'nt the example show how to actually use the transformation?

sofianehaddad: we apply it to the data (cell 15)

sofianehaddad: we apply it to the data (cell 15)

Issue #1529, opened 2020-06-19
The error of the NonLinearLeastSquaresCalibration and GaussianNonLinearCalibration are different.


mbaudin47:
The error of the NonLinearLeastSquaresCalibration is the only one which can have a non zero mean. The GaussianNonLinearCalibration error has a zero mean. This might mean that one of them has a bug.

Let us focus on the distribution error of various calibration algorithms.
* LinearLeastSquaresCalibration: The resulting distribution of the output error is a Normal and is computed from the residuals.
* NonLinearLeastSquaresCalibration: The resulting distribution of the output error is a Normal and is computed from the residuals.
* GaussianLinearCalibration: The resulting error distribution is Normal with a zero mean and with a covariance matrix equal to the errorCovariance input argument.
* GaussianNonLinearCalibration: The resulting error distribution is Normal with a zero mean and with a covariance matrix equal to the errorCovariance input argument.

The fact that linear algorithms have a zero error mean is expected, because the residuals for these method is always centered on zero by design. The fact that the mean of the error of the NonLinearLeastSquaresCalibration algorithm can be non zero seems right to me, because a non linear algorithm does not always produce a zero mean error. The Normal distribution makes sense if we want to link this result to the maximum likelihood estimator. 

From these facts, I conclude that the GaussianNonLinearCalibration has a bug: the distribution of the error should be Normal and computed from the residuals. This way, the mean of the error can be nonzero.

jschueller:
difficult to answer without any specific math argument here, deferring to @regislebrun

regislebrun:
@mbaudin47 Could you provide simple examples showing what you get and what you expected from the different algorithms?

PR #1533, opened 2020-06-28
Fixed GaussianNonLinearCalibration

regislebrun:
The estimation of distribution of the observations error was incorrect. It was created using a zero mean to compute the Cholesky factor of its (known) covariance matrix in order to compute the residual function, but was not updated once the residual function was known. It fixes issue #1529.


lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
```diff
@@ -330,10 +330,23 @@ void GaussianNonLinearCalibration::run()
 {
   // Compute the posterior MAP
   // Error distribution
-  const Normal error(Point(errorCovariance_.getDimension()), errorCovariance_);
+  // It is built in two steps to benefit from the Cholsky factorization of the
```
josephmure, line 333:
```suggestion
  // It is built in two steps to benefit from the Cholesky factorization of the
```

josephmure:
The gap between a value produced with VisualStudio and the corresponding expected value can be quite large:
[-0.0842 instead of -0.0855](https://ci.appveyor.com/project/openturns/openturns/builds/33787714/job/x5c13dhxx413kd4c#L6895) for example. Is the optimization algorithm unstable?

jschueller:
see #1588

mbaudin47:
To sum up the change the updated GaussianNonLinearCalibration class has a resulting error distribution which is Normal with a nonzero mean and with a covariance matrix equal to the errorCovariance input argument. This corresponds to the fact that the linearization of the function near the optimum makes so that the residuals may not necessarily have a zero mean: this effect can indeed be retrieved with the residuals.getMean() method. 

In order to check the code, we can use the GaussianNonLinearCalibration with a flat prior (i.e. a Normal with a large sigma). We must get the same results as the NonLinearLeastSquaresCalibration.

PR #1538, opened 2020-07-01
Update StandardEvent_doc.i.in

adutfoy:
Minor correction


jschueller:
-1, this is actually a RV

PR #1539, opened 2020-07-01
Update ImportanceSamplingExperiment_doc.i.in

adutfoy:
Corrected some wrong information in the constructors


jschueller:
moved to #1541

PR #1551, opened 2020-07-10
Grid graphs

jschueller:
- Adding GridLayout graph, has ctor(nrows, ncols), setGraph(i,j,graph), getGraph(i,j), getNbColumns, getNbRows:
```
grid = ot.GridLayout(2, 3)
for j in range(grid.getNbColumns()):
...    beta = 1.0 + j
...    grid.setGraph(0, j, ot.Gumbel(beta, 0.0).drawPDF())
...    grid.setGraph(1, j, ot.Gumbel(beta, 0.0).drawCDF())
```

GridLayout

- New VisualTest::DrawPairs(Sample) service => The pairs class is deprecated:
VisualTest_DrawPairs

- New VisualTest::DrawPairsMarginals(Sample, Distribution) service:

VisualTest_DrawPairsMarginals

- New MinimumVolumeClassifier class:
classifies data according to a minimum volume level set, defined from a distribution and a confidence level
```
# 2-d test
dists = [ot.Normal([-1.0, 2.0], [1.0]*2, ot.CorrelationMatrix(2)), ot.Normal([1.0, -2.0], [1.5]*2, ot.CorrelationMatrix(2))]
mixture = ot.Mixture(dists)
sample = mixture.getSample(1000)
distribution = ot.KernelSmoothing().build(sample)
algo = ot.MinimumVolumeClassifier(distribution, 0.8)
threshold = algo.getThreshold()
print("threshold=", threshold)
assert m.fabs(threshold - 0.0012555) < 1e-3, "wrong threshold"
cls_ref = [1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,0,1,1,1,1,1]
for i in range(35):
    x = sample[i]
    cls = algo.classify(x)
    pdf = mixture.computePDF(x)
    print(i, x, cls, pdf - threshold)
    assert cls == cls_ref[i], "wrong class"
```

MinimumVolumeClassifier

- Add CalibrationResult graphs:
drawResiduals:
res
drawParameterDistributions:
param
drawObservationsVsInputs:
obsvsin
drawObservationsVsPredictions:
obsvspred

- New SobolSimulationResult::draw to plot indices with confidence intervals
SobolSimulationResult

Relies on the enhanced static service in SobolIndicesAlgorithm.


jschueller:
cc @mbaudin47

lib/src/Base/Graph/openturns/GridLayout.hxx
```diff
@@ -0,0 +1,88 @@
...
+BEGIN_NAMESPACE_OPENTURNS
+
+
+
+/**
+ * @class GridLayout
+ *
+ * The class GridLayout
+ */
+class OT_API GridLayout : public GraphImplementation
```
regislebrun, line 37: What is the rational behind this inheritance? IMO GridLayout is essentially a matrix of graphs with some additional services, so it should be implemented this way. I may have missed something but I don't understand the interest of the proposed architecture and eg the isComposite() method.

jschueller: My point of view is that GridLayout can be considered as a Graph as it has some similar services : titles, accessors for the properties of all subgraphs (setAxes/setLegendPosition/setLogScale/setGrid/setGridColor)
Also I have to point out that not inheriting from Graph will not allow to return either a regular Graph or a GridLayout with the same prototype (covariant return types).

regislebrun: Thanks for the explanations. I am not very fond of lists allowing to change the properties of its elements by reproducing their methods. You replace `myList[i].setProperty(property)` by `myList.setProperty(i, property)`, which is not a gain. If you introduce `myList.setProperties(propertyList)` it is a different method with a different name.
Concerning the covariant return types, a `draw` method can return either a Graph or a GridLayout depending on the class and the context. I can't see a situation where the base class would return a Graph and the derived class a GridLayout.
What bothers me the most is the introduction of the `isComposite()` method. It is a way to make a confusion between a set and one of its elements. It is an error from a mathematical point of view (a set is **always** different from one of its elements) so I can't imagine it is a good thing from a code pespective.
Do you have a specific use-case where GridLayout and Graph should be on the same level?

jschueller: I dont actually need GridLayout to belong to the Graph tree, so yes it can be moved.

lib/src/Base/Graph/openturns/GraphImplementation.hxx
```diff
@@ -33,6 +33,7 @@ BEGIN_NAMESPACE_OPENTURNS
  * GraphImplementation implements graphic devices for plotting through R,
  * and manages drawables to be plotted on the same window
  */
+class Graph;
```
sofianehaddad, line 36: Why do you need this?

lib/src/Uncertainty/Algorithm/Classification/MinimumVolumeClassifier.cxx
```diff
@@ -0,0 +1,248 @@
...
+  : ClassifierImplementation()
+{
+  // Nothing to do
+}
+
+/* Default constructor */
+MinimumVolumeClassifier::MinimumVolumeClassifier(const Distribution & distribution,
+                                                         const Scalar alpha)
+  : ClassifierImplementation()
+  , distribution_(distribution)
```
sofianehaddad, line 44: You rely on `computeMinimumVolumeLevelSetWithThreshold` thus we might check continuity here, isn't it?

lib/src/Base/Graph/openturns/GraphImplementation.hxx
jschueller: it was needed when gridlayout inerited from graphimplementation I think

ChangeLog
```diff
@@ -9,20 +9,26 @@
...
 
 === Python module ===
+ * Renamed Viewer *_kwargs arguments to *_kw (matplotlib convention)
 
 === Miscellaneous ===
  * Require C99/C++11 support
 
 === Bug fixes ===
+ * #1001 (Add method SobolSimulationResult::draw)
+ * #1259 (The diagonal of a scatter plot matrix should have the histograms)
```
sofianehaddad, line 31: ok for bug fixes. We can also add these two new features to API changes

python/src/VisualTest.i
```diff
@@ -10,3 +10,7 @@
 
 %include openturns/VisualTest.hxx
 
+%pythoncode %{
+def Pairs(data, title=None, labels=None, color=None, pointStyle=None):
+    return openturns.metamodel.VisualTest.DrawPairs(data)
```
sofianehaddad, line 15: deprecated warning?

jschueller: yes, ok

PR #1555, opened 2020-07-19
Dealing with null PDFs

### LINKED ISSUE #1556: WeibullMin::computePDFGradient yields the partial derivatives in the wrong order
The following is part of the code of [WeibullMin::computePDFGradient](https://github.com/openturns/openturns/blob/f9ea0d26a7e22fdba8569c79cb108836d39e5dee/lib/src/Uncertainty/Distribution/WeibullMin.cxx#L229):

```
  const Scalar x = point[0] - gamma_;
  Point pdfGradient(3, 0.0);
  if (x <= 0.0) return pdfGradient;
  const Scalar powX = std::pow(x / beta_, alpha_);
  const Scalar factor = powX / x * std::exp(-powX);
  pdfGradient[0] = factor * (1.0 + (1.0 - powX) * std::log(powX)); // actually the formula of dPDF/dalpha
  pdfGradient[1] = factor * (powX - 1.0) * alpha_ * alpha_ / beta_; // actually the formula of dPDF/dbeta
  pdfGradient[2] = factor * (1.0 - alpha_ + alpha_ * powX) / x * alpha_;
```

The formulas for `pdfGradient[0]` and `pdfGradient[1]` are switched: 

- `pdfGradient[0]` should be the derivative of the PDF w.r.t. beta, but is the derivative w.r.t. alpha;
- `pdfGradient[1]` should be the derivative of the PDF w.r.t. alpha, but is the derivative w.r.t. beta.

### LINKED ISSUE #1558: Example estimate_multivariate_normal: FittingTest::BestModelBIC fails to compute the BIC
In the example [Estimate a multivariate distribution](http://openturns.github.io/openturns/master/examples/data_analysis/estimate_multivariate_distribution.html), generated from [estimate_multivariate_distribution.ipynb](https://github.com/openturns/openturns/blob/master/python/doc/examples/data_analysis/estimate_multivariate_distribution.ipynb), cell 8, copulas are built from 2D marginals of a 4D Sample.

The method `BestModelBIC` of the class `FittingTest` is used, but it fails to accurately compute the BIC. Let us run the notebook and paste the following code in a cell at the bottom:

```
print("-2 * LogMinScalar =", -2.0 * ot.SpecFunc.LogMinScalar)
for num, bloc in enumerate(blocs):
    print("###################################################################")
    print("BIC of", estimated_copulas[num], "for the bloc", bloc, ":")
    print(ot.FittingTest_BIC(sample.getMarginal(bloc), estimated_copulas[num]))
```

We get the following output:

```
-2 * LogMinScalar = 1416.7928370645282
###################################################################
BIC of AliMikhailHaqCopula(theta = 0.601685) for the bloc [0, 3] :
1416.7928370646807
###################################################################
BIC of ClaytonCopula(theta = 2.45478) for the bloc [1, 2] :
1416.7928370646807
```

As this code suggests, I think the issue is that we hit a threshold based on `SpecFunc::LogMinScalar`.

Moreover, I believe the only reason `BestModelBIC` still manages to build the copulas is that for both blocs ([0, 3] and [1, 2]), only one of the considered CopulaFactories actually manages to build a copula, so the BIC computation is irrelevant.

josephmure:
At the last CT, we discussed how we could try replacing `SpecFunc::LogMinScalar` by `-SpecFunc::MaxScalar` whenever possible.

This has several advantages:

1. It makes the code more in line with the mathematical intuition (exp(-inf)=0).
2. OT can deal with wider ranges in the log domain.
3. We can probably remove checks like ``if (logPDF==SpecFunc::LogMinScalar) return ...``
4. It might solve @MerlinKeller's bug regarding not being able to perform a `RandomWalkMetropolisHastings` (#1532)

So this draft PR tries to implement these changes. Along the way, I have noticed that `SpecFunc::LogMinScalar` is sometimes replaced by `-SpecFunc::LogMaxScalar`, `-SpecFunc::MaxScalar`, or even `0.0`, which seems strange. Some bugs might be fixed once this PR is completed. In particular, it fixes #1556 and could close #1558.


josephmure:
This PR is now ready for review, I think. All checks finally passed.

josephmure:
PR superseded by #1566.

PR #1559, opened 2020-07-27
Fix numpy broadcasting error in example

josephmure:
In the [Advanced Kriging](https://github.com/josephmure/openturns/blob/master/python/doc/examples/meta_modeling/kriging_advanced.ipynb) example (cell 8), in order to draw confidence intervals, Samples are converted to numpy arrays of shape (1000,) with the `np.hstack` command and then added to a numpy array with shape (1,1000). The resulting array has shape (1000,1000). This is not the desired behavior.
It turns out that this conversion is useless, because the sum of a Sample with size 1000 and dimension 1 and of a numpy array with shape (1,1000) is a Sample with size 1000 and dimension 1 which can be drawn by matplotlib.

### DIFF
--- a/python/doc/examples/meta_modeling/kriging_advanced.ipynb
+++ b/python/doc/examples/meta_modeling/kriging_advanced.ipynb
@@ -268,8 +268,8 @@
    "source": [
     "level = 0.95\n",
     "quantile = ot.Normal().computeQuantile((1-level)/2)[0]\n",
-    "borne_sup = np.hstack(krigingMeta(x_plot)) + quantile * np.sqrt(variance)\n",
-    "borne_inf = np.hstack(krigingMeta(x_plot)) - quantile * np.sqrt(variance)\n",
+    "borne_sup = krigingMeta(x_plot) + quantile * np.sqrt(variance)\n",
+    "borne_inf = krigingMeta(x_plot) - quantile * np.sqrt(variance)\n",
     "\n",
     "fig, ax = plt.subplots(figsize=(8, 8))\n",
     "ax.plot(x, y, ('ro'))\n",


josephmure:
When run, cell 8 of the notebook currently produces:
bad_confidence_interval

After the patch, it produces the expected plot:
good_confidence_interval

sofianehaddad:
I missed something : according to https://github.com/openturns/openturns/blob/master/python/doc/examples/meta_modeling/kriging_advanced.ipynb the cell 8 is correct. In fact `np.hstack` transforms kriging output to an array of shape of size `(1000,)` but there is no array of shape `(1,1000)`. Where do you see such an array? IMHO current notebook is correct

josephmure:
Sorry, I got one of the shapes wrong. Here they are:

```
print("Shape of np.hstack(krigingMeta(x_plot)):", np.hstack(krigingMeta(x_plot)).shape)
print("Shape of np.sqrt(variance):", np.sqrt(variance).shape)
print("Shape of the sum:", (np.hstack(krigingMeta(x_plot)) + np.sqrt(variance)).shape)
```

```
Shape of np.hstack(krigingMeta(x_plot)): (1000,)
Shape of np.sqrt(variance): (1000, 1)
Shape of the sum: (1000, 1000)
```

@sofianehaddad have you tried running the notebook? I am not sure the compiler does: I think it simply reads the .ipynb file (output included), which I assume was run at a time numpy (or OpenTURNS) had a different behavior.

sofianehaddad:
@josephmure I am quiet surprised. According to github history https://github.com/openturns/openturns/commits/master/python/doc/examples/meta_modeling/kriging_advanced.ipynb last update of the nb was done in 9edcef69148540518799567654aa451257ca59d6 And it runs at this time without any issue. I am wondering why now it has changed ? @jschueller any idea?

jschueller:
I dont know why it changed but for sure I've got the same error, removing the hstack fixes it.
Note that notebooks are not rerendered when the sphinx doc is built.

PR #1569, opened 2020-08-07
Use c++11 override keyword

jschueller:
Several advantages:

1. Allows to mark overloaded methods (a warning is emitted if missing).
2. Allows to avoid to mispell the virtual methods (an error is emitted)

The rule is to use virtual in parent class, override in child class


jschueller:
@regislebrun @sofianehaddad I'd like to ask you for a review here, as I want to merge it quickly before we get conflicts everywhere (this pr touches most headers)

sofianehaddad:
@jschueller indeed. This should be merged asap

jschueller:
thanks for the quick reply

PR #1566, opened 2020-08-05
Expand logPDF range below LogMinScalar

josephmure:
At the last CT, we discussed how we could try replacing `SpecFunc::LogMinScalar` by `-SpecFunc::MaxScalar` whenever possible. This PR implements this idea. It supersedes #1555 and supersedes #1532.

It has several advantages:

1. It makes the code more in line with the mathematical intuition (exp(-inf)=0).
2. OT can deal with wider ranges in the log domain.
3. We can probably remove checks like ``if (logPDF==SpecFunc::LogMinScalar) return ...``
4. It might solve @MerlinKeller's bug regarding not being able to perform a `RandomWalkMetropolisHastings`.

Note that `SpecFunc::LogMinScalar` is sometimes replaced by `-SpecFunc::LogMaxScalar`, `-SpecFunc::MaxScalar`, or even `0.0`, which seems strange. This PR fixes several undetected bugs and corrects several `.expout` files. It also closes #1558.

PR #1574, opened 2020-08-13
Kd/documentation

KieranDelamotte:
This PR is related to EDF-IMACS contract on documentation and sums up ongoing work.
======================================================================

Important : this PR uses the sphinx-gallery extensions.
---------------------------------------------------------------------

Different topics are covered : 
  1. add graphics to rst pages (see Fig1)
  2. add galleries to present examples (see Fig2)
  3. customize the search box (see Fig3)

More precisely,

Subject no. 1 consists in adding rst directives to illustrate the categories. Fig1 is a screenshot of the produced documentation.

Subject no. 2 is based on the sphinx-gallery extensions to generate thumbnails for python examples (*.py files). First the notebooks (*ipynb files) have been converted to python+markdown files and the markdown directives have been replaced by rst ones in order to meet sphinx-gallery requirements.
Eventually the notebooks are removed from the sources. A typical example html page is depicted in Fig4.
Each category of examples has its own gallery and it looks like Fig2. where each thumbnail is clickable.

Subject no. 3 is a little hack of the search box which enable to narrow down the search to 3 main categories (appearing in that order) :
  - Examples
  - API
  - Theory

Please note that due to sphinx limitation (still true?) the javascript tool is copied to the install directory through CMake instead of the conf.py configuration file. An example of a search is illustrated in figure Fig3.

Finally a new section is appearing : common use cases. The aim is to factorize redundant code  and description for common models (beam, flood model, ...) to improve examples readability and maintenance (see Fig5).

Any feedback is appreciated,

Kieran
Fig1
Fig2
Fig3
Fig4
Fig5


jschueller:
- so all notebooks will have to be converted to rst ? wasnt there a way to keep ipynbs and reference its images ?
- sphinx is not responsible for the file install, it's cmake, for js files to be copied just add the *.js extension to sphinx_sources (see python/src/CMakeLists.txt:~1011) instead of copying the js folder
- I'm adding sphinx-gallery to our docker image, so you can spare the "pip install", but you will want to add the detection logic to the root CMakeLists since its mandatory now, and mention it in the dependencies (developer_guide/architecture.rst)
- I did not succeed building it locally, unfortunately the linux ci did not appear, hopefully it will trigger at next push

josephmure:
>  so all notebooks will have to be converted to rst ? wasnt there a way to keep ipynbs and reference its images ?

Sphinx-gallery requires Python scripts with rst directives in order to generate examples. @KieranDelamotte explored another solution based on the extension [sphinx-nbexamples](https://github.com/Chilipp/sphinx-nbexamples/graphs/contributors) which emulates sphinx-gallery except with notebooks instead of scripts. However, there are [few contributors](https://github.com/Chilipp/sphinx-nbexamples/graphs/contributors) and the author did not answer @KieranDelamotte's message, so @mbaudin47 was concerned about maintenance issues. Overall, I think using scripts rather than notebooks is an improvement because it will make version control easier and all examples will be run every time the doc is compiled, which will make sure they are not broken. Notebook fans will still be able to download .ipynb files directly from the website. People who prefer scripts will also be able to download them that way. The only issue is the transition, and @KieranDelamotte has already done the heavy lifting.

> I did not succeed building it locally, unfortunately the linux ci did not appear, hopefully it will trigger at next push

Yes, there are issues remaining. I was able to build it locally, but I had to fiddle around with one of the `CMakeLists.txt` files.

KieranDelamotte:
Hi folks !

I hope this PR will be merged into the master after the next COTECH (09.17). @sofianehaddad , @regislebrun  would  you please take a cursory look at this PR ?

The main idea is to remove the notebook examples and use python scripts and use sphinx-gallery to generate nice images as shown.

Kieran

jschueller:
I think there are unresolved conflicts in the examples file that were edited in the mean time: some ipynbs are still there:
```
python/doc/examples/meta_modeling/kriging_1D.ipynb
python/doc/examples/meta_modeling/kriging_beam_arbitrary_trend.ipynb
python/doc/examples/meta_modeling/kriging_cantilever_beam.ipynb
python/doc/examples/meta_modeling/kriging_simulate.ipynb
python/doc/examples/meta_modeling/kriging_sequential.ipynb
python/doc/examples/meta_modeling/create_linear_least_squares_model.ipynb
python/doc/examples/meta_modeling/kriging_advanced.ipynb
python/doc/examples/meta_modeling/kriging_cantilever_beam_hmat.ipynb
python/doc/examples/meta_modeling/kriging_beam_trend.ipynb
python/doc/examples/meta_modeling/kriging_hyperparameters_optimization.ipynb
python/doc/examples/meta_modeling/functional_chaos.ipynb
python/doc/examples/calibration/calibration_logistic.ipynb
python/doc/examples/calibration/calibration_deflection_tube.ipynb
python/doc/examples/calibration/calibration_flooding.ipynb
python/doc/examples/calibration/calibration_chaboche.ipynb
python/doc/examples/graphs/graphs_basics.ipynb
python/doc/examples/reliability_sensitivity/sensitivity_sobol_simulation.ipynb
python/doc/examples/reliability_sensitivity/plot_design.ipynb
python/doc/examples/reliability_sensitivity/design_of_experiment_continuous_discrete.ipynb
python/doc/examples/reliability_sensitivity/estimate_probability_form.ipynb
python/doc/examples/numerical_methods/estimate_integral_Fejer.ipynb
python/doc/examples/probabilistic_modeling/user_stationary_covmodel.ipynb
python/doc/examples/data_analysis/kolmogorov_distribution.ipynb
python/doc/examples/data_analysis/visualize_clouds.ipynb
python/doc/examples/data_analysis/visualize_pairs.ipynb
python/doc/examples/data_analysis/quick_start_point_and_sample.ipynb
```
You must check that the conversion did not miss their modifications and update them if necessary.
Also we should stop editing into python/doc/examples/ untils this is merged.

python/src/viewer.py
```diff
@@ -21,6 +21,14 @@
 import re
 import warnings
 import io
+import argparse
+
+# Allows for hiding figures if needed from the command line
+parser = argparse.ArgumentParser(description='Viewer parser.')
+parser.add_argument("--noshow", action="store_true", help="Hide figures if used.", dest="noshow", default=False )
+args = parser.parse_args()
+_noshow = args.noshow
```
jschueller, line 30: use a non-interactive matplotlib backend (agg) instead
you can set the env variable MPLBACKEND=agg as a test property

CMakeLists.txt
```diff
@@ -550,19 +550,16 @@ if (BUILD_PYTHON)
...
-          if (NOT IPYTHON_FOUND OR SPHINX_VERSION_STRING VERSION_LESS 1.5.1)
-            set (NBSPHINX_FOUND FALSE CACHE BOOL "")
-            message (STATUS "Disabling notebook documentation examples (need pandoc/nbsphinx/ipython)")
-          endif ()
-        endif ()
+        # the sphinx-gallery extension
+        if (USE_SPHINX_GALLERY)
+            find_python_module (sphinx_gallery 0.7)
+        else ()
+            message (STATUS "Could NOT find shpinx-gallery extension")
```
jschueller, line 557: shpinx

CMakeLists.txt
```diff
@@ -550,19 +550,16 @@ if (BUILD_PYTHON)
...
+        endif()
+        if (NOT SPHINX_GALLERY_FOUND OR SPHINX_GALLERY_VERSION_STRING VERSION_LESS 0.7)
+            set (SPHINX_FOUND FALSE)
+            message (STATUS "Disabling documentation (need sphinx_gallery >= 0.7)")
+        endif()
```
jschueller, line 562: seems we also need all the optional dependencies the examples may use (hmat, pandas, libxml, ...) to enable the doc

python/doc/developer_guide/architecture.rst
```diff
@@ -184,65 +184,67 @@ Dependencies
...
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| ReSt to HTML (optional for doc)       | `Sphinx <http://sphinx-doc.org/>`_                                              | 1.1               |
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| Sphinx extension (optional for doc)   | `Numpydoc <https://github.com/numpy/numpydoc/>`_                                | 0.9.0             |
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| Sphinx extension (optional for doc)   | `Sphinx-Gallery <https://sphinx-gallery.github.io/stable/index.html>`_          | 0.7               |
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| Sphinx extension (optional for doc)   | `nbsphinx <http://nbsphinx.rtfd.io/>`_                                          | N/A               |
```
jschueller, line 242: we're dropping nbsphinx right ?

python/doc/developer_guide/architecture.rst
```diff
@@ -184,65 +184,67 @@ Dependencies
...
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| Optimization (optional)               | `Ipopt <https://coin-or.github.io/Ipopt/>`_                                     | 3.11.9            |
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| CSV parser (optional)                 | flex                                                                            | 2.5.33            |
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| CSV parser (optional)                 | bison                                                                           | 2.4               |
++---------------------------------------+---------------------------------------------------------------------------------+-------------------+
+| XML support (optional)                | `LibXml2 <http://xmlsoft.org/>`_                                                | 2.6.27            |
```
jschueller, line 222: beware I just merged hdf5 adding a line here support so you might get a conflict

python/doc/developer_guide/architecture.rst
jschueller: there may be other references to nbsphinx to remove (conf.y, etc)

KieranDelamotte:
Hi @jschueller 

Thanks for the comments, I will look at them in detail after a talk with @mbaudin47  and @josephmure this pm.

CMakeLists.txt
jschueller: also SPHINX_GALLERY_FOUND implies version>=0.7 when calling find_python_module (sphinx_gallery 0.7)

CMakeLists.txt
jschueller: no message is needed if we do not intend to use it

.circleci/config.yml
```diff
@@ -13,6 +13,7 @@ jobs:
           command: |
             cd /tmp && git clone --depth 1 https://github.com/swig/swig.git && cd swig && ./autogen.sh && ./configure --without-alllang && make && sudo make install
             sudo pip install git+https://github.com/sphinx-doc/sphinx.git
+            sudo pip install sphinx-gallery
```
jschueller, line 16: sphinx gallery was added to the docker image, so this can be removed

jschueller:
in python/test, the example optimization_bonmin does not exist, so this fails:
```
  #if (NOT Bonmin_FOUND)
  #  set_tests_properties (pyinstallcheck_example_plot_optimization_bonmin PROPERTIES WILL_FAIL TRUE)
  #endif()
```

jschueller:
there is still one notebook: python/doc/examples/calibration/calibration_withoutobservedinputs.ipynb

josephmure:
Travis failed because the log became too long due to all the warnings. @jschueller Is it possible to suppress warnings on a system-wide basis?

jschueller:
- there is still a reference to nbsphinx in python/src/CMakeLists.txt
- please move the use cases rst in their own directory, they are dangling are the python/doc folder, also they have no test on their own, would it be possible to add one ?
- the cmake logic in root CMakeLists.txt does not take into account the case where one of the dependencies used to run the plot script is missing (pandas, hmat, ...) : either get to be optional using the tag api (see nbsphinx, it was optional), or the documentation must be disabled when they're missing
- would it be possible to squash some commits ? there are several diffs on the same files making this hard to read

KieranDelamotte:
Hi @jschueller 

About your second point, I do not understand what you mean by  "also they have no test on their own, would it be possible to add one ?"  For the organization part I agree they have to be in a separated directory.

For your last point, if a dependency (say Hmat) is missing the related examples will be associated with a "Broken" thumbnail in the generated gallery. This seems good to me.

jschueller:
I meant the usecases module (import openturns.usescases) is not tested outside the documentation generation step, in python/test
For the last point, I thought when pandas or hmat is not available, the corresponding .py in the example will fail and so would the sphinx build

mbaudin47:
@KieranDelamotte : I guess that Julien meant to create one unit test per usecase, as the Viewer.py module has a test in python/test:

https://github.com/openturns/openturns/blob/master/python/test/t_Viewer.py

Adding a "if __name__ == "__main__" :" in the module is not a proper unit test. 
The unit test should take the form of a script using assert functions, e.g.:

https://github.com/openturns/openturns/blob/593fe48362370fbfaac13eb5ed91c778b45dc651/python/test/t_GaussianLinearCalibration_std.py#L73

KieranDelamotte:
OK I see.

jschueller:
one test file for all testcases is ok too, there's not much to test anyways

jschueller:
@KieranDelamotte I'd really like to get this merged, do you think you can get it ready for next week ?

KieranDelamotte:
Hi @jschueller 

I will surely do my best. However, some of the errors I see from the automated checks are puzzling :

- mingw has an error because of the numpy package (already used in  the example I get it from)
- appveyor and linux have errors related to the dataclass syntax they do not understand
- I have seen errors due to too much warnings. 

I see a warning in a rst generated from sphin-gallery treated as an error and that I can definitely do somethong about it.

jschueller:
- tests using numpy are disabled if numpy is not found by cmake, but here perhaps numpy can be removed ?
- dataclass is python3.7+, seems appveor uses 3.6, tests relying on these must be disabled depending on the python version (we want to run tests on conda that still provides 3.6)
- yes, on travis, it seems a plot script emits too mush of them, the script must be found and its log silenced, see openturns.Log class

mbaudin47:
It sounds astonishing: mingw VM has no numpy?

mbaudin47 (changes requested):
This is destructive. I suggest to insert the appropriate UTF8 header in the script in order to manage encoding.

https://stackoverflow.com/questions/13807748/when-to-use-utf8-as-a-header-in-py-files/13807795

jschueller:
in our context its better to avoid these characters at all

mbaudin47:
> in our context its better to avoid these characters at all

Encoding is the correct solution for it and must be used when appropriate. Furthermore, this will show an example of using UTF8 characters in future examples.

KieranDelamotte:
For the travis-cr check I have for example :

pyinstallcheck_example_plot_qqplot_graph ...................................***Timeout 100.15 sec

This seems strange because it is really quick on my laptop. What am I missing ?

jschueller:
weird, maybe agg is not available and it falls back to an interactive one ?

jschueller:
seems agg is available, and plt import fine, must be something else

jschueller:
this hangs on Travis:
```
MPLBACKEND=cairo python -c "import matplotlib.pyplot as plt;plt.plot([1,23,2,4]);plt.ylabel('some numbers');plt.show();print('ok')"
```

KieranDelamotte:
> this hangs on Travis, the import alone is ok:
> 
> ```
> MPLBACKEND=agg python -c "import matplotlib.pyplot as plt;plt.plot([1,23,2,4]);plt.ylabel('some numbers');plt.show();print('ok')"
> ```

I do not understand your comment. Is this script wrong ? It seems ok to me. 

Also, one of the test still pending of the appveyor is related to a fail of pyinstallcheck_InverseGamma_std (Timeout) which is not related to the python examples.

jschueller:
This script is ok, but it hangs on Travis, while switching to python3 works.

.travis.yml
```diff
@@ -35,12 +35,11 @@ before_install:
         brew unlink python@2 &&
         brew install swig openblas bison flex cminpack ceres-solver dlib nlopt tbb &&
         export HOMEBREW_PREFIX=`brew --prefix`;
+        pip3 install matplotlib scipy
```
jschueller, line 38: oups, missing ; at the end of the line

.travis.yml
```diff
@@ -35,12 +35,11 @@ before_install:
...
     fi
 
 install:
-# use system python
-  - export PATH=/usr/bin:$PATH
```
jschueller, line 42: I believe it was removing this export PATH=/usr/bin:$PATH that unexpectedly make the cblas detection fail

mbaudin47:
Travis fails for a missing cblas header required by hmat.

jschueller:
can you try to add to add cmake in the "brew install" command ?

jschueller:
good that it passes on the ci.

when I build a second time, I've got an error related to external_program.py not containing docstrings, it goes through if I add (external_program) to the sphinx gallery ignore_pattern

but then it fail with another error:
/home/schueller/projects/openturns/kieran/build/python/src/sphinx_build/auto_py_calibration/bayesian_calibration/plot_bayesian_calibration.rst:Le document n'est inclut dans aucune table des matières de l'arborescence.

jschueller:
My changes to the calibration examples we not taken into account
https://github.com/openturns/openturns/pull/1551
It comes down to using the new methods from CalibrationResult and get rid of the custom class

mbaudin47:
Are you sure that the new examples from #1604 and #1593 are included?

josephmure:
I think the new examples from #1604 and #1593 do not necessarily need to be included in this PR. They could be reintroduced later as part of separate PRs.

jschueller:
yes, they can be added later, I'm only concerned not to loose work already included

josephmure:
@jschueller You could perhaps reopen them. I understand they cannot be merged, but they would remind us that they still need to be included.

jschueller:
here is a patch to include changes from 1563, 1597 and 1551:
https://gist.github.com/jschueller/d59d1d2713f2fbdae8f90ba2239a0422

KieranDelamotte:
Hi, 
Thank you for the patch ! I compile it locally before pushing it.

Concerning the fails error, what have you done ? I remember having some issue (even with master) when doing a second `make install` in my `build` directory. Also the size of the `build` directory increases with each `make install`.

jschueller:
- to get the failure, I get it if I do "make install" twice
- for the build dire growing, maybe it's the images that are recreated

KieranDelamotte:
@mbaudin47 
Concerning #1593, I converted the new notebooks but :
- the function 'g' in Y = CompositeRandomVector(g,X) is not yet defined at this point. Maybe still in memory within the notebook. I have deleted this line.
- the probaSystemEvent example takes a very long time on my laptop. Do you confirm this ? Not sure if it is safe to add it (at least with the chosen parameters).
- subsetSampling is OK

On a side note, in which category do I place the `createYourOwnDist-vMB` example you sent me by email ? This one is also OK.

sofianehaddad (approved):

mbaudin47:
> @mbaudin47
> Concerning #1593, I converted the new notebooks but :
>     * the function 'g' in Y = CompositeRandomVector(g,X) is not yet defined at this point. Maybe still in memory within the notebook. I have deleted this line.
>     * the probaSystemEvent example takes a very long time on my laptop. Do you confirm this ? Not sure if it is safe to add it (at least with the chosen parameters).
>     * subsetSampling is OK
> On a side note, in which category do I place the `createYourOwnDist-vMB` example you sent me by email ? This one is also OK.

These were NB from @adutfoy : perhaps she can provide more insight on these examples. 
The  `createYourOwnDist-vMB` example might be located in Probabilistic modeling> Distributions.

josephmure:
Thanks @KieranDelamotte for including my changes to the `estimate_multivariate_example`! I can now safely delete this change from PR #1566.

PR #1562, opened 2020-07-31
SobolIndicesAlgorithm doc clarified

josephmure:
A [discussion on Gitter](https://gitter.im/openturns/community?at=5f23bd57a2be904990815418) with @christelle-git revealed that the API doc of Sobol' indices-related classes could use some improvement. This PR rewords the API doc of the following classes (see [post below](https://github.com/openturns/openturns/pull/1562#issuecomment-671188945) for details).

- [SobolIndicesExperiment](https://openturns.github.io/openturns/master/user_manual/_generated/openturns.SobolIndicesExperiment.html)
- [SobolIndicesAlgorithm](https://openturns.github.io/openturns/master/user_manual/_generated/openturns.SobolIndicesAlgorithm.html)
- [SaltelliSensitivityAlgorithm](https://openturns.github.io/openturns/master/user_manual/_generated/openturns.SaltelliSensitivityAlgorithm.html)
- [MartinezSensitivityAlgorithm](https://openturns.github.io/openturns/master/user_manual/_generated/openturns.MartinezSensitivityAlgorithm.html)
- [MauntzKucherenkoSensitivityAlgorithm](https://openturns.github.io/openturns/master/user_manual/_generated/openturns.MauntzKucherenkoSensitivityAlgorithm.html)
- [JansenSensitivityAlgorithm](https://openturns.github.io/openturns/master/user_manual/_generated/openturns.JansenSensitivityAlgorithm.html)

Also, it increases the docstring example sample sizes in order to obtain valid results (negative Sobol' indices estimates were obtained before) and makes docstring examples across (Saltelli|Martinez|MauntzKucherenko|Jansen)SensitivityAlgorithm homogeneous.

It makes one change in the library: ``SobolSequence::MaxNumberOfDimension`` becomes ``SobolSequence::MaxNumberOfDimensions``.

PR #1575, opened 2020-08-17
SobolIndicesAlgorithm doc: remove erroneous bold Y

josephmure:
Patch to PR #1562. Forgot to remove several bold letters Y.


jschueller:
moved to my #1565 branch

Issue #1581, opened 2020-08-31
LinearModelAlgorithm run() fails to parse Sample description


josephmure:
The `LinearModelAlgorithm::run` method fails to parse the description of the input sample when said description contains whitespaces or parentheses (and perhaps other special characters).

Example (using the current master 1161ce7) :

```
import pandas as pd
import numpy as np
import openturns as ot

# Create Pandas dataframe
cars = {'Brand': ['Honda Civic','Toyota Corolla','Ford Focus','Audi A4'],
        'Price (euros)': [22000,25000,27000,35000],
        'Mass (kg)': [1500, 1200, 2300, 1850]
        }

df = pd.DataFrame(cars, columns = ['Brand', 'Price (euros)', 'Mass (kg)'])

print (df)

# Remove "Brand columns" that does not contain floats
sub_df = df[['Price (euros)', 'Mass (kg)']]

# Convert to OT Sample
ech = ot.Sample(np.array(sub_df))

# Names are lost in the transition: restore them
ech.setDescription(sub_df.columns)

print(ech)

# Define linear model
lin = ot.LinearModelAlgorithm(ech[:, 0], ech[:, 1])

# Attempt to run the model
lin.run()
```

Output:
```
            Brand  Price (euros)  Mass (kg)
0     Honda Civic          22000       1500
1  Toyota Corolla          25000       1200
2      Ford Focus          27000       2300
3         Audi A4          35000       1850
    [ Price (euros) Mass (kg)     ]
0 : [ 22000          1500         ]
1 : [ 25000          1200         ]
2 : [ 27000          2300         ]
3 : [ 35000          1850         ]

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-64-f5625d982e31> in <module>()
     26 
     27 # Attempt to run the model
---> 28 lin.run()
     29 

.../install/lib/python3.7/site-packages/openturns/metamodel.py in run(self)
   6018         :class:`~openturns.MetaModelResult` structure containing all the results.
   6019         """
-> 6020         return _metamodel.LinearModelAlgorithm_run(self)
   6021 
   6022 

TypeError: InvalidArgumentException : Invalid input variable: Price (euros)
```

The following lines also fail:

```
# Even without (), the problem remains
ech.setDescription(['Price euros', sub_df.columns[1]])

print(ech)

# Try the linear model again
lin = ot.LinearModelAlgorithm(ech[:, 0], ech[:, 1])
lin.run()
```

[...]

Issue #1553, opened 2020-07-14
Optimization algorithms ignore MaxEvaluationNumber parameter in SORM


tsiplugin:
Good evening!
I try to solve reliability analysis problem using SORM with the Cobyla optimization algorithm. I've set the MaximumEvaluationsNumber to 3, but I'd get much more calls of the PythonFunction. Then I've made an experiment. After experiment l could see, that if blackbox function gets i-dimentional vector as argument and returns j-dimentional vector as result, then:
               if  j == 1 then extra_function_calls = 2(i^2) + 6i + 2
               if  j > 1 then extra_function_calls = 2(i^2) + 4i + 5

Here is my experiment code:

```
import openturns as ot
ot.RandomGenerator.SetSeed(100)
global counter
max_eval = 3
for i in range(1, 10):
    for j in range(1, 10):
        counter = 0
        input_amount = i
        output_amount = j
        def my_func(args) -> ot.Point:
            global counter
            counter+=1
            returnum = [0]*output_amount
            for n in range(output_amount):
                for m in range(input_amount):
                    returnum[n] = m*(args[m]**m)
            return returnum
        my_function = ot.PythonFunction(input_amount, output_amount, func=my_func)
        a = [200]*input_amount
        b = [10]*input_amount
        my_distribution = ot.Normal(a, b, ot.IdentityMatrix(input_amount))
        vect = ot.RandomVector(my_distribution)
        output = ot.CompositeRandomVector(my_function, vect)
        lower_bounds = [0]*output_amount
        upper_bounds = [1]*output_amount
        lower_bounds_finite = [False]*output_amount
        upper_bounds_finite = [True]*output_amount
        domain = ot.Interval(lower_bounds, upper_bounds, lower_bounds_finite, upper_bounds_finite)
        event = ot.ThresholdEvent(output, domain)
        solver = ot.Cobyla()
        #solver = ot.AbdoRackwitz()
        solver.setMaximumEvaluationNumber(max_eval)
        solver.setVerbose(True)
        solver.setMaximumAbsoluteError(1e-10)
        algo = ot.SORM(solver, event, my_distribution.getMean())
        algo.run()
        result = algo.getResult()
        print("Input args amount = %s" % str(input_amount))
        print("Output results amount = %s" % str(output_amount))
        print("Total number of function calls = %s" % str(my_function.getEvaluationCallsNumber()))
        print("counter = %s" % counter)
        print("Extra function call = %s\n" % (counter - max_eval))
```

[...]

Issue #1525, opened 2020-06-16
FieldImplementation does not use the description for the Graph


mbaudin47:
The following script:
```
import openturns as ot
import openturns.viewer as otv

numberOfIntervals = 20
interval = ot.Interval(-1.0, 1.0)
mesh = ot.IntervalMesher([numberOfIntervals - 1]).build(interval)
outputDimension = 2
univariateCovariance = ot.SquaredExponential()
covarianceCollection = [univariateCovariance] * outputDimension
multivariateCovariance = ot.TensorizedCovarianceModel(covarianceCollection)
process = ot.GaussianProcess(multivariateCovariance, mesh)
sampleSize = 100
processSample = process.getSample(sampleSize)
#
field = ot.Field(mesh, processSample[0])
field.setName("Field name")
field.setDescription(["Time", "First field", "Second field"])
graph = field.drawMarginal(0)
otv.View(graph)
```

produces:

![image](https://user-images.githubusercontent.com/31351465/84761983-8225ee80-afca-11ea-9d80-eb62701875f1.png)

We see that the YTitle of the graph does not match the description of the field: it should be "First field", but it is "Values".

The bug is at:

https://github.com/openturns/openturns/blob/a53885f6a612c2d4fb2d7c9f5aca48e08b608153/lib/src/Base/Stat/FieldImplementation.cxx#L480

where the `Graph` is generated with a constant "Values" parameter, instead of using the `1+index`-th description within the Field itself.

mbaudin47:
The draw methods of the `Mesh` class have the same issue.

The following script:
```
import openturns as ot
import openturns.viewer as otv

timesNumber = 10
times = ot.Sample(range(timesNumber), 1)
timeInterval = ot.Interval(0.0, timesNumber - 1.0)
numberOfIntervals = timesNumber - 1
mesher = ot.IntervalMesher([numberOfIntervals])
mesh = mesher.build(timeInterval)
mesh.setName("Index (H)")
graph = mesh.draw()
otv.View(graph)
```
produces:

![image](https://user-images.githubusercontent.com/31351465/84764501-1ba2cf80-afce-11ea-8264-63879567aa92.png)

There are two bugs:
* The `draw` method of the `Mesh` class do not use the description of the vertices:
https://github.com/openturns/openturns/blob/a53885f6a612c2d4fb2d7c9f5aca48e08b608153/lib/src/Base/Geom/Mesh.cxx#L520
[...]

Issue #1470, opened 2020-04-30
setNbModes is sometimes ignored


mbaudin47:
The setNbModes method of a KarhunenLoeveAlgorithm appeared in v1.15. 

* There is no example in the doc which shows it in action. 

* The option is only used in the  KarhunenLoeveP1Algorithm. Hence, we can use the setNbModes  method of KarhunenLoeveSVDAlgorithm or KarhunenLoeveQuadratureAlgorithm, but it is simply ignored. This is a design issue.

mbaudin47:
As a workaround, I created the following class.

```
"""Truncate a Karhunen-Loeve result with a given number of modes."""
import openturns as ot
import otfrasil


class TruncatedKarhunenLoeveResult():
    """Truncate a Karhunen-Loeve decomposition with given number of modes."""

    def __init__(self, karhunenLoeveResult, numberOfModes):
        """
        Truncate a Karhunen-Loève decomposition with given number of modes.

        The number of mode is the minimum of the given number of modes
        and the number of modes in the given decomposition.

        Parameters
        ----------
        karhunenLoeveResult : ot.karhunenLoeveResult
            The result of a Karhunen-Loeve decomposition
        numberOfModes : int
            The number of modes.
        """
        self.karhunenLoeveResult = karhunenLoeveResult
        self.numberOfModes = numberOfModes
        return

    def run(self):
        """Truncate the decomposition."""
        # Truncate eigenvalues
        eigenvalues = self.karhunenLoeveResult.getEigenValues()
        fullNumberOfModes = eigenvalues.getDimension()
        numberOfModes = min(self.numberOfModes, fullNumberOfModes)
        truncatedEigenvalues = eigenvalues[:numberOfModes]
        truncatedThreshold = eigenvalues[numberOfModes - 1] / eigenvalues[0]
        # Truncate modes
        modes = self.karhunenLoeveResult.getModes()
        truncatedModes = ot.FunctionCollection()
        for i in range(numberOfModes):
            truncatedModes.add(modes[i])
        # Truncate process sample modes
        modesAsProcessSample = self.karhunenLoeveResult.getModesAsProcessSample()
        truncatedModesAsProcessSample = ot.ProcessSample()
        for i in range(numberOfModes):
            truncatedModesAsProcessSample.add(modesAsProcessSample[i])
        # Truncate projection matrix
        projectionMatrix = self.karhunenLoeveResult.getProjectionMatrix()
        nbCols = projectionMatrix.getNbColumns()
[...]

PR #1588, opened 2020-09-07
Various fixes

jschueller:
- #1581 
- #1553: move lagrange multipliers computation as a post-processing of OptimResult
(moved computeLagrangeMultipliers from Algo to Result, deprecated OptimResult::getLagrangeMultipliers)
- Added ProcessSample::getField to Python interface (as [i] operator returns a Sample now we miss this accessor)
- SimulatedAnnealingLHS: default to phip instead of replacing mindist by phip, cc @mbaudin47 
- #1592 
- SubsetSampling: go back to previous step if the new threshold is too close to the global threshold
- #1515 
- #1529
- #1525
- #1470
- Add ResourceMap::FindKeys
- KL: Cap number of modes with setNbModes for svd & quadrature
- ParetoFactory: enhance least-squares estimator (numerical optimization of gamma first instead everything at once)
- #1599


lib/src/Base/Optim/OptimizationResult.cxx
```diff
@@ -433,6 +430,74 @@ Graph OptimizationResult::drawOptimalValueHistory() const
...
+   The Lagrange multipliers are stored as [l_eq, l_lower_bounds, l_upper_bounds, l_ineq], where:
+   * l_eq is of dimension 0 if no equality constraint, else of dimension the number of scalar equality constraints
+   * l_lower_bounds and l_upper_bounds are of dimension 0 if no bound constraint, else of dimension dim(x) for both of them
+   * l_ineq is of dimension 0 if no inequality constraint, else of dimension the number of scalar inequality constraints
+
+   so if there is no constraint of any kind, the Lagrange multipliers are of dimension 0.
+ */
+Point OptimizationResult::computeLagrangeMultipliers(const Point & x) const
+{
+  const Scalar maximumConstraintError = ResourceMap::GetAsScalar("OptimizationAlgorithm-DefaultMaximumConstraintError");
```
regislebrun, line 446: It could even be an argument of the method, with the ResourceMap value as a default value

jschueller: oh, this one is private, I kept getLagrangeMultipliers which calls computeLagrangeMultipliers on x^

jschueller: @regislebrun an update on this: I removed the OptimizationResult attribute and deprecated OptimizationResult::getLagrangeMultipliers, OptimizationAlgo::computeLagrangeMultipliers, now we have only OptimizationResult::computeLagrangeMultipliers(x) and OptimizationResult::computeLagrangeMultipliers()

regislebrun: The idea of having both getLagrangeMultipliers() and computeLagrangeMultipliers() was to avoid computing them twice if the user asks for them twice, in the spirit of Distribution::computeMean() and Distribution::getMean(), but I agree with you: it is overkill here.

mbaudin47:
Adding the ` ProcessSample::getField` Python interface is a great thing: this will simplify many of my scripts.

May you elaborate a bit more on the `SimulatedAnnealingLHS` choice of using phip by default? I may approve the phip on principle because mindist does not generalize well in several dimensions: lower dimension projections may be of poor quality. Does this general principle match your practical experience?

jschueller:
@mbaudin47 no, it was mainly that the code replaced mindist by phip, so I just removed the replacement and set phip by default, maybe your explanation was why we put the replace in the first place, I just made the code simpler now, and without a warning

lib/src/Uncertainty/Algorithm/MetaModel/MetaModelAlgorithm.cxx
```diff
@@ -109,4 +115,75 @@ void MetaModelAlgorithm::load(Advocate & adv)
...
+  // Recover the distribution, taking into account that we look for performance
+  // so we avoid to rebuild expensive distributions as much as possible
+  const UnsignedInteger inputDimension = inputSample.getDimension();
+  Collection< Distribution > marginals(inputDimension);
+
+  // The strategy for the marginals is to find the best continuous 1-d parametric model
+  Collection< DistributionFactory > factories(DistributionFactory::GetContinuousUniVariateFactories());
+
+  // exclude non-parametric or costly ones
+  const Description excluded = {"HistogramFactory", "TrapezoidalFactory", "FrechetFactory", "MeixnerDistributionFactory", "TruncatedNormalFactory"};
```
mbaudin47, line 130: (I fear that my comment gets lost, so I post it twice).
Using the BIC as a default corresponds exactly to one of my wishes. However, this particular code with the "excluded" variable does not correspond to what I would suggest. I understand that we want to exclude the Histogram (because it always fit very well) and the Trapezoidal, Frechet, Meixner and Truncated, because fitting them costs a lot.
* Since the BuildDistribution searches for a Distribution which fits the sample, it might be clearer if the list of factories is an explicit input argument of BuildDistribution. This would clarify that this list is used to select the distribution which best matches the sample. Adding a list of copula might also be necessary.
* I would rather create a new GetContinuousPolynomialChaosUniVariateFactories method in order to include, rather than exclude a hard-coded list. This would allow to motivate the list of distribution that we include within the help of this particular method. Moreover, adding a new exotic distribution in the lib would not force to include it, while the current code would force to exclude it (if we do not simply forget this particular point).
* If the code is unchanged in its structure, the minimum would be to add the FisherSnedecor, because, according to #1061, this costs a lot. Moreover, I would create a new entry in the ResourceMap so that we can customize the list.

mbaudin47 (changes requested):
I think that improving the changes in BuildDistribution may greatly improve this particular change. Otherwise, all this seems excellent to me.

jschueller:
maybe it's time to merge this

regislebrun (approved):
I found an additional bug which may be fixed here: KarhunenLoeveResult::lift and KarhunenLoeve::liftAsSample don't check the dimension of the input point, resulting into garbage when called with points of dimension smaller than expected. There is no problem with the project method as the check is done by the functions.

jschueller:
@regislebrun done

regislebrun (approved):
Nice, ready to go for me!

sofianehaddad (approved):

Issue #1594, opened 2020-09-21
LinearLeastSquaresCalibration and CalibrationResult


mbaudin47:
* In the current code, the `LinearLeastSquaresCalibration` prior is a Dirac, which is wrong. The correct prior is a non informative prior which can be emulated using a widespread Normal distribution. We may use the candidate as the mean of the Normal, and a fraction of MaxScalar as the standard deviation. The result Normal is (almost) flat.
* The CalibrationResult.drawParameterDistributions() will then draw a very flat Normal prior with a very large X range. 

![image](https://user-images.githubusercontent.com/31351465/93785371-4edca680-fc2e-11ea-9cee-c12e7dc59caa.png)

The code which must be adapted is:

https://github.com/openturns/openturns/blob/6cde44895358ade4033a74de0453f8131fe030ac/lib/src/Uncertainty/Bayesian/CalibrationResult.cxx#L225

It should be adapted in order to make the posterior distribution still visible. The mean flat prior can be represented with a star in on the X axis.

PR #1598, opened 2020-09-23
Fixed wrong prior distribution in LinearLeastSquaresCalibration and NonLinearLeastSquaresCalibration

regislebrun:
In the case of the least squares calibration algorithms, the prior is noninformative, it means it is equal to the Lebesgue measure instead of a Dirac distribution. This measure is emulated using a Normal distribution with a variance equal to SpecFunc.MaxScalar. Fixed issue #1594.

When the prior and posterior distributions are very different, the graphs produced by CalibrationResult are almost useless, so the relevant methods have been modified to ensure that both the posterior distribution and the candidate are always visible, even if it requires to truncate the graph of the prior pdf. The fix is not perfect as the graph remains difficult to read when the candidate is far away from the posterior distribution, but I can't see a way to fix it if both the candidate and the posterior distribution have to be on the same graph...


ChangeLog
```diff
@@ -59,7 +59,7 @@
  * #1564 (Set a Point makes ot crash)
  * #1567 (The API doc of SobolIndicesExperiment has a format issue)
  * #1573 (LinearModelAnalysis::drawQQPlot line is not the first bisector)
-
+ * #1594 (LinearLeastSquaresCalibration and CalibrationResult)
```
jschueller, line 62: you removed a blank line here, it can make merging difficult

jschueller (approved):

sofianehaddad (approved):

Issue #1547, opened 2020-07-03
Mesh does not check the simplices indices


mbaudin47:
Consider the following script where I create a 1D mesh with given vertices and simplices.

```
import openturns as ot
vertices = [[2.1], [2.8], [3.5], [4.2], [4.9], [5.6], [6.3], [7.0]]
simplices = [[3, 4], [4, 5], [5, 6], [6, 7], [7, 8], [8, 9], [9, 10]]
mesh = ot.Mesh(vertices, simplices)
print(mesh.computeWeights())
```

The simplices indices are wrong: they should begin at index 0 instead of 3. Because of this, the last vertex index in the last simplex is 10, which does not exists. Hence, an exception stating that a wrong vertex index should be raised, but it is not, which is a bug.

jschueller:
see isValid

mbaudin47:
This generates a memory crash, because invalid indices are adressed.

jschueller:
so I closed it too quickly

jschueller:
isvalid is potentially costly, at creation or load

PR #1602, opened 2020-09-30
Some fixes

### LINKED ISSUE #1590: The equation of the Fejer quadrature rule is triplicated
The equation of the 3 quadrature rules: 

http://openturns.github.io/openturns/master/user_manual/_generated/openturns.FejerAlgorithm.html

but the equations are the same in the three cases:

image

sofianehaddad:
Fix #1590 and #1547


lib/src/Base/Geom/openturns/Mesh.hxx
```diff
@@ -52,7 +52,8 @@ public:
 
   /** Parameters constructor */
   Mesh(const Sample & vertices,
-       const IndicesCollection & simplices);
+       const IndicesCollection & simplices,
+       const Bool checkMeshValidity = ResourceMap::GetAsBool("Mesh-CheckValidity"));
```
jschueller, line 56: I suggest to leave the attribute out of the ctor and keep separate accessors,
(we have ctors with several bools and its confusing)

lib/src/Base/Geom/Mesh.cxx
```diff
@@ -65,14 +65,17 @@ Mesh::Mesh(const Sample & vertices)
...
+           const IndicesCollection & simplices,
+           const Bool checkMeshValidity)
   : PersistentObject()
   , dimension_(vertices.getDimension())
   , vertices_(0, vertices.getDimension())
   , simplices_(simplices)
 {
   // Use the vertices accessor to initialize the kd-tree
   setVertices(vertices);
+  if (checkMeshValidity)
```
jschueller, line 77: maybe we could enable it by default for small meshes if < Mesh-LargeSize

lib/src/Base/Geom/openturns/Mesh.hxx
sofianehaddad: ah

lib/src/Base/Geom/Mesh.cxx
sofianehaddad: Indeed. I had this in mind

lib/src/Base/Geom/openturns/Mesh.hxx
sofianehaddad: what do you suggest? Having a key but without introducing it in ctor?

lib/src/Base/Geom/Mesh.cxx
sofianehaddad: @jschueller something like that?

lib/src/Base/Geom/openturns/Mesh.hxx
jschueller: yes, an attribute that would have an accessor, or just using the key

lib/src/Base/Geom/Mesh.cxx
jschueller: yes

lib/src/Base/Geom/openturns/Mesh.hxx
josephmure: Couldn't we do things this way:

- Always use ``ResourceMap::GetAsBool("Mesh-CheckValidity"))`` (so we don't provide the option to change the behavior in the ctor, which is @jschueller's concern).
- Set the default "Mesh-CheckValidity" to `True`.
- If a user thinks mesh checking takes too long, that user can simply set the ResourceMap key  "Mesh-CheckValidity" to `False`. Then it becomes the user's responsibility to call `isValid` whenever appropriate.

regislebrun:
The main problem is what we intend to check in the validity test. In its current form it is relatively fast, but if we want to add more checks such as "do some simplices overlap?" it will become much more costly. With this in mind, IMO the test should be triggered explicitly by the user. There are different places where the meshes are correct *by construction*, eg in LeveSetMesher or IntervalMesher, and we would have to change the value of the ResourceMap before and after the creation of the mesh in order to avoid this costly test. Another problem is with the setters: do we have to check the mesh after each vertices/simplicies modifications? Sometime a sequence of modifications starts from a valid mesh to end with a valid mesh, but the intermediate meshes are not valid (eg a vertex by vertex modification).

lib/src/Base/Geom/openturns/Mesh.hxx
sofianehaddad: Finally I think I will keep the parameter in ctor.
Now we do the check if asked only once and store this into an internal attribute. Reason is that several methods require explicitly the check (`draw` for example).

lib/src/Base/Geom/Mesh.cxx
sofianehaddad: Finally I dropped the use of several RM keys for that purpose and rely only on the specific `checkMeshValidity ` one

lib/src/Base/Geom/openturns/Mesh.hxx
regislebrun: Ok for me

regislebrun (approved):

jschueller:
+1, but you have to drop the changes on estimate_integral_Fejer.ipynb until Kieran's work is merged

josephmure (changes requested):
I agree with your solution @sofianehaddad  I have 2 alternative suggestions for `Mesh.doc.i.in`: one simply fixes a typo, and the other changes the formulation, so you can pick the one you like best.

sofianehaddad:
> +1, but you have to hold on the changes on estimate_integral_Fejer.ipynb until Kieran's work is merged

ok.

josephmure (approved):
+1, great job!

python/src/Sample.i
```diff
@@ -786,6 +786,26 @@ Sample(PyObject * pyObj)
   return new OT::Sample( OT::convert< OT::_PySequence_, OT::Sample>(pyObj) );  
 }
 
+static Sample BuildFromSequence(PyObject * pyObj, const OT::UnsignedInteger size, const OT::UnsignedInteger dimension)
+{
```
jschueller, line 790: -1
when d=1 (like I suppose this will be in most cases) simply writes `ot.Sample([[x] for x in L])` !
even when d>1 this will be `ot.Sample([L[i:i+d] for i in range(0,len(L),d)])` which is not uber difficult
so imho no need for this function when the solution is built-in the language
worst case a simple example would be enough

sofianehaddad: I move this (for discussion) into a new PR in order to not freeze this one

PR #1606, opened 2020-10-02
Various enhancements

jschueller:
- #1603
- #1605
- #1617 
- Another subset fix wrt the noise the pf estimation noise
- michaels KernelSmoothing cutoff test
- Fix in Normal: computeCDFGradient was not implemented for dim>1, computePDFGradient missed correlation parameters
- Fix in Mixture: computePDFGradient & computeCDFGradient were bogus
- Add ProcessSample::draw with GridLayout
- Add ProcessSample::drawCorrelation to plot correlation beween marginals
- Initialize glog for Ceres, else messages leak to stderr
- API fix for upcoming TBB 2021.1

PR #1613, opened 2020-10-13
Update GeneralizedExtremeValue_doc.i.in

adutfoy:
Error in the parameters transformation from a Frechet to a GEV in the mu parameter: 
mu = gamma + beta and not gamma=mu + beta.....


jschueller:
moved to #1606

Issue #1616, opened 2020-10-13
The Search engine does not work


mbaudin47:
I use the URL:

https://openturns.github.io/openturns/master/search.html

and i search for "symbolicfunction":

![image](https://user-images.githubusercontent.com/31351465/95902864-ae455680-0d95-11eb-8e6f-879ef21af3d5.png)

When I click on "Search", nothing appears:

![image](https://user-images.githubusercontent.com/31351465/95903003-dcc33180-0d95-11eb-9e46-86c720f6e6f3.png)

jschueller:
_static dir is located into python/doc/usecases instead python/doc

jschueller:
it still fails, seems some js files are misplaced now

KieranDelamotte:
@jschueller 
What do you mean by misplaced ?

jschueller:
I thought js files should go into /js 
https://github.com/openturns/openturns.github.io/tree/master/openturns/master/_static

jschueller:
seems to work after all, maybe not everything was published, sorry

PR #1618, opened 2020-10-14
Move the _static dir

KieranDelamotte:
Fixing the issue with the search box.


jschueller (approved):
+1
could you mention the issue in the changelog, and in the commit body: "Closes #1616"

jschueller:
I think it must be in the commit body instead of the title: "Closes #1616"

KieranDelamotte:
Ok I change it again with

Move the _static directory

The _static directory is moved to solve the issue #1616 with the search box.


Is it good for you ?

jschueller:
It must be explicitely "Closes #1616" so that the issue is closed automatically

josephmure (approved):
Good that you've fixed the search box error!

PR #1622, opened 2020-10-15
Update ParametricFunction_doc.i.in

adutfoy:
Hello
The previous example was not clear ... I think this one is more usefull!

Anne


jschueller:
+1, but since it fails the tests I'm moving it to my branch

PR #1623, opened 2020-10-16
Update Ev3 license

josephmure:
Ev3 has been released under the LGPL by Leo Liberti, its original creator: http://www.lix.polytechnique.fr/~liberti/Ev3-1.0.zip

This PR updates the README with a proper acknowledgement and modifies the headers of all relevant files, like in our standalone Ev3 repo: https://github.com/openturns/ev3/pull/3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -640,7 +640,6 @@ install (FILES COPYING
                 COPYING.LESSER
                 COPYING.cobyla
                 COPYING.dsfmt
-                COPYING.ev3
                 COPYING.exprtk
                 COPYING.faddeeva
                 COPYING.hmat
--- a/COPYING.ev3
+++ b/COPYING.ev3
@@ -1,7 +1,0 @@
-Extracted form the source files:
-** Author:    Leo Liberti
-...
-** Source:    GNU C++
-...
-** License:    (C) Leo Liberti, all rights reserved. Code published under the
-Common Public License.
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,7 @@ Up-to-date information can be found at http://www.openturns.org
 License
 =======
 OpenTURNS is distributed under the Lesser General Public License.
-Please see the LICENSE and the COPYING* files for details of the license of each components.
+Please see the LICENSE and the COPYING* files for details of the license of each component.
 
 Release Notes
 =============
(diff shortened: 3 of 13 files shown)


regislebrun (changes requested):
I don't understand why Ev3 has a different treatment than the other components. It should have its own COPYING file, even if it refers to COPYING.LESSER, and if you add an acknowledgement why do you limit its content to Ev3?
Otherwise well done: it's nice that you convinced Ev3 author to change the license of his library.

josephmure:
I did not need to "convince" him: he readily agreed. All he asked was for his work to be acknowledged in the README file.

I think the COPYING.ev3 file is no longer necessary, since Ev3 and OpenTURNS now share the same license. The special COPYING files are necessary for parts of the code covered by a secondary license which adds obligations not mentioned in the LGPL. For example, the license of the TNC algorithm demands that we display a specific block of text. The COPYING.tnc file is our way of complying with this obligation.

In other words, COPYING files are for lawyers, and the README file is for human beings ;)

josephmure:
I think that to comply with the LGPL version 3, we also need to state the date of the modifications. Does someone remember?

From Section 5 of the GPL version 3 (`COPYING`):

```
You may convey a work based on the Program, or the modifications to produce it from the Program, in the form of source code under the terms of section 4, provided that you also meet all of these conditions:

    a) The work must carry prominent notices stating that you modified it, and giving a relevant date.
    b) The work must carry prominent notices stating that it is released under this License and any conditions added under section 7. This requirement modifies the requirement in section 4 to “keep intact all notices”.
    c) You must license the entire work, as a whole, under this License to anyone who comes into possession of a copy. This License will therefore apply, along with any applicable section 7 additional terms, to the whole of the work, and all its parts, regardless of how they are packaged. This License gives no permission to license the work in any other way, but it does not invalidate such permission if you have separately received it.
    d) If the work has interactive user interfaces, each must display Appropriate Legal Notices; however, if the Program has interactive interfaces that do not display Appropriate Legal Notices, your work need not make them do so.
```

In this case, "the Program" is the original Ev3 library. Only a) requires work:

- b) the conditions added under Section 7 are the conditions of the LGPL version 3: OK
- c) we are licensing the whole of OpenTURNS under the LGPL: OK
- d) we have no interactive user interfaces, so there is nothing to be done: OK

jschueller:
yes, its weird to mention only ev3 here, maybe an array in the doc along in the same style that we're listing external dependencies, we could have a separate list for internal codes ?

jschueller:
thanks, I've cherry-picked this into my release branch

PR #1634, opened 2020-10-22
KrigingResult: Drop inputTransformedSample_

jschueller:
Probaby a leftover from the kriging transformation


josephmure (approved):
Good catch! Are you going to apply this patch to 1.16 as well?

regislebrun (approved):

sofianehaddad:
Indeed ... Thanks for the fix

PR #1638, opened 2020-10-26
Sample: Deprecate computeStandardDevPerComponent

jschueller:
And now computeStandardDeviation returns a Point


lib/src/Base/Stat/SampleImplementation.cxx
```diff
@@ -1232,10 +1232,12 @@ CovarianceMatrix SampleImplementation::computeCovariance() const
 /*
  * Gives the standard deviation of the sample, i.e. the square-root of the covariance matrix.
```
josephmure, line 1233:
```suggestion
 * Gives the standard deviation of each component of the sample.
```

python/doc/examples/data_analysis/manage_data_and_samples/plot_sample_manipulation.py
```diff
@@ -27,10 +27,6 @@
 # Get the mean per component
 sample.computeMean()
 
-# %%
-# Get the standard deviation per component
-sample.computeStandardDeviationPerComponent()
```
josephmure, line 32: I  think instead of deleting this, you could change it to
`sample.computeStandardDeviation()`

Of course, this means that the identical paragraph at lines 54-56 must be deleted:
```
# %%
# Get the sample standard deviation
sample.computeStandardDeviation()
```

josephmure (approved):
Just fix the 2 details mentioned below and this will be good to go IMO.

python/doc/examples/data_analysis/manage_data_and_samples/plot_sample_manipulation.py
jschueller: the result is the same isnt it ?

lib/src/Base/Stat/SampleImplementation.cxx
jschueller: ok

python/doc/examples/data_analysis/manage_data_and_samples/plot_sample_manipulation.py
josephmure: Yes, but is seems more natural to have it at this spot in the example rather than at the bottom of the file. By deleting this paragraph, you are moving the "easy" standard deviation computation among more complicated methods, like the computation of the covariance matrix. What do you think @jschueller?

jschueller: ok

jschueller:
moving to #1677

PR #1639, opened 2020-10-27
Implement Frechet computeLogPDFGradient [1.16]

josephmure:
This fixes a bug with `FrechetFactory::buildAsFrechet`.

Consider the following script. We try to fit a Frechet distribution on a sample obtained from a mixture of a Frechet and a Gumbel distribution. The optimization currently fails because the gradient of the logPDF is not implemented. Implementing it fixes the issue.

```
myFrechet = ot.Frechet(1.0, 1.0, 0.0)
myGumbel = ot.Gumbel(1.0,3.0)
# We build our mixture sample of size 2*1000=2000.
mixtureSample = ot.Sample()
sampleFrechet = myFrechet.getSample(1000)
sampleGumbel = myGumbel.getSample(1000)
mixtureSample.add(sampleFrechet)
mixtureSample.add(sampleGumbel)
# Build on the mixture sample
typedEstimatedFromMixtureSample = factory.buildAsFrechet(mixtureSample)
print('Estimated dist from mixture sample=', typedEstimatedFromMixtureSample)

print('Average logPDF of mixture sample=', typedEstimatedFromMixtureSample.computeLogPDF(mixtureSample).computeMean()[0])
```

Current output:
```
Estimated dist from mixture sample= Frechet(beta = 1.38232, alpha = 1.1757, gamma = 0.0926263)
Average logPDF of mixture sample= -93.89377428835414
```

Output after the fix:
```
Estimated dist from mixture sample= Frechet(beta = 2.57426, alpha = 1.71402, gamma = -0.833636)
Average logPDF of mixture sample= -2.283122349794321
```

### DIFF
--- a/lib/src/Uncertainty/Distribution/Frechet.cxx
+++ b/lib/src/Uncertainty/Distribution/Frechet.cxx
@@ -211,6 +211,21 @@ Point Frechet::computePDFGradient(const Point & point) const
   return pdfGradient;
 }
 
+/* Get the LogPDFGradient of the distribution */
+Point Frechet::computeLogPDFGradient(const Point & point) const
+{
+  if (point.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
+
+  const Scalar x = point[0] - gamma_;
+  Point logPdfGradient(3);
+  if (x <= 0.0) return logPdfGradient;
+  const Scalar logCdfplus1 = -std::expm1(-alpha_ * std::log(x / beta_));
+  logPdfGradient[0] = alpha_ / beta_ * logCdfplus1;
+  logPdfGradient[1] = 1.0 / alpha_ - std::log(x / beta_) * logCdfplus1;
+  logPdfGradient[2] = 1.0 / x * (1.0 + alpha_ * logCdfplus1);
+  return logPdfGradient;
+}
+
 /* Get the CDFGradient of the distribution */
 Point Frechet::computeCDFGradient(const Point & point) const
 {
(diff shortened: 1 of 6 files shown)


jschueller (approved):

lib/src/Uncertainty/Distribution/Frechet.cxx
```diff
@@ -211,6 +211,21 @@ Point Frechet::computePDFGradient(const Point & point) const
...
 
+/* Get the LogPDFGradient of the distribution */
+Point Frechet::computeLogPDFGradient(const Point & point) const
+{
+  if (point.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
+
+  const Scalar x = point[0] - gamma_;
+  Point logPdfGradient(3);
+  if (x <= 0.0) return logPdfGradient;
+  const Scalar logCdfplus1 = 1.0 - std::pow(x / beta_, -alpha_);
```
regislebrun, line 222: Here you should use:
```
const Scalar logCdfplus1 = -std::expm1(-alpha_ * std::log(x / beta_));
```
for an improved accuracy when x is close to beta. You can check it using this code:
```
int main(void)
{
  double x = 1.0+1.0/(1<<20); // exact in double precision
  double val1 = 1.0 - std::pow(x, -1.0);
  double val2 = -std::expm1(-1.0 * std::log(x));
  double ref = 9.53673406912415587982570664815e-7; // from Maple
  std::cout << std::setprecision(20) << "x=" << x
            << "\nval1=" << val1 << ", err=" << 1.0-val1/ref
            << "\nval2=" << val2 << ", err=" << 1.0-val2/ref
            << "\nref =" << ref << std::endl;
  return 0;
}
```
which gives:
```
x=1.0000009536743164062
val1=9.5367340691154822707e-07, err=9.0949470177292823792e-13
val2=9.5367340691241558881e-07, err=0
ref =9.5367340691241558881e-07
```

regislebrun (approved):
Good job, just a minor remark

lib/src/Uncertainty/Distribution/Frechet.cxx
josephmure: Done. Thanks!

PR #1641, opened 2020-10-28
UserDefinedStationaryCovarianceModel easy 1d ctor

josephmure:
I think most use cases of `UserDefinedStationaryCovarianceModel` are in settings with only one output dimension. In such cases, it seems a little heavy-handed to ask the user to create a collection of 1x1 covariance matrices.

This PR implements a new constructor for such cases: it takes a `ScalarCollection` instead of
a `CovarianceMatrixCollection` as input. As a result, the syntax of the example on the API doc page is simplified.

### DIFF
--- a/lib/src/Base/Stat/UserDefinedStationaryCovarianceModel.cxx
+++ b/lib/src/Base/Stat/UserDefinedStationaryCovarianceModel.cxx
@@ -53,8 +53,8 @@ UserDefinedStationaryCovarianceModel::UserDefinedStationaryCovarianceModel(const
 {
   const UnsignedInteger size = mesh.getVerticesNumber();
   if (size != covarianceFunction.getSize())
-    throw InvalidArgumentException(HERE) << "Error: for a non stationary covariance model, sizes are incoherents"
-                                         << " mesh size = " << size << "covariance function size = " << covarianceFunction.getSize();
+    throw InvalidArgumentException(HERE) << "Error: a covariance matrix should correspond to each distance specified in the regular grid. "
+                                         << "Regular grid size = " << size << " vs number of covariance matrices in the collection = " << covarianceFunction.getSize();
   inputDimension_ = mesh.getDimension();
   covarianceCollection_ = CovarianceMatrixCollection(size);
   // put the first element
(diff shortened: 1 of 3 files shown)


lib/src/Base/Stat/UserDefinedStationaryCovarianceModel.cxx
```diff
@@ -69,6 +69,29 @@ UserDefinedStationaryCovarianceModel::UserDefinedStationaryCovarianceModel(const
...
+  : StationaryCovarianceModel()
+  , covarianceCollection_(0)
+  , mesh_(mesh)
+  , nearestNeighbour_(mesh)
+{
+  const UnsignedInteger size = mesh.getVerticesNumber();
+  if (size != covarianceValues.getSize())
+    throw InvalidArgumentException(HERE) << "Error: a covariance value should correspond to each distance specified in the regular grid. "
+                                         << "Regular grid size = " << size << " vs number of covariance values in the collection = " << covarianceValues.getSize();
+  covarianceCollection_ = CovarianceMatrixCollection(size);
```
regislebrun, line 84: Replace it by:
```
covarianceCollection_ = CovarianceMatrixCollection(size, CovarianceMatrix(1));
 for (UnsignedInteger k = 0; k < size; ++k)
  {
    if (covarianceValues[k] < 0)
      throw InvalidArgumentException(HERE) << k << "-th covariance value is " << covarianceValues[k] << ", should be nonnegative.";
    covarianceCollection_[k](0, 0) = currentCovarianceMatrix;
  }
```

regislebrun (approved):

lib/src/Base/Stat/UserDefinedStationaryCovarianceModel.cxx
josephmure: I have adapted your code (`currentCovarianceMatrix` is not needed).

josephmure:
I have also made a few changes to the API doc page (see commit message).
UserDefinedStationaryCovarianceModel


There is something that confuses me, though: when we use the constructor with the collection of covariance matrices, why do we need them to *be* covariance matrices? When the output dimension is 1, we are simply giving the covariances between the GP value at time 0 and its value at other times: this should not be a collection of 1x1 covariance matrices since some of them could be negative (see `ExponentiallyDampedCosineModel` for example). Am I missing something?

regislebrun (approved):

josephmure:
This easy constructor is actually obsolete because of @jschueller's [StationaryFunctionalCovarianceModel](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.StationaryFunctionalCovarianceModel.html). The doc fix was moved to PR #1664.

PR #1648, opened 2020-10-30
Correct the formula of MCMC::computeLogLikelihood [1.16]

### LINKED ISSUE #1642: Big white space at the beginning of examples
Here an example of the phenomenon: https://openturns.github.io/openturns/latest/auto_graphs/plot_graphs_basics.html?highlight=composeddistribution

The analysis of the html code shows that a css file "_Static/Basic.css" on the server is interfering.
More precisely, the line 697 in Basic.css "clear: both;" that should be replaced by "clear: left;"

My understanding is that "Basic.css" is added on the server side during the build of the doc. The phenomenon is not observed on local installation.
@jschueller : can you try to change this line at the source to see if it corrects the bug?

### LINKED ISSUE #1647: MCMC::computeLogLikelihood does not compute the log-likelihood
`MCMC::computeLogLikelihood` and its derivative `RandomWalkMetropolisHastings::computeLogLikelihood` do not actually compute the log-likelihood but the sum of the log-likelihood and of the logarithm of the prior density.

The latter quantity is, up to an additive constant, the logarithm of the posterior density. It is the target density of the MCMC algorithm, so we cannot simply correct the formula of the `computeLogLikelihood` method.

I will propose a fix for inclusion in 1.16. This ticket is for reference only.

josephmure:
- ~~Turn the old `computeLogLikelihood` public method
into a `computeLogLikelihoodPlusLogPrior` protected method.~~ Compute the sum of loglikelihood and logprior every time the class needs it.
- ~~Create a new~~ Make sure the public `computeLogLikelihood` method ~~that~~ actually
computes the log-likelihood. Fixes #1647.
- Deprecate `currentLogLikelihood_` attribute and replace it with
`currentLogLikelihoodPlusLogPrior_`.
- Update ChangeLog and TODO.

~~The method  `computeLogLikelihoodPlusLogPrior` is necessary, otherwise the class cannot work. Since 1.16rc1 is already out, however, I have chosen not to expose it. That way, if we decide to expose it in 1.17, we can change the name of the method without deprecating anything.~~

From the point of view of a user, nothing changes with this PR except that `computeLogLikelihood` now acts as advertised.

EDIT : added a commit to fix #1642 i.e. the big whitespace bug on doc pages.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -99,6 +99,7 @@
  * #1605 (MaximumLikelihoodFactory cannot be used with FittingTest.Kolmogorov)
  * #1624 (The graphs_loglikelihood_contour example has a bug)
  * #1643 (Problem in MaximumDistribution PDF)
+ * #1647 (MCMC::computeLogLikelihood does not compute the log-likelihood)
  * #1654 (StationaryFunctionalCovarianceModel::discretize segfaults)
  * #1660 (Cannot extract continuous modes from KLResult when dimension>1)
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -98,6 +98,7 @@
  * #1603 (FieldToPointConnection generates an invalid exception)
  * #1605 (MaximumLikelihoodFactory cannot be used with FittingTest.Kolmogorov)
  * #1624 (The graphs_loglikelihood_contour example has a bug)
+ * #1642 (Big white space at the beginning of examples)
  * #1643 (Problem in MaximumDistribution PDF)
  * #1647 (MCMC::computeLogLikelihood does not compute the log-likelihood)
  * #1654 (StationaryFunctionalCovarianceModel::discretize segfaults)
--- a/lib/src/Uncertainty/Bayesian/RandomWalkMetropolisHastings.cxx
+++ b/lib/src/Uncertainty/Bayesian/RandomWalkMetropolisHastings.cxx
@@ -42,7 +42,7 @@ RandomWalkMetropolisHastings::RandomWalkMetropolisHastings()
   , calibrationStrategy_(0)
   , samplesNumber_(0)
   , acceptedNumber_(0)
-  , currentLogLikelihood_(0.)
+  , currentPenalizedLogLikelihood_(0.)
 {
   // Nthing to do
 }
@@ -58,7 +58,7 @@ RandomWalkMetropolisHastings::RandomWalkMetropolisHastings( const Distribution &
   , calibrationStrategy_(proposal.getSize())
   , samplesNumber_(0)
   , acceptedNumber_(initialState.getDimension())
-  , currentLogLikelihood_(0.)
+  , currentPenalizedLogLikelihood_(0.)
 {
   setProposal(proposal);
 }
@@ -76,7 +76,7 @@ RandomWalkMetropolisHastings::RandomWalkMetropolisHastings( const Distribution &
   , calibrationStrategy_(proposal.getSize())
   , samplesNumber_(0)
   , acceptedNumber_(initialState.getDimension())
-  , currentLogLikelihood_(0.)
+  , currentPenalizedLogLikelihood_(0.)
 {
   setProposal(proposal);
 }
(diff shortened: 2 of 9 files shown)


jschueller:
this method might as well just be moved to private, can you think of a use-case ?
If not I dont mind having an extra term but with a readable name.

josephmure:
`computeLogLikelihoodPlusLogPrior` is now a private method.

josephmure:
@jschueller Making `MCMC:computeLogLikelihoodPlusLogPrior` private makes compilation fail. I guess that is because `RandomWalkMetropolisHastings::getRealization()` calls `MCMC:computeLogLikelihoodPlusLogPrior`. Is there a way to get `RandomWalkMetropolisHastings` to access it despite its "private" status or should I make it "protected" again?

jschueller:
I mostly meant these methods are useless outside the class so I wonder if its even useful to have one public, but yes protected is the way to go for inherited classes.

josephmure:
I think `computeLogLikelihood` is useful as a public method. `computeLogLikelihoodPlusLogPrior` not so much, because it is easily computed.

jschueller:
but why is it useful ?

josephmure:
As a diagnostic tool, or for plotting purposes. You may want to plot the log-likelihood as a function of a specific parameter when all others are fixed, for example.

MerlinKeller:
As a matter of fact, the log-likelihood function is extremely useful in the context of MCMC, and most MCMC packages provide this service. The most popular use is to compute selection model criteria, such as the deviance information criterion (DIC), or the Watanabe Akaike information criterion (wAIC) to cite only a few. But it can have oser uses, such as if you want to re-cycle your MCMC output as as a proposal sample for posterior-guided Importance sampling. Hence I strongly recommend to keep it!

jschueller:
Isn't it simpler to just move the prior term outside the function ?

josephmure:
I don't understand what you mean @jschueller. This is exactly what this PR does: it removes the prior term from `computeLogLikelihood`.

jschueller:
we could do it without renaming anything

josephmure:
@jschueller Do you like this version better?

lib/src/Uncertainty/Bayesian/RandomWalkMetropolisHastings.cxx
```diff
@@ -284,7 +284,13 @@ void RandomWalkMetropolisHastings::load(Advocate & adv)
   adv.loadAttribute("calibrationStrategy_", calibrationStrategy_);
   adv.loadAttribute("samplesNumber_", samplesNumber_);
   adv.loadAttribute("acceptedNumber_", acceptedNumber_);
-  adv.loadAttribute("currentLogLikelihood_", currentLogLikelihood_);
+  if (adv.hasAttribute("currentLogLikelihood_"))
+  {
+    adv.loadAttribute("currentLogLikelihood_", currentLogLikelihoodPlusLogPrior_);
+    LOGWARN(OSS() << "The internals of the class RandomWalkMetropolisHastings have changed in version 1.16. Please save this object again to ensure compatibility with future versions.");
```
jschueller, line 290: please drop this warning, it should be silent

jschueller:
I propose penalizedLogLikelihood instead of logLikelihoodPlusLogPrior, as mentioned in the test

lib/src/Uncertainty/Bayesian/RandomWalkMetropolisHastings.cxx
josephmure: Done.

jschueller (approved):

Jean-A:
This last commit made by Joseph (custom CSS) allows to correct issue #1642 
The new css rule overwrite the bad one defined in basic.css

jschueller:
@regislebrun @sofianehaddad this should go to 1.16, please review

ChangeLog
```diff
@@ -98,9 +98,10 @@
  * #1603 (FieldToPointConnection generates an invalid exception)
  * #1605 (MaximumLikelihoodFactory cannot be used with FittingTest.Kolmogorov)
  * #1624 (The graphs_loglikelihood_contour example has a bug)
+ * #1642 (Big white space at the beginning of examples)
+ * #1647 (MCMC::computeLogLikelihood does not compute the log-likelihood)
  * #1654 (StationaryFunctionalCovarianceModel::discretize segfaults)
 
-
```
jschueller, line 103: you're dropping a line here

regislebrun (approved):

PR #1653, opened 2020-11-03
Fix trapezoidal factory

### LINKED ISSUE #1651: Cobyla freezes in 0T1.16rc1
See https://stackoverflow.com/questions/64605271/meta-model-of-the-field-function-openturns-1-16rc1

regislebrun:
When cobyla find a point with NaNs it freezes after a few iterations. Fixes #1651.

I also adapted the verbosity level of the optimization algorithm to the verbosity level of Log in some factories.

### DIFF
--- a/lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
@@ -227,7 +227,16 @@ Sample EmpiricalBernsteinCopula::getSample(const UnsignedInteger size) const
   return sample;
 }
 
-/* Get the PDF of the EmpiricalBernsteinCopula */
+/* Get the PDF of the EmpiricalBernsteinCopula
+   The empirical Bernstein copula is a mixture of products of Beta distributions:
+   c_m(u_1,...,u_d)=1/n\sum_{i=1}^n\prod_{j=1}^d \beta_{r_j^i,s_j^i}(u_j)
+   where r_j^i=\lceil mU_j^i\rceil, s_j^i=m+1-r_j^i
+   and \beta_{r,s}(x)=x^{r-1}(1-x)^{s-1}/B(r,s)
+   Here \log(r_j^i) is stored into logFactors_(i,j)=LF(i,j),
+   B(r_j^i,s_j^i) is stored into logBetaMarginalFactors_(i,j)=LBMF(i,j) and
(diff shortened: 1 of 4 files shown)


lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
```diff
@@ -335,6 +344,210 @@ Scalar EmpiricalBernsteinCopula::computeProbability(const Interval & interval) c
...
+  const UnsignedInteger conditioningDimension = y.getDimension();
+  if (conditioningDimension >= getDimension()) throw InvalidArgumentException(HERE) << "Error: cannot compute a conditional PDF with a conditioning point of dimension greater or equal to the distribution dimension.";
+  if (x <= 0.0 || x >= 1.0) return 0.0;
+  for (UnsignedInteger i = 0; i < y.getDimension(); ++i)
+    if (y[i] <= 0.0 || y[i] >= 1.0) return 0.0;
+  const UnsignedInteger size = copulaSample_.getSize();
+  // Special case for no conditioning or independent copula
+  if ((conditioningDimension == 0) || (hasIndependentCopula()))
+    {
+      if (isCopula()) return 1.0;
```
josephmure, line 374: I do not understand what it would mean for an `EmpiricalBernsteinCopula` not to be a copula.

josephmure:
Also, I am not sure what the `EmpiricalBernsteinCopula` has to do with the `TrapezoidalFactory`.

lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
regislebrun: This commit should never have landed here. I messed things yesterday. Concerning your question, a better place would be to discuss it on PR#1646. What is called a Bernstein copula is a Bernstein approximation of a copula function over a regular grid. When applied to the empirical copula (which is NOT a copula despite its name: the marginal distributions are not uniform over [0,1]), one get an absolutely continuous function which is an actual copula if and only if the marginal grid size is a proper divider of the sample size. So yes, an empirical Bernstein copula can be something else than a copula, and it is useful to have it e.g. for estimation purpose such as the MIIC algorithm in otagrum.

regislebrun:
This PR is superseded by PR #1657

lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
josephmure: All right thanks for your explanation.

PR #1664, opened 2020-11-09
Allow CovarianceModel operator() to yield an asymmetric SquareMatrix

### LINKED ISSUE #1662: Operator() of a CovarianceModel with multidimensional output should yield object of type SquareMatrix
... and not `CovarianceMatrix` as is currently the case:

https://github.com/openturns/openturns/blob/6de2a00118458512c4cbbf179648073c92a44d6d/lib/src/Base/Stat/openturns/TensorizedCovarianceModel.hxx#L60

There exist covariance kernels that can [produce asymmetric matrices](https://openturns.discourse.group/t/covariance-kernels-for-random-processes-with-multidimensional-output-do-not-output-symmetric-matrices/55).

josephmure:
Makes all `CovarianceModelImplementation` derivatives yield a `SquareMatrix` when `operator()` is called.
Fixes #1662.


regislebrun (approved):
Only a minor suggestion. Did you checked the several discretize() methods?

jschueller:
we cannot change the api in the release branch

regislebrun:
@josephmure @jschueller I agree on the fact that we cannot change the API in RC phase, but we have to check that the bug has no impact on e.g. KrigingAlgorithm, GeneralLinearModelAlgorithm and GaussianProcess. It resort to a careful reading of the many discretize() methods and derivatives (discretizeAndFactorize() aso). I started to sample randomly such methods and it looks right, but I can't do it alone.
The only way to get wrong results is when the model returns a symmetric matrix when it should not, and as most of the use of the covariance models is to feed the preceding algorithms, it means a wrong implementation of the discretization. IMO it should be checked before the release. I feel so ashamed to have let this bug grow during so much time! I was really convinced that the output should be a SPD matrix in any case :-(!

jschueller:
could we use joseph's branch, and, at each operator(), throw an exception if the squarematrix is symmetric, and run that on all the testsuite ?

sofianehaddad:
It seems impacted classes are `KrigingAlgorithm`, `GeneralLinearModelAlgorithm` or `GaussianProcess` when using `TensorizedCovarianceModel`

Probably we might fix some issues if we reimplement `TensorizedCovarianceModel::discretize`.

regislebrun:
Yep. Someone working on it or I give it a try? BTW, I suffered a lot when working with large meshes and such covariance models: it takes ages to discretize, a specific implementation is definitely needed for the most common case of 1D covariance models with no interactions. Is it acceptable to let such an implementation sneak into the bug-fix ;-)?

sofianehaddad:
After a quick reading of the code, with a focus on the link covariance models/gaussian processes, I have some remarks : 
 - For scalar models (`AbsoluteExponential`, `SquaredExponential`,...) current operator() has negligible/null impact. `discretize` & co methods seem accurate.
- For `ExponentialModel`, the operator() should return a `SymmetricMatrix` by its definition. However method provides a `CovarianceMatrix`  (which is symmetric) but the definite or semi-definite positive behaviour is not checked and has minor impact. Also `discretize` method is well defined as we rely on the `computeStandardRepresentative` method.
- Finally, for `TensorizedCovarianceModel`, we currently rely on a list of covariance models and suppose **there is no spaatial (output) correlation**. Thus if we have a look at the list of covariance models we got in the library, output result for operator() is a `SymmetricMatrix` --> `discretize` provides an accurate result.

In other words, even if operator() might return an asymmetric square matrix, we might fix this after the release as we already planned to make several changes in `CovarianceModel` API

josephmure:
@jschueller @sofianehaddad I can move this to the master branch if you want.

josephmure:
I know for sure it has an impact on `UserDefinedStationaryCovarianceModel`, because we should allow non-symmetric matrices in the collection.

sofianehaddad:
@josephmure I only mentioned impact on `GP`. `UserDefinedStationaryCovarianceModel` is not used/applied in that context

lib/src/Base/Stat/openturns/CovarianceModelImplementation.hxx
```diff
@@ -199,6 +197,9 @@ public:
   /** Method load() reloads the object from the StorageManager */
   void load(Advocate & adv) override;
 
+  /** outputCovariance_ accessor*/
+  CovarianceMatrix getOutputCovariance() const;
```
jschueller, line 201: this new method is not documented

lib/src/Base/Stat/UserDefinedStationaryCovarianceModel.cxx
```diff
@@ -24,8 +24,8 @@
 
 BEGIN_NAMESPACE_OPENTURNS
 
-// TEMPLATE_CLASSNAMEINIT(PersistentCollection< CovarianceMatrix >)
-// static const Factory<PersistentCollection<CovarianceMatrix> > Factory_PersistentCollection_CovarianceMatrix;
+// TEMPLATE_CLASSNAMEINIT(PersistentCollection< SquareMatrix >)
+// static const Factory<PersistentCollection<SquareMatrix> > Factory_PersistentCollection_SquareMatrix;
```
jschueller, line 28: you can remove those, it's already instanciated in ARMACoefficients

jschueller:
so if I understand correctly Sofiane pov, there's a problem with the TensorizedCM on it's own, but not through the Kriging as discretize is ok.

lib/src/Base/Stat/openturns/CovarianceModelImplementation.hxx
josephmure: Ah, sorry, I added this method to debug the changes and forgot to remove it. I will do it shortly.

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -154,23 +154,26 @@ UnsignedInteger CovarianceModelImplementation::getInputDimension() const
...
   const Scalar rho = computeStandardRepresentative(s, t);
   if (outputDimension_ == 1)
   {
-    CovarianceMatrix result(1);
+    SquareMatrix result(1);
     result(0, 0) = rho * outputCovariance_(0, 0);
     return result;
   }
-  return CovarianceMatrix((outputCovariance_ * rho).getImplementation());
+  // outputCovariance_ * rho is a SymmetricMatrix.
```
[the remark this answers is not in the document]
josephmure: Could you please check this code again? I think I have done what you suggest but I may have made a mistake.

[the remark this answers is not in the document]
josephmure: The discretize() methods produce CovarianceMatrices (as they should) so I don't think anything needs to be changed in their code.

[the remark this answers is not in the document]
regislebrun: @josephmure Looks good to me

jschueller:
a rebase is needed

josephmure:
Apparently, EGO failed when CircleCI build_linux ran [plot_ego.py](https://openturns.github.io/openturns/master/auto_numerical_methods/optimization/plot_ego.html). But theoretically, Gaussian processes should not be affected by the fix, right?

jschueller:
I think it's okay, maybe a one-time glitch.

@sofianehaddad I think we should merge this prior to your work on cov models, is it good to go ?

ChangeLog
```diff
@@ -65,7 +66,6 @@
  * Deprecated FittingTest::BestModelKolmogorov(Sample, DistributionFactoryCollection, TestResult), use BestModelLilliefors
  * Deprecated FittingTest::Kolmogorov(Sample, DistributionFactory, TestResult, level), use Lilliefors
  * MetamodelValidation: now computePredictivityFactor returns Point, drawValidation return GridLayout
- * Deprecated coupling_tools.execute is_shell/workdir/shell_exe/check_exit_code arguments
```
jschueller, line 68: Please restore that line

josephmure: Oops. Restored.

jschueller (approved):

sofianehaddad (approved):

jschueller:
looks like there's an issue with ego here after all

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingResult.cxx
```diff
@@ -541,7 +541,7 @@ CovarianceMatrix KrigingResult::getConditionalCovariance(const Point & point) co
   const Point data(point);
   // 1) compute \sigma_{x,x}
   LOGINFO("Compute interactions Sigma_xx");
-  const CovarianceMatrix sigmaXX(covarianceModel_(Point(inputDimension, 0.0)));
+  const SquareMatrix sigmaXX(covarianceModel_(Point(inputDimension, 0.0)));
```
sofianehaddad, line 544: Wrong here should be `CovarianceMatrix`

lib/src/Base/Stat/StationaryCovarianceModel.cxx
```diff
@@ -72,16 +72,19 @@ StationaryCovarianceModel * StationaryCovarianceModel::clone() const
...
 
-CovarianceMatrix StationaryCovarianceModel::operator() (const Point & tau) const
+SquareMatrix StationaryCovarianceModel::operator() (const Point & tau) const
 {
   const Scalar rho = computeStandardRepresentative(tau);
-  return CovarianceMatrix((outputCovariance_ * rho).getImplementation());
+  // outputCovariance_ * rho is a SymmetricMatrix.
+  // Its implementation only contains terms in the lower half.
+  // Conversion to a SquareMatrix is required.
+  return SquareMatrix(outputCovariance_ * rho);
```
sofianehaddad, line 87: maybe we need to specify the `symmetrize`

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingResult.cxx
jschueller: I merged this too quickly, how do we fix this ?

sofianehaddad: Yep.

josephmure: @jschueller  I have added a commit on my branch josephmure:covmodel_bugfix that applies Sofiane's corrections (this one and the one below). If you want, you can pick it in your branch.

jschueller: if you already got this in a branch, could you submit it ?

sofianehaddad:
@jschueller I have no trouble with `EGO`
How to reproduce the bug?

jschueller:
it happened twice in circleci, but locally it runs fine, maybe it was bad luck

sofianehaddad:
Ok that is why I can not reproduce

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingResult.cxx
josephmure: OK.

josephmure:
We will have to see whether this problem with EGO occurs more frequently now. I do not understand how it can be "bad luck", isn't the same seed always used by the random generator?
Anyway, we can continue this discussion on the new PR #1687.

sofianehaddad:
From my experience, using `LD_LBFGS` sometimes provides such an error (not only with `EGO`)
If we get again this kind of error, we should probably update the test

josephmure:
The issue is not with a test but with the example [plot_ego.py](https://openturns.github.io/openturns/master/auto_numerical_methods/optimization/plot_ego.html)

sofianehaddad:
@josephmure yep I know. That is why I explicitly mentioned the used `LD_LBFGS` algorithm 😉

sofianehaddad:
@jschueller did you update `nlopt` in `docker` ?

jschueller:
it should still be at 2.6.2 for linux

PR #1675, opened 2020-11-18
Update Gamma_doc.i.in

adutfoy:
Fixed variance formula


jschueller:
added to my misc branch, I cant merge it without a proper commit message

PR #1684, opened 2020-11-27
Remove IsoProbabilisticTransform in Kriging examples

josephmure:
It is possible to play with the bounds of the optimization algorithm instead.
This is more elementary in my opinion.

### DIFF
--- a/python/doc/examples/reliability_sensitivity/reliability/plot_subsetSampling.py
+++ b/python/doc/examples/reliability_sensitivity/reliability/plot_subsetSampling.py
@@ -32,7 +32,8 @@ Subset Sampling
 # First, import the python modules: 
 
 # %%
-from  openturns import *
+import openturns as ot
+from openturns.viewer import View
 
 # %%
 # Create the probabilistic model :math:`Y = g(X)`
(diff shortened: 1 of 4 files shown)


jschueller (approved):

python/doc/examples/reliability_sensitivity/reliability/plot_subsetSampling.py
```diff
@@ -220,7 +220,7 @@
 graph.setXTitle(r'$x_1$')
 graph.setYTitle(r'$x_2$')
 
-Show(graph)
+ot.Show(graph)
```
jschueller, line 223: could we use viewer instead ? that's what we promote in all other examples except this one (maybe we should consider deprecating it someday)

josephmure: Done. What does Show do differently than the viewer anyway?

jschueller: it just calls viewer :)

jschueller:
you must rebase for tests to pass

sofianehaddad:
IMHO it is not only a matter of optimisation bounds. In fact, the algorithm should set in practice the accurate bounds.
It will be interesting if, for a particular example or someones, you illustrate the difference in terms of precision/convergence between algo performed with/without transformation (basic affine transformation)

josephmure:
In the cantilever beam example (the one modified by this PR), the bounds set automatically define an empty set.
This happens because the lower bound is automatically [set to 0.01](https://github.com/openturns/openturns/blob/7e839ff97f10adbc6dc7a4ab2b4620b2fd5bdd2d/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx#L289) while the lower bound [depends on the design](https://github.com/openturns/openturns/blob/7e839ff97f10adbc6dc7a4ab2b4620b2fd5bdd2d/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx#L303). The moment of inertia of the cantilever beam example is of order 10^(-7), so the dynamically set upper bound is lower than 10^(-2).

I do not think these examples are well suited to accuracy comparisons between affine transformations and resetting of the optimization bounds because the Q2 is extremely high. After implementing the reset of the optimization bounds, I actually had to lower the size of the input sample from 20 to 10 to get a Q2 strictly lower than 1.

sofianehaddad (approved):
I see your point.

regislebrun (approved):

PR #1690, opened 2020-12-03
StationaryFunctionalCovarianceModel thread safety

jschueller:
Here the goal is to fix StationaryFunctionalCovarianceModel thread safety in CovarianceModel::discretize, so we need:
- Function::isParallel 
- CovarianceModel::isParallel
- Add TBB::ParallelForCondition to reuse the parallel path, but to disable the actual parallelization according to a boolean (eg the value of CovarianceModel::isParallel)

### DIFF
--- a/lib/src/Base/Algo/ClassifierImplementation.cxx
+++ b/lib/src/Base/Algo/ClassifierImplementation.cxx
@@ -92,29 +92,13 @@ struct ClassifyPolicy
 }; /* end struct ClassifyPolicy */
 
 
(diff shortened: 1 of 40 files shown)


sofianehaddad (approved):

sofianehaddad:
@jschueller any plan for integration into master ?

jschueller:
yes!

sofianehaddad:
Hey @jschueller what is the issue in `circle ci` with `cminpack?`

jschueller:
cminpack 1.3.7 adds cmake config files, so  we have to adapt

Issue #1681, opened 2020-11-24
LevelSetClassifier


jschueller:
we could extend MinimumVolumeClassifier and define a classifier defined on only the levelset
also it could be extended to multi-class with several quantile levels (reuse samples ?)

PR #1691, opened 2020-12-04
Misc enhancements

jschueller:
- MultiStart: use the same base budget for all starting points instead of a global budget
- MinimumVolumeClassifier: Generalize to n confidence levels (#1681)
- #1710
- #1713
- #1721
- #1729
- #1737 
- #1742
- fix for an issue in lapack exiting on nan: https://github.com/Reference-LAPACK/lapack/issues/469


regislebrun (changes requested):
Why do you want to drop Show? I find it very useful and use it quite extensively...

jschueller:
I dont want 2 apis to do the same thing, another issue is that its not documented.

regislebrun:
ok but IMO the difference between 
```
import openturns as ot
ot.Show(ot.Normal().drawPDF())
```
and
```
import openturns as ot
g = ot.Normal.drawPDF()
v = otv.View(g)
v.ShowAll()
```
is significant enough to keep Show() in the module. I don't know where to add the documentation but I am ready to add it if needed. But if you *really* want to remove it and if this choice is shared by others, it's ok for me.

jschueller:
> is significant enough to keep Show() in the module. I don't know where to add the documentation but I am ready to add it if needed. But if you _really_ want to remove it and if this choice is shared by others, it's ok for me.


point taken, I keep Show for now

lib/src/Uncertainty/Algorithm/Classification/MinimumVolumeClassifier.cxx
```diff
@@ -87,24 +100,34 @@ Indices MinimumVolumeClassifier::classify(const Sample & inS) const
...
-LevelSet MinimumVolumeClassifier::getLevelSet() const
+LevelSet MinimumVolumeClassifier::getLevelSet(const UnsignedInteger j) const
 {
   if (!isLevelSetComputed_)
   {
-    levelSet_ = distribution_.computeMinimumVolumeLevelSetWithThreshold(alpha_, threshold_);
+    levelSetCollection_.resize(alpha_.getSize());
+    threshold_.resize(alpha_.getSize());
+    for (UnsignedInteger i = 0; i < alpha_.getSize(); ++ i)
+      levelSetCollection_[i] = distribution_.computeMinimumVolumeLevelSetWithThreshold(alpha_[i], threshold_[i]);
```
regislebrun, line 110: The `computeMinimumVolumeLevelSetWithThreshold()` should be complemented by a `computeMinimumVolumeLevelSetCollectionWithThreshold()` version taking two points as arguments, in order to speed-up computation e.g. when the levels are computed by sampling. The default implementation can easily be adapted:
```
Collection<LevelSet> DistributionImplementation::computeMinimumVolumeLevelSetCollectionWithThreshold(const Point & prob,
    Point & threshold) const
{
  if (!isContinuous()) throw NotYetImplementedException(HERE) << "In DistributionImplementation::computeMinimumVolumeLevelSet()";
  // 1D special case here to avoid a double construction of minimumVolumeLevelSetFunction
  if ((dimension_ == 1) && (ResourceMap::GetAsBool("Distribution-MinimumVolumeLevelSetBySampling")))
  {
    LOGINFO("Compute the minimum volume level set by sampling (QMC)");
    return result(computeUnivariateMinimumVolumeLevelSetCollectionByQMC(prob, threshold));
  }
  Function minimumVolumeLevelSetFunction(MinimumVolumeLevelSetEvaluation(clone()).clone());
  minimumVolumeLevelSetFunction.setGradient(MinimumVolumeLevelSetGradient(clone()).clone());
  // If dimension_ == 1 the threshold can be computed analyticaly
  Point minusLogPDFThreshold;
  const UnsignedInteger probSize = prob.getSize();
  if (dimension_ == 1)
  {
    const CompositeDistribution composite(minimumVolumeLevelSetFunction, *this);
    minusLogPDFThreshold = composite.computeQuantile(prob).asPoint();
    LOGINFO("Compute the minimum volume level set by using a composite distribution quantile (univariate general case)");
  } // dimension == 1
  else
  {
    LOGINFO("Compute the minimum volume level set by sampling (Monte Carlo)");
    const UnsignedInteger size = ResourceMap::GetAsUnsignedInteger("Distribution-MinimumVolumeLevelSetSamplingSize");
    const Sample xSample(getSample(size));
    const Sample logPDFSample(computeLogPDF(xSample));
    minusLogPDFThreshold = -logPDFSample.computeQuantile(Point(probSize, 1.0) - prob).asPoint();
  } // dimension > 1
  threshold = Point(probSize);
  Collection<LevelSet> levelSets(probSize);
  for (UnsignedInteger i = 0; i < probSize; ++i)
  {
    threshold[i] = std::exp(-minusLogPDFThreshold[i]);
    levelSets[i] =  LevelSet(minimumVolumeLevelSetFunction, LessOrEqual(), minusLogPDFThreshold[i]);
  }
  return levelSets;
}
```
This way, the costly computeLogPDF() samples are reused for all the levels. Note that you have to implement a `computeUnivariateMinimumVolumeLevelSetCollectionByQMC()` method the same way the Monte Carlo version is computed. For the few distributions with a specific implementation of the `computeMinimumVolumeLevelSetWithThreshold()` method (EllipticalDistribution, CompositeDistribution, Frechet, GeneralizedPareto, IndependentCopula, InverseNormal, ParametrizedDistribution, Pareto, RandomMixture, TruncatedNormal, Uniform) some additional work is needed, but not so much.

regislebrun (changes requested):
IMO the changes I propose are not so complex and would provide a huge speed-up of the computation when many levels are considered.

lib/src/Uncertainty/Algorithm/Classification/MinimumVolumeClassifier.cxx
jschueller: I propose to postpone this change to another pr if that's ok for you, I've created #1734 for this

jschueller:
@sofianehaddad @regislebrun good to go ?

jschueller:
@sofianehaddad @regislebrun can we merge this ?

regislebrun (approved):

josephmure:
@jschueller I am a little late for this PR but I just wanted to thank you for commit 699b32d in particular: it is going to make CI much easier for me now that build_linux no longer automatically fails at the "Publish documentation" step.

PR #1702, opened 2020-12-17
Improved ProcessSample

regislebrun:
Now it has all the methods of Sample returning Point or Scalar. It is only a proof of concept, as much more efficient implementations exist, but it illustrates the idea I exposed during our meeting on KL decomposition.


mbaudin47:
So the key is to implement the `getSampleAtVertex()` method, which returns a `Sample` of all values in the process sample for a given vertex. The other methods use `getSampleAtVertex()` to get the stats. This seems as simple as possible.

jschueller (commented):
the documentation is missing

regislebrun:
I added some documentation. It should obviously be more extensive, and tests & examples are still missing. I will probably not have much more time to spend on this work. Any contribution will be greatly appreciated.

jschueller:
We need at least some minimal testing, like "print(sample.computeRange())"

jschueller (approved):

sofianehaddad (approved):

PR #1715, opened 2021-01-12
Fix Pareto & GeneralizedPareto

### LINKED ISSUE #1714: The log-PDF of the Pareto is wrong
The code for the LogPDF of the Pareto has a bug:
https://github.com/openturns/openturns/blob/bc1ba291d2377bdd2bbefce9009f1790e3443305/lib/src/Uncertainty/Distribution/Pareto.cxx#L146

The correct code would be:
```
std::log(alpha_) + alpha_ * std::log(beta_) + (-alpha_ - 1.0) * std::log(x)
```

regislebrun:
It solves issue #1714


lib/src/Uncertainty/Distribution/GeneralizedPareto.cxx
```diff
@@ -2,7 +2,7 @@
 /**
  *  @brief The GeneralizedPareto distribution
  *
- *  Copyright 2005-2021 Airbus-EDF-IMACS-ONERA-Phimeca
+ *  Copyright 2005-2020 Airbus-EDF-IMACS-ONERA-Phimeca
```
jschueller, line 5: wanna stay in 2020 ?

jschueller (approved):

lib/src/Uncertainty/Distribution/GeneralizedPareto.cxx
regislebrun: Do you see any change since last month ;-)?

sofianehaddad (approved):

jschueller:
that test passes only in 64 bits, so it should be skipped in mingw 32 bits, just do like the Trapezoidal test:
```
if (CMAKE_SIZEOF_VOID_P EQUAL 8)
ot_pyinstallcheck_test (TrapezoidalFactory_std)
endif ()
```

PR #1726, opened 2021-01-22
Nan-robust tests in Base

josephmure:
Convert tests to make them robust to NaNs.
All .cxx files in the Base directory were reviewed, but not .c files.

The idea is to use the following structure for tests leading to an exception:
`if (!good) throw Exception`
instead of
`if (bad) throw Exception`

### DIFF
--- a/lib/src/Base/Func/SpecFunc/SpecFunc.cxx
+++ b/lib/src/Base/Func/SpecFunc/SpecFunc.cxx
@@ -520,7 +520,7 @@ Complex SpecFunc::Dawson(const Complex & z)
 Scalar SpecFunc::Debye(const Scalar x,
                        const UnsignedInteger n)
 {
-  if ((n == 0) || (n > 20)) throw InvalidArgumentException(HERE) << "Error: cannot compute Debye function of order outside of {1,...,20}";
+  if (!(n > 0 && n <= 20)) throw InvalidArgumentException(HERE) << "Error: cannot compute Debye function of order outside of {1,...,20}, here order=" << n;
   if (x < 0.0) return Debye(-x, n) - n * x / (n + 1.0);
   // The threshold is such that the overall error is less than 1.0e-16
   if (x < 1.0e-8) return 1.0 - n * x / (2.0 * (n + 1.0));
@@ -535,7 +535,7 @@ Scalar SpecFunc::DiLog(const Scalar x)
   // Special case for 1
   if (x == 1.0) return PI2_6;
   // No real value on (1, \infty)
-  if (!(x <= 1.0)) throw InvalidArgumentException(HERE) << "Error: the DiLog function does not take real values for arguments greater than 1.";
+  if (!(x <= 1.0)) throw InvalidArgumentException(HERE) << "Error: the DiLog function does not take real values for arguments greater than 1, here argument=" << x;
   // Special case for x close to 1
   if (x >= 0.999997)
   {
@@ -1276,7 +1276,7 @@ Complex SpecFunc::Log1MExp(const Scalar x)
 // Integer log2
 UnsignedInteger SpecFunc::Log2(const Unsigned64BitsInteger n)
 {
-  if (n == 0) throw InvalidArgumentException(HERE) << "Error: n must be positive";
+  if (!(n > 0)) throw InvalidArgumentException(HERE) << "Error: n must be positive";
 
   // De Bruijn sequence
   const UnsignedInteger tab64[64] =
@@ -1362,7 +1362,7 @@ UnsignedInteger SpecFunc::BitCount(const Unsigned64BitsInteger n)
 // Missing functions in cmath wrt math.h as of C++98
 Scalar SpecFunc::Acosh(const Scalar x)
 {
-  if (!(x >= 1.0)) throw InvalidArgumentException(HERE) << "Error: acosh is not defined for x<1, here x=" << x;
+  if (!(x >= 1.0)) throw InvalidArgumentException(HERE) << "Error: acosh is only defined for x>=1, here x=" << x;
   return 2.0 * std::log(sqrt(0.5 * (x + 1.0)) + sqrt(0.5 * (x - 1.0)));
 }
 
--- a/lib/src/Base/Algo/BoundingVolumeHierarchy.cxx
+++ b/lib/src/Base/Algo/BoundingVolumeHierarchy.cxx
@@ -256,7 +256,7 @@ inline Bool is_point_inside_bounds(const Point & point, const Point & lowerBound
 {
   for(UnsignedInteger i = 0; i < point.getSize(); ++i)
   {
-    if (point[i] < lowerBounds[i] || point[i] > upperBounds[i])
+    if (!(point[i] >= lowerBounds[i] && point[i] <= upperBounds[i]))
       return false;
   }
   return true;
--- a/lib/src/Base/Algo/CholeskyMethod.cxx
+++ b/lib/src/Base/Algo/CholeskyMethod.cxx
@@ -176,7 +176,7 @@ void CholeskyMethod::update(const Indices & addedIndices,
   {
     Indices newBasis(conservedIndices);
     newBasis.add(addedIndices);
-    if (removedIndices.getSize() > 0) throw NotYetImplementedException(HERE) << " in CholeskyMethod::update removedIndices.getSize() > 0";
+    if (!(removedIndices.getSize() == 0)) throw NotYetImplementedException(HERE) << " in CholeskyMethod::update removedIndices.getSize() is not 0";
 
     // Update decomposition
     if (addedIndices.getSize() > 0)
--- a/lib/src/Base/Algo/CorrectedLeaveOneOut.cxx
+++ b/lib/src/Base/Algo/CorrectedLeaveOneOut.cxx
@@ -80,7 +80,7 @@ Scalar CorrectedLeaveOneOut::run(LeastSquaresMethod & method,
   const Scalar variance = y.computeVariance()[0];
 
   const UnsignedInteger basisSize = method.getImplementation()->currentIndices_.getSize();
-  if (sampleSize < basisSize) throw InvalidArgumentException(HERE) << "Not enough samples (" << sampleSize << ") required (" << basisSize << ")";
+  if (!(sampleSize >= basisSize)) throw InvalidArgumentException(HERE) << "Not enough samples (" << sampleSize << ") required (" << basisSize << ")";
 
   // Build the design of experiments
   LOGINFO("Build the design matrix");
--- a/lib/src/Base/Algo/DesignProxy.cxx
+++ b/lib/src/Base/Algo/DesignProxy.cxx
@@ -78,7 +78,7 @@ void DesignProxy::initialize() const
   // allocate cache
   UnsignedInteger cacheSize = ResourceMap::GetAsUnsignedInteger("DesignProxy-DefaultCacheSize");
   UnsignedInteger nbRows = x_.getSize();
-  if (nbRows == 0) throw InvalidArgumentException(HERE) << "Cannot initialize a DesignProxy with an empty sample";
+  if (!(nbRows > 0)) throw InvalidArgumentException(HERE) << "Cannot initialize a DesignProxy with an empty sample";
 
   UnsignedInteger nbCols = cacheSize / nbRows;
   // The cache stores at least the first function values
--- a/lib/src/Base/Algo/EnclosingSimplexAlgorithmImplementation.cxx
+++ b/lib/src/Base/Algo/EnclosingSimplexAlgorithmImplementation.cxx
@@ -283,9 +283,7 @@ Bool EnclosingSimplexAlgorithmImplementation::checkPointInSimplex(const Point & 
   // Exit if point is outside simplex bounding box
   for (UnsignedInteger i = 0; i < dimension; ++i)
   {
-    if(point[i] < lowerBoundingBoxSimplices_(index, i))
-      return false;
-    if(point[i] > upperBoundingBoxSimplices_(index, i))
+    if (!(point[i] >= lowerBoundingBoxSimplices_(index, i) && point[i] <= upperBoundingBoxSimplices_(index, i)))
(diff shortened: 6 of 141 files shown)


jschueller:
no need to complicate code for integer types: nan/inf are only for floatting point values

josephmure:
Parts of the code (perhaps newer code by @regislebrun or @mbaudin47?) seem to do this for integer types too.  Although it is not required, I think it is better to have a uniform policy for tests. Besides, it is not really more complicated, just a different mindset: "which conditions need to be met in order to make the code work?" instead of "which conditions make the code _not_ work?".

regislebrun (approved):
It would have been better not to change the code working on integers (much less files to check!) but I agree with you: I prefer to have the same way to test things regardless of their type even if it somehow hides the true difference between the way real numbers are approximated by floats and the way integers are approximated by ints.

josephmure:
@mbaudin47 asked me to provide an example where NaNs are blocked by this new code. Let us take the example of the Box-Cox transform.
```
import openturns as ot
import numpy as np
myBoxCox = ot.BoxCoxTransform(1.5)
```
### Transform a `Point` containing a nonpositive scalar.
```
print(myBoxCox([0.0]))
```
In version 1.16, this produces:
```
TypeError: InvalidArgumentException : Can not apply the Box Cox evaluation function to a negative shifted value x=0
```
With the fix, this produces a slightly more accurate message:
```
TypeError: InvalidArgumentException : Can not apply the Box Cox evaluation function to a nonpositive shifted value x=0
```
### Let us now input a Nan.
```
print(myBoxCox([np.nan]))
```
Version 1.16 thinks NaN is a valid input value and prints:
```
[nan]
```
With the fix, an Exception is thrown:
```
TypeError: InvalidArgumentException : Can not apply the Box Cox evaluation function to a nonpositive shifted value x=nan
```
### With a `Sample` as input

Unfortunately, unlike `BoxCoxEvaluation::operator() (const Point & inP)`, `BoxCoxEvaluation::operator() (const Sample & inS)` does not implement the positivity check.
```
print(myBoxCox([[-1.0]]))
```
Both versions produce NaN in this case:
```
    [ y0  ]
0 : [ nan ]
```

josephmure:
Following @regislebrun's advice in https://github.com/openturns/openturns/pull/1788#discussion_r588942150, I have tried to formalize the coding rule regarding exception handling that I have been following in this PR.

Since PR #1783 is not yet merged, here is a screenshot of the addition to `coding_rules.rst`:

image

regislebrun:
Nice work. Their is just a little glitch "under which the which the...". You should also include Julien's remark and your answer: integral values cannot generate NaNs, but we still code tests the same way for the uniformity of the code.

josephmure:
OK. I have updated the screenshot to reflect these changes.

regislebrun:
Ok for me (the rules, I didn't checked the code yet but I'm sure its right)

jschueller:
I'm still not comfortable applying these to integers, I find something like "if (size == 0) throw" is still very clear to understand.

regislebrun:
@josephmure @jschueller All in all I don't care how we write tests for integers, as long as they are correct ;-). Between the ease of reading and the uniformity, I hope to be still able to read and check the tests without referring to a norm. But in the floating point context I find it valuable to have guidelines to write NaN robust tests, which is precisely the purpose of this PR  (and integer arithmetic does not produces NaN)

jschueller:
I propose to remove the last part with the examples about integers then.

josephmure:
I have updated the rule to state that it does not apply to integers and rebased on master.

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -1065,6 +1069,7 @@ Scalar SpecFunc::HyperGeom_2_2(const Scalar p1,
   const Scalar logX = std::log(std::abs(x));
   Scalar signX = x > 0.0 ? 1.0 : -1.0;
   Scalar signTerm = 1.0;
+  UnsignedInteger k = 0;
```
josephmure, line 1072:
```suggestion
```

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -745,6 +745,8 @@ Scalar SpecFunc::RegularizedIncompleteGammaInverse(const Scalar a,
 // Derivative of a Lanczos approximation of log(Gamma)
 Scalar SpecFunc::DiGamma(const Scalar x)
 {
+  // Throw if x is NaN
+  if (x != x) throw InvalidArgumentException(HERE) << "Error: the argument of the DiGamma function is " << x;
```
regislebrun, line 749: You have !IsNormal(x) for that.

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -790,6 +792,8 @@ Scalar SpecFunc::DiGammaInv(const Scalar a)
 // Trigamma function: TriGamma(x) = ((d^2gamma/dx^2) / gamma)(x)
 Scalar SpecFunc::TriGamma(const Scalar x)
 {
+  // Throw if x is NaN
+  if (x != x) throw InvalidArgumentException(HERE) << "Error: the argument of the TriGamma function is " << x;
```
regislebrun, line 796: Same here

lib/src/Base/Optim/Cobyla.cxx
```diff
@@ -269,6 +269,8 @@ int Cobyla::ComputeObjectiveAndConstraint(int n,
         return 1;
       }
   Point outP(problem.getObjective().operator()(inP));
+  // Detect the presence of a NaN
+  if (outP[0] != outP[0]) throw InvalidArgumentException(HERE) << "Error : the current output value is " << outP[0];
```
regislebrun, line 273: Same here

regislebrun (changes requested):
You shuold use IsNormal(), IsNaN() and IsInf() as much as possible in such a PR!

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
josephmure:
```suggestion
  if IsNaN(x) throw InvalidArgumentException(HERE) << "Error: the argument of the DiGamma function is " << x;
```

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
josephmure:
```suggestion
  if IsNaN(x) throw InvalidArgumentException(HERE) << "Error: the argument of the TriGamma function is " << x;
```

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
josephmure:
```suggestion
  if (IsNaN(x)) throw InvalidArgumentException(HERE) << "Error: the argument of the DiGamma function is " << x;
```

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
josephmure:
```suggestion
  if (IsNaN(x)) throw InvalidArgumentException(HERE) << "Error: the argument of the TriGamma function is " << x;
```

lib/src/Base/Optim/Cobyla.cxx
josephmure:
```suggestion
  if (IsNaN(outP[0])) throw InvalidArgumentException(HERE) << "Error : the current output value is " << outP[0];
```

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -745,6 +745,8 @@ Scalar SpecFunc::RegularizedIncompleteGammaInverse(const Scalar a,
 // Derivative of a Lanczos approximation of log(Gamma)
 Scalar SpecFunc::DiGamma(const Scalar x)
 {
+  // Throw if x is NaN
+  if (IsNaN(x)) throw InvalidArgumentException(HERE) << "Error: the argument of the DiGamma function is " << x;
```
jschueller, line 749: Use IsNormal to check inf too

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -790,6 +792,8 @@ Scalar SpecFunc::DiGammaInv(const Scalar a)
 // Trigamma function: TriGamma(x) = ((d^2gamma/dx^2) / gamma)(x)
 Scalar SpecFunc::TriGamma(const Scalar x)
 {
+  // Throw if x is NaN
+  if (IsNaN(x)) throw InvalidArgumentException(HERE) << "Error: the argument of the TriGamma function is " << x;
```
jschueller, line 796: same

lib/src/Base/Optim/Cobyla.cxx
```diff
@@ -269,6 +269,8 @@ int Cobyla::ComputeObjectiveAndConstraint(int n,
         return 1;
       }
   Point outP(problem.getObjective().operator()(inP));
+  // Detect the presence of a NaN
+  if (SpecFunc::IsNaN(outP[0])) throw InvalidArgumentException(HERE) << "Error : the current output value is " << outP[0];
```
jschueller, line 273: here too

josephmure: Infinite values are already caught.

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
josephmure: It already throws for invalid infinite values.

jschueller: where ? IsNaN does not checks for inf values

josephmure: I don't know, I simply ran the function on infinite values.

josephmure: NaN, on the other hand, does not throw with the current master so I patched this problem.

regislebrun: As the code is valid only for normal numbers, you should use (!IsNormal(x)) even if it already throw for inf values somewhere else (in the x == round(x) test, which primary intention is to check if x is an integer, not an infinite number, see the message in the exception). It makes things much more readable.

jschueller: it's mostly best to catch invalid values early instead of wasting computations

josephmure: I disagree. *If it ain't broken, don't fix it!*

josephmure:
For the sake of getting this PR merged, I will use remove the litigious lines and keep OpenTURNS' current behavior. Let me simply note that R and Scipy implement the polygamma functions differently.

# The DiGamma function

## DiGamma function In R:

```
> digamma(Inf)
[1] Inf

> digamma(0)
[1] NaN
Warning message:
In digamma(0) : production de NaN

> digamma(-1)
[1] NaN
Warning message:
In digamma(-1) : production de NaN

> digamma(-Inf)
[1] NaN
Warning message:
In digamma(-Inf) : production de NaN
```

## DiGamma function in Scipy:

```
from scipy.special import digamma
from numpy import inf

digamma(inf)
Out[27]: inf

digamma(0) # positive 0
Out[29]: -inf

digamma(-1.0/inf) # negative 0
Out[33]: inf

digamma(-1)
Out[34]: nan

digamma(-inf)
Out[35]: nan
```

## DiGamma function in OpenTURNS:

```
from openturns import SpecFunc_DiGamma as DiGamma
from numpy import inf

DiGamma(inf)
Out[39]: inf

DiGamma(0)
TypeError: InvalidArgumentException : Error: the argument of the DiGamma function cannot be a non positive integer.

DiGamma(-1)
TypeError: InvalidArgumentException : Error: the argument of the DiGamma function cannot be a non positive integer.

DiGamma(-inf)
TypeError: InvalidArgumentException : Error: the argument of the DiGamma function cannot be a non positive integer.
```

## Summary of DiGamma behaviors

- Scipy chooses to return limits whenever possible (at positive infinity, at positive 0 and at negative 0) and returns nan when this is not possible.
- R does the same except it does not distinguish between positive and negative 0: since the limits are different, it returns nan and issues a warning.
- OpenTURNS has the same behavior as R except it throws instead of issuing warnings.

josephmure:
Let us examine TriGamma behavior the same way.

# The TriGamma function

## TriGamma function in R:

```
> trigamma(Inf)
[1] 0

> trigamma(0)
[1] Inf

> trigamma(-1)
[1] Inf

> trigamma(-Inf)
[1] Inf
```

## TriGamma in Scipy

```
from scipy.special import polygamma
from numpy import inf

polygamma(1, inf)
Out[47]: array(0.)

polygamma(1, 0)
Out[50]: array(inf)

polygamma(1, -1)
Out[51]: array(inf)

polygamma(1, -inf)
Out[53]: array(inf)
```

## TriGamma function in OpenTURNS:

```
TriGamma(inf)
Out[55]: 0.0

TriGamma(0)
TypeError: InvalidArgumentException : Error: the argument of the TriGamma function cannot be a non positive integer.

TriGamma(-1)
TypeError: InvalidArgumentException : Error: the argument of the TriGamma function cannot be a non positive integer.

TriGamma(-inf)
TypeError: InvalidArgumentException : Error: the argument of the TriGamma function cannot be a non positive integer.
```

## Summary of TriGamma behaviors

- Since limits are always well defined, R and Scipy return them.
- OpenTURNS only returns the limit at positive infinity, and returns nan in all other cases.

regislebrun:
So no more checks for NaN/Inf for SpecFunc functions arguments? Or did I missed something these last days?

josephmure:
No, you are correct. They can be added later.

josephmure:
Is this branch OK now?

regislebrun (approved):

PR #1732, opened 2021-01-25
Doc: Fixes  #1731

### LINKED ISSUE #1731: The Extreme value example may be improved.
The Extreme value example may be improved:

http://openturns.github.io/openturns/master/auto_data_analysis/distribution_fitting/plot_fit_extreme_value_distribution.html

* The actual distribution does not fit the comment ; the comment says "GumbelMax", but the script says "Frechet":

image

or later:

image

* There are unexpected blank spaces in front of colons:
```
The distributions used :
```

* Comments may be regarded as inappropriate:
```
For educational purpose we compare different fitting strategies for this sample :
```
Hence, 
```
We compare different fitting strategies for this sample :
```
would sound simpler, regardless of the intent of the user.

mbaudin47:


jschueller:
I suggest a more relevant commit message, the issue should be mentioned in the commit message body.
The Changelog should be updated.

josephmure:
I have cherry-picked this in #1736 so this PR can be closed. @jschueller the new commit in #1736 applies your suggestions.

PR #1743, opened 2021-01-31
Update the ExprTk library

ArashPartow:


jschueller (approved):

regislebrun (approved):
Thanks for your contribution. Could it be possible to have a short description of what has changed since the previous version?

PR #1766, opened 2021-02-10
Misc

jschueller:
- rework LinearModelStepwiseAlgorithm api
- #1772 
- #1776 
- #1784 DistributionFactory: Reject constant/nan samples
- #1787
- #1790
- #1793
- #1794
- #1795
- Sample: Drop FromCSV->FromText fallback
- started generic Distribution test (https://openturns.discourse.group/t/automated-consistency-tests-of-a-distribution/82/7)


jschueller:
maybe it's time

sofianehaddad (approved):

sofianehaddad:
@jschueller how could we check `locally` that #1794 is fixed?

jschueller:
I guess rebuild the doc

sofianehaddad:
Hmm I have also troubles with #1795 
History convergences is "constant" and axis incoherent 
image

I don't think this is what we are expecting....

lib/src/Base/Geom/IntervalMesher.cxx
```diff
@@ -192,6 +192,10 @@ Mesh IntervalMesher::build(const Interval & interval,
   } // dimension == 2
   if (dimension == 3)
   {
+    // FIXME: https://github.com/openturns/openturns/issues/1670
+    if (diamond)
+      throw NotYetImplementedException(HERE) << "In IntervalMesher::build with 3-d/diamond=true";
```
regislebrun, line 197: I started a branch to fix this one, but it is not so easy...

lib/src/Base/Stat/CovarianceMatrix.cxx
```diff
@@ -45,6 +45,13 @@ CovarianceMatrix::CovarianceMatrix(const MatrixImplementation & i)
   // Nothing to do
 }
 
+/* Constructor from SymmetricMatrix */
+CovarianceMatrix::CovarianceMatrix(const SymmetricMatrix & sym)
```
regislebrun, line 49: Do we really need that? Isn't it supposed to be done by inheritance?

lib/src/Base/Stat/CorrelationMatrix.cxx
```diff
@@ -46,6 +46,13 @@ CorrelationMatrix::CorrelationMatrix(const MatrixImplementation & i)
   // Nothing to do
 }
 
+/* Constructor from SymmetricMatrix */
+CorrelationMatrix::CorrelationMatrix(const SymmetricMatrix & sym)
```
regislebrun, line 50: Do we really need that? Isn't it supposed to be done by inheritance?

lib/src/Base/Stat/SampleImplementation.cxx
```diff
@@ -313,9 +313,8 @@ CLASSNAMEINIT(SampleImplementation)
 SampleImplementation SampleImplementation::BuildFromCSVFile(const FileName & fileName,
     const String & csvSeparator)
 {
-  if (csvSeparator == " ") throw InvalidArgumentException(HERE) << "Error: the space separator is not compatible for CSV file.";
-
-
+  if (csvSeparator == " ")
```
regislebrun, line 316: Maybe we should also add tab in the list of forbidden separators?

lib/src/Uncertainty/Algorithm/MetaModel/LinearModel/LinearModelStepwiseAlgorithm.cxx
```diff
@@ -155,21 +128,25 @@ LinearModelStepwiseAlgorithm::Direction LinearModelStepwiseAlgorithm::getDirecti
 }
 
 /* Penalty accessors */
+void LinearModelStepwiseAlgorithm::setPenalty(const Scalar penalty)
+{
+  penalty_ = penalty;
```
regislebrun, line 133: Any value is allowed for penalty?

python/test/t_Distribution_std.py
```diff
@@ -0,0 +1,68 @@
...
+    CDF_up = distribution.computeCDF(interval.getUpperBound())
+    CDF_low = distribution.computeCDF(interval.getLowerBound())
+    computed_probability = CDF_up - CDF_low
+    if distribution.getName() != "Histogram":
+        ott.assert_almost_equal(probability, computed_probability, 1e-5, 1e-5, str(distribution))
+    
+    # MinimumVolumeLevelSet
+    probability = 0.9
+    levelSet, threshold = distribution.computeMinimumVolumeLevelSetWithThreshold(probability)
+    x = distribution.computeQuantile(1.0 - (1.0 - probability) / 2.0)
```
regislebrun, line 57: IT IS WRONG! It assumes that the minimum volume level set is a segment symmetrical wrt the median.I already fixed it and sent the fix to @mbaudin47. I don't remember on which channel, probably as a comment on the repo where you picked the code.

regislebrun (changes requested):
Only one serious issue, otherwise only surface comments.

python/test/t_Distribution_std.py
sofianehaddad: image

lib/src/Base/Stat/CovarianceMatrix.cxx
jschueller: Yes, I tried a the c++ level

lib/src/Uncertainty/Algorithm/MetaModel/LinearModel/LinearModelStepwiseAlgorithm.cxx
jschueller: I think I should be positive

jschueller:
> Hmm I have also troubles with #1795
> History convergences is "constant" and axis incoherent
> image
> 
> I don't think this is what we are expecting....

I fixed that too

sofianehaddad:
For #1795 we can improve last graph in `ego` and illustrate the usage of `kwargs` in `View`
```
view = viewer.View(graph, axes_kw={"xticks": range(0, result.getIterationNumber(), 5)})
plt.show()
```
This avoids having non integer values for iteration number as illustrated below:
image

jschueller:
ok, applied

lib/src/Base/Geom/IntervalMesher.cxx
jschueller: I dont think its urgent

PR #1773, opened 2021-02-15
Minor bugfix

### LINKED ISSUE #1771: GridLayout hides the titles in the graph
When we create a grid, each graph in the grid may (or may not) have its own title. However, it seems that the class hides them by default. Moreover, it seems that it cannot be shown.

The following script, for example, 
```
import openturns as ot

sample_array = [100, 200, 400, 1000]
grid = ot.GridLayout(1, len(sample_array))
for i in range(len(sample_array)):
    sample = ot.Normal().getSample(sample_array[i])
    distribution = ot.KernelSmoothing().build(sample)
    graph = distribution.drawPDF()
    graph.setTitle("n=%d" % (sample_array[i]))
    graph.setLegends([""])
    grid.setGraph(0, i, graph)
```
produces:

image

We see that the titles in the graphs are not shown.

sofianehaddad:
Fix #1771

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -42,6 +42,7 @@
  * #1752 (Error while estimating the reduced log-likelihood when using a StationaryFunctionalCovarianceModel)
  * #1759 (Kriging model with StationaryFunctionalCovarianceModel might provide bad results)
  * #1768 (Bug in ExponentiallyDampedCosineModel & SphericalModel)
+ * #1771 (GridLayout hides the titles in the graph)
 
 
 == 1.16 release (2020-11-16) == #release-1.16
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -42,7 +42,7 @@
  * #1752 (Error while estimating the reduced log-likelihood when using a StationaryFunctionalCovarianceModel)
  * #1759 (Kriging model with StationaryFunctionalCovarianceModel might provide bad results)
  * #1768 (Bug in ExponentiallyDampedCosineModel & SphericalModel)
-
+ * #1771 (GridLayout hides the titles in the graph)
```
jschueller, line 45: you're deleting a line there, we want to avoid conflicts in that file

jschueller (approved):

ChangeLog
sofianehaddad: hmm strange. I will fix it.

PR #1777, opened 2021-02-17
Update chaboche_model.py

mbaudin47:
Added description of the model to fill the labels of the plots.


python/src/usecases/chaboche_model.py
```diff
@@ -9,7 +9,7 @@
 
 def g(X):
     strain, R, C, gamma = X
-    stress = R + C * (1 - np.exp(-gamma * strain)) / gamma
+    stress = R + C * (1.0 - np.exp(-gamma * strain)) / gamma
```
regislebrun, line 12: Maybe np.expm1 should be preferred here?

josephmure:
```suggestion
    stress = R - C * np.expm1(-gamma * strain) / gamma
```

mbaudin47: This is technically OK, but that does not solve any actual problem in this particular case, because we generally are not interested in *extremely small* strains :)

regislebrun: just a good habit, nothing more ;-)

mbaudin47: Ok. I'll brush my teeth twice, just to make sure!

josephmure (approved):

python/src/usecases/chaboche_model.py
```diff
@@ -9,7 +9,7 @@
 
 def g(X):
     strain, R, C, gamma = X
-    stress = R + C * (1 - m.exp(-gamma * strain)) / gamma
+    stress = R - C * np.expm1(-gamma * strain) / gamma
```
sofianehaddad, line 12: seems numpy is not imported?

sofianehaddad: By the way, math has the same function thus you can use `m.exp1`

jschueller:
moved to #1802

PR #1783, opened 2021-02-20
CircleCI saves doc pages as artifacts

### LINKED ISSUE #1781: The link_to_an_external_code example has small bugs
The example:

http://openturns.github.io/openturns/master/auto_functional_modeling/link_to_an_external_code/plot_link_computer_code_coupling_tools.html

contains an invalid link to an image:

image

It has format issues:

image

image

image

josephmure:
As suggested by @tupui on [Discourse](https://openturns.discourse.group/t/build-doc-without-code/93/3?u=josephmure), I have modified the `.circleci/config.yml` file in order to make CircleCI `build_linux` save all doc pages as artifacts. This way the doc will be temporarily saved (for a duration of 30 days) every time new commits are pushed. This should make doc reviews easier.

EDIT: Since @jschueller has pushed updates to the CircleCI config, I have rebased the branch and updated my proposal. The change w.r.t. the master is that artifacts are uploaded between build and test. This prevents issues affecting tests making the compiled doc unavailable. I have also tried to fix the issue reported by @jschueller concerning the `utils/build_locally.sh` script, but since I do not know how to use docker, I cannot check if it works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,8 +9,14 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Build and test
+          name: Build
           command: ./.circleci/run_docker_linux.sh
+      - store_artifacts:
+          path: ~/.local/share/openturns/doc/html
+          destination: html
+      - run:
+          name: Test
+          command: ./.circleci/run_docker_linux_test.sh
       - run:
           name: Publish documentation
           command: ./.circleci/upload_github_io.sh
--- a/.circleci/run_docker_linux.sh
+++ b/.circleci/run_docker_linux.sh
@@ -2,9 +2,6 @@
 
 set -xe
 
-uid=$1
-gid=$2
-
 env
 
 # build with frozen date unless on release for reproducible builds
@@ -28,11 +25,4 @@ cmake -DCMAKE_INSTALL_PREFIX=~/.local \
       -DCMAKE_CXX_FLAGS="-Wall -Wextra -Werror -D_GLIBCXX_ASSERTIONS" -DSWIG_COMPILE_FLAGS="-O1 -Wno-unused-parameter -Wno-missing-field-initializers" \
       ${source_dir}
 make install
-if test -n "${uid}" -a -n "${gid}"
-then
-  zip -r openturns-doc.zip ~/.local/share/openturns/doc/html/*
-  sudo chown ${uid}:${gid} openturns-doc.zip && sudo cp openturns-doc.zip ${source_dir}
-fi
-ctest -R pyinstallcheck --output-on-failure --timeout 100 ${MAKEFLAGS} --repeat after-timeout:2 --schedule-random
-make tests
-ctest -R cppcheck --output-on-failure --timeout 100 ${MAKEFLAGS} --repeat after-timeout:2 --schedule-random
+
--- a/.circleci/run_docker_linux_test.sh
+++ b/.circleci/run_docker_linux_test.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+set -xe
+
+uid=$1
+gid=$2
+
+env
+
+## copy build variables
+# build with frozen date unless on release for reproducible builds
+if test "${CIRCLE_BRANCH}" = "master"
+then
+  export SOURCE_DATE_EPOCH=1514764800
+fi
+
+if test -d "/io"
+then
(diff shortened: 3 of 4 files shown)


josephmure:
## New `CovarianceModel.discretize` doc:

image

tupui:
You could also have a direct link to the generated doc from the checks. Looking at SciPy's config, I believe you would need to add something like this:

```yaml
      - persist_to_workspace:
          root: ~/.local/share/openturns/doc/
          paths:
            - doc_html
```
And a file named `artifact_path` with this line: `0/doc_html/index.html`.

josephmure:
Thanks again @tupui for pointing this out! It turns out that we do not need the `attach_workspace` and `persist_to_workspace` steps when all we want to do is to save artifacts.

To get this we need to either use the GitHub action [run-circleci-artifacts-redirector](https://github.com/marketplace/actions/run-circleci-artifacts-redirector) or the GitHub app [circleci-artifacts-redirector](https://github.com/apps/circleci-artifacts-redirector/). The latter needs to read the `.circleci/artifact_path` file and both need the CircleCI job to be named either "build", "doc", or "build-doc". This is why I have renamed "build_linux" as "build". I have tried out the GitHub app and it works on my fork:

image

I will try to set up the GitHub action in my next commit. It is supposedly more ["future-proof"](https://github.com/larsoner/circleci-artifacts-redirector#circleci-artifacts-redirector) than the app according to its creator.

josephmure:
I am not sure my commit will be enough to trigger the action. On the [run-circleci-artifacts-redirector](https://github.com/marketplace/actions/run-circleci-artifacts-redirector) marketplace page, there is a note:

>Note: The standard PR-to-main-repo-from-branch-in-a-fork workflow might not activate the action. For changes to take effect, changes might need to be made to to the default branch in a repo with the action enabled. For example, you could iterate directly in master, or in master of a fork. This seems to be a limitation of the fact that CircleCI uses the status (rather than app) API and that this is always tied to the master/default branch of a given repository.

If it does not trigger, I will try to set it up directly in the master branch of my fork and see what happens.

josephmure:
The GitHub action  [run-circleci-artifacts-redirector](https://github.com/marketplace/actions/run-circleci-artifacts-redirector) triggered on my fork after I pushed commit https://github.com/josephmure/openturns/commit/fc4f8f954330119f3460ed56b127613b12ff7e5a to its master branch. The action now triggers in both "master" and "doc" branches of my fork, so you can see the result by clicking on the "josephmure:doc" button at the top of this PR.

I guess that we will have to wait until this PR is merged to see the GitHub action be triggered by PRs.

jschueller:
Looks good.
Can you drop the circleci commits ? I dont see a clean way to write the ci scripts.

josephmure:
We really need this to work, though. It is an important feature.

josephmure:
Have you tested the build_locally fix I proposed?

jschueller:
> We really need this to work, though. It is an important feature.
Current master already works (not as fast though)

jschueller:
> Have you tested the build_locally fix I proposed?
No, but I see it's incomplete

josephmure:
Is there a tutorial on how to run docker with openturns?

jschueller:
> Is there a tutorial on how to run docker with openturns?



> Is there a tutorial on how to run docker with openturns?
no

josephmure:
Here is my proposal: merge this PR as is, and then we can work on fixing build_locally.sh

jschueller:
> Here is my proposal: merge this PR as is, and then we can work on fixing build_locally.sh
I prefer the other way around

josephmure:
Does anybody else have an idea about how to fix this? @sofianehaddad @regislebrun @mbaudin47 @KieranDelamotte

jschueller:
The current master is already good enough, you can already see the PR doc, you just have to wait a bit long longer

josephmure:
And hope no test fails or timeouts.

jschueller:
Yes, I pushed more fixes for that.

josephmure:
Here is what I will do: I will move the doc fixes to a new PR so you can merge them. This PR stays open to deal with CircleCI.

jschueller:
ok

josephmure:
Closing because this is no longer needed: CircleCI now allows the job to continue even after one step has failed.

PR #1798, opened 2021-03-18
CI: Store doc pages artifacts

jschueller:
simpler implementation of #1783

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,9 @@ jobs:
       - run:
           name: Publish documentation
           command: ./.circleci/upload_github_io.sh
+      - store_artifacts:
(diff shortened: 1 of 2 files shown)


josephmure:
The drawback of this implementation is that a single Python or C++ unit test timeout is enough to prevent the doc being stored, and these timeouts happen often. There will be cases where we will use the GitHub web interface to make a small change in the doc and won't be able to see the result because an unrelated test will have timed out.

jschueller:
I've made several tests shorter rencently to adress these timeouts.

josephmure:
Still, what is the issue with splitting the build and the test phase?

jschueller:
It it breaking the build_locally script.

josephmure:
Can you tell me how to fix it?

PR #1799, opened 2021-03-19
CI: fix timeouts

jschueller:


jschueller:
@josephmure @regislebrun there are still some timeouts, I believe this PR could help

Issue #1751, opened 2021-02-03
DrawCorrelationCoefficients has a wrong Text height


mbaudin47:
The height of the labels in `DrawCorrelationCoefficients` is wrong:

http://openturns.github.io/openturns/master/auto_data_analysis/manage_data_and_samples/plot_sample_correlation.html?highlight=drawcorrelationcoefficients

This is seen in the figure:

![image](https://user-images.githubusercontent.com/31351465/106780074-ed026380-6647-11eb-92f6-95d82d357d5f.png)

In the previous figure, the "X1" label is much too low. 
This is because of:

https://github.com/openturns/openturns/blob/bc1ba291d2377bdd2bbefce9009f1790e3443305/lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx#L831

I guess that:

```
data(k, 1) = values[k] - 0.1;
```

would be good, because the correlation indices are scaled in [-1.0, 1.0].

Issue #1725, opened 2021-01-21
The BetaFactory help page has wrong equations


mbaudin47:
Once the bound parameters $\hat{a}$ and $\hat{b}$ are estimated by the `BetaFactory`, the shape parameters $\alpha$ and $\beta$ are computed by the method of moments. However, the equations are stated depending on $t$ and $r$:

![image](https://user-images.githubusercontent.com/31351465/105415924-b38a2b00-5c39-11eb-8835-02f4d5c187ce.png)

which was the parametrization used in older versions of OpenTURNS. More precisely, the new parametrization uses the shape parameters $\alpha$ and $\beta$ instead of the old parameters $t$ and $r$.

These equations should be replaced with the ones presented in the `BetaMuSigma` class:

![image](https://user-images.githubusercontent.com/31351465/105416239-24314780-5c3a-11eb-994c-6446a9e6806c.png)

which are correct, but have an exceeding number of parentheses before and after the fraction in the equation for $\beta$. Notice that, in the doc of the `BetaFactory`, hats must be placed onto $\alpha$ and $\beta$ since these are estimates and are, in general, different from the true, but unknown, parameters.

mbaudin47:
@josephmure : this one is an excellent candidate for a doc-bug-smashing-session.

Issue #1257, opened 2019-09-17
Cannot create a SimulatedAnnealingLHS without specifying the temperature profile


regislebrun:
We cannot create a SimulatedAnnealingLHS without specifying the temperature profile. This is a pity, since it would be interesting to configure the optimization criteria (PhiP, minDist, C2) without specifying the temperature profile.

Hence, the following script fails :

```
import openturns as ot
from openturns.viewer import View

# Considering 3 independent Uniform distributions
myU = ot.Uniform()
distribution = ot.ComposedDistribution([myU]*3)

# LHS
N = 100
lhs = ot.LHSExperiment(distribution, N)
lhs.setAlwaysShuffle(True) # randomized
crit = ot.SpaceFillingMinDist()
algo = ot.SimulatedAnnealingLHS(lhs,crit)
sample = algo.generate()
```
and produces :
```
NotImplementedError: Wrong number or type of arguments for overloaded function 'new_SimulatedAnnealingLHS'.
  Possible C/C++ prototypes are:
    OT::SimulatedAnnealingLHS::SimulatedAnnealingLHS()
etc...
```
The missing constructor is :
```
SimulatedAnnealingLHS(lhsDesign, spaceFilling)
```
This possibility would be interesting for the following reasons :
  * it is difficult to configure the temperature profile because the parameters of such a temperature profile are difficult to set,
  * it is easy to select a spaceFilling criteria among three choices (PhiP, minDist, C2). 

Linked to [http://trac.openturns.org/ticket/988](http://trac.openturns.org/ticket/988)

Issue #1229, opened 2019-09-16
Better encapsulation of optional 3rd-party headers


mbaudin47:
Hello,

When compiling an application using OT headers, one has currently to be in an environment very similar to the one which had been used to build OT. If OT had been built with muParser, TBB or XML support, then those header files are required to compile anything #including OT headers.

It would be much better if OT headers did not depend on those optional 3rd-party headers. One could then build OT and link against an XML static library, dependent applications would then not require XML to link against OT.

A proof-of-concept is attached for muParser; it needs some polishing. XML and TBB are probably much harder.

Reported by: | barbier@…

Reference: http://trac.openturns.org/ticket/727

mbaudin47:
Changed 5 years ago by barbier@…
Attachment encapsulate-muparser.patch​ added
[encapsulate-muparser.patch.txt](https://github.com/openturns/openturns/files/3617173/encapsulate-muparser.patch.txt)

comment:1 Changed 11 months ago by lebrun
Still true?

comment:2 Changed 11 months ago by schueller
Still true for xml, tbb, even boost when using msvc.

jschueller:
tbb encaspulation is not possible on msvc

PR #1802, opened 2021-03-22
Misc

jschueller:
- make boost optional on msvc (and public headers do not depend on it anymore)
it was required with old versions for math functions, but now we require c++11 anyways this can be simplified
- update install doc to mention openturns is not compatible with the anaconda installer and recommend miniforge
- remove public dependency to libxml2 to ease compilation for FMI (#1229)
- #1257 
- #1725 
- #1751
- #1810 
- #1815
- boost.multiprecision with mpc is only available from 1.68, so disable it with boost<1.68
- set linearmodelstepwise default penalty value to 2 instead of -1 cc @JPelamatti


lib/test/CMakeLists.txt
```diff
@@ -260,7 +260,9 @@ ot_check_test (TimeSeries_std)
 ot_check_test (FilteringWindows_std)
 ot_check_test (WelchFactory_std)
 ot_check_test (ExponentialModel_std)
+if (Boost_FOUND)
 ot_check_test (CovarianceModel_std IGNOREOUT)
+endif ()
```
sofianehaddad, line 265: Why `boost` is required here?

jschueller: MaternModel uses some bessel functions and an assert fails

sofianehaddad: Thus maybe we need to update the assert? there is an alternative to `Bessel` functions inside the library and it would be pity if test is not performed because `boost` is missing. What do you think?

jschueller: it's something else: this test randomly fails unless I disable tbb

jschueller: yes: LogBesselK uses SymbolicFunction without boost so it's not thread-safe

regislebrun: I can provide an implementation without SymbolicFunction if needed. Would it help?

jschueller: I already implemented MaternModel::isParallel

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -312,6 +312,30 @@ Scalar SpecFunc::DeltaLogBesselI10(const Scalar x)
...
+    else
+      outP[0] = exp(-x_ * cosh(t)) * std::pow(sinh(t), 2.0 * nu_);
+    return outP;
+  }
+
+private:
+  Scalar nu_ = 0.0;
+  Scalar x_ = 0.0;
+};
+
```
sofianehaddad, line 338: 👍

sofianehaddad (approved):

PR #1813, opened 2021-04-13
LinearModelResult tests

josephmure:
Unit tests were missing. Thanks to @sofianehaddad for explaining how Cook's distance is implemented.

### DIFF
--- a/lib/test/t_LinearModelAlgorithm_std.cxx
+++ b/lib/test/t_LinearModelAlgorithm_std.cxx
@@ -70,6 +70,15 @@ int main(int, char *[])
     LinearModelAlgorithm test(X, Y);
     LinearModelResult result(test.getResult());
     fullprint << "trend coefficients = " << result.getCoefficients() << std::endl;
+
+    // Test various attributes
+    const Point cook(result.getCookDistances());
+    const Point cook_reference = {0.0233296,0.0360369,0.00178903,0.0502183,0.0966701,0.00562596};
+    assert_almost_equal(cook, cook_reference, 1e-5, 0.0);
+
+    const Point leverages(result.getLeverages());
+    const Point leverages_reference = {0.0864939,0.0797831,0.0735447,0.0677578,0.0624023,0.0574582};
+    assert_almost_equal(leverages, leverages_reference, 1e-6, 0.0);
   }
 
   return ExitCode::Success;
--- a/python/test/t_LinearModelAlgorithm_std.py
+++ b/python/test/t_LinearModelAlgorithm_std.py
@@ -2,6 +2,7 @@
 
 from __future__ import print_function
 import openturns as ot
+import openturns.testing as ott
 from math import sin
 
 ot.TESTPREAMBLE()
@@ -44,6 +45,15 @@ for i in range(size):
 test = ot.LinearModelAlgorithm(X, Y)
 result = test.getResult()
 print ("trend coefficients = ", result.getCoefficients())
+
+# Test various attributes
+cook = result.getCookDistances()
+cook_reference = [0.0233296,0.0360369,0.00178903,0.0502183,0.0966701,0.00562596]
+ott.assert_almost_equal(cook[0:6], cook_reference, 1e-5, 0.0)
+
+leverages = result.getLeverages()
+leverages_reference = [0.0864939,0.0797831,0.0735447,0.0677578,0.0624023,0.0574582]
+ott.assert_almost_equal(leverages[0:6], leverages_reference, 1e-6, 0.0)
 
 # with invalid sample description
 X.setDescription(['X0', 'price (euros)'])


josephmure:
Normally, unit tests should fail because the formula for Cook's distance has changed.

sofianehaddad:
@josephmure current implementation is good (see Cook distance as function [of studentized residual](https://en.wikipedia.org/wiki/Cook%27s_distance#Relationship_to_other_influence_measures_(and_interpretation)))

lib/test/t_LinearModelAlgorithm_std.cxx
```diff
@@ -70,6 +70,23 @@ int main(int, char *[])
...
+    assert_almost_equal(cook[4], 0.0966701, 1e-5, 0.0);
+    assert_almost_equal(cook[4], 0.00562596, 1e-5, 0.0);
+
+    Point leverages(result.getLeverages());
+    assert_almost_equal(leverages[0], 0.0864939, 1e-6, 0.0);
+    assert_almost_equal(leverages[1], 0.0797831, 1e-6, 0.0);
+    assert_almost_equal(leverages[2], 0.0735447, 1e-6, 0.0);
+    assert_almost_equal(leverages[3], 0.0677578, 1e-6, 0.0);
+    assert_almost_equal(leverages[4], 0.0624023, 1e-6, 0.0);
+    assert_almost_equal(leverages[4], 0.0574582, 1e-6, 0.0);
```
jschueller, line 89: I'd rather issue one single call to assert_almost_equal on a Point
you can use "cont Point cook_reference = {0.086, 0.079, ...}"

josephmure: Thanks for the tip! Done.

jschueller (approved):

sofianehaddad (approved):

python/test/t_LinearModelAnalysis_std.py
```diff
@@ -52,6 +53,17 @@
...
+
+# Test various attributes
+cook = result.getCookDistances()
+cook_reference = [0.0233296,0.0360369,0.00178903,0.0502183,0.0966701,0.00562596]
+ott.assert_almost_equal(cook[0:6], cook_reference, 1e-5, 0.0)
+
+leverages = result.getLeverages()
+leverages_reference = [0.0864939,0.0797831,0.0735447,0.0677578,0.0624023,0.0574582]
+ott.assert_almost_equal(leverages[0:6], leverages_reference, 1e-6, 0.0)
+
```
jschueller, line 65: I think the right file to test these would be t_LinearModelAlgorithm not t_LinearModelAnalysis

josephmure: Test moved to t_LinearModelAlgorithm

lib/test/t_LinearModelAlgorithm_std.cxx
```diff
@@ -70,6 +70,15 @@ int main(int, char *[])
...
+
+    // Test various attributes
+    const Point cook(result.getCookDistances());
+    const Point cook_reference = {0.0233296,0.0360369,0.00178903,0.0502183,0.0966701,0.00562596};
+    assert_almost_equal(cook, cook_reference, 1e-5, 0.0);
+
+    const Point leverages(result.getLeverages());
+    const Point leverages_reference = {0.0864939,0.0797831,0.0735447,0.0677578,0.0624023,0.0574582};
+    assert_almost_equal(leverages, leverages_reference, 1e-6, 0.0);
   }
```
sofianehaddad, line 82: I just noticed this.
I have some troubles with.  IMHO `cook` data should be of size `100` (same as data sample)
How did you pass this?

sofianehaddad: It is very strange
```
(base) haddad_s@becane:~/dev/openturns/build$ ./lib/test/t_LinearModelAlgorithm_std
Fit y ~ 3 - 2 x + 0.05 * sin(x) model using 20 points (sin(x) ~ noise)
trend coefficients = class=Point name=Unnamed dimension=2 values=[3.01168,-2.00025]
Fit y ~ 1 + 0.1 x + 10 x^2 model using 100 points
trend coefficients = class=Point name=Unnamed dimension=3 values=[0.978992,0.110565,9.99924]
libc++abi.dylib: terminating with uncaught exception of type OT::InvalidArgumentException: A and B must have the same dimension 100 vs 6
Abort trap: 6
(base) haddad_s@becane:~/dev/openturns/build$ ctest -R LinearModelAlgorithm_std
Test project /Users/haddad_s/dev/openturns/build
    Start  451: cppcheck_GeneralLinearModelAlgorithm_std
1/6 Test  #451: cppcheck_GeneralLinearModelAlgorithm_std ..............   Passed    0.12 sec
    Start  453: cppcheck_LinearModelAlgorithm_std
2/6 Test  #453: cppcheck_LinearModelAlgorithm_std .....................   Passed    0.06 sec
    Start  457: cppcheck_GeneralLinearModelAlgorithm_std_hmat
3/6 Test  #457: cppcheck_GeneralLinearModelAlgorithm_std_hmat .........   Passed    0.07 sec
    Start 1001: pyinstallcheck_GeneralLinearModelAlgorithm_std
4/6 Test #1001: pyinstallcheck_GeneralLinearModelAlgorithm_std ........   Passed    0.84 sec
    Start 1003: pyinstallcheck_GeneralLinearModelAlgorithm_std_hmat
5/6 Test #1003: pyinstallcheck_GeneralLinearModelAlgorithm_std_hmat ...   Passed    0.57 sec
    Start 1008: pyinstallcheck_LinearModelAlgorithm_std
6/6 Test #1008: pyinstallcheck_LinearModelAlgorithm_std ...............   Passed    0.58 sec

100% tests passed, 0 tests failed out of 6
```
It strange because it fails but pass through the `ctest`

jschueller: the test can still segfault if the diff is ok as ctest sees the return code of diff

sofianehaddad: Nothing except the error message
OK Understand why this generates troubles
I will drop this part and add try/catch

josephmure: Wow. Epic fail on my part! Have you figured out why the CI passes anyway?

sofianehaddad: Don't know. I added try/catch but no changes.
By the way I drop this block

josephmure: Are you replacing it with something else? Otherwise, I think it would be better to fix it by taking only the first 6 elements of `leverages_reference`.

PR #1816, opened 2021-04-14
Github Actions to check osx config

sofianehaddad:
Travis is limited to 50 mn and build + cppcheck seem to require more
Thus this last one will be ignored or used for other specific check

Now we will use this continuous integration platform to check osx
configuration, including both cxx/py tests (without doc checking)

Here we are trying to use all dependencies. For the moment:
 - dlib (from brew) has compilation issues (bad cxx flags/includes)
 - bonmin is not used as there is no brew formula and requires some work
 to integrate it.


.github/workflows/build.yml
```diff
@@ -0,0 +1,25 @@
+on: push
+jobs:
+  build:
+    runs-on: macos-latest
+    env:
+      LDFLAGS: -L/usr/local/opt/openblas/lib
+      CPPFLAGS: -I/usr/local/opt/openblas/include
```
jschueller, line 7: you can set MAKEFLAGS onc in env instead of calling sysctl in each command to set the number of jobs

jschueller: maybe then repush to retrigger as I've just enabled actions in PRs from forks in openturns settings

sofianehaddad: indeed

jschueller: or maybe the yml must already be present, like what we had for the sphinx artifacts

jschueller:
you can drop the travis config file then

.github/workflows/build.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - run: brew upgrade
+    - run: brew install openblas swig boost python3 tbb nlopt cminpack ceres-solver bison flex hdf5 ipopt r
```
jschueller, line 11: did you try mpfr/mpc ?

.github/workflows/build.yml
```diff
@@ -0,0 +1,25 @@
...
+    - run: cd $HOME/work/openturns && git clone https://github.com/yixuan/spectra.git spectra
+    - run: cd $HOME/work/openturns/spectra && git checkout v0.9.0
+    - run: cd $HOME/work/openturns/spectra && cmake -DBUILD_TESTS=TRUE -DCMAKE_INSTALL_PREFIX=$HOME/ .
+    - run: cd $HOME/work/openturns/spectra && make all && make install
```
jschueller, line 15: with cmake you can do "make install" without need for "make all"

.github/workflows/build.yml
sofianehaddad: Have a look at https://github.com/sofianehaddad/openturns/runs/2341501238?check_suite_focus=true
Both `MPFR`, `MPC` are available

jschueller: ah ok

jschueller:
... and the travis link in the README

.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+    - run: cd $HOME/work/openturns/spectra && git checkout v0.9.0
+    - run: cd $HOME/work/openturns/spectra && cmake -DBUILD_TESTS=TRUE -DCMAKE_INSTALL_PREFIX=$HOME/ .
+    - run: cd $HOME/work/openturns/spectra && make install ${MAKEFLAGS}
+    - run: cd $HOME/work/openturns && git clone https://github.com/jeromerobert/hmat-oss.git hmat-oss
+    - run: cd $HOME/work/openturns/hmat-oss/ && git checkout 1.6.1
+    - run: cd $HOME/work/openturns/hmat-oss/ && cmake -DBUILD_EXAMPLES=OFF -DCBLAS_INCLUDE_DIR=/usr/local/opt/openblas/include -DCMAKE_MACOSX_RPATH=ON .
+    - run: cd $HOME/work/openturns/hmat-oss/ && make install ${MAKEFLAGS}
+    - run: pip3 install matplotlib scipy chaospy pandas dill
+    - run: cmake -DR_EXECUTABLE="dummyR" -DPYTHON_EXECUTABLE=/usr/local/bin/python3 -DFLEX_EXECUTABLE=/usr/local/opt/flex/bin/flex -DBISON_EXECUTABLE=/usr/local/opt/bison/bin/bison -DSPECTRA_INCLUDE_DIR=/Users/runner/include .
+    - run: make ${MAKEFLAGS}
```
jschueller, line 23: no need to specify ${MAKEFLAGS} for make commands

README.rst
```diff
@@ -1,9 +1,6 @@
 .. image:: https://circleci.com/gh/openturns/openturns.svg?style=shield
     :target: https://circleci.com/gh/openturns/openturns
 
-.. image:: https://travis-ci.org/openturns/openturns.svg?branch=master
-    :target: https://travis-ci.org/openturns/openturns
-
```
jschueller, line 6: also we mention travis in the dev doc

.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+      LDFLAGS: -L/usr/local/opt/openblas/lib
+      CPPFLAGS: -I/usr/local/opt/openblas/include
+      MAKEFLAGS: "-j $(sysctl -n hw.logicalcpu)"
+    steps:
+    - uses: actions/checkout@v2
+    - run: brew upgrade
+    - run: brew install openblas swig boost python3 tbb nlopt cminpack ceres-solver bison flex hdf5 ipopt r
+    - run: cd $HOME/work/openturns && git clone https://github.com/yixuan/spectra.git spectra
+    - run: cd $HOME/work/openturns/spectra && git checkout v0.9.0
+    - run: cd $HOME/work/openturns/spectra && cmake -DBUILD_TESTS=TRUE -DCMAKE_INSTALL_PREFIX=$HOME/ .
```
jschueller, line 15: seems you can write a long command in a single run instruction, with | and \ to skip lines if needed

.github/workflows/build.yml
```diff
@@ -0,0 +1,31 @@
+on: push
```
jschueller, line 1: seems it must be explicitely enabled for pull requests:
"pull_request:"
at least that's what I see in hmat

jschueller:
I think you should enable unity builds to speed up the compilation, see run_docker_linux

.github/workflows/build.yml
sofianehaddad: yes but we might have both (`pull_request` +` push`)
However I am wondering if there is some jobs do not run concurrently in that case. I can add specific push on `master`

.github/workflows/build.yml
sofianehaddad: Done

README.rst
sofianehaddad: Done

.github/workflows/build.yml
sofianehaddad: seem that `makeflags` is generating some troubles

sofianehaddad:
@jschueller Some cosmetics are needed but the main important is that some tests are failing.
Any idea on how to solve https://github.com/openturns/openturns/runs/2352281215?check_suite_focus=true#step:12:478 and https://github.com/openturns/openturns/runs/2352281215?check_suite_focus=true#step:12:1655

I can not reproduce on my `osx` laptop

jschueller:
the generalizedpareto can be skipped (ctest -E) now and debugged later
cppcheck_EmpiricalBernsteinCopula_std can be fixed (only a digit change),

jschueller:
it was "ctest -E GeneralizedParetoFactory_std", my bad

jschueller:
cxx tests take forever: 1h for 50%

sofianehaddad:
strange. Platform sometimes hangs and I do not know why

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+           make install -j$(sysctl -n hw.logicalcpu)
+    - run:  |
+        cmake -DR_EXECUTABLE="dummyR" -DPYTHON_EXECUTABLE=/usr/local/bin/python3 \
+        -DFLEX_EXECUTABLE=/usr/local/opt/flex/bin/flex \
+        -DBISON_EXECUTABLE=/usr/local/opt/bison/bin/bison \
+        -DSPECTRA_INCLUDE_DIR=/Users/runner/include \
+        -DCMAKE_UNITY_BUILD=ON -DCMAKE_UNITY_BUILD_BATCH_SIZE=32 .
+    - run: make install -j$(sysctl -n hw.logicalcpu)
+    - run: make tests -j$(sysctl -n hw.logicalcpu)
+    - run: ctest -E GeneralizedParetoFactory_std --output-on-failure --timeout 100
```
jschueller, line 36: the tests dont run in parallel anymore ?

sofianehaddad: indeed
I will fix it

sofianehaddad:
@jschueller seem that all troubles have been fixed

jschueller (approved):

jschueller:
i'll let it run one last time and merge it

sofianehaddad:
ok. Maybe we can try later to seem how it works with `ubuntu` platform?
For info, last `osx` run (without doc of course) took 1h (install of dep + compilation of spectra + hmat + openturns + ctest)
It might be interesting to benchmark for `ubuntu`

jschueller:
why not

PR #1819, opened 2021-04-15
LinearModelAlgorithm: Allow null DOF

jschueller:
Allows cases whese basis size == sample size, (not just sample size > basis size).
In that case, we throw when user asks for the noise distribution which is not defined, also for the R^2

### DIFF
--- a/lib/src/Uncertainty/StatTests/LinearModelTest.cxx
+++ b/lib/src/Uncertainty/StatTests/LinearModelTest.cxx
@@ -58,7 +58,9 @@ TestResult LinearModelTest::LinearModelFisher(const Sample & firstSample,
   if (size < 3) throw InvalidArgumentException(HERE) << "Error: sample too small. Sample should contains at least 3 elements";
   // As we rely on a linear model result, we should be very generic
   // Instead of using input dimension, one should use parameter size
-  const UnsignedInteger df = linearModelResult.getDegreesOfFreedom();
+  const SignedInteger dof = linearModelResult.getDegreesOfFreedom();
+  if (dof <= 0)
+    throw InvalidArgumentException(HERE) << "Cannot perform linear model test when DOF is null";
 
   // Regression coefficient
   const Function fHat(linearModelResult.getMetaModel());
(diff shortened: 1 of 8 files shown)


lib/src/Uncertainty/StatTests/LinearModelTest.cxx
```diff
@@ -58,7 +58,9 @@ TestResult LinearModelTest::LinearModelFisher(const Sample & firstSample,
   if (size < 3) throw InvalidArgumentException(HERE) << "Error: sample too small. Sample should contains at least 3 elements";
   // As we rely on a linear model result, we should be very generic
   // Instead of using input dimension, one should use parameter size
-  const UnsignedInteger df = linearModelResult.getDegreesOfFreedom();
+  const SignedInteger dof = linearModelResult.getDegreesOfFreedom();
+  if (dof <= 0)
+    throw InvalidArgumentException(HERE) << "Cannot perform linear model test when DOF is null";
```
sofianehaddad, line 64:
```suggestion
  const UnsignedInteger p = linearModelResult_.getCoefficients().getSize();
```

lib/src/Uncertainty/StatTests/LinearModelTest.cxx
```diff
@@ -76,12 +78,12 @@ TestResult LinearModelTest::LinearModelFisher(const Sample & firstSample,
   // H0 : Beta_i = 0
   // H1 : Beta_i < 0 or Beta_i > 0
   // The statistics follows a Fisher distribution
-  const Scalar numerator = sumSquaredExplained / (size - df - 1);
-  const Scalar denomerator = sumSquaredResiduals / df;
+  const Scalar numerator = sumSquaredExplained / (size - dof - 1);
```
sofianehaddad, line 81:
```suggestion
  const Scalar numerator = sumSquaredExplained / (p - 1);
```

sofianehaddad: It is difficult to read. Usually we have `SSM/(p-1)` in literature

lib/src/Uncertainty/StatTests/LinearModelTest.cxx
```diff
@@ -76,12 +78,12 @@ TestResult LinearModelTest::LinearModelFisher(const Sample & firstSample,
...
+  const Scalar denomerator = sumSquaredResiduals / dof;
 
   const Scalar statistic = numerator / denomerator;
   Log::Debug(OSS() << "F-statistic = " << statistic);
-  const Scalar pValue =  FisherSnedecor(size - df - 1, df).computeComplementaryCDF(statistic);
+  const Scalar pValue =  FisherSnedecor(size - dof - 1, dof).computeComplementaryCDF(statistic);
```
sofianehaddad, line 86:
```suggestion
  const Scalar pValue =  FisherSnedecor(p - 1, dof).computeComplementaryCDF(statistic);
```

sofianehaddad (commented):
Only minor cosmetics for easy reading/debugging
Thanks to this I catched a small bug in `LinearModelAnalysis` that I will fix once this PR is merged

lib/src/Uncertainty/StatTests/LinearModelTest.cxx
jschueller: here you want me to rewrite code I didnt touch, that's for another PR

sofianehaddad: ok I will do the job

sofianehaddad (approved):

sofianehaddad:
Could we merge this?

Issue #1818, opened 2021-04-15
Problem with wilksNumber 


Jean-A:
This code

````
import openturns as ot
import numpy as np

sample = ot.Normal(0,10).getSample(10000)
alpha = 0.995
confidence_level = np.linspace(0.001,0.999, 10000)
plt.plot(confidence_level, [ot.Wilks.ComputeSampleSize(alpha, beta) for beta in confidence_level])
````
shows the evolution of wilksNumber according to the confidence level. The function is expected to be monotonous but it isn't.

![image](https://user-images.githubusercontent.com/52526981/114847146-ae262200-9ddd-11eb-8adb-9467ae193900.png)

regislebrun:
I fixed it, now I get:
![Wilks](https://user-images.githubusercontent.com/8220933/114849745-5806ae00-9de0-11eb-835d-1e0d63d6d93c.png)

PR #1821, opened 2021-04-17
Fixed Wilks

regislebrun:
The initializtion of the sequential search for the sample size was done using an asymptotic approximation not valid for low quantile levels. It fixes issue #1818.


lib/src/Uncertainty/Algorithm/Simulation/Wilks.cxx
```diff
@@ -86,6 +96,7 @@ Point Wilks::computeQuantileBound(const Scalar quantileLevel,
 {
   // Compute the needed sample size
   const UnsignedInteger size = ComputeSampleSize(quantileLevel, confidenceLevel, marginIndex);
+  std::cerr << "quantileLevel=" << quantileLevel << ", confidenceLevel=" << confidenceLevel << ", marginIndex=" << marginIndex << ", size=" << size << std::endl;
```
jschueller, line 99: drop the cerr

regislebrun: I am rewriting the code, I found a numerical instability thanks to the plot_quantile_estimation_wilks.py example.

jschueller: ah ok, sorry

lib/src/Uncertainty/Algorithm/Simulation/Wilks.cxx
```diff
@@ -45,38 +48,120 @@ Wilks::Wilks(const RandomVector & vector)
...
+  {
+    return 1;
+  }
+
+  UnsignedInteger getOutputDimension() const
+  {
+    return 1;
+  }
+
+  Description getInputDescription() const
```
jschueller, line 86: description and str converter methods are useless if the functions is local

python/test/t_Wilks_std.py
```diff
@@ -5,20 +5,11 @@
...
-    import sys
-    print("t_Wilks_std.py", sys.exc_info()[0], sys.exc_info()[1])
+alpha = [0.05 * i for i in range(1, 20)]
+beta = [0.05 * i for i in range(1, 20)]
+for a in alpha:
+    for b in beta:
+        print("alpha=%.2f" % a, "beta=%.2f" % b)
+        for i in range(5):
+            print("n(%.2f" % a, ",%.2f" %b, ",", i, ")=",
+                  Wilks.ComputeSampleSize(a, b, i))
```
jschueller, line 15: seems it fails on win32

regislebrun: They were there for debugging purpose. They helped me during the development, maybe they would ease to catch future bugs ;-)!

jschueller: I mean the code of the test is good but they fail to pass on win32

PR #1823, opened 2021-04-19
 master to use actions

sofianehaddad:


jschueller:
I think you have to keep these swig flags and add wno-unused-functions if you still want to use Werror

sofianehaddad:
The problem is that compilation fails with these falgs (see [actions](https://github.com/openturns/openturns/runs/2383157709?check_suite_focus=true) for more details)
However no time for the moment to investigate the reason. Note that these flags are used with `circleci`

jschueller:
yes, because Werror enable additional errors in clang wrt gcc, and Wunused-functions was the one you needed to disable
if you dont use Werror, no need to disable any warnings

jschueller (approved):

Issue #1801, opened 2021-03-21
The HDF5 format is undocumented


mbaudin47:
The `XMLH5StorageManager` class was created, but no example allows to use it. @YomGarcia please create an API example both in `XMLH5StorageManager` and in its actual use-case, i.e. the `Study` class:

http://openturns.github.io/openturns/master/user_manual/_generated/openturns.Study.html

and

http://openturns.github.io/openturns/master/auto_numerical_methods/general_methods/plot_study_save_load.html

PR #1824, opened 2021-04-19
Storage manager

YomGarcia:
Minor fixes to StorageManagers:
- removed unused BinaryStorageManager
- added check on buffer size to allocate size-appropriate dataspace
- added xml attribute associated to the type of StorageManager used to generate the xml


jschueller (approved):

YomGarcia:
closes #1801

python/doc/examples/numerical_methods/general_methods/plot_study_save_load.py
```diff
@@ -71,6 +71,7 @@
 #
 # - its name: it is useful to give names to the objects we want to save. If no name has been given by the user, we can use the name by default. The name of each object saved can be checked in the file.XML created or by printing the study in the python interface (with the command print).
 # - its id number : this id number is unique for each object. It is useful to distinguish objects of same type which names are identical, equal to the default name given ('Unnamed'). This id number may be checked by printing the study loaded in the python interface (with the command print): be careful, this print operation must be performed after having loaded the study (the id number may be different from the one indicated in the file.XML associated to the study).
+# - for HDF5 storage : the id is used to name the HDF5 dataset. This way, the various floating point/integer collections are stored and accessed through the name of their respective datasets.
```
josephmure, line 74:
```suggestion
# - for HDF5 storage: the id is used to name the HDF5 dataset. This way, the various floating point/integer collections are stored and accessed through the name of their respective datasets.
```
Does this mean that the id serves as both id and name, or that the HDF5 dataset has no name?

python/doc/examples/numerical_methods/general_methods/plot_study_save_load.py
```diff
@@ -81,6 +82,12 @@
 fileName = 'study.xml'
 study.setStorageManager(ot.XMLStorageManager(fileName))
 
+# %%
+# Alternatively, large amount of data can be stored in binary HDF5 file.; XML file ('study_h5.xml') serves as header for binary data (stored in 'study_h5.h5' which is automatically created).
```
josephmure, line 86:
```suggestion
# Alternatively, large amounts of data can be stored in binary HDF5 file. An XML file (`study_h5.xml`) serves as header for binary data, which are stored in the automatically created `study_h5.h5` file.
```

josephmure (commented):
This PR is a good opportunity to improve the wording of the *Save/load a study* example. If you agree with my suggestions, you can commit them directly from the GitHub interface.

YomGarcia:
> This PR is a good opportunity to improve the wording of the _Save/load a study_ example. If you agree with my suggestions, you can commit them directly from the GitHub interface.

Sure your changes are more than welcome

YomGarcia:
I Included comments from Joseph and added an attribute to the xml file to check wether the appropriate StorageManager is being used when loading a Study. If the attribute is missing (backward compatibility issue), it is added on the fly when the xml is read

python/doc/examples/numerical_methods/general_methods/plot_study_save_load.py
josephmure: So does the HDF5 dataset have no name?

YomGarcia: The dataset gets named after the id, as shown in the code snippet below. id role in the xml remains unchanged
```  
<object class="Point" id="9" size="1" StudyVisible="false" StudyLabel="">
    <string>/Study_0.h5:/9</string>
  </object>```

Comment (YomGarcia) [reply python/doc/examples/numerical_methods/general_methods/plot_study_save_load.py]: I updated this part to better reflect this

Comment (josephmure) [reply python/doc/examples/numerical_methods/general_methods/plot_study_save_load.py]: Thanks for your explanation.

File: python/doc/examples/numerical_methods/general_methods/plot_study_save_load.py [hunk 3]
```diff
@@ -17,7 +17,7 @@
 Save/load a study
 =================
 """
-# %% 
+# %%
```
josephmure, line 21:
```suggestion
```
This empty cell can be removed.

josephmure (approved):

jschueller:
@sofianehaddad you'll probably want to look into it even if we tested this already in persalys

regislebrun (approved):

regislebrun (approved):
The parenthesis of the if condition is so tightly packed to the keyword that it makes it looks like a function. A unique space would be appreciated to conform to the other parts of the code.

josephmure (changes requested):
Upon further reflexion, I don't think this PR can close #1801 unless you also create an example in the API doc page:

https://github.com/openturns/openturns/blob/f1fce95b1485bd48a246a8f1d41351e4aa957891/python/src/Study_doc.i.in#L28
Here is the current output:

http://openturns.github.io/openturns/1.16/user_manual/_generated/openturns.Study.html

YomGarcia:
> The parenthesis of the if condition is so tightly packed to the keyword that it makes it looks like a function. A unique space would be appreciated to conform to the other parts of the code.

It is a convention I tend to forget about, I have put some breathing room around those parenthesis

YomGarcia:
> Upon further reflexion, I don't think this PR can close #1801 unless you also create an example in the API doc page:
> 
> https://github.com/openturns/openturns/blob/f1fce95b1485bd48a246a8f1d41351e4aa957891/python/src/Study_doc.i.in#L28
> 
> Here is the current output:
> 
> http://openturns.github.io/openturns/1.16/user_manual/_generated/openturns.Study.html

I wanted to avoid duplicate examples. Apart from setting a different instance of StorageManager, the rest is transparent to the user as far a reading and writing study files go. That is why I just added these lines (during the initial hdf5 PR):

https://github.com/openturns/openturns/blob/f1fce95b1485bd48a246a8f1d41351e4aa957891/python/src/Study_doc.i.in#L39

josephmure:
Makes sense to me. What do you think @mbaudin47?

josephmure:
@jschueller @YomGarcia I now run into an issue with `plot_study_save_load.py`. Curiously, this does not seem to happen to CI builds. Perhaps it depends on the installed dependencies?

```
WARNING: python/src/sphinx_build/examples/numerical_methods/general_methods/plot_study_save_load.py failed to execute correctly: Traceback (most recent call last):
  File "python/src/sphinx_build/examples/numerical_methods/general_methods/plot_study_save_load.py", line 87, in <module>
    study_h5.setStorageManager(ot.XMLH5StorageManager(fileName_h5))
TypeError: XMLH5StorageManager() takes no arguments
```

jschueller:
yes, it fails without hdf5, I'm fixing it

josephmure:
I still get the error, even after your fix.

jschueller:
strange, it fixed the error for me

josephmure:
Yes I did.

jschueller:
do other hdf5 tests pass ?

josephmure:
How can I see that?

josephmure:
Here is the output of the CMake command:
```
devel@OT-VM:~/GIT/openturns/build-master-doc$ cmake .. -DCMAKE_INSTALL_PREFIX=$PWD/install -DPYTHON_EXECUTABLE=/usr/bin/python3 -DSPHINX_FLAGS="-j6"
-- A library with LAPACK API found.
-- Could NOT find Spectra (missing: SPECTRA_INCLUDE_DIRS) 
-- Found HMAT: /usr/local/lib/cmake/hmat
-- Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS CXX) (found version "")
-- Boost version: 1.67.0
-- Could NOT find MPFR (missing: MPFR_LIBRARY MPFR_INCLUDE_DIRS) 
-- Could NOT find MPC (missing: MPC_LIBRARY MPC_INCLUDE_DIRS) 
-- nlopt::GN_AGS: fails to build
-- Found dlib: /usr/lib/x86_64-linux-gnu/cmake/dlib (found version 19.10.0)
-- Found installed version of Eigen: /usr/lib/cmake/eigen3
-- Found required Ceres dependency: Eigen version 3.3.7 in /usr/include/eigen3
-- Found required Ceres dependency: glog
-- Found installed version of gflags: /usr/lib/x86_64-linux-gnu/cmake/gflags
-- Detected gflags version: 2.2.2
-- Found required Ceres dependency: gflags
-- Found Ceres version: 1.14.0 installed in: /usr with components: [EigenSparse, SparseLinearAlgebraLibrary, LAPACK, SuiteSparse, CXSparse, SchurSpecializations, OpenMP, Multithreading]
-- Could NOT find Bonmin (missing: COIN_BONMIN_LIBRARY COIN_INCLUDE_DIR) 
-- Could NOT find Ipopt (missing: IPOPT_LIBRARIES IPOPT_INCLUDE_DIRS) 
-- French locale found
-- std::regex found
-- Could NOT find chaospy (missing: CHAOSPY_LOCATION _CHAOSPY_VERSION_MATCH) 
-- Could NOT find dill (missing: DILL_LOCATION _DILL_VERSION_MATCH) 
-- The following OPTIONAL packages have been found:

 * BISON (required version >= 2.4)
 * FLEX
 * TBB
 * HMAT (required version >= 1.6)
 * LibXml2
 * Boost (required version >= 1.46)
 * NLopt
 * dlib
 * Ceres
 * CMinpack
 * OpenMP
 * SWIG (required version >= 3)
 * PythonInterp
 * PythonLibs (required version == 3.7.3)

-- The following REQUIRED packages have been found:

 * BLAS
 * LAPACK
 * Threads

-- The following OPTIONAL packages have not been found:

 * Spectra
 * HDF5
 * MPFR
 * MPC
 * Bonmin
 * Ipopt

-- Configuring done
-- Generating done
```

josephmure:
After that, the compilation of the doc fails with the error I reported.

jschueller:
ah, I fixed only the test, not the doc

PR #1825, opened 2021-04-20
Misc

jschueller:
last fixes before the release! (hopefully)
- RandomMixture setParameter fix cc @adumasphi 
- MemoizeEvaluation compact cache serialization cc @YomGarcia


sofianehaddad (approved):

python/test/t_RandomMixture_std.py
```diff
@@ -329,15 +309,20 @@
...
     # For ticket 1129
-    dist = RandomMixture([Uniform()]*200)
+    dist = ot.RandomMixture([ot.Uniform()]*200)
     print("CDF(0)=%.5g" % dist.computeCDF([0]))
-except:
-    import sys
-    print("t_RandomMixture_std.py", sys.exc_info()[0], sys.exc_info()[1])
+
+# check parameter accessors
+dist = ot.Gumbel() + ot.Normal(0, 0.1)
```
josephmure, line 323: Are there examples of the construction with the overloaded `+` in the doc? It's great!

josephmure (approved):

python/test/t_RandomMixture_std.py
jschueller: there is no dedicated example, but 2 mention it, look for "arithmetic" in python/doc/examples

josephmure: Thanks! Maybe we could also add an example to the [RandomMixture](https://openturns.github.io/openturns/1.16/user_manual/_generated/openturns.RandomMixture.html) API page.

lib/src/Uncertainty/Distribution/RandomMixture.cxx
```diff
@@ -2870,6 +2870,15 @@ void RandomMixture::setParameter(const Point & parameter)
 {
   if (parameter.getSize() != getParameter().getSize()) throw InvalidArgumentException(HERE) << "Error: expected " << getParameter().getSize() << " values, got " << parameter.getSize();
   const Scalar w = getWeight();
+  const UnsignedInteger size = distributionCollection_.getSize();
```
regislebrun, line 2873: Beware of the simplification mechanism! If you write `Y = Uniform(1.0, 2.0) + Uniform(2.0, 4.0)` it becomes `Y = Triangular(4.0, 5.0, 6.0)` so you have 3 parameters instead of the 4 initial ones. This point can be mentioned in the documentation instead of trying to catch it in the code.

regislebrun (approved):
Be careful with the simplification mechanism of RandomMixture. The number of parameters may not be what the user would have expected when he created the RandomMixture instance.

lib/src/Uncertainty/Distribution/RandomMixture.cxx
jschueller: I can use setDistributionCollectionAndWeights with simplify=false so that no further simplification is done

PR #1831, opened 2021-04-27
t_KrigingAlgorithm_std_hmat.py runs tests

josephmure:
The file `t_KrigingAlgorithm_std_hmat.py` runs no tests in the current 1.17 and master branches: all it does is define 2 functions which are never run.

### DIFF
--- a/python/test/t_KrigingAlgorithm_std_hmat.py
+++ b/python/test/t_KrigingAlgorithm_std_hmat.py
@@ -1,14 +1,12 @@
 #! /usr/bin/env python
 
-from __future__ import print_function
-from openturns import *
-from openturns.testing import *
-from math import sqrt
+import openturns as ot
+import openturns.testing as ott
 
-TESTPREAMBLE()
+ot.TESTPREAMBLE()
 
 # Set hmat
-ResourceMap.Set("KrigingAlgorithm-LinearAlgebra", "HMAT")
+ot.ResourceMap.SetAsString("KrigingAlgorithm-LinearAlgebra", "HMAT")
 
 # Test 1
 # Test 1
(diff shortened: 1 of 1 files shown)


josephmure:
@jschueller @KieranDelamotte Any ideas?

sofianehaddad:
My bad seems is due to [this](https://github.com/openturns/openturns/commit/9edcef69148540518799567654aa451257ca59d6#diff-5db5a828ad74ae3abd5355db1404219b12acc752a3f1fda909bd446c08722c1c).

I just forgot the lines you added. Thanks @josephmure for the fix

josephmure:
I do not understand why the tests from [t_KrigingAlgorithm_std_hmat.py](https://github.com/josephmure/openturns/blob/ac7cc3f1497282af87af620bd77ad4bd9ac79f96/python/test/t_KrigingAlgorithm_std_hmat.py) fail in the first commit. They look identical to their equivalents from [t_KrigingAlgorithm_std_hmat.cxx](https://github.com/josephmure/openturns/blob/ac7cc3f1497282af87af620bd77ad4bd9ac79f96/lib/test/t_KrigingAlgorithm_std_hmat.cxx) which pass.

sofianehaddad:
Probably because in hmat,the regularization of the covariance matrix is more important. The spectrum is shifted using `epsilon * lambda_max`, `lambda_max` being the maximal eigenvalue of the discretized cov function.
Thus we loose a bit the interpolation property...

josephmure:
But why do C++ tests pass? Unless I missed something, they are identical.

sofianehaddad:
use cases yes, requirements for validation no
The script you fixed is a copy/paste from `lapack` implementation one

regislebrun (approved):

PR #1834, opened 2021-05-03
Release fixes

jschueller:


lib/test/CMakeLists.txt
```diff
@@ -678,7 +678,9 @@ endif ()
 ot_check_test (ProbabilitySimulationAlgorithm_sensitivity)
 ot_check_test (PostAnalyticalControlledImportanceSampling_std)
 ot_check_test (PostAnalyticalImportanceSampling_std)
+if (Boost_FOUND)
 ot_check_test (Wilks_std)
```
regislebrun, line 682: If boost is not present, OT should switch to its own algorithms for pBeta. Is it a numerical problem, e.g. a discrepancy in the results, or an algorithmic problem (infinite loop) or something else?

jschueller: yes, it throws in brent

jschueller: I opened an issue for this and removed this commit

regislebrun:
@jschueller Unfortunately it does not solve #1837: the timings remain the same. IMO it is due to the special path taken by pickle to store collections and the like: they are first converted into Python equivalents, then stored as Python objects if I understood correctly what appends in the SWIG layer.

jschueller:
weird, on my side I have some nice improvements:
```
Pickle save t=5.79 s
Pickle size= 276924791
Pickle load t=2.02 s
--------------------------------------------------
XML     no compression save t=98.9 s
XML     size= 2585260699
XML     no compression load t=226 s
--------------------------------------------------
XML    max compression save t=275 s
XML     size= 456588234
XML    max compression load t=417 s
--------------------------------------------------
XML/H5  no compression save t=36.8 s
XML     size= 14321
H5     size= 276909192
XML/H5  no compression load t=37.7 s
--------------------------------------------------
XML/H5 max compression save t=5.22 s
XML     size= 1607
H5     size= 276909192
XML/H5 max compression load t=6.32 s
```

Issue #1836, opened 2021-05-05
Hypergeometric results differ without boost


jschueller:
fyi, the hypergeometric special function boost fallback methods provides somewhat different results and the test is currently disabled:
```
 Continuous =  False
 oneRealization= [2]
 oneSample first= [3]  last= [3]
-mean= [3.956]
-covariance= [[ 1.3246 ]]
+mean= [3.9551]
+covariance= [[ 1.32262 ]]
 Chi2 test for the generator, sample size= 100  is accepted
 Chi2 test for the generator, sample size= 1000  is accepted
 Point=  [5]
-log pdf= -1.47376350877
-pdf     = 0.229061784897
-pdf (FD)= 0.229061784897
-cdf= 0.92414187643
-ccdf= 0.0758581235698
-characteristic function= (0.112691702824+0.314426247514j)
-log characteristic function= (-1.09658163526+1.22665413395j)
-generating function= (-0.0767726758078-0.123723016677j)
-log generating function= (-1.92684335624-2.12616798622j)
-quantile= [6]
-cdf(quantile)= 0.992860411899
-entropy=1.546540
+log pdf= -1.47422634283
+pdf     = 0.228955791832
+pdf (FD)= 0.0835723288137
+cdf= 0.992118636512
+ccdf= 0.00788136348777
+characteristic function= (0.112617215751+0.314218417743j)
+log characteristic function= (-1.09724283479+1.22665413395j)
+generating function= (-0.0767219305288-0.123641238116j)
+log generating function= (-1.92750455577-2.12616798622j)
+quantile= [5]
+cdf(quantile)= 0.992118636512
+entropy=1.546179
 mean= [3.92]
 standard deviation= [1.13737]
 skewness= [-0.0504599]
@@ -28,9 +28,9 @@
 covariance= [[ 1.2936 ]]
 parameters= [[n : 25, k : 7, m : 14]]
 standard moment n= 0 , value= [1]
-standard moment n= 1 , value= [3.92]
-standard moment n= 2 , value= [16.66]
-standard moment n= 3 , value= [75.3748]
-standard moment n= 4 , value= [358.978]
-standard moment n= 5 , value= [1785.54]
+standard moment n= 1 , value= [3.91741]
+standard moment n= 2 , value= [16.649]
+standard moment n= 3 , value= [75.325]
+standard moment n= 4 , value= [358.741]
+standard moment n= 5 , value= [1784.36]
[...]

PR #1838, opened 2021-05-05
Minor fixes

regislebrun:
In this PR I propose a fix for issue #1836, and I improve the initial scaling in the regularization loop of GLMAlgo & GaussianProcess in the dense case (it was already done in the compressed case). It allows to save costly regularization loops. I also fixed a bug in the try/catch clauses where the type of exception was too narrow.


ChangeLog
```diff
@@ -110,6 +110,7 @@
  * #1815 (1.16 fails with dlib-cpp-19.22)
  * #1818 (Problem with wilksNumber)
  * #1820 (Incoherent results in LinearModelAnalysis)
+ * #1836 (Hypergeometric results differ without boost)
```
jschueller, line 113: the hypergeometric tests should be re-enabled in the CmakeLists files then

jschueller (approved):

jschueller (commented):
oh, can you retarget the PR to the 1.17 branch instead of master ?
(I believe you have to open another one for this)

Issue #1843, opened 2021-05-16
Ali-Mikhail-Haq copula parameter value


MLasserre:
When I define an Ali-Mikhail-Haq copula with theta=1:
```
import openturns as ot  
amh_copula = ot.AliMikhailHaqCopula(1)
```
I have the following error:
`
InvalidArgumentException : Theta MUST be in (-1, 1), here theta=1
`
According to documentation it is 1 by default and theta is in [-1;1].

MLasserre:
I just checked the file `AliMikhailHaqCopula.cxx` and it is set to 0.5 by default.

MLasserre:
There is also the same problem with the Farlie-Gumbel-Morgenstern copula.
The parameter theta is said (according to documentation) to be set to 1 by default while it is set to 0.5.

regislebrun:
@MLasserre You are right. I checked all the copula and found other discrepancies between the documentation and the code:
  * in EmpiricalBernsteinCopula, the default sample is *[[0.0],[0.0]]* and not *[[0.0]]*
  * in FarlieGumbelMorgensternCopula, the default theta is 0.5 and not 1
  * in NormalCopula, the default dimension is 2 (was not in the doc)

Issue #1844, opened 2021-05-25
MaximumDistribution::computePDF is wrong when all the marginals are equal and independent


regislebrun:
The following script should output the same value two times:
```
from openturns import *
from math import log

d_max = MaximumDistribution(Gumbel(1.0, 0.0), 10)
d_ref = Gumbel(1.0, log(10.0))
print("pdf_max=", d_max.computePDF(3.0))
print("pdf_ref=", d_ref.computePDF(3.0))
```
but the output is:
```
pdf_max= 0.03026175170960397
pdf_ref= 0.30261751709603957
```
There is a missing factor of n in the code, where n is the number of iid distributions in the MaximumDistribution. The workaround is to write:
```
d_max = MaximumDistribution(ComposedDistribution([Gumbel(1.0, 0.0)]*10))
```

Issue #1845, opened 2021-05-25
MemoizeFunction does not propagate to finite difference gradient&hessian


regislebrun:
Let's assume you define a `Function` this way:
```
import openturns as ot

def py_function(x):
    return x

ot_function = ot.MemoizeFunction(ot.PythonFunction(1, 1, py_function))
```
then the `ot_function.gradient()` and `ot_function.hessian()` method don't benefit from the cache of the `MemoizeFunction` evaluation. A workaround is to write:
```
ot_function = ot.PythonFunction(1, 1, py_function)
ot_function = ot.MemoizeFunction(ot.Function(ot.MemoizeEvaluation(ot_function.getEvaluation())))
```
but it is cumbersome...

PR #1846, opened 2021-05-25
Minor fixes

regislebrun:
This PR fixes issues #1843 (wrong default value for theta in AliMikhailHaqCopula documentation) and issue #1844 (bug in MaximumDistribution::computePDF()).
It also provides a fix for other discrepancy between the documentation of default copula parameters and the actual code.
Two evolutions are proposed for discussion:
  * the MemoizeFunction now propagates the MemoizeEvaluation through gradient and hessian if they are computed using finite differences
  * the ParametricFunction now computes the gradient() and parameterGradient() methods using only the needed part of the full gradient

The implementation is open to discussion because it introduces the special case of finite differences in several classes, which is the sign of a bad design. Nevertheless, the gain in terms of number of evaluations is worth considering.


mbaudin47:
This should fix #1847.

ChangeLog
```diff
@@ -20,6 +20,9 @@
 === Miscellaneous ===
 
 === Bug fixes ===
+ * #1843 (Ali-Mikhail-Haq copula parameter value)
+ * #1844 (MaximumDistribution::computePDF is wrong when all the marginals are equal and independent)
```
jschueller, line 24: there's also #1845

python/test/t_ParametricFunction_std.py
```diff
@@ -37,3 +38,50 @@
...
+ot_f = ot.MemoizeFunction(ot.PythonFunction(3, 3, py_f))
+param_f = ot.ParametricFunction(ot_f, [0, 1], [1.0, 2.0])
+
+x = [3.0]
+y = [1.5]
+n_calls_0 = ot_f.getCallsNumber()
+par_grad = param_f.parameterGradient(x)
+n_calls_1 = ot_f.getCallsNumber()
+assert_almost_equal(n_calls_1 - n_calls_0, 4, 0.0, 0.0)
+assert_almost_equal(n_calls, 4, 0.0, 0.0)
```
jschueller, line 59: it makes no sense to use assert_almost_equal for ints, just use `assert n_calls == 4, "message"`

jschueller (approved):
the parametric & memoize changes are ok

sofianehaddad (approved):

Issue #1842, opened 2021-05-13
libOT.so.0.0.0 doesn't have a SONAME


yurivict:
```
Warning: /disk-samsung/freebsd-ports/math/openturns/work/stage/usr/local/lib/libOT.so.0.0.0 doesn't have a SONAME.
Warning: pkg(8) will not register it as being provided by the port.
Warning: If another port depend on it, pkg will not be able to know where it comes from.
```

Version 1.17
OS: FreeBSD 12.2

jschueller:
Does it go away with: ```cmake -DLIB_SOVERSION=0.17 -DLIB_VERSION=0.18.0``` ?

yurivict:
I didn't try this. Setting SONAME should be done inside of cmake scripts and shouldn't be user-visible.

jschueller:
could you test this so we can make it the default for the next release ?

yurivict:
It works, except it should be:
```
cmake -DLIB_SOVERSION=0 -DLIB_VERSION=0.18.0
```

ldconfig(8) manpage says:
```
Filenames must conform to the lib*.so.[0-9] pattern in order to be added to the hints file.
```
and therefore SONAME can only have one number in it.

yurivict:
I take back my previous comments.

No, ```-DLIB_SOVERSION=0.18 -DLIB_VERSION=0.18.0``` doesn't work.
It creates symlinks but doesn't set the SONAME.

You needs to add something like this:
```
SET_TARGET_PROPERTIES(
  libOT
  PROPERTIES
  VERSION 0.18.0
  SOVERSION 0.18
)
```

Multiple numbers in SOVERSION are ok.

PR #1849, opened 2021-06-01
Misc

jschueller:
- Fix for BlockIndependent::getStandardDeviation by overriding getShiftedMoment to avoid integration and incorect values in some cases
- #1842 
- Point == None crash
- Fix for MetaModelAlgorithm::BuildDistribution possibly returning invalid distributions (undefined mean with Frechet)


regislebrun (approved):
IMO we should spend time thinking about how MemoizeFunction and the history mechanism of the algorithms could interact in order to reduce the memory footprint. In penstock, all the functions are memoized, the net result is a massive memory consumption and a very significant performance impact when producing large output samples

lib/src/Base/Stat/GeneralizedExponential.cxx
```diff
@@ -78,6 +78,7 @@ Scalar GeneralizedExponential::computeAsScalar(const Point & tau) const
     tauOverThetaNorm += dx * dx;
   }
   tauOverThetaNorm = sqrt(tauOverThetaNorm);
+  const CovarianceMatrix outputCovariance(outputCovariance_);
   return tauOverThetaNorm <= SpecFunc::ScalarEpsilon ? outputCovariance_(0, 0) * (1.0 + nuggetFactor_) : outputCovariance_(0, 0) * exp(-pow(tauOverThetaNorm, p_));
```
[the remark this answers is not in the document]
jschueller: indeed!
this commit avoids a crash in the KLQuadrature example, and also unnecessary copyonwrite, because the compiler chose the non-const accessor (this attribute is mutable)

jschueller:
yes, that's why I'm removing when unnecessary in algorithms, it should be only used at top-level
I also suspect it's unneeded in calibration

lib/src/Base/Optim/openturns/BonminProblem.hxx
```diff
@@ -179,7 +178,8 @@ public:
 private:
   const OptimizationProblem optimProblem_;
   const Point startingPoint_;
-  MemoizeFunction objectiveFunction_;
+  Sample evaluationInputHistory_;
```
regislebrun, line 181: So if I understand well, the logic is to avoid the creation of MemoizeFunction in favor of algorithms histories? Is there a way to reuse the function history if the given objective function is a MemoizeFunction? For me, it is not clear if we have to automate things or to explain them in the documentation, but it can be a huge penalty to store everything in multiple places, both in terms of memory footprint and CPU performance

lib/src/Base/Stat/HMatrixImplementation.cxx
```diff
@@ -178,6 +178,14 @@ HMatrixClusterTree::~HMatrixClusterTree()
...
 }
 
+struct CDeleter
+{
+  void operator ()(void * p)
+  {
+    free(p);
+  }
+};
+
```
regislebrun, line 188: A kind of HMAT black magic?

lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
```diff
@@ -334,7 +333,7 @@ void GaussianNonLinearCalibration::run()
   const TriangularMatrix errorInverseCholesky(error.getInverseCholesky());
   const Point thetaStar(run(inputObservations_, outputObservations_, getCandidate(), parameterInverseCholesky, errorInverseCholesky));
   // Build the residual function this way to benefit from the automatic Hessian
-  const MemoizeFunction residualFunction(NonLinearLeastSquaresCalibration::BuildResidualFunction(model_, inputObservations_, outputObservations_));
+  const Function residualFunction(NonLinearLeastSquaresCalibration::BuildResidualFunction(model_, inputObservations_, outputObservations_));
```
regislebrun, line 336: So for this algorithm and the other calibration algorithms, it is the user's responsibility to activate a cache on his function. Does it allow to reuse the cache if the algorithm is run twice?

lib/src/Uncertainty/Distribution/BlockIndependentDistribution.cxx
```diff
@@ -671,6 +671,22 @@ BlockIndependentDistribution::InverseIsoProbabilisticTransformation BlockIndepen
...
+/* Get the shifted moments of the distribution */
+Point BlockIndependentDistribution::getShiftedMoment(const UnsignedInteger n, const Point & shift) const
+{
+  Point shiftedMoment;
+  UnsignedInteger index = 0;
+  for (UnsignedInteger i = 0; i < distributionCollection_.getSize(); ++i)
+  {
+    const UnsignedInteger localDimension = distributionCollection_[i].getDimension();
+    Point localShift(localDimension);
+    std::copy(shift.begin() +index,shift.begin() +index + localDimension,  localShift.begin());
```
regislebrun, line 683: '+index' -> '+ index' (add a space after plus), ',  localShift'->', localShift' (remove a space after the coma)

regislebrun (approved):
Ok for me. Only minor remarks.

lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
jschueller: yes it should

lib/src/Base/Optim/openturns/BonminProblem.hxx
jschueller: yes, here we choose to avoid storring everything in a memoize

lib/src/Base/Stat/HMatrixImplementation.cxx
jschueller: regular c++ magic ^^

regislebrun: My mother told me I would never master C++. She was right (as any mother is)

jschueller: CDeleter helps std::shared_ptr to free the C pointers given by hmat instead of using c++'s delete

sofianehaddad (approved):

PR #1851, opened 2021-06-03
Kriging cantilever beam HMAT: sensible epsilon

josephmure:
Following technical committee meeting.


regislebrun (approved):

jschueller:
cc @KieranDelamotte

jschueller:
cool, the q2 is much better

Issue #1857, opened 2021-06-16
ProductCovarianceModel fails when constructed with DiracCovarianceModel


regislebrun:
The following script:
```
import openturns as ot
c = ot.ProductCovarianceModel([ot.DiracCovarianceModel(1)])
```
enters an (almost) infinite loop. It is due to the way extra parameters are computed in `ProductCovarianceModel.cxx`:
https://github.com/openturns/openturns/blob/master/lib/src/Base/Stat/ProductCovarianceModel.cxx#L115
For a DiracCovarianceModel it is not true that the number of parameters is greater than the input dimension as the scale parameter has been removed (and not simply deactivated) from the list of parameters.

sofianehaddad:
ah my bad! Probably the same issue for `IsotropicCovarianceModel`, `KroneckerCovarianceModel`...
I will check

regislebrun:
I already have a PR to fix it and other little glitches (e.g. one of the constructors of MaternModel was missing an isStationary_ = true instruction). I will push it in a few minutes.

sofianehaddad:
ok 
thanks

Issue #1856, opened 2021-06-12
Speed up Normal computeSurvivalFunction


jaronowitz:
I have some code that needs to call `computeSurvivalFunction` function around 100,000 times for a 20-dimensional multivariate normal. Is there any way to do this in a reasonable amount of time? I've been trying to parallelize the function calls but they still are very time intensive. Below is the function that needs to be sped up.

```
def Phi(x, sig, mu=None):
    """Computes survival function of MVN at point x"""
    if mu is None:
        mu = np.zeros(sig.shape[0])

    assert sig.shape[0] == mu.shape[0]

    mvn = ot.Normal(mu, ot.CovarianceMatrix(sig))
    prob = mvn.computeSurvivalFunction(x)

    return prob

```

Thank you so much in advance for any guidance.

regislebrun:
Hi,
First of all your problem is hard if the covariance matrix is general, i.e. if there is no structure allowing to reduce the problem to a lower dimensional one. If you track the calls you see that Normal::computeSurvivalFunction() is implemented as EllipticalDistribution::computeSurvivalFunction(), which uses the symmetry of the elliptical distributions to make a call to computeCDF(). So the only way you can speed-up the computation is by tweaking the numerical parameters of Norma::computeCDF(). In ResourceMap.cxx, you have the following keys:
```
  // Normal parameters //
  addAsScalar("Normal-MaximumCDFEpsilon", 5.0e-6);
  addAsScalar("Normal-MinimumCDFEpsilon", 5.0e-2);
  addAsUnsignedInteger("Normal-MarginalIntegrationNodesNumber", 16);
  addAsUnsignedInteger("Normal-MaximumNumberOfPoints", 10000000);
  addAsUnsignedInteger("Normal-MinimumNumberOfPoints", 100000);
  addAsUnsignedInteger("Normal-SmallDimension", 6);
```
[...]

Issue #1858, opened 2021-06-17
Normal::computeCDF(sample) crash for large dimensions


regislebrun:
When dimension>ResourceMap::GetAsUnsignedInteger("Normal-SmallDimension"), computeCDF() uses a Monte Carlo estimation which makes many calls to RandomGenerator. By default, when called over a sample, computeCDF() computes its result in parallel, but RandomGenerator is not thread-safe, leading to random crashes:
```
from openturns import *
from time import time

size = 16

dim = 7

C = Normal(dim).getSample(2*dim).computeCovariance()

d = Normal([0.0]*dim, C)
points = d.getSample(size)
t0 = time()
d.computeCDF(points)
t1 = time()
print("speed=", size / (t1 - t0), "cdf/s")
```

PR #1859, opened 2021-06-17
Minor improvements

regislebrun:
This PR proposes several bug fixes and improvements:
  * The Normal and Student distribution had buggy parallel computeCDF() methods, calling the random generator in parallel. It has been fixed, and the getSample() and computeCDF() methods have been significantly improved. This work has been triggered by issue #1856 
  * The DiracCovarianceModel excluded the scale parameter from its parameters, resulting into a crash when used in the ProductCovarianceModel, see issue #1857 
  * The MaternModel was not always flagged as stationary depending on the constructor
  * The MarginalDistribution::computeCDF() method fails for some arguments due to a circular call in ContinuousDistribution::computeCDF()


regislebrun:
I get a failure on a bonmin test, strange:
python: BonHeuristicDiveMIP.cpp:133: virtual int Bonmin::HeuristicDiveMIP::solution(double&, double*): Assertion `isNlpFeasible(minlp, primalTolerance)' failed.

Any idea on how to fix it?

jschueller:
yes, its due to ipopt that got updated to 3.14.0, investigating

jschueller:
turns out they changed the default of honor_original_bounds in ipopt, so we just have to change it back on our side, see #1849

sofianehaddad (approved):

lib/src/Uncertainty/Distribution/Normal.cxx
```diff
@@ -379,6 +373,13 @@ Scalar Normal::computeCDF(const Point & point) const
   return value;
 } // computeCDF
 
+Sample Normal::computeCDF(const Sample & sample) const
+{
```
jschueller, line 377: you forgot to mention #1858

jschueller:
There are commits from master here, it should probably be rebased.
I had already a commit about the missing KL include in #1766.

regislebrun:
The remaining errors I get during the continuous integration are timeouts on other tests than mine...

jschueller:
I rerun the CI, it should be ok.
I see the Changelog is untouched though.

Issue #1864, opened 2021-06-22
FORM - IMPORTANCE FACTOR


HAFID-Anas:
Hi, 

I do not understand why deterministic variable modelled with Dirac distribution, have high importance factor values.

Example Bellow : 

```
from __future__ import print_function
import openturns as ot
import numpy as np
import openturns.viewer as viewer
from matplotlib import pylab as plt
#ot.Log.Show(ot.Log.NONE)
```

#X = [Hs, Delta, Dn, Kd, cot_a, A]
```
def Hudson(X):
    Hs, Delta, Dn, Kd, cot_a, A = X
    Z = A*Delta*Dn*np.power(Kd*cot_a,1.0/3.0)-Hs
    return [Z]
```
```
ot_Hudson = ot.PythonFunction(6, 1, Hudson)
dist_Hs = ot.Gumbel(1/1.164,2.74/1.164)
dist_Delta = ot.Normal(1.233, 0.047)
dist_Dn = ot.Normal(2.06, 0.113)
dist_Kd = ot.Dirac(8) # Deterministic parameter
dist_cot_a = ot.Normal(2,0.10)
dist_A = ot.Normal(1,0.18)
```

```
distribution = ot.ComposedDistribution([dist_Hs, dist_Delta, dist_Dn, dist_Kd, dist_cot_a, dist_A])
vect = ot.RandomVector(distribution)
output = ot.CompositeRandomVector(ot_Hudson, vect)
Faillure_Hudson = ot.ThresholdEvent(output, ot.Less(), 0.0)
```

# Estimate the probability with FORM - Define a solver.
```
optimAlgo = ot.Cobyla()
optimAlgo.setMaximumEvaluationNumber(1000)
optimAlgo.setMaximumAbsoluteError(1.0e-10)
optimAlgo.setMaximumRelativeError(1.0e-10)
optimAlgo.setMaximumResidualError(1.0e-10)
optimAlgo.setMaximumConstraintError(1.0e-10)
```

# Run  FORM
```
startingPoint = distribution.getMean()
algo = ot.FORM(optimAlgo, Faillure_Hudson, startingPoint)
algo.run()
result = algo.getResult()
ImportanceFactors = result.getImportanceFactors()
print('ImportanceFactors =', ImportanceFactors)
```

# Results : 
[...]

Issue #1854, opened 2021-06-10
Sample indexing does not work on np.int64 type


JPelamatti:
Hello,
I have noticed that the indexing of ot.Sample objects (in the sense of accessing specific elements of a Sample) only works with int indices, and not with np.int64 indices. This issue can be problematic when we are given the indices in the form a numpy array, as is briefly shown in the example below.

```
import openturns as ot
import numpy as np

sample = ot.Uniform().getSample(5)

index_list = [1,3,4] #Indices to be accessed under the form of list
print(type(index_list))  # <class 'list'>
print(type(index_list[0])) # <class 'int'>
print(sample[index_list])

index_array = np.array([1,3,4])  #Indices to be accessed under the form of a numpy array
print(type(index_array))
print(sample[index_array]) #This does not work

index_array2list = list(index_array)  #The array is transformed into a list in order to access the sample elements
print(type(index_array2list)) # <class 'list'> , as expected
print(type(index_array2list[0])) # <class 'numpy.int64'> , this is different than the working example above
print(sample[index_array2list]) # This does not work either
```

The solution I found would be to build the list of indices from an array in a sequential manner in order to be able to call the int() command, as follows:
```
index_fromarray = [int(i) for i in index_array]
print(type(index_fromarray)) #<class 'list'>
print(type(index_fromarray[0])) # <class 'int'>
print(sample[index_fromarray]) #This second workaround works
```
Given the good compatibility between openturns and numpy objects, I think it might be worth extending the indexing of openturns objects to np.int64 indices, especially as the difference between the two types of values is not particularly easy to spot and users might get stuck.

EDIT: the np.array().tolist() command also seems to be a viable solution
EDIT2 : This issue is also acknowledged in issue #1427

jschueller:
Sample.__getitem__ should check for an `__int__()` method, unfortunately its only wired when asking a single Point

sofianehaddad:
@jschueller I am lost. how to solve this issue?

PR #1867, opened 2021-06-23
Misc

jschueller:
- #1854
- #1864
- #1870
- Move linear calibration evaluation into run method (https://openturns.discourse.group/t/calculations-in-linearleastsquarescalibration-are-run-before-the-run-call/128/2)
- #1874 
- #1877 
- #1878
- #1879  
- #1880 
- #1882
- #1883 
- #1884


lib/src/Uncertainty/Algorithm/Analytical/Analytical.cxx
```diff
@@ -43,7 +43,10 @@ Analytical::Analytical(const OptimizationAlgorithm & nearestPointAlgorithm,
     physicalStartingPoint_(physicalStartingPoint)
 {
   const UnsignedInteger dimension = event.getImplementation()->getFunction().getInputDimension();
-  if (physicalStartingPoint.getDimension() != dimension) throw InvalidArgumentException(HERE) << "Starting point dimension (" << physicalStartingPoint.getDimension() << ") does not match event dimension (" << dimension << ").";
+  if (physicalStartingPoint.getDimension() != dimension)
+    throw InvalidArgumentException(HERE) << "Starting point dimension (" << physicalStartingPoint.getDimension() << ") does not match event dimension (" << dimension << ").";
+  if (!event_.getImplementation()->getAntecedent().getDistribution().isContinuous())
```
regislebrun, line 48: Be careful, discrete is not the opposite of being continuous. FORM/SORM methods are dedicated to absolutely continuous distributions, if only one marginal distribution is not continuous it will fail. Discrete is when **all** the marginals are discrete.

jschueller: yes, isContinuous will fail if at least one marginal is discrete

regislebrun: exactly

python/test/t_TruncatedDistribution_std.py
```diff
@@ -206,3 +206,8 @@ def cleanCovariance(inCovariance):
 bounds = ot.Interval([-2, -3], [2, 3.0], [True, False], [True, True])
 dist = ot.TruncatedDistribution(ot.Normal(2), bounds)
 print('proba=%.6f' % dist.computeCDF([3.0, -3.0]))
+
+# relative range wrt quantile epsilon issue
+unif = ot.Uniform(0.0, 1e12)
+trunc = ot.TruncatedDistribution(unif, 0.25, 2.0)
+print('q@0.1=', trunc.computeQuantile(0.1))
```
josephmure, line 213: Couldn't you use an `assert_almost_equal` here?

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -276,6 +276,9 @@ void TruncatedDistribution::computeRange()
     const Scalar probability = distribution_.computeProbability(range);
     if (!(probability > 0.0)) throw InvalidArgumentException(HERE) << "Error: the truncation interval does not contain a non-empty part of the support of the distribution";
     normalizationFactor_ = 1.0 / probability;
+
+    // scale quantile epsilon of the inner distribution
+    distribution_.getImplementation()->setQuantileEpsilon(distribution_.getImplementation()->getQuantileEpsilon() * probability);
```
josephmure, line 281: This solution to #1870 seems reasonable to me.

josephmure (approved):

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
josephmure: On second thought, isn't there a risk of `computeRange` begin called multiple times (say, by `setBounds`) and lowering `quantileEpsilon` every time?

josephmure: Use `ResourceMap.GetAsScalar('Distribution-DefaultQuantileEpsilon')` instead of `distribution_.getImplementation()->getQuantileEpsilon()`

lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx
```diff
@@ -913,7 +913,10 @@ void SobolIndicesAlgorithmImplementation::setDesign(const Sample & inputDesign,
     if (!(referenceVariance_[j] > 0.0))
       throw InvalidArgumentException(HERE) << "Null output sample variance";
 
+  // reset intermediate results
   alreadyComputedIndicesDistribution_ = false;
+  varianceI_.clear();
```
mbaudin47, line 918: You chose to clear `varianceI_` in SobolIndicesAlgorithmImplementation.cxx, which fixes the bug. But why not using `alreadyComputedIndicesDistribution_` instead of `varianceI_` at:
https://github.com/openturns/openturns/blob/186499f86b16a03b976b0d8b578f9903430d1fb7/lib/src/Uncertainty/Algorithm/Sensitivity/SobolIndicesAlgorithmImplementation.cxx#L131
to detect the change in the data? I think that it would make the intent much clearer.

jschueller: that's only for the distribution

jschueller:
hello @regislebrun @sofianehaddad, can I request a review ?

sofianehaddad (approved):

PR #1887, opened 2021-08-10
IndicatorFunction takes a Domain as input

josephmure:
# Extend [`IndicatorFunction`](https://7940-40717169-gh.circle-artifacts.com/0/html/user_manual/_generated/openturns.IndicatorFunction.html#openturns.IndicatorFunction) to all domains

Previously, it reimplemented code from `LevelSet` and did not work for other `Domains`.

# New functionality
Coded in tandem with @JPelamatti 
- Domains get a new [`computeDistance`](https://7940-40717169-gh.circle-artifacts.com/0/html/user_manual/_generated/openturns.Domain.html#openturns.Domain.computeDistanceToDomain) method (only actually implemented for `Intervals` and `DomainUnions`)
- New [`DistanceToDomainFunction`](https://7940-40717169-gh.circle-artifacts.com/0/html/user_manual/_generated/openturns.DistanceToDomainFunction.html#openturns.DistanceToDomainFunction) coded the same way as `IndicatorFunction`.


lib/src/Base/Func/openturns/IndicatorEvaluation.hxx
```diff
@@ -63,18 +60,7 @@ public:
   /** Operator () */
   using EvaluationImplementation::operator();
```
josephmure, line 61: Is this line still necessary, now that ` Sample IndicatorEvaluation::operator() (const Sample & inSample)` has been overridden?

lib/src/Base/Geom/LevelSet.cxx
```diff
@@ -134,6 +139,26 @@ Bool LevelSet::contains(const Point & point) const
...
+/* Compute the Euclidean distance from a given point to the LevelSet */
+Scalar LevelSet::computeDistanceToDomain(const Point & point) const
+{
+  if (contains(point)) return 0.0;
+  const UnsignedInteger dim(point.getDimension());
+  const TranslationFunction translate(point);
+  const ComposedFunction levelFunction(function_, translate);
+  const NearestPointProblem problem(levelFunction, level_);
+  AbdoRackwitz abdoRackwitz(problem);
+  Sample startingSample(2, dim); // AbdoRackwitz fails if levelFunction has null gradient at its starting point
```
mbaudin47, line 151: If a non-convex LevelSet is used, I guess that there is only a small chance of getting the minimum starting only with the points [0, ..., 0] and [1, ..., 1]. Would a random sample be useful?

lib/src/Base/Geom/LevelSet.cxx
```diff
@@ -134,6 +139,26 @@ Bool LevelSet::contains(const Point & point) const
   return operator_(function_(point)[0], level_);
 }
 
+/* Compute the Euclidean distance from a given point to the LevelSet */
+Scalar LevelSet::computeDistanceToDomain(const Point & point) const
```
mbaudin47, line 143: Would it be useful to set a norm different from the Euclidian norm, e.g. 1-norm or infinity-norm?

lib/src/Base/Geom/LevelSet.cxx
josephmure: I guess, but how many sample points would we need? Which standard deviation do we need? Should we use an LHSExperiment?

lib/src/Base/Geom/LevelSet.cxx
josephmure: Perhaps. I do not think other norms are needed in the HSIC context, but we could add them in the future.

lib/src/Base/Func/DistanceToDomainEvaluation.cxx
```diff
@@ -0,0 +1,115 @@
...
+DistanceToDomainEvaluation * DistanceToDomainEvaluation::clone() const
+{
+  return new DistanceToDomainEvaluation(*this);
+}
+
+
+/* Comparison operator */
+Bool DistanceToDomainEvaluation::operator ==(const DistanceToDomainEvaluation & ) const
+{
+  return true;
```
josephmure, line 61: @regislebrun mentioned yesterday that this is incorrect. Since it is copied from `IndicationEvaluation::operator ==`, the original should also be changed.

Could we perhaps do it this way?

```suggestion
Bool DistanceToDomainEvaluation::operator ==(const DistanceToDomainEvaluation & other) const
{
  return domain_ == other.getDomain();
```
Of course, to do this, I also need to implement the accessor `getDomain()`.

jschueller: not necessarily, you can access the attribute from there

josephmure: OK. Done.

lib/src/Base/Geom/openturns/Domain.hxx
```diff
@@ -52,6 +52,12 @@ public:
   /** Check if the closed domain contains given points */
   virtual BoolCollection contains(const Sample & sample) const;
 
+  /** Compute the Euclidean distance from a given point to the domain */
+  virtual Scalar computeDistanceToDomain(const Point & point) const;
+
+  /** Compute the Euclidean distance from given points to the domain */
+  virtual Sample computeDistanceToDomain(const Sample & sample) const;
+
```
jschueller, line 60: This method name would suggest that we compute a distance from a domain (as class is Domain)  to another domain (as argument)
but from the class point of view it is actually a distance to point(s) as we'll likely write something like:
```domain.computeDistance(point)```
So I would suggest just 'computeDistance' (or computeDistanceToPoint)

Comment (josephmure) [reply lib/src/Base/Geom/openturns/Domain.hxx:60]: OK, point taken. Do you think the name of the class `DistanceToDomainFunction` should also be changed?

Comment (jschueller) [reply lib/src/Base/Geom/openturns/Domain.hxx:60]: this one is fine, no need to explicit that it is a domain-point distance

another possibility would be 'DomainDistanceFunction'

File: lib/src/Base/Func/DistanceToDomainFunction.cxx [hunk 6]
```diff
@@ -0,0 +1,66 @@
...
+
+BEGIN_NAMESPACE_OPENTURNS
+
+CLASSNAMEINIT(DistanceToDomainFunction)
+
+/* Default constructor */
+DistanceToDomainFunction::DistanceToDomainFunction ()
+  : Function()
+{
+}
```
regislebrun, line 34: Add // Nothing to do

lib/src/Base/Func/IndicatorEvaluation.cxx
```diff
@@ -22,35 +22,32 @@
...
-  , evaluation_()
-  , comparisonOperator_(comparisonOperator)
-  , threshold_(threshold)
+  , domain_(domain)
 {
-  setEvaluation(evaluation);
-  setDescription(evaluation.getDescription());
+  const UnsignedInteger domainDimension = domain_.getDimension();
+  const Sample dummySample(1, domainDimension);
+  EvaluationImplementation::setInputDescription(dummySample.getDescription());
```
regislebrun, line 50: Remove `dummySample` and use `Description::BuildDefault(domainDimension, "v")` instead to build the description.

lib/src/Base/Func/IndicatorEvaluation.cxx
```diff
@@ -72,26 +69,33 @@ String IndicatorEvaluation::__repr__() const
...
   callsNumber_.increment();
   return result;
 }
 
+Sample IndicatorEvaluation::operator() (const Sample & inSample) const
+{
+  const UnsignedInteger size = inSample.getSize();
+  const UnsignedIntegerCollection belong(domain_.contains(inSample));
+  Sample result(size, 1);
+  for (UnsignedInteger i = 0; i < size; ++i) result[i][0] = belong[i];
```
regislebrun, line 90: use `result(i, 0)` here

lib/src/Base/Geom/DomainUnion.cxx
```diff
@@ -93,6 +93,16 @@ DomainUnion::BoolCollection DomainUnion::contains(const Sample & sample) const
   return result;
 }
 
+/* Compute the Euclidean distance from a given point to the domain */
+Scalar DomainUnion::computeDistance(const Point & point) const
+{
+    const Scalar distanceLeft = left_.computeDistance(point);
+    if (distanceLeft==0.0) return 0.0;
```
regislebrun, line 100: Leave one space around `==`

lib/src/Base/Geom/Interval.cxx
```diff
@@ -177,6 +177,32 @@ Bool Interval::contains(const Point & point) const
...
+    Point upperBound = getUpperBound();
+    BoolCollection finiteLowerBound = getFiniteLowerBound();
+    BoolCollection finiteUpperBound = getFiniteUpperBound();
+
+    Scalar squaredDistance = 0.0;
+
+    for (UnsignedInteger i = 0; i < intervalDimension; ++i)
+{
+    if (finiteLowerBound[i] && point[i]<lowerBound[i]) squaredDistance += (lowerBound[i] - point[i])*(lowerBound[i] - point[i]);
+    else if (finiteUpperBound[i] && point[i]>upperBound[i]) squaredDistance += (point[i] - upperBound[i])*(point[i] - upperBound[i]);
```
regislebrun, line 200: Leave one space around `<` and `>` in this line and the preceding one, as well as around `*`

lib/src/Base/Geom/LevelSet.cxx
regislebrun: A robust way to compute this distance is to rely on `LevelSetMesher` then to build a `MeshDomain` using the resulting `Mesh` and you either return the result of `MeshDomain.computeDistance()` or use the nearest vertex to initialize your algorithm. It all depends on how robust you want to make this method and the price you are ready to pay for this robustness...

regislebrun (changes requested):
Nice work.

lib/src/Base/Geom/LevelSet.cxx
josephmure: @regislebrun How could we set the `boundingBox` argument of [LevelSetMesher.build](https://openturns.github.io/openturns/1.17/user_manual/_generated/openturns.LevelSetMesher.html#openturns.LevelSetMesher.build)?

regislebrun: You can use the bounding box of the LevelSet!

josephmure: I don't think so. In the example code from the [LevelSetMesher doc page](https://openturns.github.io/openturns/1.17/user_manual/_generated/openturns.LevelSetMesher.html), the bounding box is manually specified:

```
import openturns as ot
mesher = ot.LevelSetMesher([5, 10])
level = 1.0
function = ot.SymbolicFunction(['x0', 'x1'], ['x0^2+x1^2'])
levelSet = ot.LevelSet(function, ot.LessOrEqual(), level)
mesh = mesher.build(levelSet, ot.Interval([-2.0]*2, [2.0]*2))
```
Moreover, ``levelSet.getLowerbound()``  and ``levelSet.getUpperbound()``both yield
``
class=Point name=Unnamed dimension=0 values=[]
``

regislebrun: Nope. If you define a LevelSet object in the perspective of using it in this context, you have to define the bounds of the LevelSet. You can check the dimension of the bounds and throw the appropriate exception here. There is no magic here: you have to help the system a little bit...

josephmure: In the interest of getting this PR merged, I have removed the commit introducing LevelSet::computeDistance.

josephmure:
Since HSIC-related developments depend on the new class, is it possible to rapidly merge this PR? @jschueller @regislebrun @sofianehaddad @mbaudin47

regislebrun (approved):

sofianehaddad (approved):
Fine with me.
You can potentially squash some commits to make it more readable

josephmure:
@sofianehaddad Done. For some reason, the fourth commit does not appear on the conversation page, but it does appear in the "commits" tab.

jschueller:
It needs a rebase before merge

PR #1890, opened 2021-09-02
Improve covariance parameters optimization. Set local solver.

mbaudin47:
This PR improves the example for kriging. 
- It shows how to configure the local optimization algorithm.
- It correctly configures the optimization of the covariance model.

### DIFF
--- a/python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_hyperparameters_optimization.py
+++ b/python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_hyperparameters_optimization.py
@@ -3,29 +3,29 @@ Kriging :configure the optimization solver
 ==========================================
 """
 # %%
-# The goal of this example is to show how to fine-tune the optimization solver used to estimate the hyperparameters of the covariance model of the kriging metamodel. 
+# The goal of this example is to show how to fine-tune the optimization solver used to estimate the hyperparameters of the covariance model of the kriging metamodel.
 #
 # Introduction
 # ------------
 #
-# In a kriging metamodel, there are various types of parameters which are estimated from the data. 
+# In a kriging metamodel, there are various types of parameters which are estimated from the data.
 #
-# * The parameters :math:`{\bf \beta}` associated with the deterministic trend. These parameters are computed based on linear least squares. 
-# * The parameters of the covariance model. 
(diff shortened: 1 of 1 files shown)


python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_hyperparameters_optimization.py
```diff
@@ -88,38 +89,53 @@
...
-covarianceModel = ot.SquaredExponential([1.]*dimension, [1.0])
+
+# Trick B, v2
+x_range = X_train.getMax() - X_train.getMin()
+print("x_range:")
+print(x_range)
+scale_max_factor = 4.0  # Must be > 1, tune this to match your problem
+scale_min_factor = 0.1  # Must be < 1, tune this to match your problem
+maximum_scale_bounds = scale_max_factor * x_range
+minimum_scale_bounds = scale_min_factor * x_range
```
JPelamatti, line 105: I personally do not agree with this. I think that the minimum_scale_bounds value for each dimension should be proportional to the smallest distance within the data set for the given dimension, not the largest.
I have suggested an alternative initialization in the relative discussion on the Discourse channel ( https://openturns.discourse.group/t/normalization-of-the-input-sample-in-krigingalgorithm-in-ot1-16/101/2 )

mbaudin47: I do not agree :) This might be discussed more izili tomorrow at the TC.

python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_hyperparameters_optimization.py
```diff
@@ -258,31 +286,57 @@
...
+# Configure the local optimization solver
+# ---------------------------------------
+
+# %%
+# The following example shows how to set the local optimization solver.
+# We choose the SLSQP algorithm from NLOPT.
+
+# %%
+problem = solver.getProblem()
+local_solver = ot.NLopt(problem, "LD_SLSQP")
```
mbaudin47, line 318: Using the problem from a previous kriging is not required. We may as well use directly:
```Python
local_solver = ot.NLopt("LD_SLSQP")
```
This is because only the optimization class is required: the problem is created by the KrigingAlgorithm class.

Issue #1899, opened 2021-10-08
Inconsistency in the way ResourceMap keys are referenced in the documentation


regislebrun:
The reference to ResourceMap keys is not uniform within the documentation:
```
python/src/ExpectationSimulationAlgorithm_doc.i.in:The default type is set by the `ExpectationSimulationAlgorithm-DefaultCoefficientOfVariationCriterionType` key of the :class:`~openturns.ResourceMap`.
python/src/MultiStart_doc.i.in:    Default value is *MultiStart-KeepResults* in :class:`~openturns.ResourceMap`"
python/src/SobolIndicesExperiment_doc.i.in:The type of :class:`~openturns.WeightedExperiment` depends on the value of ``'SobolIndicesExperiment-SamplingMethod'`` in the :class:`~openturns.ResourceMap`:
python/src/BernsteinCopulaFactory_doc.i.in:    The name of thebin number selection method. Possible choices are *AMISE*, *LogLikelihood* and *PenalizedCsiszarDivergence*. Default is *LogLikelihood*, given by the *'BernsteinCopulaFactory-BinNumberSelection'* entry of :class:`~openturns.ResourceMap`.
```
How should we refer to a key (both the wording and the presentation) and its value?

jschueller:
it seems to be mostly ``` `Key` ```

PR #1904, opened 2021-10-14
Improved VisualTest::DrawPairs

regislebrun:
This PR provides the following improvements:
* Now the user can control the point style and the color of VisualTest::DrawPairs through ResourceMap.
* The binning algorithm in KernelSmoothing could lead to slightly negative weights due to tound-off, which is fixed here.

### DIFF
--- a/lib/src/Uncertainty/Distribution/KernelSmoothing.cxx
+++ b/lib/src/Uncertainty/Distribution/KernelSmoothing.cxx
@@ -373,25 +373,46 @@ Mixture KernelSmoothing::buildAsMixture(const Sample & sample,
   // 2D binning?
   if (dimension == 2)
   {
-    const Point xMin(sample.getMin());
-    const Point xMax(sample.getMax());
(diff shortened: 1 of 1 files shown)


josephmure:
The typography of the names ‘Drawable-DefaultPointStyle’ and ‘Drawable-DefaultColor’  does not look great in the CI-compiled doc:

https://8133-40717169-gh.circle-artifacts.com/0/html/user_manual/_generated/openturns.VisualTest_DrawPairs.html?highlight=visualtest#openturns.VisualTest_DrawPairs

Possible alternatives:

1. *Drawable-DefaultPointStyle*
2. ``Drawable-DefaultPointStyle``
3. "Drawable-DefaultPointStyle"
4. **Drawable-DefaultPointStyle**

What do you think? IMO option 2 is best, but option 3 emphasizes it is a string.

regislebrun:
@josephmure Your remark on the typography is exactly the scope of issue #1899. IMO it should be fixed in a dedicated PR.

PR #1932, opened 2021-12-17
Update sobol validation

mbaudin47:
This PR updates Sobol' validation scripts to OT v1.17.


validation/src/sobol_estimator_variance/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.png
+__pycache__
+graphe_validation/
+graphe_validation_mean_distribution/
+
```
jschueller, line 5: perhaps a .gitignore in /validation would be best, to avoid having one per folder in the future

validation/src/sobol_estimator_variance/convergence_stddev_old.py
```diff
@@ -0,0 +1,342 @@
...
+import openturns as ot
+from gsobollib import gsobolDistribution, gsobol
+from ishigamilib import (
+    ishigamiDistribution,
+    ishigamiGSymbolic,
+)
+from numpy import sqrt, array
+import numpy as np
+import matplotlib as mpl
+
```
jschueller, line 18: this one looks similar the new isnt it ? why keep it ?

jschueller:
a rebase is in order to go green

validation/src/sobol_estimator_variance/.gitignore
mbaudin47: Ok. I create a new .gitignore in the parent directory with that particular content. This makes sense because there are Python scripts here, e.g. ValidPareto.py

validation/src/sobol_estimator_variance/convergence_stddev_old.py
mbaudin47: You are right: I remove it.

josephmure (approved):

PR #1933, opened 2021-12-17
Improve covariance parameters optimization. Set local solver.

mbaudin47:
The goal of this PR is to configure the optimization solver of this kriging example so that it correctly find the scale hyperparameters of the covariance model. It sets the initial value of the scale parameter and sets the bounds of the optimization algorithm. 
With the old settings, the optimized covariance model is:
```
SquaredExponential(scale=[1,1,1,1], amplitude=[5.08434])
```
which shows that the optimization did not perform correctly on the scale parameters, leaving the parameters unmodified with respect to their default initial values.
With the updated settings, the optimized covariance model is:
```
SquaredExponential(scale=[5.26006e+07,44804,32.7808,255.395], amplitude=[17.4415])
```
which is much better.


regislebrun (commented):
+1

jschueller:
+1, but I cannot merge it as it has conflicts, please rebase

regislebrun (approved):

sofianehaddad (approved):

mbaudin47:
I fixed the conflict.

PR #1936, opened 2021-12-19
Python getImplementation performs dynamic casting

### LINKED ISSUE #1931: Interface class getImplementation methods are useless in Python
Suppose I get a `KrigingResult` and want to access the underlying covariance kernel. The method `KrigingResult::getCovarianceModel` yields an object of the interface class `CovarianceModel`. Let us call this object `cov`.

Even if the implementation is actually a `MaternModel`, I will not be able to call `cov.getNu()`, since that method is specific to the `MaternModel` class. Now, I could access it indirectly through the generic `getParameter` method, but this is impractical.

In C++, `cov.getImplementation()` yields a pointer and I could write `cov.getImplementation()->getNu()`.
In Python however `cov.getImplementation()` yields a useless `CovarianceModelImplementationPointer` object, but that "pointer" cannot access `getNu`.

Would it be possible to make the Python version of `getImplementation` yield a reference to the actual implementation instead of a pointer?

EDIT: This problem is not limited to the case of `CovarianceModel` objects. Here is a (probably not exhaustive) list of situations where the lack of Python `dynamic_cast` is an issue.

## CovarianceModel

See above. I recently coded a Python class with @efekhari27 that relied on OT covariance kernels. We were able to derive analytical formulas for a specific kernel (Matérn with nu=5/2), so the class needed to inspect the `CovarianceModel` object in order to see whether analytical formulas were available or not. It went like this: with `kernel` being some `CovarianceModel`:

```python
 if kernel.getImplementation().getClassName()=="MaternModel"
    if kernel.getParameter()[-1]==2.5
        _analytical = True
```
Of course, what I would have liked to write is:
```python
 if kernel.getImplementation().getClassName()=="MaternModel"
    if kernel.getNu()==2.5
        _analytical = True
```
However, whenever `kernel` is obtained from a `KrigingResult`, its type is `CovarianceModel` so Matérn-specific methods are unavailable.

## Distribution

Many classes have `getDistribution` methods, and all return... objects of type `Distribution`. As Sofiane pointed out below, factories often provide `buildAsXXX` methods to circumvent the issue, but non-factory classes do not have such methods. @mbaudin47 may be able to provide more specific examples.

## RandomVector

The `RandomVector` interface class has many methods that can only used by one or some of the classes derived from `RandomVectorImplementation`. Here is a list:
Method | Description
-- | --
getAntecedent() | Accessor to the antecedent RandomVector in case of a composite RandomVector.
getCovariance() | Accessor to the covariance of the RandomVector.
getDescription() | Accessor to the description of the RandomVector.
getDistribution() | Accessor to the distribution of the RandomVector.
getDomain() | Accessor to the domain of the Event.
getFunction() | Accessor to the Function in case of a composite RandomVector.
getId() | Accessor to the object's id.
getMarginal(*args) | Get the random vector corresponding to the  marginal component(s).
getMean() | Accessor to the mean of the RandomVector.
getOperator() | Accessor to the comparaison operator of the Event.
getParameter() | Accessor to the parameter of the distribution.
getParameterDescription() | Accessor to the parameter description of the distribution.
getThreshold() | Accessor to the threshold of the Event.
intersect(other) | Intersection of two events.
isComposite() | Accessor to know if the RandomVector is a composite one.
isEvent() | Whether the random vector is an event.
join(other) | Union of two events.
setName(name) | Accessor to the object's name.
setParameter(parameters) | Accessor to the parameter of the distribution.

These methods exist in the interface but are implemented as throwing `NotYetImplementedException` in `RandomVectorImplementation`. As a result, they clutter both the API and the doc of all `RandomVectorImplementation` derived classes that do not implement them. If we could dynamic-cast in Python, we probably would not need to do this.

## Conclusion

Several strategies exist to circumvent the lack of dynamic-cast (or some substitute) in Python:

- The `CovarianceModel` strategy: since all derived classes are similar, we can get away with losing some model-specific methods because they are just more convenient accessors, but it can be a hassle for users.
- The `Factory` strategy: write as many `buildAsXXX` methods as necessary. This is more convenient for users, but is a lot of work for developers.
- The `RandomVector` strategy: put everything in the interface. Users can access everything but must understand that most methods available are dummy methods, this makes the API harder to understand.

I believe we need to come up with some way of achieving in Python what is done with dynamic cast in C++. Maybe there are SWIG tools that could be helpful?

josephmure:
An alternative version of `TypedInterfaceObject.getImplementation` is implemented as a SWIG extension of the class`TypedInterfaceObject`.
It returns a handle with type the output of the C++ command  `getImplementation()->getClassName()`.

Fixes #1931 since it lets objects obtained from the Python version of `getImplementation` access all methods of the actual implementation class.

As an example, here is a code snippet:

```python
import openturns as ot

mat = ot.MaternModel()
cov = ot.CovarianceModel(mat)
imp = cov.getImplementation()
print(type(imp))
```

With OT version 1.18, the output is
```python
<class 'openturns.statistics.CovarianceModelImplementationPointer'>
```

With this PR, it is
```python
<class 'openturns.statistics.MaternModel'>
```

As a result, the following code now works:

```python
imp.setNu(0.75) # fails under OT 1.18 because CovarianceModelImplementation has no setNu method
print(imp)
```
Which prints:
```python
MaternModel(scale=[1], amplitude=[1], nu=0.75)
```

### DIFF
--- a/python/src/TypedInterfaceObject.i
+++ b/python/src/TypedInterfaceObject.i
@@ -4,8 +4,6 @@
 #include "openturns/TypedInterfaceObject.hxx"
 %}
 
-%ignore OT::TypedInterfaceObject::swap;
-
 %define TypedInterfaceObjectImplementationHelper(Namespace, Interface, Implementation)
 
 %template(Implementation ## TypedInterfaceObject) OT::TypedInterfaceObject<Namespace::Implementation>;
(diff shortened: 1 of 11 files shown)


josephmure:
Finally managed to implement the third approach from the [SWIG documentation](http://www.swig.org/Doc4.0/Python.html#Python_memory_management_member_variables). This PR is now ready to be reviewed, even though we need to decide what to do with the problematic `ProductFunction(FunctionImplementation, FunctionImplementation)` constructor.

To summarize, each of the last 3 commits of the PR proposes a different way to prevent premature collection of `TypedInterfaceObjects` when executing a line like `ot.CovarianceModel().getImplementation()`. Each of them passes all unit tests:

1. e0061e9 uses Python code to create an attribute in the returned Implementation object that refers to the original `TypedInterfaceObject` and prevents its deletion by the garbage collector.
2. 54c207d puts all the code transforming the behavior of `getImplementation` in a `typemap`. The advantage is that the code that creates the attribute referencing the original ``TypedInterfaceObject`` is also in the `typemap` and is written in C++ rather than Python.
3. 8a7b53e does the same, except part of the work is done at module initialization, which according to the doc is "thread-safe".

jschueller:
Nice.
I think we want to avoid getImplementation to return a dangling pointer here, would returning a clone be enough ? (this would avoid any garbage collection trick)

josephmure:
I do not see how that could happen but then again, I am no expert either.

regislebrun (approved):
Cool

python/src/TypedInterfaceObject.i
```diff
@@ -26,13 +26,38 @@
...
+// https://github.com/swig/swig/blob/90cdbee6a69d13b39d734083b9f91069533b0d7b/Examples/test-suite/template_private_assignment.i
+%template() OT::Pointer<Namespace::ParentImplementation>;
+
+// Different typemaps are generated for all ParentImplementation classes.
+%typemap(out) OT::TypedInterfaceObject< Namespace::ParentImplementation >::Implementation
+{
+  // Construct the SWIG identifier of the derived class we want to get (say ChildImplementation).
+  OT::String childname = "Namespace::";
+  childname.append($1->getClassName());
+  childname.append("*");
```
jschueller, line 50: `const OT::String childname = "Namespace::" + $1->getClassName() + "*";`

jschueller (approved):
you could add a test, like your MaternModel snippet

josephmure:
@jschueller I have kept all commits to keep the history of this page, but you are welcome to squash all commits when you merge the PR.

PR #1942, opened 2022-01-31
MetaModel: deprecate Function ctors

jschueller:
Rationale: FCA doesnt actually implement anything from a Function argument that cannot be done with the X/Y Sample counterparts.
This drops the Function argument from MetaModelResult/MetaModelAlgorithm ctors (we dont care about backwards compatibility there)
At the FunctionalChaosAlgorithm they are only deprecated (the attribute model_ has to stay until they are removed).


sofianehaddad (approved):

jschueller:
deferring this

Issue #1935, opened 2021-12-18
Some help pages have content and formatting issues


mbaudin47:
## TODO

- [x] The following cantilever beam help page has issues: [The cantilever beam model](https://openturns.github.io/openturns/latest/usecases/use_case_cantilever_beam.html)

The sentence should not be in bold case.

![image](https://user-images.githubusercontent.com/31351465/155025798-c4bf242c-0690-4af8-821a-7b149c7e1a00.png)

Anne Dutfoy

Edit (30/03/2023). Attempt to fix this bug in PR #2262 (but seem to failed?).

- [x] The [importance_form](http://openturns.github.io/openturns/latest/theory/reliability_sensitivity/importance_form.html) theory help page has a blank space.

![image](https://user-images.githubusercontent.com/31351465/218273254-40e72252-8b95-46cc-a21a-dad13f876b3a.png)

The variables E and F_1, ..., F_n are undefined before or after Y* is suddenly introduced:

![image](https://user-images.githubusercontent.com/31351465/218273499-f5e74fc3-35d4-4b3c-b728-0fd91f64d03f.png)

The variable \tilde{g} is undefined:

![image](https://user-images.githubusercontent.com/31351465/218273845-fbdb1261-369c-4ac9-8536-c4e878adbb66.png)

I assume that this is the linear model which approximates g in the standard space, but this is undefined in the text. 

Anne Dutfoy

Edit (30/03/2023). Attempt to fix in PR #2262.

- [x] The following help page has issues: [Polynomial chaos graphs](http://openturns.github.io/openturns/master/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.html)

The following sentences has no meaning.

![image](https://user-images.githubusercontent.com/31351465/146645555-9e98b122-1cec-4f76-9cab-4f40b77de7b4.png)

~~Loic Brevault~~

- [ ] The following page has issues: [Polynomial chaos over database](http://openturns.github.io/openturns/master/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_database.html)

The following sentence is not formatted.

![image](https://user-images.githubusercontent.com/31351465/146645399-78c4e26e-1deb-4e4f-bc9f-ff4c5c00a9d4.png)

The following sentence has no capital letter, and no point.

![image](https://user-images.githubusercontent.com/31351465/146645421-e0361b28-a7bf-486f-a81d-a3401c195dcd.png)

~~Julien Floquet~~

[...]

PR #1968, opened 2022-02-16
Fix formating plot_functional_chaos_graphs.py

l-brevault:
#1935

### DIFF
--- a/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.py
+++ b/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.py
@@ -31,7 +31,7 @@ def drawFamily(factory, degreeMax=5):
     # Fix the number of points for the graph
     pointNumber = 101
 
-    # Bounds of the graph
+    # Define the bounds of the graph
     xMinJacobi = -1.0
     xMaxJacobi = 1.0
 
@@ -50,11 +50,11 @@ def drawFamily(factory, degreeMax=5):
 # Draw the 5-th first members of the Jacobi family.
 
 # %%
-# Create the Jacobi polynomials family using the default Jacobi.ANALYSIS
+# Create the Jacobi polynomials family using the default analysis  
 # parameter set
 alpha = 0.5
 beta = 1.5
-jacobiFamily = ot.JacobiFactory(alpha, beta)
+jacobiFamily = ot.JacobiFactory(alpha, beta, ot.JacobiFactory.ANALYSIS)
 graph = drawFamily(jacobiFamily)
 view = viewer.View(graph)
 


python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.py
```diff
@@ -50,11 +50,11 @@ def drawFamily(factory, degreeMax=5):
 # Draw the 5-th first members of the Jacobi family.
 
 # %%
-# Create the Jacobi polynomials family using the default Jacobi.ANALYSIS
+# Create the Jacobi polynomials family using the default analysis  
 # parameter set
```
josephmure, line 54: The reader should understand that we're choosing one of two available parametrizations for the Beta distribution. It might be helpful to create a link to the API doc page of `JacobiFactory`, which contains this explanation.

jschueller: I think its okay to merge as is because it fulfills the initial goal.
Imho no need to add a link to the API, it will be tedious to add these for each example.

jschueller (approved):

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.py
```diff
@@ -31,7 +31,7 @@ def drawFamily(factory, degreeMax=5):
     # Fix the number of points for the graph
     pointNumber = 101
 
-    # Bounds of the graph
+    # Define the bounds of the graph
```
mbaudin47, line 34: Please add a point to the end of the line, so that it makes a plain sentence.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.py
mbaudin47: The correction is somewhat worse than the original, but cryptic, wording, because the word "analysis" is unrelated to the content of the script. At this point, we must explain that the `JacobiFactory` has two different parametrizations. An additional bug is that:

https://github.com/openturns/openturns/blob/9adf87d2687d5dc956214bb66326dac0688d69b8/python/src/JacobiFactory_doc.i.in#L25

does not state which of either ot.JacobiFactory.ANALYSIS or ot.JacobiFactory.PROBABILITY is the default setting.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.py
```diff
@@ -50,11 +50,11 @@ def drawFamily(factory, degreeMax=5):
 # Draw the 5-th first members of the Jacobi family.
 
 # %%
-# Create the Jacobi polynomials family using the default Jacobi.ANALYSIS
+# Create the Jacobi polynomials family using the default analysis  
 # parameter set
 alpha = 0.5
 beta = 1.5
-jacobiFamily = ot.JacobiFactory(alpha, beta)
+jacobiFamily = ot.JacobiFactory(alpha, beta, ot.JacobiFactory.ANALYSIS)
```
mbaudin47, line 57: If the default parametrization is ot.JacobiFactory.PROBABILITY, then we must explain that we use ot.JacobiFactory.ANALYSIS to specifically use that parametrization. If the default parametrization is ot.JacobiFactory.ANALYSIS, therefore setting this option has no interest. I suggest use both parametrizations (e.g. with a loop over the two parametrizations) to show the difference in the plot. This way, the meaning of this parameter will be clear and the feature will have an comprehensive example.

mbaudin47 (changes requested):

mbaudin47:
The whole example has more issues than the ones identified in the initial bug report. An option would be to drop the PR and fix the content with more depth, which is beyond the initial motivation.

jschueller:
agreed, sorry loic

PR #1973, opened 2022-02-16
Fixed typos in field surrogate model example

JPelamatti:
Minor corrections in plot_fieldfunction_metamodel.py. Refers  to #1935

### DIFF
--- a/python/doc/examples/meta_modeling/fields_metamodels/plot_fieldfunction_metamodel.py
+++ b/python/doc/examples/meta_modeling/fields_metamodels/plot_fieldfunction_metamodel.py
@@ -6,14 +6,14 @@ Metamodel of a field function
 #
 # In this example we are going to create a metamodel of a field function following these steps:
 #
-# - Creation of a field model over an 1-d mesh
-# - Creation of a Gaussian process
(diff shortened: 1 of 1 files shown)


python/doc/examples/meta_modeling/fields_metamodels/plot_fieldfunction_metamodel.py
```diff
@@ -6,14 +6,14 @@
...
 #
-# - Creation of a field model over an 1-d mesh
-# - Creation of a Gaussian process
-# - Karhunen-Loeve decomposition of a process with known covariance function
-# - Karhunen-Loeve decomposition of a process with known trajectories
-# - Projection of Fields
-# - Functional chaos decomposition between the coefficients of the input and output processes
-# - Build a metamodel of the whole field model
-# - Validate the metamodel
+# Step 1: Creation of a field model over an 1-d mesh.
```
josephmure, line 9: Maybe use an enumerated list: https://docutils.sourceforge.io/docs/user/rst/quickref.html#enumerated-lists

I think the rendering would be better.

josephmure (approved):

PR #1982, opened 2022-02-27
Create a Sample.argsort method

mbaudin47:
This PR creates a `argsort` method for the `Sample` class. The purpose is to be able to simultaneously sort the rows of a `Sample` and another object like a `Point` or any other object. This can be used, for example, to simultaneously sort the nodes and weights of a `WeighteExperiment`.

One of the goals of this PR is to simplify the testing of e.g. the suggested `TensorProductExperiment` at PR #1963, and more precisely:

https://github.com/openturns/openturns/blob/7dbb251c089edc05f0c5b6ee32694d16150acdd8/lib/test/t_TensorProductExperiment_std.cxx#L55


lib/src/Base/Stat/SampleImplementation.cxx
```diff
@@ -1573,6 +1573,25 @@ void SampleImplementation::sortUniqueInPlace()
...
 
+Indices SampleImplementation::argsort() const
+{
+  Collection< std::pair<Point, UnsignedInteger> > pointsPairs(size_);
+
+  for (UnsignedInteger i = 0; i < size_; ++i)
+  {
+    Point point_i(dimension_);
+    for (UnsignedInteger j = 0; j < dimension_; ++j) point_i[j] = operator()(i, j);
+    pointsPairs[i] = std::pair<Point, UnsignedInteger>(point_i, i);
```
regislebrun, line 1584: Useless. You can replace the whole loop by:
```
pointsPairs[i] = std::pair<Point, UnsignedInteger>((*this)[i], i);
```
as a cont_

mbaudin47: Much simpler indeed. Thank you! Fixed.

lib/test/t_Sample_std.cxx
```diff
@@ -32,6 +32,89 @@ class TestObject : public Sample
...
+  fullprint << "+ Print the nodes and weights" << std::endl;
+  const UnsignedInteger size(weights.getDimension());
+  const UnsignedInteger dimension(nodes.getDimension());
+  fullprint << "  size = " << size << std::endl;
+  fullprint << "  dimension = " << dimension << std::endl;
+  fullprint << "weight, node" << std::endl;
+  for (UnsignedInteger i = 0; i < size; ++i)
+  {
+    fullprint << weights[i] << " : [";
+    for (UnsignedInteger j = 0; j < dimension; ++j)
```
regislebrun, line 48: a trick is to write:
```
String sep("");
for (UnsignedInteger j = 0; j < dimension; ++j, sep=", ")
  fullprint << sep << nodes(i, j);
```

mbaudin47: Quite nice! Thank you. Fixed.

josephmure (approved):

jschueller:
a rebase is needed

jschueller (approved):
a rebase is needed, otherwise ok for me

jschueller:
ok for you @regislebrun ?

lib/src/Base/Stat/SampleImplementation.cxx
```diff
@@ -1573,6 +1573,32 @@ void SampleImplementation::sortUniqueInPlace()
   if (last + 1 < size_) erase(last + 1, size_);
 }
 
+Indices SampleImplementation::argsort(Bool isIncreasing) const
+{
+  Collection< std::pair<Point, UnsignedInteger> > pointsPairs(size_);
+  for (UnsignedInteger i = 0; i < size_; ++i)
```
regislebrun, line 1579: You should swap the test and the for loop. It will save you many tests even if the drawback is to write two for loops.

lib/src/Base/Stat/SampleImplementation.cxx
```diff
@@ -1573,6 +1573,32 @@ void SampleImplementation::sortUniqueInPlace()
...
+    if (isIncreasing)
+    {
+      pointsPairs[i] = std::pair<Point, UnsignedInteger>((*this)[i], i);
+    }
+    else
+    {
+      const Point pointI(static_cast<Point>((*this)[i]));
+      const Point zero(dimension_);
+      const Point oppositePoint(zero - pointI);
+      pointsPairs[i] = std::pair<Point, UnsignedInteger>(oppositePoint, i);
```
regislebrun, line 1590: And something like
```
pointsPairs[i] = std::pair<Point, UnsignedInteger>(pointI * (-1.0), i);
```
This way, no need for two additional points. You can also provide an alternate comparison operator for this case.

regislebrun (approved):
Minor suggestions, good to go after your feedback.

lib/src/Base/Stat/SampleImplementation.cxx
mbaudin47: Of course. Done.

lib/src/Base/Stat/SampleImplementation.cxx
mbaudin47: I created a unitary minus operator for Points. Fixed.

mbaudin47:
I did the rebase, which droped the inclusion of the unit tests. This should be good to go for merging.

PR #1987, opened 2022-03-14
Added permutation parallelisation

JPelamatti:
This PR implements the factorization and parallelization of the computation of permutation-based p-values for the HSIC indices

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
+++ b/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
@@ -57,7 +57,7 @@ void HSICEstimatorConditionalSensitivity::computePValuesAsymptotic() const
 /* Is it safe to compute the weight matrices in parallel? */
 Bool HSICEstimatorConditionalSensitivity::isWeightMatrixParallel() const
 {
-  return weightFunction_.getImplementation()->isParallel();;
+  return weightFunction_.getImplementation()->isParallel();
 }
 
 /* Compute the weight matrix from the weight function */
--- a/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
+++ b/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
@@ -54,6 +54,12 @@ void HSICEstimatorConditionalSensitivity::computePValuesAsymptotic() const
     throw NotYetImplementedException(HERE) << "HSICEstimatorConditionalSensitivity cannot compute asymptotic p-values.";
 }
 
+/* Is it safe to compute the weight matrices in parallel? */
+Bool HSICEstimatorConditionalSensitivity::isWeightMatrixParallel() const
+{
+  return weightFunction_.getImplementation()->isParallel();;
+}
+
 /* Compute the weight matrix from the weight function */
 SquareMatrix HSICEstimatorConditionalSensitivity::computeWeightMatrix(const Sample & Y) const
 {
--- a/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
+++ b/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
@@ -332,7 +332,12 @@ Point HSICEstimatorImplementation::getPValuesPermutation() const
 /* Is it safe to compute the permutation p values in parallel? */
 Bool HSICEstimatorImplementation::isParallel() const
 {
-  return true;
+  bool isParallel_ = true;
+  for (UnsignedInteger i = 0; i < covarianceList_.getSize(); ++i)
+  {
+    isParallel_ = isParallel_ && covarianceList_[i].getImplementation()->isParallel();
+  }
+  return isParallel_;
 }
 
 /* Draw the HSIC indices */
--- a/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
+++ b/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
@@ -176,8 +176,8 @@ struct PValuesPermutationPolicy
   Point & output_;
   const UnsignedInteger dim_;
   const Collection<Sample> shuffleCollection_;
-  const HSICEstimatorImplementation & estimator_;
   const Collection<SquareMatrix> weightMatrixCollection_;
+  const HSICEstimatorImplementation & estimator_;
   Sample xdim_;
   CovarianceModel inputCovariance_;
   CovarianceModel outputCovariance_;
--- a/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
+++ b/lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
@@ -31,6 +31,7 @@
 #include "openturns/Curve.hxx"
 #include "openturns/Pie.hxx"
 #include "openturns/Text.hxx"
+#include "openturns/TBBImplementation.hxx"
 
 BEGIN_NAMESPACE_OPENTURNS
 CLASSNAMEINIT(HSICEstimatorImplementation)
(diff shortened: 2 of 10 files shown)


JPelamatti:
@regislebrun, do you see any reason the parallelization of the computeWeightMatrix method would return slightly different results than the sequential alternative in a few rare cases? 
I have noticed slight differences in results when computing the conditional HSIC indices over large numbers of permutations, and after a bit of digging, it comes down to the collection of weight matrices, that contains a few of the matrices that differ with respect some of their elements (if compared to the previous non-parallel implementation).

jschueller:
results can differ because of ordering of computations
you should still get the same results when you force 1 thread though, else it's a bug

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
```diff
@@ -147,24 +148,96 @@ UnsignedInteger HSICEstimatorImplementation::getPermutationSize() const
...
+  }
+}; /* end struct ComputeWeightMatrixPolicy */
+
+struct PValuesPermutationPolicy
+{
+  Point & output_;
+  const UnsignedInteger dim_;
+  const Collection<Sample> shuffleCollection_;
+  const HSICEstimatorImplementation & estimator_;
+  const Collection<SquareMatrix> weightMatrixCollection_;
```
JPelamatti, line 180:
```suggestion
  const Collection<SquareMatrix> weightMatrixCollection_;
  const HSICEstimatorImplementation & estimator_;
```

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
```diff
@@ -256,6 +329,17 @@ Point HSICEstimatorImplementation::getPValuesPermutation() const
...
 
+/* Is it safe to compute the permutation p values in parallel? */
+Bool HSICEstimatorImplementation::isParallel() const
+{
+  bool isParallel_ = true;
+  for (UnsignedInteger i = 0; i < covarianceList_.getSize(); ++i)
+  {
+    isParallel_ = isParallel_ && covarianceList_[i].getImplementation()->isParallel();
+  }
+  return isParallel_;
```
josephmure, line 340:
```suggestion
  Bool isParallel = true;
  for (UnsignedInteger i = 0; i < covarianceList_.getSize(); ++i)
  {
    isParallel = isParallel && covarianceList_[i].getImplementation()->isParallel();
  }
  return isParallel;
```
 - Bool needs to be written with an uppercase B (because "Bool" it the OT name for "bool", just like "Scalar" is the OT name for "float")
 - `isParallel` is no attribute, so no underscore is needed.

lib/src/Uncertainty/Algorithm/Sensitivity/openturns/HSICEstimatorImplementation.hxx
```diff
@@ -104,6 +106,9 @@ public:
   /** Compute all indices at once */
   virtual void run() const;
 
+  /* Is it safe to compute the permutation p values in parallel? */
+  Bool isParallel() const;
```
jschueller, line 110: it can stay protected since no other class needs it

jschueller:
do you have some benchmarking results ?

JPelamatti:
> do you have some benchmarking results ?

Do you mean regarding the computation time performance (compared to the previous implementation), or regarding the difference in actual numerical results? Either way I do, but apart from actually providing some numbers, they are hard to show as they require compiling and running 2 modified versions of the old and new implementation.

However, I think I might have a possible explanation for the aforementioned issue. The tests that I've been running rely on a SymbolicFunction, which if I remember correctly is not 'thread safe'. Running other tests with a similar weight function but defined as a PythonFunction seems to produce correct results. However, trying to combine PythonFunctions and ComposedFunctions makes the python kernel crash, so I have a new Mistery to solve

JPelamatti:
I have created 2 separate 'isParallel()' methods, one for the actual permutation parallelization, and one for the parallelization of the weight matrices computation. This allows to independently check whether the covariance model and the weight functions are thread safe.

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
```diff
@@ -54,6 +54,12 @@ void HSICEstimatorConditionalSensitivity::computePValuesAsymptotic() const
     throw NotYetImplementedException(HERE) << "HSICEstimatorConditionalSensitivity cannot compute asymptotic p-values.";
 }
 
+/* Is it safe to compute the weight matrices in parallel? */
+Bool HSICEstimatorConditionalSensitivity::isWeightMatrixParallel() const
+{
+  return weightFunction_.getImplementation()->isParallel();;
```
jschueller, line 60: double semicolon

jschueller:
it starts to look good, does the tests still fail ?

JPelamatti:
For now it surely will, as by factorizing the generation of a single set of permutations for all output dimensions, the new version of the code will always produce slightly different results (due to the randomness of the permutation generation). I am currently validating the latest version on a hybrid solution to check whether all tests would pass, and I will then update the new tests on this branch.

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
JPelamatti, line 60:
```suggestion
  return weightFunction_.getImplementation()->isParallel();
```

JPelamatti:
I updated the tests, which now run without errors. However, they appear to time out on the MacOS integration machine. Any idea regarding what might be causing this? These tests should take a fraction of second when everything works smoothly.

jschueller:
I dont know, some kind of deadlock, maybe something else is not thread-safe here ?

JPelamatti:
Apart from the weight function and the covariance models methods, which we made sure are thread safe, the only other 'non trivial' computations that are contained in the parallel loop are the `computeHSICIndex` methods of the HSICUStat and HSICVstat classes, which again should only involve the covariance models methods as well as linear algebra computations.

Moreover, it seems weird that some non-thread safe operation we missed would only cause troubles on the MacOS tests.

JPelamatti:
After some digging, there is 1 of the 6 tests on HSICEstimator, `pyinstallcheck_HSICEstimatorGlobalSensitivity_std`, that actually goes through on MacOS (github). However it takes 88 seconds. The exact same python test takes roughly 2 seconds on Windows (github) and roughly 1 second on Linux (CircleCI), so there clearly is some issue there. Could it be related to a different handling of the TBB/openblas conflicts?
@sofianehaddad , maybe you have some insight regarding parallelization on MacOS? 
Cheers

jschueller:
so not a deadlock but only a timeout ?
it might be related to the way TBBContext disables openblas threading in a nested tbb parallel region since the new loops wrap the covariancemodel::discretize which is already parallelized with tbb
maybe we should spare the innermost openblas calls ?

JPelamatti:
> so not a deadlock but only a timeout ?

Yes, I believe it is an actual timeout due to a parallelization malfunction.

josephmure:
@sofianehaddad @jschueller @regislebrun @mbaudin47 If this works on all platforms except macOS, can it be merged? Is there a way to deactivate this on macOS specifically?

jschueller:
this is likely not macos-specific but rather due to the openblas code (we'll see this on linux too), see my previous comment

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
```diff
@@ -256,6 +329,12 @@ Point HSICEstimatorImplementation::getPValuesPermutation() const
   return PValuesPermutation_;
 }
 
+/* Is it safe to compute the permutation p values in parallel? */
+Bool HSICEstimatorImplementation::isParallel() const
+{
+  return true;
```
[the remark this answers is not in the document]
sofianehaddad: worst! you have at least 2 covariance models (Xi & Y)

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
```diff
@@ -256,6 +329,23 @@ Point HSICEstimatorImplementation::getPValuesPermutation() const
...
+}
+
+/* Is it safe to compute the permutation p values in parallel? */
+Bool HSICEstimatorImplementation::isPvaluesParallel() const
+{
+  Bool isParallel = true;
+  for (UnsignedInteger i = 0; i < covarianceList_.getSize(); ++i)
+  {
+    isParallel = isParallel && covarianceList_[i].getImplementation()->isParallel();
```
sofianehaddad, line 344: `isParallel && covarianceList_[i].getImplementation()->isParallel() && covarianceList_[inputDimension_].getImplementation()->isParallel();` probably
But not sure it is sufficient

sofianehaddad:
Agreed with @jschueller as mac-os implementation uses `openblas` in `github-action`. In your parallelisation process, you embed 2 covariance models and the `discretize` method of each is already parallel!

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
JPelamatti: I guess you mean:
`isParallel && covarianceList_[dim].getImplementation()->isParallel() && covarianceList_[inputDimension_].getImplementation()->isParallel();`.  (this line would replace the entire for cycle)
Indeed; I believe this is a sufficient check, which would eventually allow to perform part of the permutation computations in parallel, and the remaining ones sequentially.

JPelamatti:
Could you suggest me a way of fixing this? thanks

PR #1995, opened 2022-03-25
Misc 3

jschueller:
- ProcessSample: erase/clear methods
- Sample.ImportFromCSVFile to support nan/inf values
- #1997 
- #1994 
- #1993 
- Sample.asDataFrame/FromDataFrame pandas conversion functions

PR #2004, opened 2022-03-30
FindHDF5: `C` Component is required.

MehdiChinoune:
Either search for both components C and CXX or link to hdf5::hdf5_cpp


jschueller:
may I ask what this is supposed to solve ?

jschueller:
moved to #1995, thanks

PR #2005, opened 2022-04-01
Fix default `OPENTURNS_PYTHON_MODULE_PATH` value on MINGW

MehdiChinoune:


CMakeLists.txt
```diff
@@ -662,7 +662,7 @@ if (BUILD_PYTHON)
     if (DEFINED PYTHON_SITE_PACKAGES)
       set (OPENTURNS_PYTHON_MODULE_PATH "${PYTHON_SITE_PACKAGES}")
     else ()
-     if (WIN32)
+     if (MSVC)
```
jschueller, line 665: we're using mingw too, that's going to break our packages as on windows the install path is always "Lib/site-packages"

MehdiChinoune: Python doesn't even support MINGW, which distribution/variant are you using?
Or do you use Python modules built on MINGW with official Python, If so I don't think It's a reliable/wise choice.
In general MINGW uses the same UNIX packaging structure with some exceptions like installing shared libraries in bin directory.

jschueller: we ship mingw-built binaries for use with official cpython, that worked pretty well for us so far ;)

MehdiChinoune: Fedora: mingw64-python https://fedora.pkgs.org/35/fedora-updates-testing-x86_64/mingw64-python3-3.10.4-1.fc35.noarch.rpm.html
ArchLinux: mingw-w64-python https://archlinux.pkgs.org/rolling/jlk-x86_64/mingw-w64-python-3.7.2-1-any.pkg.tar.xz.html

> we ship mingw-built binaries for use with official cpython, that worked pretty well for us so far ;)

That's not guaranteed.

I see you cross-build them. Couldn't you observe that the mingw-python that you are using is not structured like the official MSVC one! https://github.com/openturns/openturns/blob/master/.circleci/run_docker_mingw.sh#L22 (Just to say It's not MSYS specific)

jschueller: no, we dont use mingw-python but the official python msvc binaries repackaged into archlinux

jschueller:
This will yield an unexpected result with non-msvc compiler on native windows platform (and also break our scripts).
IMHO msys is a special case where you want the linux path for a windows target.
You can still define PYTHON_SITE_PACKAGES in the cmake invocation to override the default windows path.
What do you think ?

MehdiChinoune:
>  break our script

I told you msys/msys2 is not a special case, read my comment above. https://github.com/openturns/openturns/pull/2005#discussion_r840301140
Also, It is is not a linux path, Every platform has its own structure.
While MINGW build native Windows binaries, It doesn't mean It should follow MSVC rules. For example its static and imported libraries are suffixed with '.a' and `.dll.a` instead of `.lib`.

I think you should build python modules on Windows using MSVC, instead of this unguaranteed tricks.
I propose to use vcpkg, It contains hundreds of packages and It integrates well with CMake.

I am closing this PR, As It's unlikely to get merged.
Thanks

PR #2006, opened 2022-04-04
Factorization and improvement of permutation-based HSIC p-value computation (sequential)

JPelamatti:
Given that I do not seem able to solve the issues with the fully parallelized computation of the permutation based HSIC p-values (see PR 1987 https://github.com/openturns/openturns/pull/1987 , which was now reverted to a draft) before the next release, I propose these much simpler modifications.

The underlying idea of this PR is to use the same set of permutations of the output vector in order to estimate the p-values for every input dimension. This results in a fair comparison between different inputs p-values, especially for smaller data set, and as an added bonus it shaves off a bit of computation time (while still not solving the computation time issue).

Cheers


lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
```diff
@@ -152,25 +152,37 @@ void HSICEstimatorImplementation::computePValuesPermutation() const
...
+    Scalar HSIC_loc;
     UnsignedInteger count = 0;
+
     for( UnsignedInteger b = 0; b < permutationSize_; ++b)
     {
-      const Sample Yp(shuffledCopy(outputSample_));
-      const SquareMatrix W(computeWeightMatrix(Yp));
-      const Scalar HSIC_loc = computeHSICIndex(xdim, Yp, covarianceList_[dim], covarianceList_[inputDimension_], W);
+	  const Sample Yp(shuffleCollection[b]);
+      const SquareMatrix W(weightMatrixCollection[b]);
```
josephmure, line 177:
```suggestion
      const Sample Yp(shuffleCollection[b]);
      const SquareMatrix W(weightMatrixCollection[b]);
```

josephmure (approved):
Interesting improvement. Thanks Julien!

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
```diff
@@ -152,25 +152,37 @@ void HSICEstimatorImplementation::computePValuesPermutation() const
...
     const Scalar HSIC_obs = computeHSICIndex(xdim, outputSample_, covarianceList_[dim], covarianceList_[inputDimension_], Wobs);
-
+    Scalar HSIC_loc;
     UnsignedInteger count = 0;
+
     for( UnsignedInteger b = 0; b < permutationSize_; ++b)
     {
-      const Sample Yp(shuffledCopy(outputSample_));
-      const SquareMatrix W(computeWeightMatrix(Yp));
-      const Scalar HSIC_loc = computeHSICIndex(xdim, Yp, covarianceList_[dim], covarianceList_[inputDimension_], W);
```
jschueller, line 166: you can keep variable initialization inside loop, this is no c89 :)

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
```diff
@@ -152,25 +152,37 @@ void HSICEstimatorImplementation::computePValuesPermutation() const
 {
   const SquareMatrix Wobs(computeWeightMatrix(outputSample_));
   PValuesPermutation_ = Point(inputDimension_);
+  Sample ShuffledSample;
+  Collection<Sample> shuffleCollection(permutationSize_);
+  Collection<SquareMatrix> weightMatrixCollection(permutationSize_);
+
+  for( UnsignedInteger b = 0; b < permutationSize_; ++b)
+  {
+    ShuffledSample = shuffledCopy(outputSample_);
```
jschueller, line 161: same here, the variable can be declared inside the loop
it should begin with with lowercase

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorImplementation.cxx
josephmure: Is this recommended? Is the goal to have the variables exist only within the scope of the for loop?

josephmure (approved):

regislebrun (approved):

PR #2011, opened 2022-04-07
Hsic fixes

sofianehaddad:


lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
```diff
@@ -33,12 +34,11 @@ HSICEstimatorConditionalSensitivity::HSICEstimatorConditionalSensitivity()
...
   const CovarianceModelCollection & covarianceList
-  , const Sample & X, const Sample & Y
-  , const HSICStat & estimatorType
+  , const Sample & X
+  , const Sample & Y
   , const Function & weightFunction)
-  : HSICEstimatorImplementation(covarianceList, X, Y, estimatorType)
+  : HSICEstimatorImplementation(covarianceList, X, Y, HSICVStat())
 {
-  if (!estimatorType_.isCompatibleWithConditionalAnalysis()) throw InvalidArgumentException(HERE) << "You must use an appropriate HSICStat object for ConditionalSensitivity";
```
jschueller, line 41: seems isCompatibleWithConditionalAnalysis method is useless now

josephmure:
-1 :  Requiring an HSICStat object even in the HSICEstimatorConditionalSensitivity class was a conscious decision, because we wanted the APIs of the various HSICEstimator classes to differ as little as possible.

This way the user is forced to realize that (currently) only the V-stat is compatible with conditional sensitivity analysis.

Tagging @JPelamatti and @vchabri who may have further insight on this subject.

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
sofianehaddad: indeed

sofianehaddad:
@josephmure sometimes forcing to have the same API is useless, especially when what you offer have no sense!
When new estimators will be implemented we can allow again an `HStat` in the ctor

lib/src/Uncertainty/Algorithm/Sensitivity/HSICEstimatorConditionalSensitivity.cxx
sofianehaddad: But I will not drop it

JPelamatti:
Indeed the choice of having the HSICEstimatorConditionalSensitivity constructor take a HSICStat object in the constructor was conscious in order to ensure the coherence between the various HISCEstimator classes. However I also understand that having unused parameters in the constructor is not a good practice. 
I'm overall OK with these changes, I'll just wait to see the compiled doc before approving :)

josephmure:
I agree with the "cosmetic" changes of your PR (aligning the method parameters in the .cxx files for example), but I disagree with the proposed API change.

There was a logic to the order of the arguments in the constructors of the HSICEstimator classes:

- first the essential arguments: input sample, output sample, list of covariance models
- then the choice of estimator (an estimator is always needed, although only one works with conditional analysis)
- finally the class-dependent argument: the function (weight for conditional analysis, none for global analysis, filter for target analysis)

Your change breaks this logic. Imagine that a user starts with some target analysis (which inputs determine the occurrence of a given event)

```python
target = ot.HSICEstimatorTargetSensitivity(X, Y, covlistXY, ot.HSICVStat(), function)
```

Then, they decide they want to see the residual influence of the input variables assuming the event of interest has occurred:

```python
conditional = ot.HSICEstimatorConditionalSensitivity(X, Y, covlistXY, ot.HSICVStat(), function)
```

As you can see, under the current API, the line is virtually the same. With your proposal, the user needs to remember to artificially remove the V-stat from the constructor.

Could you please remove the API change from your commit so we can merge the coding style improvements?

Alternatively, you could **add** a new constructor that works without the HStat argument, without removing the current constructor. But this can wait until 1.20 IMO (it won't break any scripts).

JPelamatti:
After some thought, I agree with @josephmure

regislebrun:
If I sum up things, having an argument only when it is used is artificial, and having it systematically even if it is used only in one configuration is **not** artificial? Apparently wee are not living in the same world...

sofianehaddad:
@josephmure I can not catch your logic.

First you mentioned that an estimator is always needed. As this is not dependent on the essential arguments (data/cov models), one can imagine setting a default one for each kind of study, having a setter in the class and dropping them from the ctor.

Second the function (weight, filtering) seems more important than the estimator (UStat, VStat) for the kind of analysis you mentioned. The influence on results is for sure more likely to depend on the function (than the estimator). Thus in terms of ranking, why doesn't it appear before ?

Third consider then by symmetry that a user starts with some target analysis (which inputs determine the occurrence of a given event)
```python
target = ot.HSICEstimatorTargetSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
```
Then, they decide they want to see the residual influence of the input variables assuming the event of interest has occurred:
```python
conditional = ot.HSICEstimatorConditionalSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
```
Because you allowed the usage of the estimator, you are just pushing user to an exception!

Fourth if you wanted to have the same behaviour, imagine that user might implements this:
```python
target = ot.HSICEstimatorGlobalSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
```
Here because you are aware that function is not used, you **break** the logic above.

All in all, I strongly believe that the argument should be dropped. It is more important to break "a logic" when it makes no sense.

regislebrun (approved):

jschueller:
thoughts on sofiane's answer @josephmure / @JPelamatti ?

josephmure:
> @josephmure I can not catch your logic.

Then I will try to make it clearer.

> First you mentioned that an estimator is always needed. As this is not dependent on the essential arguments (data/cov models), one can imagine setting a default one for each kind of study, having a setter in the class and dropping them from the ctor.

To do that, we would need to know which statistic is generally the best, but @JPelamatti and @vchabri tell me the literature is unclear on this point.

> Second the function (weight, filtering) seems more important than the estimator (UStat, VStat) for the kind of analysis you mentioned. The influence on results is for sure more likely to depend on the function (than the estimator). Thus in terms of ranking, why doesn't it appear before ?

Maybe I shouldn't have used the word "essential argument" in my previous post. All arguments are mandatory and thus "essential", but the first 4 are the same for all HSICEstimator classes. The function (5th argument) is used differently by Target and Conditional, and is not required by Global, so in contrast to all other arguments it is class-specific. So it is rather a question of generic vs class-specific than a question of essential vs non-essential arguments.

> Third consider then by symmetry that a user starts with some target analysis (which inputs determine the occurrence of a given event)
> 
> ```python
> target = ot.HSICEstimatorTargetSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
> ```
> 
> Then, they decide they want to see the residual influence of the input variables assuming the event of interest has occurred:
> 
> ```python
> conditional = ot.HSICEstimatorConditionalSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
> ```
> 
> Because you allowed the usage of the estimator, you are just pushing user to an exception!

That's not a bug, it's a feature! With your changes the same lines would produce a different, more confusing, error message: too many arguments. Then some users would simply remove the `ot.HSICUStat()` argument without understanding that they are implicitely switching to `ot.HSICVStat()`.

> Fourth if you wanted to have the same behaviour, imagine that user might implements this:
> 
> ```python
> target = ot.HSICEstimatorGlobalSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
> ```
> 
> Here because you are aware that function is not used, you **break** the logic above.

What meaning could a "function" argument have for Global sensitivity analysis? The HSICStat argument always has meaning, even in the Conditional class, because we really are using the only HSICStat available, whereas Global sensitivity analysis has no use for any "function" we could pass to it.

I really feel any API change is premature. If we get complaints from users about this, I could change my mind.

Moreover, if you really need your 4-arguments HSICEstimatorConditionalSensitivity constructor, I am fine with you adding it in 1.20 (not 1.19, the freeze was supposed to happen days ago anyway). I just don't want it to erase the 5-arguments constructor.

jschueller:
ok, hard to draw a line here, I wont merge it for 1.19

sofianehaddad:
> > @josephmure I can not catch your logic.
> 
> Then I will try to make it clearer.
> 
> > First you mentioned that an estimator is always needed. As this is not dependent on the essential arguments (data/cov models), one can imagine setting a default one for each kind of study, having a setter in the class and dropping them from the ctor.
> 
> To do that, we would need to know which statistic is generally the best, but @JPelamatti and @vchabri tell me the literature is unclear on this point.
It is a fact. However we can imagine to arbitrarily fix `v-stat` as default one as it might be used everywhere! It is an arbitrary  choice.   
> 
> > Second the function (weight, filtering) seems more important than the estimator (UStat, VStat) for the kind of analysis you mentioned. The influence on results is for sure more likely to depend on the function (than the estimator). Thus in terms of ranking, why doesn't it appear before ?
> 
> Maybe I shouldn't have used the word "essential argument" in my previous post. All arguments are mandatory and thus "essential", but the first 4 are the same for all HSICEstimator classes. The function (5th argument) is used differently by Target and Conditional, and is not required by Global, so in contrast to all other arguments it is class-specific. So it is rather a question of generic vs class-specific than a question of essential vs non-essential arguments.
> 
> > Third consider then by symmetry that a user starts with some target analysis (which inputs determine the occurrence of a given event)
> > ```python
> > target = ot.HSICEstimatorTargetSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, they decide they want to see the residual influence of the input variables assuming the event of interest has occurred:
> > ```python
> > conditional = ot.HSICEstimatorConditionalSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Because you allowed the usage of the estimator, you are just pushing user to an exception!
> 
> That's not a bug, it's a feature! With your changes the same lines would produce a different, more confusing, error message: too many arguments. Then some users would simply remove the `ot.HSICUStat()` argument without understanding that they are implicitely switching to `ot.HSICVStat()`.
> 
> > Fourth if you wanted to have the same behaviour, imagine that user might implements this:
> > ```python
> > target = ot.HSICEstimatorGlobalSensitivity(X, Y, covlistXY, ot.HSICUStat(), function)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Here because you are aware that function is not used, you **break** the logic above.
> 
> What meaning could a "function" argument have for Global sensitivity analysis? The HSICStat argument always has meaning, even in the Conditional class, because we really are using the only HSICStat available, whereas Global sensitivity analysis has no use for any "function" we could pass to it.
> 
In fact, all estimators share these parameters :
```python
 estim = ot.HSICEstimatorXXX(X, Y, covlistXY, ...)
 ```
When I say essential I mean within Global & Target cases where both `u-stat` and `v-stat` might be used, final results should not be (in theory) so different. It is less sensitive than the impact of the weight/filter function. It is related to the first point : why don't we arbitrarily use a `v-stat` as default one? or `u-stat` as default one except for ``conditional? 

> I really feel any API change is premature. If we get complaints from users about this, I could change my mind.
> 
> Moreover, if you really need your 4-arguments HSICEstimatorConditionalSensitivity constructor, I am fine with you adding it in 1.20 (not 1.19, the freeze was supposed to happen days ago anyway). I just don't want it to erase the 5-arguments constructor.

It is not **my need**. The "feature" you are mentioning is not a feature, it is a big mistake!

Whatever now it is too late for `1.19` and this PR will bee strongly updated for `1.20`. In my sense:
 - We drop U/V stat from ctor arguments to keep the same interface
 - We select arbitrarily a default method (`v-stat` when possible otherwise `u-stat` for example)
 - We add setter/getter to the class if not existing  (`setHSICStat`, `getHSICStat`) --> everyone is happy!
 
This will at least keep some coherence between classes, fix a default stat (I think people who handle correctly the impact of `v-stat` vs `u-stat` usage is too small) and allows more advanced users tune the algorithm by fixing the kind of stat)

jschueller:
@josephmure or @JPelamatti do you confirm ?

josephmure (approved):
I still disagree with the PR, but since I was outvoted at the technical committee meeting, the PR should indeed be merged.

jschueller:
thanks

PR #2013, opened 2022-04-07
Improved ConditionedGaussianProcess

### LINKED ISSUE #2012: Cannot sample a ConditionedGaussianProcess on a mesh containing conditioning points
Based on the (slightly shortened) `plot_kriging_simulate.py` example [here](https://openturns.github.io/openturns/latest/auto_meta_modeling/kriging_metamodel/plot_kriging_1d.html#sphx-glr-auto-meta-modeling-kriging-metamodel-plot-kriging-1d-py):
```
import openturns as ot
import openturns.viewer as viewer
ot.Log.Show(ot.Log.NONE)

g = ot.SymbolicFunction(['x'], ['sin(x)'])

x_train = ot.Sample([[x] for x in [1., 3., 4., 6., 7.9, 11., 11.5]])
y_train = g(x_train)
n_train = x_train.getSize()

dimension = 1
basis = ot.ConstantBasisFactory(dimension).build()
covarianceModel = ot.MaternModel([1.]*dimension, 1.5)
algo = ot.KrigingAlgorithm(x_train, y_train, covarianceModel, basis)
algo.run()
krigingResult = algo.getResult()
krigeageMM = krigingResult.getMetaModel()
y_test_MM = krigeageMM(x_test)

xmin = 0.
xmax = 12.
n_test = 101
step = (xmax-xmin)/(n_test-1)
myRegularGrid = ot.RegularGrid(xmin, step, n_test)
process = ot.ConditionedGaussianProcess(krigingResult, myRegularGrid)

trajectories = process.getSample(10)
```
I get, using master:
```
RuntimeError: NotSymmetricDefinitePositiveException : Error: the matrix is not definite positive.
```
This problem was already identified and mentioned in the comments in the script, but with no associated issue.

regislebrun:
Now it implements the same regularization as in GaussianProcess, allowing to use a mesh containing some of the conditioning points. It is an attempt to fix #2012.
With this fix, the cumulated relative error is of order 1e-6 on the known points in the example given in issue #2012.


jschueller:
looks good, what more does it need ? some testing ?

lib/src/Uncertainty/Process/ConditionedGaussianProcess.cxx
```diff
@@ -98,7 +98,47 @@ void ConditionedGaussianProcess::initialize()
...
+        maxEV = covarianceMatrix.computeLargestEigenValueModule();
+        LOGDEBUG(OSS() << "maxEV=" << maxEV);
+        scaling *= maxEV;
+      }
+      cumulatedScaling += scaling ;
+      LOGDEBUG(OSS() << "scaling=" << scaling << ", cumulatedScaling=" << cumulatedScaling);
+      // Unroll the regularization to optimize the computation
+      for (UnsignedInteger i = 0; i < covarianceMatrix.getDimension(); ++i) covarianceMatrix(i, i) += scaling;
+      scaling *= 2.0;
+      continuationCondition = scaling < maxEV * maximalScaling;
```
sofianehaddad, line 134: startingScaling is very small (1e-13) whereas maximalScaling is huge (~1e5)
it is not relevant  your scaling is around 1e5 * maxEV.
Same for GP, GLM 
Is it better to go for a maximumIteration number (like in HMat) ?

regislebrun: Looks like 15 should be replaced by 1e-5 in ResourceMap! Good catch! We can also do some arithmetic and say that we do 27 iterations at max. @jschueller the point is to discuss and agree if an approximate method giving a relative error of order 1e-6 is good enough for a sampling method for a process or if we need something else. @sofianehaddad do you prefer to have a starting scaling and a number of iterations (in which case the maximum scaling has to be computed) or a starting scaling and a final scaling (in which case the number of iterations has to be computed)?

IMO this PR is intended to be part of the 1.19 release as it fix a long standing issue, even if the fix is not perfect. A more in-depth rework is needed to simplify the EllipticalDistribution class, in order to have a lightweight Normal class, and use it in any place one want to sample a Gaussian random vector (i.e. GaussianProcess, ConditionalGaussianProcess and some other classes I missed). It should also include an H-mat version, in order to be able to manipulate huge Normal distributions (and also huge Student distributions).

sofianehaddad: I'd go for your first choice. But given time before release, we can for the moment just fix accurately the respective values

sofianehaddad (approved):

lib/src/Uncertainty/Process/ConditionedGaussianProcess.cxx
```diff
@@ -98,7 +98,47 @@ void ConditionedGaussianProcess::initialize()
...
+      continuationCondition = false;
+    }
+    // If the factorization failed regularize the matrix
+    // Here we use a generic exception as different exceptions may be thrown
+    catch (Exception &)
+    {
+      // If the largest eigenvalue module has not been computed yet...
+      if (maxEV < 0.0)
+      {
+        maxEV = covarianceMatrix.computeLargestEigenValueModule();
```
josephmure, line 125: I am not sure the maximum eigenvalue should be the reference quantify used for to scale lambda. The maximum eigenvalue of the *correlation* matrix necessarily belongs to [1, n], where n is the number of columns in the matrix (while all other eigenvalues can be arbitrarily small). The largest eigenvalue of the *covariance* matrix is also unlikely IMO to give much insight into the proper value of lambda. When the conditioning of the covariance matrix is poor, it is rather because the smallest eigenvalue is too small than because the largest eigenvalue is too large.

Theoretically, I think the criterion should rather be built around the smallest eigenvalue. For example, we could start with lambda equal to the smallest eigenvalue and stop the algorithm when it reaches some multiple of the smallest eigenvalue.

Failing that, we could base the criterion on the determinant of the covariance matrix (det^{1/n} for example) : it is still more informative than the largest eigenvalue.

Regardless, your proposal is a clear improvement so I am in favor of it being merged.

josephmure (approved):

lib/src/Uncertainty/Process/ConditionedGaussianProcess.cxx
josephmure: If a conditioning point is used, then I think it should be excluded from the computation of the covariance matrix: after all, the value of the conditioned process at observation points is deterministic.

regislebrun: @josephmure The use off the maximum eigenvalue as a reference scale is to avoid cancellation of the regulartization during the floating point operations. The condition number is a ratio between the largest and the smallest eigenvalues, so it cannot be reduced to any of these eigenvalues. The way I perform the regularization is to ensure that the matrix is perturbed as slightly as possible in the most important direction, the one associated to the largest eigenvalue, but with an actual perturbation also in this direction.

Your idea of excluding training points from the sampling sounds good to me. I will see if I can implement it on time.

regislebrun: In any case, a regularization loop is a good thing as the covariance matrix may be ill conditioned even if no training point is present in the mesh vertices...

regislebrun: I started a quite complex implementation before to realize that it was trivial to detect if training points were among the mesh vertices. BTW a geometrical check would not have been enough as training points can be in the set of vertices without creating degenerated covariance matrices if the nugget factor is positive. Here are the main comments in ConditionedGaussianProcess::initialize():
```  // Now check if there is any point both in the input sample and in the mesh vertices.
  // They are characterized by
  // a zero cross: a null row and a null column which cross at a zero diagonal element.
  // The trick is to replace this value by the maximum marginal variance, then to
  // remember to set to zero the value at the corresponding index during the sampling phase
```
and:
```
        // Then put the maximum variance on the diagonal.
        // In theory any positive number should work but
        // this way the condition number should be greatly improved.
```

josephmure (changes requested):
In principle, I agree with the changes, however I noticed that the trajectories in the example move a lot between the different iterations of this PR:

# Pushes in this PR

## Last push (commit https://github.com/openturns/openturns/pull/2013/commits/0b1d5e45d5cfa6b6df05d586e2d4db1963b3ae7b)

https://output.circle-artifacts.com/output/job/4b09eec8-6a6c-48df-8ea4-a7f414a5c18e/artifacts/0/html/auto_meta_modeling/kriging_metamodel/plot_kriging_simulate.html?highlight=conditional%20gaussian

image



## Previous push (commit https://github.com/openturns/openturns/commit/4804b8fddf8fc322af510a6f85de648b274d7404)

https://output.circle-artifacts.com/output/job/6f3a8df5-dd49-47ff-b14f-481f93816497/artifacts/0/html/auto_meta_modeling/kriging_metamodel/plot_kriging_simulate.html

image

# OT official branches

The trajectories do not change much between successive releases

## Current master (future 1.19 RC1 - commit https://github.com/openturns/openturns/commit/d3fe0649e198ba120f8d49e418a111989ad2bc60)

https://openturns.github.io/openturns/master/auto_meta_modeling/kriging_metamodel/plot_kriging_simulate.html#sphx-glr-auto-meta-modeling-kriging-metamodel-plot-kriging-simulate-py


image

## OT 1.18

https://openturns.github.io/openturns/1.18/auto_meta_modeling/kriging_metamodel/plot_kriging_simulate.html#sphx-glr-auto-meta-modeling-kriging-metamodel-plot-kriging-simulate-py

image

## OT 1.17

https://openturns.github.io/openturns/1.17/auto_meta_modeling/kriging_metamodel/plot_kriging_simulate.html#sphx-glr-auto-meta-modeling-kriging-metamodel-plot-kriging-simulate-py

image

While it is true that the random generation is altered (more values are generated in the PR because the mesh contains initial sampling points), I am surprised that trajectories vary that much between successive iterations of the PR. This suggests the impact of the regularization is not fully understood, so I no longer think this PR is ready to be merged.

regislebrun:
@josephmure The change is purely due to the shift in the RandomGenerator stream.

Just to calm your fears regarding regularization, with the last commit, there is **no** regularization loop involved in the simulation. You can activate the logs to check this.

One remark regarding your methodology. You cannot just extract pictures from previously published results in the documentation, you have to run the exact same script with different releases of OT on your computer and see if the results are identical. It is what I have done right now, using the master version of `plot_kriging_simulate.py`. If you can see a difference between the figures please tell me where are the moving pixels 

I made significant changes in the example, simplifying the construction of the input sample, removing the duplicated mesh, removing the loop in charge of catching points in the input sample. As a result you cannot expect the curve to be the same, and not even close one to the other. You have to check if the statistical properties of the trajectories are correct using e.g. estimates of the mean, the marginal variance or the covariance function. As a general rule, you cannot rely on the visual appearance of Gaussian process trajectories to see if something is going wrong, except in very degenerated cases which is not the case here!

regislebrun:
Some additional pictures, showing that the trajectories reproduce the correct conditional covariance. Here is the script:
```
import numpy as np
import openturns as ot
import openturns.viewer as viewer
ot.Log.Show(ot.Log.NONE)

# %%
g = ot.SymbolicFunction(['x'], ['sin(x)'])

# %%
x_train = ot.Sample([[x] for x in [1., 3., 4., 6., 7.9, 11., 11.5]])
y_train = g(x_train)
n_train = x_train.getSize()
n_train

# %%
xmin = 0.
xmax = 12.
n_test = 251
step = (xmax-xmin)/(n_test-1)
myRegularGrid = ot.RegularGrid(xmin, step, n_test)
x_test = myRegularGrid.getVertices()
y_test = g(x_test)

# %%
dimension = 1
basis = ot.ConstantBasisFactory(dimension).build()
covarianceModel = ot.MaternModel([1.]*dimension, 1.5)
algo = ot.KrigingAlgorithm(x_train, y_train, covarianceModel, basis)
algo.run()
krigingResult = algo.getResult()
krigingResult

# %%
krigeageMM = krigingResult.getMetaModel()
y_test_MM = krigeageMM(x_test)

vertices = x_test

# %%
def deleteCommonValues(x_train, x_test):
    '''
    Delete from x_test the values which are in x_train so that
    values in x_test have no interect with x_train.
    '''
    x_test_filtered = x_test  # Initialize
    for x_train_value in x_train:
        print("Checking %s" % (x_train_value))
        indices = np.argwhere(x_test == x_train_value)
        if len(indices) == 1:
            print("   Delete %s" % (x_train_value))
            x_test_filtered = np.delete(x_test_filtered, indices[0, 0])
        else:
            print("   OK")
    return x_test_filtered


# %%
vertices_filtered = deleteCommonValues(
    np.array(x_train.asPoint()), np.array(vertices.asPoint()))

# %%
evaluationMesh = ot.Mesh(ot.Sample([[vf] for vf in vertices_filtered]))

# %%
process = ot.ConditionedGaussianProcess(krigingResult, evaluationMesh)

# %%

covMatrix = krigingResult.getConditionalCovariance(evaluationMesh.getVertices())
cov = ot.UserDefinedCovarianceModel(evaluationMesh, covMatrix)
g = cov.draw(0, 0, xmin, xmax)
g.setTitle("Conditional covariance\n" + "OT version=" + ot.PlatformInfo.GetVersion() + " rev." + ot.PlatformInfo.GetRevision())
view = viewer.View(g, (800, 800), square_axes=True)
view.save("cond_cov" + "OT version=" + ot.PlatformInfo.GetVersion() + " rev." + ot.PlatformInfo.GetRevision() + ".png")
view.close()

N = 1000000
trajectories = process.getSample(N)

cov_est = ot.NonStationaryCovarianceModelFactory().build(trajectories)
g = cov_est.draw(0, 0, xmin, xmax)
g.setTitle("Estimated covariance, " + str(N) + " trajectories\n" + "OT version=" + ot.PlatformInfo.GetVersion() + " rev." + ot.PlatformInfo.GetRevision())
view = viewer.View(g, (800, 800), square_axes=True)
view.save("est_cov" + "OT version=" + ot.PlatformInfo.GetVersion() + " rev." + ot.PlatformInfo.GetRevision() + ".png")
view.close()
```
Exact and estimated conditional covariance (PR):
cond_covOT version=1 19dev rev 9e0f292b1
est_covOT version=1 19dev rev 9e0f292b1
The same with master:
cond_covOT version=1 19dev rev 43b3cad80
est_covOT version=1 19dev rev 43b3cad80
The same with OT 1.18:
cond_covOT version=1 18 rev a7240776b
est_covOT version=1 18 rev a7240776b

and I get the exact same figures using OT 1.17 & OT 1.16.

PR #2015, opened 2022-04-08
Misc

jschueller:
Few more commits before the release including what can be included from the KL sensitivity branch:
- Matrix::resize
- check KL algo threshold argument 

Also applied spelling fixes patch from #2016


sofianehaddad (approved):

lib/src/Base/Optim/OptimizationProblemImplementation.cxx
```diff
@@ -1,6 +1,6 @@
 //                                               -*- C++ -*-
 /**
- *  @brief OptimizationProblemImplementation allows to describe an optimization problem
+ *  @brief OptimizationProblemImplementation allows one to describe an optimization problem
  *
  *  Copyright 2005-2021 Airbus-EDF-IMACS-ONERA-Phimeca
```
sofianehaddad, line 5: Could you also update the headers please?

jschueller: the text is the same in both

sofianehaddad: I mean `2005-2021` --> `2005-2022`

jschueller: yes, I'll do that in the release branch to avoid too many diffs here

sofianehaddad: ok thanks

PR #2023, opened 2022-04-13
Release

jschueller:


python/doc/theory/numerical_methods/optimization_algorithm.rst
```diff
@@ -81,9 +81,6 @@ where :math:`\alpha_k` is the parameter *stepmx*.
     - See :doc:`/auto_numerical_methods/optimization/plot_optimization_rosenbrock`
     - See :doc:`/auto_numerical_methods/optimization/plot_optimization_constraints`
     - See :doc:`/auto_numerical_methods/optimization/plot_optimization_nlopt`
-    - See :doc:`/auto_numerical_methods/optimization/plot_optimization_dlib`
```
regislebrun, line 84: No possibility to have conditional inclusion here I suppose?

jschueller: I dont think so, maybe we could try later, or just link to the optimization section

PR #2028, opened 2022-04-14
Add Gibbs example

jschueller:


python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,603 @@
...
+# distribution of all uncertain parameters, including the vector
+# continuous intensities :math:`Y,` seen here as a latent variable.
+# Indeed, adding :math:`Y` to the vector of sampled variables yields a
+# posterior density which is available in closed form, up to an unkown
+# multiplicative factor
+# 
+# .. math::
+# 
+#    \pi(Y,\theta,\tau|Y^{obs})
+#    \propto
```
MerlinKeller, line 209: From @mbaudin47 
> Please make the `propto` symbols aligned, to make the equations more readable: image

MerlinKeller (changes requested):

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,603 @@
...
+
+# %%
+# Plot the simulated dataset.
+plt.plot(X[:,1], Yobs_sim, '.')
+plt.plot(X[:,1], mean_true)
+plt.legend(("Simulations", "Simulation mean"))
+plt.xlabel("X", fontsize=16)
+plt.ylabel("$Y^{obs}$", fontsize=16)
+_ = plt.title("Simulated data")
+
```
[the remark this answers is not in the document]
jschueller: I doubt this will make the code simpler

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,602 @@
...
+# simplified form:
+# 
+# .. math::
+# 
+#     \mu_n =  (\mat{X}^\top (\mat{I_n} + \tau \mat{Q}^{-1})^{-1} \mat{X})^{-1}(X^\top (I_n + \tau \mat{Q}^{-1})^{-1} Y)\\
+#     \Sigma_n = \tau^{-1}(\mat{X}^\top (\mat{I_n} + \tau \mat{Q}^{-1})^{-1} \mat{X})^{-1}.
+# 
+# Explicit simulation from a box-constrained multivariate normal can be
+# done following the algorithm in [koch2019]_:
+#
```
regislebrun, line 312: Have you seen the following comment on arxiv?

"There is an error in the the technical proofs for the proposed algorithms' validity. While the proposed algorithms can in many cases produce approximate draws from the described target distributions, they do not produce exact draws"

regislebrun: The article has been retracted from arxiv. IMO the algorithm is wrong and should not be used as an example.

MerlinKeller: Thanks for your feedback Régis! I hadn't noticed the remark, but as the paper presents several algorithms, it may be aimed to a different method. I checked the proof for the box-constrained multivariate normal algorithm, which is quite straightforward, and it seemed ok to me. I will double check and if it is fine, maybe add it as a theory section.

regislebrun:
@MerlinKeller Could you please send me the article? As it is no more available on arxiv, I can't check anything by myself...

MerlinKeller:
@regislebrun, the paper is still available here: https://arxiv.org/pdf/1907.00057v1.pdf

josephmure:
Hi everybody! I don't think we need Koch et al.'s method at all in the example because the multivariate normal distribution in question has an independent copula. Sampling from a box constrained version of this distribution simply amounts to sampling from the product of 1-dimensional truncated normal distributions.

MerlinKeller:
@regislebrun: After a more rigorous verification, it turns out there IS a big problem with the technical proof! The authors decompose the truncated mvn as a product of 1D conditionals, but neglect the normalizing constant of each conditional, which are hidden in the "\propto" symbol, even though they depend on the conditioning variables... This is quite subtle and easily overlooked, plus experimentally the algorithm gives credible results. So many thanks for pointing this out.
@josephmure: while the parameters are independent a priori by construction, they are still dependent a posteriori. The simplest way to fix that is to update each component one by one, yielding 1D truncated normals.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,602 @@
...
+# 1. Model formulation
+# --------------------
+#
+# We consider the following linear model:
+#
+# .. math::
+#
+#    \vect{Y} = \mat{X} \vect{\theta} + \vect{\varepsilon},
+#
+# where the observation vector :math:`Y \in \mathbb R^{n}` is modeled as
```
josephmure, line 16: Please check the example thoroughly for typos. All vectors and matrices should be bold (which is done with the \vect and \mat macros).

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,602 @@
...
+# Make the precision matrix :math:`\mat{Q}` a diagonal matrix and sample
+# its diagonal coefficients from an :class:`~openturns.Exponential` distribution.
+Q = np.ones((n,1)) + ot.Exponential().getSample(n)
+
+
+# %%
+# Choose values for the parameters :math:`\theta` and :math:`\tau`.
+theta_true = np.array([0.0, 1.0]).reshape(-1,1)
+p = len(theta_true)
+tau_true = .1
```
josephmure, line 115: Please use Black on the script. I think it should turn ".1" into "0.1" in this instance.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,602 @@
...
+# By having all hyperparameters go to :math:`0,` we obtain the following
+# simplified form:
+# 
+# .. math::
+# 
+#     \mu_n =  (\mat{X}^\top (\mat{I_n} + \tau \mat{Q}^{-1})^{-1} \mat{X})^{-1}(X^\top (I_n + \tau \mat{Q}^{-1})^{-1} Y)\\
+#     \Sigma_n = \tau^{-1}(\mat{X}^\top (\mat{I_n} + \tau \mat{Q}^{-1})^{-1} \mat{X})^{-1}.
+# 
+# Explicit simulation from a box-constrained multivariate normal can be
+# done following the algorithm in [koch2019]_:
```
josephmure, line 311:
```suggestion
# done with a simple rejection sampling scheme:
```
To address @regislebrun's remark.

python/doc/bibliography.rst
```diff
@@ -123,6 +123,8 @@ Bibliography
 .. [koay2006] Koay C.G., Basser P.J.,
     *Analytically exact correction scheme for signal extraction from noisy magnitude MR signals*,
     Journal of magnetics Resonance 179, 317-322, 2006.
+.. [koch2019] Koch, H., & Bopp, G. P., *Fast and exact simulation of multivariate normal and Wishart random variables with box constraints*,
+    arXiv preprint arXiv:1907.00057, 2019.
```
josephmure, line 127:
```suggestion
```
We're no longer using the algorithm proposed by Koch et al following @regislebrun's comment:

https://github.com/openturns/openturns/pull/2028/files#r860588932

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,602 @@
...
+        z[:, 0] = ot.TruncatedNormal(0., 1.0, (r[0]-mu[0]) / L[0,0], (s[0]-mu[0]) / L[0,0]).getSample(nsimu)
+        # simulate remaining coordinates
+        for i in range(1, D):
+            #print("simulating coordinate %s out of %s"%(i+1, D))
+            xi = np.dot(L[i:i+1,:i], z[:,:i].T).ravel() + mu[i]
+            z[:, i] = ot.TruncatedNormal(0.0, 1.0, (r[i]-xi[0]) / L[i,i], (s[i]-xi[0]) / L[i,i]).getSample(nsimu)
+        return np.dot(z, L.T) + mu.reshape(1, -1)
+
+    def getRealization(self):
+        return self.truncated_mvn_rvs(self.mu.ravel(), self.Sigma, self.r, self.s)[0]
```
josephmure, line 389:
```suggestion
    def getRealization(self):
        accept = False
        proposaldist = ot.Normal(self.mu, ot.CovarianceMatrix(self.Sigma))
        while not accept:
            proposal = proposaldist.getRealization()
            accept = self.interval.contains(proposal)
        return proposal
```

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,602 @@
...
+
+    def getRange(self):
+        return ot.Interval(self.r, self.s)
+
+    def setParameter( self, parameter ):
+        d = self.getDimension()
+        self.mu = np.array(parameter[:d])
+        self.Sigma = np.array(parameter[d:d+d*d]).reshape(d,d)
+        self.r = np.array(parameter[-2*d:-d])
+        self.s = np.array(parameter[-d:])
```
josephmure, line 348:
```suggestion
        self.s = np.array(parameter[-d:])
        self.interval = ot.Interval(self.r, self.s)
```

josephmure (changes requested):
This review proposes a change to make the sampling of the box-constrained multivariate normal random vector rigorous.

Please address the comments from the older review.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,585 @@
...
+# .. math::
+#
+#     \vect{\mu}_n =  (\mat{X}^\top (\mat{I}_n + \tau \mat{Q}^{-1})^{-1} \mat{X})^{-1}(X^\top (\mat{I}_n + \tau \mat{Q}^{-1})^{-1} \vect{Y})\\
+#     \mat{\Sigma}_n = \tau^{-1}(\mat{X}^\top (\mat{I}_n + \tau \mat{Q}^{-1})^{-1} \mat{X})^{-1}.
+#
+# Explicit simulation from a box-constrained multivariate normal can be
+# done with a simple rejection sampling scheme:
+
+# %%
+# Step 1 : Create associated :class:`~openturns.PythonRandomVector`
```
josephmure, line 313:
```suggestion
# Step 1 : Create associated :class:`~openturns.RandomVector`
```
Sorry for this mistake.

jschueller:
is it ok @josephmure ?

josephmure:
I would like a review from @mbaudin47, who requested most of the changes that were implemented.

josephmure (changes requested):

josephmure:
@regislebrun The sampler issue you raised is solved (we now do a simple rejection scheme). Do you have other concerns or is the example OK for you?

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,584 @@
...
+marginals_trunc = [
+    ot.TruncatedNormal(Yobs_sim[i], 1.0, lower[i], upper[i]) for i in range(len(X))
+]
+
+trunc_cond_Y = ot.ComposedDistribution(marginals_trunc)
+RV_Y = ot.RandomVector(trunc_cond_Y)
+
+
+# %%
+# Extract nominal values of the parameters.
```
MerlinKeller, line 243:
```suggestion
# Step 2 : Link function, giving the parameters of the univariate truncated normals in function of the chain's current state
```

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,584 @@
...
+#
+# .. math::
+#
+#     \widehat\tau = \arg\max_{\tau\in[\tau_{\min};\tau_{\max}]} \mathcal N(\vect{Y}^{obs}|\vect{\mu}_n(\tau) \mat{X}; \tau^{-1} \mat{I}_n + \mat{Q}^{-1}).
+#
+# The optimal value of :math:`\vect{\theta}` is then given by:
+# :math:`\widehat{\vect{\theta}} = \vect{\mu}_n(\widehat\tau).`
+
+# %%
+# Initialization
```
MerlinKeller, line 476:
```suggestion
```

MerlinKeller:
@jschueller, @josephmure : Thanks for all your work! I've corrected some last minor bugs, otherwise I think this PR is ripe for merging.

jschueller:
ok, thanks

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,582 @@
...
+
+ot.Log.Show(ot.Log.NONE)
+ot.RandomGenerator.SetSeed(11123)
+import numpy as np
+
+n = 10
+delta = 0.5
+
+
+X = ot.ComposedDistribution([ot.Dirac(1.0), 10.0 * ot.Normal()]).getSample(n)
```
mbaudin47, line 102: Setting X as the sample from a joint (Dirac, Normal) distribution deserves a comment.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,582 @@
...
+                raise ValueError
+        if Sigma.shape != (d, d):
+            print("covariance matrix not have dimensions (%s,%s)" % (d, d))
+            raise ValueError
+        super(BoxConstrainedNormal, self).__init__(d)
+        self.mu = mu
+        self.Sigma = Sigma
+        self.r = r
+        self.s = s
+
```
mbaudin47, line 336: The line 
```
self.interval = ot.Interval(self.r, self.s)
```
must be inserted there. Otherwise, using `getSample()` fails, because `setParameter()` did not set the `interval` attribute.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,582 @@
...
+post_sample = sample.getMarginal([i for i in range(p + 1)])
+post_sample.setDescription(["$\\theta_1$", "$\\theta_2$", "$\\tau$"])
+
+ks = ot.KernelSmoothing()
+posterior = ks.build(post_sample)
+
+# Optional: change the bandwidth
+bandwidth = ks.getBandwidth()
+for k in range(len(bandwidth)):
+    bandwidth[k] = 1.0 * bandwidth[k]
```
mbaudin47, line 558: I suggest this comment so as the content is as clear as possible. 
```suggestion
    bandwidth[k] = 1.0 * bandwidth[k]  # Change 1.0 to 2.0 to smooth the fitted distribution
```

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,582 @@
...
+
+# %%
+# Draw pairplots of the posterior sample.
+
+# sphinx_gallery_thumbnail_number = 3
+graph = ot.Graph("Cloud pairs graph", " ", " ", True, "")
+graph = ot.VisualTest.DrawPairs(post_sample)
+_ = View(graph)
+
+
```
mbaudin47, line 581: What should we see in this plot? Was the calculation correct or not?

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,582 @@
...
+grid.setTitle("Bayesian inference")
+xlabs = [r"$\theta_0$", r"$\theta_1$", r"$\tau$"]
+for parameter_index in range(3):
+    graph = posterior.getMarginal(parameter_index).drawPDF()
+    graph.setColors(ot.Drawable.BuildDefaultPalette(2))
+    graph.setLegends(["Posterior"])
+    graph.setXTitle(xlabs[parameter_index])
+    grid.setGraph(0, parameter_index, graph)
+_ = View(grid)
+
```
mbaudin47, line 571: I cannot see what is interesting in this plot. Should'nt the posterior be compared to the prior? What comment may we write so that the reader can see what is the meaning of this graphics?

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
```diff
@@ -0,0 +1,582 @@
...
+mi_theta = [i for i in range(p)]
+link_function_theta = ot.PythonFunction(len(x), p + (p + 2) * p, py_link_function_theta)
+rvmh_theta = ot.RandomVectorMetropolisHastings(
+    RV_theta, initialState, mi_theta, link_function_theta
+)
+
+log_pdf_tau = ot.PythonFunction(len(x), 1, py_log_density)
+rwmh_tau = ot.RandomWalkMetropolisHastings(
+    log_pdf_tau, support, initialState, proposal, [p]
```
mbaudin47, line 526:
```suggestion
    log_pdf_tau, support, initialState, proposal_tau, [p]
```

mbaudin47 (commented):
Thank you for the huge improvement. I feel that a conclusion should be added to the example, because the reader cannot currently see what is interesting in the results of this example. The PR was merged before I could review the improved PR.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
josephmure: Good idea! Could you change your suggestion to "Change 1.0 to 2.0 to make the fitted distribution smoother". I don't think "smooth" can be used as a verb.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
josephmure: In addition to this, the column of ones should explicitly be created as such rather through an artificial Dirac distribution.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
josephmure: Maybe we can add a vertical line corresponding to the true value of the parameters (as the data are simulated, the true value is known).

josephmure:
One of the interesting elements about this example is that it combines 2 MetropolisHastings classes (RandomWalkMH and RandomVectorMH) within a Gibbs sampler.

python/doc/examples/calibration/bayesian_calibration/plot_gibbs_simus.py
mbaudin47: I agree: this should be simply deleted.

PR #2037, opened 2022-05-06
Fix HSIC doc

sofianehaddad:
Gaussian kernel expression


sofianehaddad:
@jschueller can you pick this and add it to 1.19 fixes ? Doc should be fixed within this also

jschueller:
ok, edit the right branch next time

sofianehaddad:
Oh I intentionally did this as this is a (too small) fix and I wanted that to be in a more large PR
Next time I will create a PR directly on `jschueller/XXX` for such purpose

PR #2038, opened 2022-05-06
cosmetics

sofianehaddad:
In HSIC, sample size is `n`, not `m`


jschueller:
trivial, there is no "m", merging

PR #2042, opened 2022-05-12
Update Changelog (new class GalambosCopula missing)

vchabri:


jschueller:
yes, sorry we missed that one

Issue #2043, opened 2022-05-12
Some attributes of PythonDistribution are changed during the lifecycle of the object


regislebrun:
The algoCDF attribute of the following class changes depending on the method called:
```
import openturns as ot

class LegendrePolynomialDistribution(ot.PythonDistribution):

    def __init__(self, pdf, xmin, xmax, N):
        super(LegendrePolynomialDistribution, self).__init__(pdf.getInputDimension())
        self.pdf = pdf
        dim = pdf.getInputDimension()
        self.dim = dim
        self.xmin = xmin
        self.xmax = xmax
        self.N = N
        self.algoCDF = ot.GaussLegendre([N]*dim)
        print('init', self.algoCDF.getDiscretization())
        
    def getRange(self):
        print('range', self.algoCDF.getDiscretization())
        return ot.Interval(self.xmin, self.xmax)

    def computePDF(self, X):
        print("PDF", self.algoCDF.getDiscretization())
        return max(0.0, self.pdf(X)[0])

    def computeCDF(self, X):
        print("CDF", self.algoCDF.getDiscretization())
        return self.algoCDF.integrate(self.pdf, ot.Interval(self.xmin, X))[0]

dist =  ot.Distribution(LegendrePolynomialDistribution(ot.SymbolicFunction(['x'], ['x']), [0.0], [1.0], 20))
x = [0.5]
dist.computePDF(x)
dist.computeCDF(x)
dist.getMean()
```
Depending of the method, self.algoCDF is of dimension 2 or 1

PR #2044, opened 2022-05-13
Misc

jschueller:
- #2043


sofianehaddad (approved):

josephmure:
@sofianehaddad I'm going to do it this afternoon.

python/doc/user_manual/response_surface/low_rank_approximation.rst
```diff
@@ -1,18 +0,0 @@
...
-
-.. autosummary::
-    :toctree: _generated/
-    :template: class.rst_t
-
-    CanonicalTensorEvaluation
-    CanonicalTensorGradient
-
-    :template: classWithPlot.rst_t
-    TensorApproximationAlgorithm
```
josephmure, line 15: I don't think you removed src/TensorApproximationAlgorithm_doc.i.in or the other low-rank approximation-related API doc files. Is there a reason why? Do we need to wait until the classes are removed, not just deprecated?

python/test/t_RandomWalkMetropolisHastings_std.py
```diff
@@ -106,3 +106,19 @@
 print("With 504 observations, getRealization() produces {!r}".format(
     real_504[0]))
 ott.assert_almost_equal(real_504[0], 2.0)
+
+
+# this example throws in ot 1.19 as it tries to evaluate the likelihood outside the prior
```
josephmure, line 111:
```suggestion
# this example throws in ot 1.19 as it tries to evaluate the likelihood outside the support of the prior
```

python/doc/user_manual/response_surface/low_rank_approximation.rst
jschueller: yes, lets remove all file at once

jschueller:
@josephmure is it ok ?

josephmure:
Did you remove the src/TensorApproximationAlgorithm_doc.i.in and the other low-rank approximation-related API doc files in the end?

jschueller:
no, I'll remove everything at once

PR #2054, opened 2022-05-20
added IndependentCopulaFactory

YomGarcia:
added IndependentCopulaFactory so we can use it in persalys when infering marginals


lib/src/Uncertainty/Distribution/IndependentCopulaFactory.cxx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+Distribution IndependentCopulaFactory::build() const
+{
+  return buildAsIndependentCopula().clone();
+}
+
+IndependentCopula IndependentCopulaFactory::buildAsIndependentCopula(const Sample &) const
+{
+  return buildAsIndependentCopula();
```
regislebrun, line 63: It should be return IndependentCopula(sample.getDimension). At least you have to return something with a dimension which matches your input sample! BTW I assume that sample is the (missing) name of the argument of the method.

lib/src/Uncertainty/Distribution/IndependentCopulaFactory.cxx
```diff
@@ -58,14 +58,14 @@ Distribution IndependentCopulaFactory::build() const
...
 {
-  return buildAsIndependentCopula();
+  return IndependentCopula(sample.getDimension());
 }
 
-IndependentCopula IndependentCopulaFactory::buildAsIndependentCopula(const Point &) const
+IndependentCopula IndependentCopulaFactory::buildAsIndependentCopula(const Point & parameters) const
 {
-  return buildAsIndependentCopula();
+  return IndependentCopula(parameters.getDimension());
```
regislebrun, line 68: Here you have to check that parameters is of size 0 as there is no parameter for the independent copula. It is thus impossible to specify the dimension using the vector of parameters

python/test/t_IndependentCopulaFactory_std.py
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/env python
+
+from openturns import *
+
+TESTPREAMBLE()
+
+factory = IndependentCopulaFactory()
+factory.build()
+factory.build(Sample(3, 2))
+factory.build(Point())
```
JPelamatti, line 10: Could we remove the expout file and add an IGNOREOUT flag in the test cmakelist?

python/test/t_IndependentCopulaFactory_std.py
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/env python
+
+from openturns import *
```
josephmure, line 3: Could you also write t_IndependentCopulaFactory as a C++ test?

josephmure (changes requested):
Only a few small requests on my part.

For some reason CircleCI did not trigger, so we cannot see the compiled doc. Do you know why? @jschueller ?

python/test/t_IndependentCopulaFactory_std.py
josephmure: Also, "from openturns import *" should be replaced with "import openturns as ot" which is cleaner.

sofianehaddad (approved):
if possible, `squash` the commits. Otherwise @jschueller might do it when merging
ok for me to go

mbaudin47:
What happened with this PR? Is it taken into account in another PR?

PR #2055, opened 2022-05-21
Calibration doc

### LINKED ISSUE #2030: There is no example which shows how to calibrate a model without observed inputs
After several requirements from users, I created an example which shows how to calibrate a parametric model which has no observed input. This was done in the PR #1490. 
The PR was not merged because we decided not to change the API. The example was first integrated in the PR #1565:

https://github.com/openturns/openturns/pull/1565/files#diff-1e1d40da1fa1d29c8eec7774f7bde1e0b3cc53425aa95ea2768bb687e3106b2a

at:

https://github.com/mbaudin47/openturns/blob/a60832bce370880580d320191fae503e52256ade/python/doc/examples/calibration/calibration_withoutobservedinputs.ipynb

but the example has then disappeared.

mbaudin47:
The goal of this PR is to create a new calibration example without observed inputs:
image

This PR fixes #2030.

~~The changelog should be updated before merging.~~

### DIFF
--- a/python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
+++ b/python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
@@ -167,7 +167,7 @@ plt.show()
 # Still, we notice a small discrepancy between the true mode and the model 
 # after calibration, but this discrepancy is very small.
 # Since the model is linear with respect to the parameters :math:`a`, :math:`b`, :math:`c`, 
-# therefore the LinearLeastSquares method performs well. 
-# If the modal was non linear w.r.t. :math:`a`, :math:`b`, :math:`c`, therefore the linear least 
+# the LinearLeastSquares method performs well. 
+# If the model were non linear w.r.t. :math:`a`, :math:`b`, :math:`c`, then the linear least 
 # squares method would not work that well and the parameters would be estimated 
 # with less accuracy.
--- a/python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
+++ b/python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python
+
+"""
+Calibration without observed inputs
+===================================
+"""
+# %%
+# The goal of this example is to present the calibration of a parametric model which 
+# does not have observed inputs.
+# We are going to see how to use the :class:`~openturns.Sample` class 
+# and create an empty sample. 
+# Indeed, this is mandatory in order to fit within the calibration 
+# framework that is used in the library. 
+# In this example there are, however, several outputs to be calibrated.
+
+import openturns as ot
+from matplotlib import pylab as plt
+import openturns.viewer as viewer
+
+# %%
+# The vector of parameters is 
+# :math:`\boldsymbol{\theta} = (a, b, c)^T \in \mathbb{R}^3`. 
+# This model is linear in :math:`\boldsymbol{\theta}` and identifiable.
+# It is derived from the equation:
+# 
+# .. math::
+#     y(x) = a + b x + c x^2
+# 
+# at :math:`x=-1.0, -0.6, -0.2, 0.2, 0.6, 1.0`. 
+# In the model, however, the abscissas are fixed and constant. 
+# Therefore, the parametric model has 3 parameters, no input and 6 outputs 
+# :math:`y_1, ..., y_6`.
+# This produces a set of 5 observations for each output, leading to a total 
+# of 5 (observations per output) * 6 (outputs) = 30 observations.
+g = ot.SymbolicFunction(
+    ["a", "b", "c"],
+    [
+        "a +  -1.0  * b +  1.0  * c",
+        "a +  -0.6  * b +  0.36  * c",
+        "a +  -0.2  * b +  0.04  * c",
+        "a +  0.2  * b +  0.04  * c",
+        "a +  0.6  * b +  0.36  * c",
+        "a +  1.0  * b +  1.0  * c",
+    ],
+)
+outputDimension = g.getOutputDimension()
+print(outputDimension)
+
+# %%
+# We set the true value of the parameters.
+trueParameter = ot.Point([12.0, 7.0, -8.0])
+print(trueParameter)
+
+# %%
+# In order to generate the observed outputs, we create a distribution 
+# in dimension 3, with Dirac (i.e. constant) marginals.
+inputRandomVector = ot.ComposedDistribution([ot.Dirac(theta) for theta in trueParameter])
+
+# %%
+# The candidate value is chosen to be different from the true parameter value. 
+candidate = ot.Point([8.0, 9.0, -6.0])
+calibratedIndices = [0, 1, 2]
+model = ot.ParametricFunction(g, calibratedIndices, candidate)
+
+# %%
+# We consider a multivariate gaussian noise with zero mean, standard deviation 
+# equal to 0.05 and independent copula.
+# The independent copula implies that the errors of the 6 outputs are independent.
+outputObservationNoiseSigma = 1.0
+meanNoise = ot.Point(outputDimension)
+covarianceNoise = [outputObservationNoiseSigma] * outputDimension
+R = ot.IdentityMatrix(outputDimension)
+observationOutputNoise = ot.Normal(meanNoise, covarianceNoise, R)
+print(observationOutputNoise)
+
+# %%
+# Finally, we generate the outputs by evaluating the exact outputs of the 
+# function and adding the noise.
+# We use a sample with size equal to 5.
+size = 5
+# Generate exact outputs
+inputSample = inputRandomVector.getSample(size)
+outputStress = g(inputSample)
+# Add noise
+sampleNoise = observationOutputNoise.getSample(size)
+outputObservations = outputStress + sampleNoise
+
+# %%
+# Now is the important part of this script : there are no input observations. 
+# This is why we create a sample with size equal to 5 and dimension equal to 0. 
+# Even if the calibration model has no input observations, observed inputs 
+# are required by the current programming interface and this is why we have to 
+# create this object. 
+inputObservations = ot.Sample(size, 0)  # Trick
+
+# %%
+# We are now ready to perform the calibration.
+algo = ot.LinearLeastSquaresCalibration(
+    model, inputObservations, outputObservations, candidate, "SVD"
+)
+algo.run()
+calibrationResult = algo.getResult()
+parameterMAP = calibrationResult.getParameterMAP()
+print(parameterMAP)
+
+
+# %%
+# We observe that the estimated parameter is relatively close to 
+# the true parameter value.
+print(parameterMAP - trueParameter)
+
+
+# %%
+# Graphical validation
+# --------------------
+#
+# We now validate the calculation by drawing the exact function and compare 
+# it to the function with estimated parameters.
+fullModel = ot.SymbolicFunction(["a", "b", "c", "x"], ["a + b * x + c * x^2"])
+parameterIndices = [0, 1, 2]
+trueFunction = ot.ParametricFunction(fullModel, parameterIndices, trueParameter)
+print(trueFunction)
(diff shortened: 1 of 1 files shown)


python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+# coding: utf-8
+
+"""
+Calibration without observed inputs
+===================================
+"""
+# %%
+# The goal of this example is to present the calibration of a parametric model which 
+# does not have observed inputs.
+# We are going to see how to use the :meth:`~openturns.Sample` class
```
josephmure, line 11:
```suggestion
# We are going to see how to use the :class:`~openturns.Sample` class 
```
I am surprised that "meth" works as a keyword here, but it should not be used in this context because it creates inappropriate parentheses.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+===================================
+"""
+# %%
+# The goal of this example is to present the calibration of a parametric model which 
+# does not have observed inputs.
+# We are going to see how to use the :meth:`~openturns.Sample` class 
+# and create an empty sample. 
+# Indeed, this is mandatory in order to fit within the calibration 
+# framework that is used in the library. 
+# In this example, there are, however, several outputs to be calibrated.
```
josephmure, line 15:
```suggestion
# In this example there are, however, several outputs to be calibrated.
```
A question of taste, I guess, but I dislike the overuse of commas.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+# %%
+# In order to generate the observed outputs, we create a distribution 
+# in dimension 3, with Dirac (i.e. constant) marginals.
+Theta1 = ot.Dirac(trueParameter[0])
+Theta2 = ot.Dirac(trueParameter[1])
+Theta3 = ot.Dirac(trueParameter[2])
+inputRandomVector = ot.ComposedDistribution([Theta1, Theta2, Theta3])
+
+# %%
+# The candidate value is chosen to be different to the true parameter value.
```
josephmure, line 64:
```suggestion
# The candidate value is chosen to be different from the true parameter value. 
```

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+size = 5
+# Generate exact outputs
+inputSample = inputRandomVector.getSample(size)
+outputStress = g(inputSample)
+# Add noise
+sampleNoise = observationOutputNoise.getSample(size)
+outputObservations = outputStress + sampleNoise
+
+# %%
+# Now is the important part of this script : there is no input observations.
```
josephmure, line 93:
```suggestion
# Now is the important part of this script : there are no input observations. 
```
to be consistent with the plural "observation*s*"

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+inputSample = inputRandomVector.getSample(size)
+outputStress = g(inputSample)
+# Add noise
+sampleNoise = observationOutputNoise.getSample(size)
+outputObservations = outputStress + sampleNoise
+
+# %%
+# Now is the important part of this script : there is no input observations. 
+# This is why we create a sample with size equal to 5 and dimension equal to 0. 
+# Even if the calibration model has no inputs observations, observed inputs
```
josephmure, line 95:
```suggestion
# Even if the calibration model has no input observations, observed inputs 
```
The first "input" is an adjective, not a noun, and I think adjectives are invariable in English.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+plt.show()
+
+
+# %%
+# We notice that the calibration produces a good fit to the data. 
+# Still, we notice a small discrepancy between the true mode land the model 
+# after calibration, but this discrepancy is very small.
+# 
+# Actually, this was made much easier from the fact that the model is 
+# linear with respect to the calibrated parameters a, b and c.
```
josephmure, line 175: I am not sure this remark is appropriate. Also, a b and c should be written as math expressions to keep notations consistent. What is easier? And easier compared to what?

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+        cloud.setLegend("Observations")
+    graph.add(cloud)
+viewer = viewer.View(graph)
+
+plt.show()
+
+
+# %%
+# We notice that the calibration produces a good fit to the data. 
+# Still, we notice a small discrepancy between the true mode land the model
```
josephmure, line 171:
```suggestion
# Still, we notice a small discrepancy between the true mode and the model 
```

josephmure (changes requested):
Good job! This example was sorely missing. I have only a few minor remarks.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+trueParameter = ot.Point([12.0, 7.0, -8.0])
+print(trueParameter)
+
+# %%
+# In order to generate the observed outputs, we create a distribution 
+# in dimension 3, with Dirac (i.e. constant) marginals.
+Theta1 = ot.Dirac(trueParameter[0])
+Theta2 = ot.Dirac(trueParameter[1])
+Theta3 = ot.Dirac(trueParameter[2])
+inputRandomVector = ot.ComposedDistribution([Theta1, Theta2, Theta3])
```
jschueller, line 61: this is less verbose:
`inputRandomVector = ot.ComposedDistribution([ot.Dirac(theta) for theta in trueParameter])`

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+sampleNoise = observationOutputNoise.getSample(size)
+outputObservations = outputStress + sampleNoise
+
+# %%
+# Now is the important part of this script : there is no input observations. 
+# This is why we create a sample with size equal to 5 and dimension equal to 0. 
+# Even if the calibration model has no inputs observations, observed inputs 
+# are required by the current programming interface and this is why we have to 
+# create this object. 
+inputObservations = ot.Sample(size, 0)  # Trick
```
jschueller, line 98: I Calibration classes should allow for an `inputObservations` argument set via `Sample()` : ie no need to set the size, any 0-dim sample should be considered as empty. If the class wants to set the size to simplify the computations it can do so internally.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
...
+calibratedIndices = [0, 1, 2]
+model = ot.ParametricFunction(g, calibratedIndices, candidate)
+
+# %%
+# We consider a multivariate gaussian noise with zero mean, standard deviation 
+# equal to 0.05 and independent copula.
+# The independent copula implies that the errors of the 6 outputs are independent.
+outputObservationNoiseSigma = 1.0
+meanNoise = ot.Point(outputDimension)
+covarianceNoise = ot.Point(outputDimension, outputObservationNoiseSigma)
```
jschueller, line 75: we can avoid Point here, just use plain lists, ie `covarianceNoise=[outputObservationNoiseSigma] * outputDimension`

regislebrun (approved):

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
mbaudin47: I must add the following comment: "Since the model is linear with respect to the parameters a, b, c, therefore the LinearLeastSquares method performs well. If the modal was non linear w.r.t. a, b, c, therefore the linear least squares method would not work that well and the parameters would be estimated with less accuracy."

ChangeLog
```diff
@@ -19,7 +19,7 @@
 === Miscellaneous ===
 
 === Bug fixes ===
-
+ * #2030 (There is no example which shows how to calibrate a model without observed inputs)
```
jschueller, line 22: can drop the bugfix entry

sofianehaddad (approved):
Minor comments left unsolved. Otherwise fine with me

ChangeLog
mbaudin47: Done.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env python
+# coding: utf-8
```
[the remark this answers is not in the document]
mbaudin47: Fixed.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
mbaudin47: Fixed.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
mbaudin47: Fixed.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
mbaudin47: I agree that this is not satisfactory. Your suggestion, however, is not related to this example, but suggests a (relevant) change in the API. So I suggest to do this in a separate PR.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
mbaudin47: Fixed.

python/doc/examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_withoutobservedinputs.py
```diff
@@ -0,0 +1,173 @@
...
+plt.show()
+
+
+# %%
+# We notice that the calibration produces a good fit to the data. 
+# Still, we notice a small discrepancy between the true mode and the model 
+# after calibration, but this discrepancy is very small.
+# Since the model is linear with respect to the parameters :math:`a`, :math:`b`, :math:`c`, 
+# therefore the LinearLeastSquares method performs well. 
+# If the modal was non linear w.r.t. :math:`a`, :math:`b`, :math:`c`, therefore the linear least
```
josephmure, line 171:
```suggestion
# the LinearLeastSquares method performs well. 
# If the model were non linear w.r.t. :math:`a`, :math:`b`, :math:`c`, then the linear least 
```

regislebrun (approved):

josephmure (approved):

PR #2056, opened 2022-05-27
Convert CorrelationAnalysis to a class

### LINKED ISSUE #1753: The SRC function is a misnomer.
In the library, the `CorrelationAnalysis_SRC` static method actually computes the squared SRC. This corresponds to an importance factor of a linear model. 

More precisely, the definition for Standardized Regression Coefficient is:

image

(from https://www.sciencedirect.com/topics/mathematics/standardized-regression-coefficient) or:

image

(from uqftp.ethz.ch/uqlab_doc_pdf/1.4.0/UserManual_Sensitivity.pdf, UQLab User Manual. Sensitivity Analysis, page 4).

However, the function `CorrelationAnalysis_SRC()` from:

http://openturns.github.io/openturns/master/user_manual/_generated/openturns.CorrelationAnalysis_SRC.html

computes :

image

which is the squared SRC coefficient.

The `CorrelationAnalysis_SignedSRC` function computes the SRC:

http://openturns.github.io/openturns/master/user_manual/_generated/openturns.CorrelationAnalysis_SignedSRC.html

And this is why the `SignedSRC` was asked by users: it provides the classical feature that most users know about. I admit that the current feature returns an importance factor which, in some situations, is more interesting. But the content does not match the name, and this is a bug. (In practice, we should consider both indices).  

Changing the content of the method without changing its name would break backward compatibility and lead to silent bugs in users' scripts.

Here is why I suggest the following update:
*  `CorrelationAnalysis_SRC` is tagged deprecated and removed in a future release
* `CorrelationAnalysis_SignedSRC` is tagged deprecated and removed in a future release
* a new `CorrelationAnalysis_StandardizedRegressionCoefficient` is created, which maps to `CorrelationAnalysis_SignedSRC`
* a new `CorrelationAnalysis_LinearImportanceFactor` is created, which maps to `CorrelationAnalysis_SRC`
* Both new pages would define their equations, so that the update content would be clear-cut.

@sofianehaddad : would you agree?

josephmure:
Upgrade documentation (API, theory, example).
Fixes #1753

## Notable API changes

- `SRC` becomes `computeSquaredSRC`.
- `SignedSRC` becomes `computeSRC`.
- `SRRC` is dropped: it called `SRC` (which computed the *squared* SRC) on the ranked samples.
- A new `computeSRRC` method is created: it calls `computeSRC` on the ranked samples.

## Following Vincent's review https://github.com/openturns/openturns/pull/2056#issuecomment-1141906413

- Create a new `CorrelationAnalysis::computeKendallTau` method based on the `CorrelationAnalysis::computePearsonCorrelation` blueprint.


lib/src/Base/Stat/openturns/CorrelationAnalysis.hxx
```diff
@@ -34,37 +35,69 @@ BEGIN_NAMESPACE_OPENTURNS
...
+                      const Sample & outputSample);
+
+  /** Virtual constructor */
+  CorrelationAnalysis * clone() const override;
+
+  /** String converter */
+  String __repr__() const override;
+
+  /** Compute the Pearson correlation coefficient between the input sample and the output sample */
+  const Point computePearsonCorrelation();
```
jschueller, line 58: avoid const for return types when its not a pointer

jschueller:
there should be backward-compatibility paths

lib/src/Base/Stat/openturns/CorrelationAnalysis.hxx
josephmure: Oh yes, that's what I was trying to do.

josephmure: Fixed.

josephmure:
@jschueller This PR does factorize some code (but not much, I agree):

- the dimension checks at the start of each of the old functions are now done in the constructor only,
- the code of the old "SRC" function is considerably simplified in the new "computeSquaredSRC" function because it now relies on "computeSRC" instead of rewriting the same lines of code.

Concerning backward compatibility, I believe @mbaudin47 argued in favor of breaking compatibility in order to make users realize that "computeSRC" does not do what "SRC" previously did. I don't think we reached a consensus on the subject.

regislebrun:
According to the minutes of the meeting, we decided to perform the following changes:
+ ComputeSRC->ComputeSquaredSRC
+ ComputeSignedSRC->ComputeSRC

josephmure:
@regislebrun  the original names were simply "SRC" and "SignedSRC". I added "compute" to the method names in order to make them more consistent with the rest of the OT API.

Concerning SRC, the new "computeSRC" method does what "SignedSRC" previously did, and "computeSquaredSRC" does what the old "SRC" did, so I think I implemented technical committee decision correctly.

For me the question is: should we maintain backwards compatibility despite this change? And if yes, then how? Should I create additional static methods "SRC" and "SignedSRC" to keep the old bad names in 1.20?

vchabri:
Hi all,
Thanks Joseph for this nice PR.

By the way, I have a suggestion: would it be possible to add a method "computeKendallTau" to the CorrelationAnalysis class?

This would be a nice additional feature to this class.

lib/src/Base/Stat/CorrelationAnalysis.cxx
```diff
@@ -26,12 +26,57 @@
...
+}
+
+/* Standard constructor */
+CorrelationAnalysis::CorrelationAnalysis(const Sample & inputSample,
+                                         const Sample & outputSample)
+  : PersistentObject()
+  , inputSample_(inputSample)
+  , outputSample_(outputSample)
 {
   if (outputSample.getDimension() != 1) throw InvalidDimensionException(HERE) << "Error: output sample must be 1D";
```
JPelamatti, line 49: Could we perform a single test here? i.e., `if (outputSample.getDimension() != 1 || inputSample.getDimension() != 1 ) throw InvalidDimensionException(HERE) << "Error: both samples must be 1D";`

JPelamatti, line 49: Could we perform a single test here? i.e., `if (outputSample.getDimension() != 1 || inputSample.getDimension() != 1 ) throw InvalidDimensionException(HERE) << "Error: both samples must be 1D";`

lib/src/Base/Stat/CorrelationAnalysis.cxx
```diff
@@ -43,63 +88,34 @@ Point CorrelationAnalysis::PearsonCorrelation(const Sample & inputSample,
...
+{
+  return computeSRC(inputSample_, outputSample_);
+}
+
+// Compute the Standard Regression Coefficients (SRC) with arguments
+Point CorrelationAnalysis::computeSRC(const Sample & inputSample,
+                                      const Sample & outputSample) const
 {
   const UnsignedInteger dimension = inputSample.getDimension();
   if (!(dimension >= 2)) throw InvalidDimensionException(HERE) << "Error: input sample must have dimension > 1, here dimension=" << dimension;
```
JPelamatti, line 118: An SRC value could be computed between 2 1-dimensional samples, could it not?

JPelamatti, line 118: An SRC value could be computed between 2 1-dimensional samples, could it not?

python/doc/theory/reliability_sensitivity/ranking_src.rst
```diff
@@ -1,70 +1,78 @@
...
+:math:`Y`.
 
-The contribution to the variance :math:`C_i` is sometimes described in
+The square :math:`(C_i)^2`, which is the contribution of :math:`X^i`
+to the variance of :math:`Y`,
+is sometimes described in
 the literature as the “importance factor”, because of the similarity
 between this approach to linear regression and the method of cumulative
-variance quadratic which uses the term importance factor.
+variance which uses the term importance factor.
```
mbaudin47, line 65: There are two missing points here: 
- If the physical model g is linear, then squared SRC coefficients are equal to the Sobol' indices. This is what is important to see in the squared SRC coefficient. On this point, a new pointer to the [Sobol'](https://openturns.github.io/openturns/latest/theory/reliability_sensitivity/sensitivity_sobol.html) help page should be added.
- The quality of the linear regression should be checked before estimating the SRC indices. This is because if the linear regression model is a poor fit to the data, then the SRC coefficients are useless. This is why a new pointer to the [Regression](https://openturns.github.io/openturns/latest/theory/data_analysis/linear_regression.html) help page should be added, with a suggestion: compute the R2 and check it before estimating the SRC indices.

python/doc/theory/reliability_sensitivity/ranking_src.rst
```diff
@@ -1,70 +1,78 @@
...
 
+Standard *Rank* Regression Coefficients (SRRC) are SRC coefficients
+computed on the ranked input variables
+:math:`r\vect{X} = \left( rX^1,\ldots,rX^{n_X} \right)`
+and the ranked output variable :math:`rY`.
```
mbaudin47, line 70: There is a missing point here: SRRC coefficient are useful when the function is monotonic but not necessarily linear. In that case, SRC coefficients cannot be used (because the function is not necessarily linear).

python/test/t_CorrelationAnalysis_std.py
```diff
@@ -23,31 +24,30 @@
...
-src = ot.CorrelationAnalysis.SRC(inputSample, outputSample)
-print("src=", src)
+# Create the CorrelationAnalysis object
+corr_analysis = ot.CorrelationAnalysis(inputSample, outputSample)
 
-# Src with normalize
-src_normalize = ot.CorrelationAnalysis.SRC(inputSample, outputSample, True)
-print("src with normalize=", src_normalize)
+squared_src = corr_analysis.computeSquaredSRC()
+ott.assert_almost_equal(squared_src, [0.901238,0.105409], 1e-5, 0.0)
```
mbaudin47, line 31: The same exact reference values as used in the C++ unit test should be used at this point.

mbaudin47 (changes requested):
Thank your for this good bugfix. :)

It seems that there is still some work to go from _good_ to _great_, as the current code and doc seems to lack of performance, unit test quality and cross-pointers. Perhaps a new meeting could help to clarify the API before going further.

lib/test/t_CorrelationAnalysis_std.cxx
```diff
@@ -61,34 +61,33 @@ int main(int, char *[])
...
-    Point src_normalize(CorrelationAnalysis::SRC(inputSample, outputSample, true));
-    fullprint << "src with normalize=" << src_normalize.__str__() << std::endl;
+    Point squared_src = corr_analysis.computeSquaredSRC();
+    assert_almost_equal(squared_src, Point({0.901238,0.105409}), 1e-5, 0.0);
 
-    Point signed_src(CorrelationAnalysis::SignedSRC(inputSample, outputSample));
-    fullprint << "signed src=" << signed_src.__str__() << std::endl;
+    // Squared SRC with normalize
+    Point squared_src_normalize(corr_analysis.computeSquaredSRC(true));
+    assert_almost_equal(squared_src_normalize, Point({0.895287,0.104713}), 1e-5, 0.0);
```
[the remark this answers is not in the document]
josephmure: This cannot be true: by definition the sum of the normalized squared SRCs must be 1.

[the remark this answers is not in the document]
josephmure: I don't understand: this test concerns the normalized squared SRC (in this case equal to the squared SRC), not the "signed" SRC.

lib/src/Base/Stat/CorrelationAnalysis.cxx
josephmure: Actually, there is no problem at all. I don't know why why this check was made in the first place. Removed now.

lib/test/t_CorrelationAnalysis_std.cxx
```diff
@@ -61,34 +61,33 @@ int main(int, char *[])
...
-    Point src(CorrelationAnalysis::SRC(inputSample, outputSample));
-    fullprint << "src=" << src.__str__() << std::endl;
+    // Create the CorrelationAnalysis object
+    CorrelationAnalysis corr_analysis(inputSample, outputSample);
 
-    // Taking into account normalize
-    Point src_normalize(CorrelationAnalysis::SRC(inputSample, outputSample, true));
-    fullprint << "src with normalize=" << src_normalize.__str__() << std::endl;
+    Point squared_src = corr_analysis.computeSquaredSRC();
+    assert_almost_equal(squared_src, Point({0.901238,0.105409}), 1e-5, 0.0);
```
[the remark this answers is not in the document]
josephmure: OK, I have increased the sample size and used the theoretical values whenever possible. I have added comments stating whether the reference value is exact or approximate.

lib/test/t_CorrelationAnalysis_std.cxx
```diff
@@ -61,34 +61,33 @@ int main(int, char *[])
     Sample inputSample(randomVector.getSample(sampleSize));
     Sample outputSample(analytical(inputSample));
```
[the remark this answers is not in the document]
josephmure: At our meeting, we agreed to leave this to a future PR.

lib/test/t_CorrelationAnalysis_std.cxx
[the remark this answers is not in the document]
josephmure: Done, but I had to increase the tolerance.

lib/test/t_CorrelationAnalysis_std.cxx
```diff
@@ -61,34 +61,33 @@ int main(int, char *[])
...
+    assert_almost_equal(pcc, Point({1.0, 1.0}), 1e-5, 0.0);
 
-    Point prcc(CorrelationAnalysis::PRCC(inputSample, outputSample));
-    fullprint << "prcc=" << prcc.__str__() << std::endl;
+    Point prcc(corr_analysis.computePRCC());
+    assert_almost_equal(prcc, Point({0.98879,0.905344}), 1e-5, 0.0);
 
-    Point pearson(CorrelationAnalysis::PearsonCorrelation(inputSample, outputSample));
-    fullprint << "pearson=" << pearson.__str__() << std::endl;
+    Point pearson(corr_analysis.computePearsonCorrelation());
```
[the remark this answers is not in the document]
josephmure: At our meeting, we agreed to leave this to a future PR.

lib/src/Base/Stat/CorrelationAnalysis.cxx
```diff
@@ -43,63 +88,34 @@ Point CorrelationAnalysis::PearsonCorrelation(const Sample & inputSample,
...
-  // Compute the output variance from the regression coefficients
-  Point src(inputSample.computeVariance());
-  for (UnsignedInteger i = 0; i < dimension; ++ i) src[i] *= linear[i] * linear[i];
-  Scalar varOutput = 0.0;
-  for (UnsignedInteger i = 0; i < dimension; ++ i) varOutput += src[i];
-  if (varOutput > 0.0)
-    src /= varOutput;
-  else
-    throw InvalidArgumentException(HERE) << "No output variance";
+  Point src(computeSRC());
```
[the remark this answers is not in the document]
josephmure: At our meeting, we agreed to leave this to a future PR.

lib/test/t_CorrelationAnalysis_std.cxx
```diff
@@ -61,34 +61,33 @@ int main(int, char *[])
...
-    Point signed_src(CorrelationAnalysis::SignedSRC(inputSample, outputSample));
-    fullprint << "signed src=" << signed_src.__str__() << std::endl;
+    // Squared SRC with normalize
+    Point squared_src_normalize(corr_analysis.computeSquaredSRC(true));
+    assert_almost_equal(squared_src_normalize, Point({0.895287,0.104713}), 1e-5, 0.0);
 
-    Point srrc(CorrelationAnalysis::SRRC(inputSample, outputSample));
-    fullprint << "srrc=" << srrc.__str__() << std::endl;
+    Point src(corr_analysis.computeSRC());
+    assert_almost_equal(src, Point({0.949336,0.324667}), 1e-5, 0.0);
```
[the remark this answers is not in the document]
josephmure: Done.

python/test/t_CorrelationAnalysis_std.py
josephmure: Done.

python/doc/theory/reliability_sensitivity/ranking_src.rst
josephmure: OK, I have added this to the doc page.

python/doc/theory/reliability_sensitivity/ranking_src.rst
josephmure: OK, I have added this to the doc page.

lib/src/Base/Stat/openturns/CorrelationAnalysis.hxx
```diff
@@ -34,37 +35,92 @@ BEGIN_NAMESPACE_OPENTURNS
...
+  Point computePearsonCorrelation(const Sample & sampleX,
+                                  const Sample & sampleY) const;
+
+  // computePCC with arguments
+  Point computePCC(const Sample & sampleX,
+                   const Sample & sampleY) const;
+
+  // computeSRC with arguments
+  Point computeSRC(const Sample & sampleX,
+                   const Sample & sampleY) const;
```
jschueller, line 119: these 3 functions do not use any attributes so they can be marked static

jschueller (approved):
ok, pending static private methods

lib/src/Base/Stat/openturns/CorrelationAnalysis.hxx
```diff
@@ -107,16 +107,16 @@ public:
...
-  Point computePCC(const Sample & sampleX,
-                   const Sample & sampleY) const;
+  static Point computePCC(const Sample & sampleX,
+                          const Sample & sampleY);
 
   // computeSRC with arguments
-  Point computeSRC(const Sample & sampleX,
-                   const Sample & sampleY) const;
+  static Point computeSRC(const Sample & sampleX,
+                          const Sample & sampleY);
```
jschueller, line 119: usually they also begin with uppercase

josephmure: Done.

mbaudin47 (approved):

lib/test/t_CorrelationAnalysis_std.cxx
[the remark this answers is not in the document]
josephmure: OK, I took the reference values for the normalized squared SRC in the test.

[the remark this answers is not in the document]
josephmure: TODO : add the theoretical computation as a comment or a pointer to the appropriate doc page.

josephmure:
TODO : fix API "See also" in Theory.

regislebrun (approved):

lib/src/Base/Stat/CorrelationAnalysis.cxx
```diff
@@ -129,7 +129,7 @@ Point CorrelationAnalysis::computeSRC(const Sample & sampleX,
                                       const Sample & sampleY) const
 {
   const UnsignedInteger dimension = sampleX.getDimension();
-  if (!(dimension >= 2)) throw InvalidDimensionException(HERE) << "Error: input sample must have dimension > 1, here dimension=" << dimension;
```
sofianehaddad, line 100: You mean `sampleY`?

`sampleX` might have dimension >= 2!

josephmure: I don't understand your point. I removed this check after @JPelamatti pointed out it was useless:

https://github.com/openturns/openturns/pull/2056#discussion_r886621668

lib/src/Base/Stat/CorrelationAnalysis.cxx
```diff
@@ -157,7 +157,7 @@ Point CorrelationAnalysis::computePCC(const Sample & sampleX,
...
+  //if (!(dimension >= 2)) throw InvalidDimensionException(HERE) << "Error: input sample must have dimension > 1, here dimension=" << dimension;
```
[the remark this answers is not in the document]
josephmure: The PCC may not be very interesting in the case where `sampleX` is 1-dimensional because it is simply equal to the SRC, but the computation is correct.

lib/src/Base/Stat/CorrelationAnalysis.cxx
sofianehaddad: I did not notice the `!` in `!(dimension >= 2)`. Sorry
Everything is ok

sofianehaddad (approved):

lib/src/Base/Stat/CorrelationAnalysis.cxx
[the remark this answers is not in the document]
vchabri: Relevant remark which might have a strong impact if we consider (in the future) Shapley indices for linear models (LMG, PMVD). Minimizing the number of regressions might drastically improve the computational burden.

lib/src/Base/Stat/CorrelationAnalysis.cxx
```diff
@@ -26,105 +26,163 @@
 
 BEGIN_NAMESPACE_OPENTURNS
 
-/* Compute the Pearson correlation coefficient between the component number index of the input sample and the 1D outputSample */
+/**
```
vchabri, line 29: Sorry for the minor remark...
Following what @JPelamatti proposed, I was more fond of the names "firstSample / secondSample" as these metrics are "two-sample statistics". Using "sampleX / sampleY" still denotes (implicitly) an input-output relationship which might be confusing. Again, this is a tiny detail.

python/doc/theory/reliability_sensitivity/ranking_pcc.rst
```diff
@@ -1,51 +1,52 @@
...
 
-Uncertainty ranking: PCC
-------------------------
+Uncertainty ranking: PCC and PRCC
+---------------------------------
 
-This method deals with analyzing the influence the random vector
+Partial Correlation Coefficients
+deal with analyzing the influence the random vector
 :math:`\vect{X} = \left( X^1,\ldots,X^{n_X} \right)` has on a random
```
vchabri, line 8: @jschueller / @mbaudin47 : why do we use here a superscript-based notation for the input variables? Regarding the remainder of the documentation, one should use a subscript (e.g., X_{j}^{(i)}, for the i-th sample of the j-th variable).

python/doc/theory/reliability_sensitivity/ranking_src.rst
```diff
@@ -1,70 +1,88 @@
...
 the literature as the “importance factor”, because of the similarity
 between this approach to linear regression and the method of cumulative
-variance quadratic which uses the term importance factor.
+variance which uses the term importance factor.
 
+It is a good idea to check the quality of the :ref:`linear regression <linear_regression>`
+before estimating the SRC coefficients: if the linear regression model
+is a poor fit to the data, then the SRC coefficients are useless.
+
+Note that if there exists a map :math:`g` such that :math:`Y=g(X^1, ..., X^n)`,
```
vchabri, line 71: X^n should be X_{n_X}

vchabri (changes requested):
Good job @josephmure ! Thanks!

lib/src/Base/Stat/CorrelationAnalysis.cxx
josephmure: Done. Also removed further references to input/output variables in the `CorrelationAnalysis` API doc.

python/doc/theory/reliability_sensitivity/ranking_src.rst
josephmure: Changed to X^{n_X} for the time being. I'm waiting for @mbaudin47 or @jschueller's answer to your other comment https://github.com/openturns/openturns/pull/2056#discussion_r911165155 before taking further action.

python/doc/theory/reliability_sensitivity/ranking_pcc.rst
jschueller: I guess its a mistake, feel free to change it.

josephmure: OK, done.

josephmure:
For some reason, continuous integration did not trigger with my latest commit but I have taken care of the super/subscript related stuff. If you want to see the result for yourself @vchabri you can try to visualize the relevant files only:

1. Pull the branch locally
2. Inside the root openturns folder, run:

```
python3 utils/docfast.py --nodiff build-onlydoc ranking_src.rst ranking_pcc.rst
```

Then you should find the resulting HTML at build-onlydoc/install/theory/reliability_sensitivity/reliability_sensitivity.html

vchabri (approved):

JPelamatti (approved):

PR #2062, opened 2022-06-08
getSobolGroupedIndex does not document the marginal index

mbaudin47:
This PR documents the `FunctionalChaosSobolIndices` class.

The doc of the `getSobolIndex` and `getSobolTotalIndex` methods document the optional `out_marginal` input argument:

image

but `getSobolGroupedIndex` and `getSobolGroupedTotalIndex` methods do not:

image

This PR fixes that issue.
- Adds doc of `marginalIndex` for grouped Sobol' indices. 
- Uses consistent notations across the methods.

### DIFF
--- a/python/src/FunctionalChaosSobolIndices_doc.i.in
+++ b/python/src/FunctionalChaosSobolIndices_doc.i.in
@@ -94,14 +94,14 @@ Parameters
 i : int or sequence of int, :math:`0 \leq i < d - 1`
     Indice(s) of the variable(s) we want the associated Sobol indices. :math:`d`
     is the dimension of the input variables.
-out_marginal : int
-    Output marginal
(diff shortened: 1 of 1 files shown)


jschueller (approved):

mbaudin47:
> the getGroupedSobolIndex with the int argument is questionable since it is equivalent to getSobolIndex (same for total) do you think we should drop it ?

I agree. The `getGroupedSobolIndex(int)` is mathematically and numerically equivalent to `getSobolIndex(int)`. It makes the API unconsistent. Removing these should be done both at the doc and C++ levels. Do you think that I should integrate these removals into the current PR?

regislebrun (approved):

jschueller:
lets do it in another PR

Issue #1473, opened 2020-05-05
FunctionalChaosResult should provide the used input/output samples


sofianehaddad:
`FunctionalChaosResult` does not provide `get{Input, Output}Sample` methods.
This might be useful if we relied on a DOE + Function.

As an example, for some post-processing such as LOO estimate, we need these methods.

jschueller:
+1, the samples are always known or generated at some points, even with the weird FCE ctors

PR #2079, opened 2022-06-28
misc chaos

jschueller:
reflect in the api that metamodels are built on data, not from functions, following yesterdays meeting:
- Deprecate FunctionalChaos(Function, ...) ctors, FunctionalChaosResult.getComposedModel, MetaModelResult.getModel
- Add sample ctors, accessors to MetaModelResult & MetaModelAlgorithm


sofianehaddad (approved):

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_ishigami.py
```diff
@@ -95,8 +95,7 @@ def plotXvsY(sampleX, sampleY, figsize=(15, 3)):
 multivariateBasis = ot.OrthogonalProductPolynomialFactory(
     [im.X1, im.X2, im.X3])
 selectionAlgorithm = ot.LeastSquaresMetaModelSelectionFactory()
-projectionStrategy = ot.LeastSquaresStrategy(
-    inputTrain, outputTrain, selectionAlgorithm)
+projectionStrategy = ot.LeastSquaresStrategy(selectionAlgorithm)
```
josephmure, line 98: Unless I am mistaken, you are using an documented constructor here (and at other places):
image

You seem to be using `LeastSquaresStrategy(approxAlgoImpFact)`.

jschueller: yes, added this one to the list

lib/src/Uncertainty/Algorithm/MetaModel/MetaModelAlgorithm.cxx
```diff
@@ -43,14 +44,39 @@ MetaModelAlgorithm::MetaModelAlgorithm()
...
+  : MetaModelAlgorithm(inputSample, outputSample, Dirac(Point(inputSample.getDimension(), 0.0)))
+{
+  // Nothing to do
+}
+
+/* Constructor with parameters */
+MetaModelAlgorithm::MetaModelAlgorithm(const Sample & inputSample,
+                                       const Sample & outputSample,
+                                       const Distribution & distribution,
+                                       const Function & model)
```
regislebrun, line 59: I don't see the point of keeping a model if we have an output sample.

regislebrun (changes requested):
If this PR is supposed to remove the model based parametrization of the meta-model algorithms, then I don't understand why we would have to specify both an output sample and a model in several places.

lib/src/Uncertainty/Algorithm/MetaModel/MetaModelAlgorithm.cxx
jschueller: The model accessors are deprecated; we keep them for this release then remove them (and the attributes/ctors) in the new release.

jschueller:
> If this PR is supposed to remove the model based parametrization of the meta-model algorithms, then I don't understand why we would have to specify both an output sample and a model in several places.

The model accessors are deprecated; we keep them for this release then remove them (and the attributes/ctors) in the new release.

regislebrun:
Ok but why do we need to actually store the model? The setModel() accessor could simply emit a warning, and the getModel() accessor could build a DatabaseFunction on demand, no?

sofianehaddad:
@jschueller is there a way to fix #1473 with this PR?

jschueller:
Indeed it fixes #1473

josephmure (approved):

lib/src/Uncertainty/Algorithm/MetaModel/MetaModelAlgorithm.cxx
```diff
@@ -43,14 +44,39 @@ MetaModelAlgorithm::MetaModelAlgorithm()
   // Nothing to do
 }
 
+/* Constructor with parameters */
+MetaModelAlgorithm::MetaModelAlgorithm(const Sample & inputSample,
+                                       const Sample & outputSample)
+  : MetaModelAlgorithm(inputSample, outputSample, Dirac(Point(inputSample.getDimension(), 0.0)))
```
regislebrun, line 50: Why did you choose a Dirac distribution here? A UserDefined one would have been more adapted from a statistical point of view. Is it related to the cost of constructing such a distribution?

regislebrun (approved):
Only a minor question, otherwise good to go for me

lib/src/Uncertainty/Algorithm/MetaModel/MetaModelAlgorithm.cxx
jschueller: I dont think its actually used, but UserDefined is ok too

mbaudin47 (commented):
I am sorry, but I did not find the time to perform the requested review.

PR #2092, opened 2022-07-13
Distribution quantiles

### LINKED ISSUE #2084: computeScalarQuantile returns inf
```
import openturns as ot
coll = ot.DistributionFactory.GetContinuousUniVariateFactories()
coll.add(ot.DistributionFactory.GetDiscreteUniVariateFactories())
for factory in coll:
    distribution = factory.build()
    for p in [0.0,1.0]:
        q = distribution.computeQuantile(p)[0]
        try:
            sq = distribution.computeScalarQuantile(p)
        except Exception as exc:
            print(distribution.getImplementation().getClassName(), p, str(exc))
        if abs(q-sq)> 1e-15:
            print(distribution.getImplementation().getClassName(), p, q, sq, distribution)

99999999999999.42 inf Burr(c = 1, k = 1)
7.739256319504374 8.209536151601387 Chi(nu = 1)
59.89608837898838 67.39648382445012 ChiSquare(nu = 1)
32.23619130191664 inf Exponential(lambda = 1, gamma = 0)
100079991719343.84 -inf Frechet(beta = 1, alpha = 1, gamma = 0)
32.23619130191664 36.04365338911715 Gamma(k = 1, lambda = 1, gamma = 0)
36.04365338911715 inf GeneralizedPareto(sigma = 1, xi=0, u=0)
32.236191301916634 inf Gumbel(beta = 1, gamma = 0)
31.543843718786892 inf Laplace(mu = 0, lambda = 1)
32.23619130191663 inf Logistic(mu = 0, beta = 1)
2101.965404514274 3380.8778196094727 LogNormal(muLog = 0, sigmaLog = 1, gamma = 0)
7.650628092935269 8.125890664701906 Normal(mu = 0, sigma = 1)
4503599627370496.0 inf Pareto(beta = 1, alpha=1, gamma=0)
8.029469634031459 inf Rayleigh(beta = 1, gamma = 0)
47952.7571796676 170608.3076215258 Student(nu = 3, mu = 0, sigma = 1)
32.23619130191664 inf WeibullMin(beta = 1, alpha = 1, gamma = 0)
```

also happens for some discrete ones or at the lower bound

computeQuantile checks for quantileEpsilon_ then calls computeScalarQuantile:
```
Point DistributionImplementation::computeQuantile(const Scalar prob,
    const Bool tail,
    Scalar & marginalProb) const
{
  const Scalar q = tail ? 1.0 - prob : prob;
  marginalProb = q;
  // Special case for bording values
  if (prob < quantileEpsilon_) return (tail ? getRange().getUpperBound() : getRange().getLowerBound());
  if (prob >= 1.0 - quantileEpsilon_) return (tail ? getRange().getLowerBound() : getRange().getUpperBound());
  // Special case for dimension 1
  if (dimension_ == 1) return Point(1, computeScalarQuantile(prob, tail));
```

should it be considered unsafe and made private ?

an interesting thing is that if I replace computeScalarQuantile by computeQuantile in MarginalTransformationEvaluation.cxx
it makes t_AdaptiveDirectionalStratification_flood.py fails as the TruncatedDistribution gives a realization at exactly to the zero lowerbound (ok if I specify a small positive bound).

jschueller:
Attempt to fix #2084

In many cases we just need to clip the proba level  in [MinScalar, 1-ScalarEpsilon] because the quantile formula involves log.
In some methods we also use log1p instead of log for the tail case.
Now with computeScalarQuantile we can safely explore the full range of quantiles and not be limited by quantileEpsilon_.

```
Point DistributionImplementation::computeQuantile(const Scalar prob,
{
  const Scalar q = tail ? 1.0 - prob : prob;
  marginalProb = q;
  
  // Special case for bording values
  if (prob < quantileEpsilon_) return (tail ? getRange().getUpperBound() : getRange().getLowerBound());
  if (prob >= 1.0 - quantileEpsilon_) return (tail ? getRange().getLowerBound() : getRange().getUpperBound());

  // Special case for dimension 1
  if (dimension_ == 1) return Point(1, computeScalarQuantile(prob, tail));
...
```

For Frechet the formula is invalid at 0 as it does pow(inf) and still return 0, may need a special case to get the old bound instead the one at MinScalar.
When we enable it at the computeQuantile level it seems to fix a minimum volume interval testcase for the truncated normal.


jschueller:
I added a new `SpecFunc::Clip01(Scalar p, Bool tail)` that does the clipping and (1-p), helps when formulas are symmetric.

I went down several rabbit holes:
- enhanced computeUnivariateMinimumVolumeInterval: use Cobyla + correct optimization interval + a small heuristic to decide the starting point on the other side of the tail from the value of the skewness (helps TruncatedDistribution)
- A corner-case issue in cobyla where x* is not the optimum and even not an actual called x point (but close)
- An inf problem in invincgam for the Chi distribution (used by RegularizedIncompleteGammaInverse when boost is not available)
now the distributions computeScalarQuantile methods should be now much more robust to continuity problems

josephmure (changes requested):
-1 concerning the "Distribution quantiles" commit, see the discussion in https://github.com/openturns/openturns/issues/2084

I haven't looked at the rest of the PR yet.

jschueller:
it looks we are going to disagree again whether or not openturns should handle inf/nan :]

one practical reason to avoid inf (or nan) here is that its unsuitable for computation by the higher level algorithms (computeScalarQuantile is very low-level), like optimization to compute minimal proba intervals

also some of these inf resulted from a completely out of range computation, where the limit quantile is finite,
but we just evaluated a formula with an unclipped proba

mbaudin47:
One difficult issue is to have reference numerical values. This is because that particular information is costly to get.

For your information, I heavily tested distributions and, in particular, quantiles of distributions in [distfun](https://gitlab.com/scilab/forge/distfun). Each unit test is based on a CSV data file which contains reference values. For example, the reference file for the truncated normal is available at [tnorm.dataset.csv](https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/tnorm/tnorm.dataset.csv). The execution of the unit tests is performed in [tnorminv.tst](https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/tnorm/tnorminv.tst). Its content is very simple : it loops over the rows and compare the computed value and the reference value from the file. This allows to access to output from e.g. symbolic calculations computed from Mapl, without actually evaluating Maple. The actual unit test uses only the data file.

Many distributions have their reference file:
- Beta : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/beta/beta.dataset.csv
- Binomiale : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/bino/binomiale.dataset.csv
- Chi2 : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/chi2/chi2.dataset.csv
- Extreme Value : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/ev/ev.dataset.csv
- Exponential : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/exp/exp.dataset.csv
- F : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/f/f.dataset.csv
- Gamma : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/gam/gamma.dataset.csv
- Geometric : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/geo/geo.dataset.csv
- Hypergeometric : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/hyge/hypergeometric.dataset.csv
- Kolmogorov-Smirnov : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/ks/ks.dataset.csv
- Log-Normal : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/logn/logn.dataset.csv
- Log-Uniform : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/logu/logu.dataset.csv
- Multivariate normal : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/mvn/mvn.dataset.csv
- Negative binomial : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/nbin/nbin.dataset.csv
- Non central F : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/ncf/ncf.dataset.csv
- Non central T : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/nct/nct.dataset.csv
- Non central chi2 : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/ncx2/ncx2.dataset.csv
- Normal : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/norm/normal.R.dataset.csv
- Poisson : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/poiss/poisson.dataset.csv
- T : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/t/t.dataset.csv
- Uniform : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/unif/uniform.dataset.csv
- Weibul : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/wbl/wbl.dataset.csv
- Truncated Normal : https://gitlab.com/scilab/forge/distfun/-/blob/master/tests/unit_tests/tnorm/tnorm.dataset.csv

For some distributions, other reference files are available, including Yalta's benchmark.

A straightforward method would be to get the reference files into OT's source code and use them to check our distriubutions. This could be done during the coding party for example. There is no INF/NAN data in the references, so that no discussion on that point is necessary.

PR #2099, opened 2022-07-21
Nais update

l-brevault:
NAIS update: modification of the auxiliary density to account for multivariate Mixture of gaussians


l-brevault:
Sorry manipulation error

PR #2100, opened 2022-07-21
Update NAIS algorithm

l-brevault:
Hello,

We propose a modification of the auxiliary density in NAIS algorithm in order to use a multivariate gaussian mixture instead of a product of univariate gaussian mixture. It improves the generation of relevant samples with respect to the failure domains.
We modified two files: NAIS.cxx with the proposed enhancements and the corresponding test file.
The documentation file stays valid. 

Regards,
Mathieu & Loïc

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
@@ -96,11 +96,11 @@ Distribution NAIS::computeAuxiliaryDistribution(const Sample & sample,
   Collection<Distribution> collectionOfDistribution(numberOfSample);
   for (UnsignedInteger i = 0; i < numberOfSample ; ++i)
   {
-    const Point  mean_normal =  sample[i];
+    const Point meanNormal(sample[i]);
     
-    collectionOfDistribution[i] = Normal(mean_normal, silverman);
+    collectionOfDistribution[i] = Normal(meanNormal, silverman);
   }
-  Distribution auxiliaryDistribution = Mixture(collectionOfDistribution, weights);
+  const Mixture auxiliaryDistribution(collectionOfDistribution, weights);
 
   return auxiliaryDistribution;
 }
--- a/lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
@@ -92,17 +92,17 @@ Distribution NAIS::computeAuxiliaryDistribution(const Sample & sample,
   Collection<Distribution> margins(dimensionSample);
 
   // Computation of auxiliary distribution using ot.Mixture
-  for (UnsignedInteger k = 0; k < dimensionSample ; ++k)
+  const UnsignedInteger numberOfSample = getMaximumOuterSampling() * getBlockSize();
+  Collection<Distribution> collectionOfDistribution(numberOfSample);
+  for (UnsignedInteger i = 0; i < numberOfSample ; ++i)
   {
-    const UnsignedInteger numberOfSample = getMaximumOuterSampling() * getBlockSize();
-    Collection<Distribution> collectionOfDistribution(numberOfSample);
-    for (UnsignedInteger i = 0; i < numberOfSample ; ++i)
-    {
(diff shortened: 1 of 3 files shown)


lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
```diff
@@ -92,17 +92,17 @@ Distribution NAIS::computeAuxiliaryDistribution(const Sample & sample,
...
-    const UnsignedInteger numberOfSample = getMaximumOuterSampling() * getBlockSize();
-    Collection<Distribution> collectionOfDistribution(numberOfSample);
-    for (UnsignedInteger i = 0; i < numberOfSample ; ++i)
-    {
-      collectionOfDistribution[i] = Normal(sample(i, k), silverman[k]);
-    }
-    margins[k] = Mixture(collectionOfDistribution, weights);
-  } // for k
-  return ComposedDistribution(margins);
+    const Point  mean_normal =  sample[i];
```
regislebrun, line 99: For complex types (such as Point) use the constructor. Also use camelCase for the variable names:
``` 
const Point meanNormal(sample[i]);
```

lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
```diff
@@ -92,17 +92,17 @@ Distribution NAIS::computeAuxiliaryDistribution(const Sample & sample,
...
+    
+    collectionOfDistribution[i] = Normal(mean_normal, silverman);
+  }
+  Distribution auxiliaryDistribution = Mixture(collectionOfDistribution, weights);
```
regislebrun, line 103: Same ramrk here, plus const correctness. If `auxiliaryDistribution` is not modified afterward, use:
```
const Distribution auxiliaryDistribution(Mixture(collectionOfDistribution, weights));
```
or even better:
```
const Mixture auxiliaryDistribution(collectionOfDistribution, weights);
```

regislebrun (changes requested):
Good work. Only minor remarks regarding the coding rules.

lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
l-brevault: Thank you, modification has been done accordingly

lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
l-brevault: Thank you, modification has been done accordingly

jschueller (approved):
you forgot to update the c++ test value
else all good

python/test/t_NAIS_std.py
```diff
@@ -72,4 +72,4 @@ def four_branch(x):
 Nais_algo.run()
 NAIS_result = Nais_algo.getResult()
 
-assert_almost_equal(NAIS_result.getProbabilityEstimate(), 0.00215688)
+assert_almost_equal(NAIS_result.getProbabilityEstimate(), 0.00227031)
```
jschueller, line 75: you reverted the python test

l-brevault: Yes indeed, thank you !
The correction has been made

regislebrun (approved):

sofianehaddad (approved):

lib/test/t_NAIS_std.cxx
```diff
@@ -89,6 +89,6 @@ int main()
   algoNais.run();
 
   const NAISResult resultNais(algoNais.getResult());
-  OT::Test::assert_almost_equal(resultNais.getProbabilityEstimate(), 0.00227031);
+  OT::Test::assert_almost_equal(resultNais.getProbabilityEstimate(), 0.00215688);
```
josephmure, line 92: It would be better to manually specify the tolerances IMO:

`assert_almost_equal(tested, reference, relative_tolerance, absolute_tolerance)`

I never know the default values!

josephmure (changes requested):

lib/test/t_NAIS_std.cxx
jschueller: by default rtol = 1.0e-5, atol = 1.0e-8; these are fine in many cases
seems everyone approved and and there's only this minor remark left then lets merge it

Issue #2101, opened 2022-07-21
drop FunctionalChaosSobolIndices::summary


jschueller:
FunctionalChaosSobolIndices::summary method can be moved to the `__str__` method, like LinearModelAnalysis does

Issue #2098, opened 2022-07-21
LinearModelStepwiseAlgorithm null basis


jschueller:
if a null basis term is passed to LMSA then the maxX_ matrix has a column of zeros
then in buildCurrentMatrixFromIndices it may be copied to currentX_ depending on the indices chosen
finallly computeQR yields nan from currentX_ in computeLogLikelihood that propagate until the end
or, it can also crash if DBG logs are enabled because the algo tries to go forward when there is no longer indices to add

```
import openturns as ot
#ot.Log.Show(ot.Log.DBG) # glibcxx assertion
dist = ot.ComposedDistribution([ot.Uniform(0., 10.)]*2)
f = ot.SymbolicFunction(['x1', 'x2'], ["0.5*x1 + x2+ 7"])
X = ot.LHSExperiment(dist, 10).generate()
X = ot.MonteCarloExperiment(dist, 10).generate()
Y = f(X)
i_min = [0]
coll = []
coll.append(ot.SymbolicFunction(['x1', 'x2'], ["0"]))
coll.append(ot.SymbolicFunction(['x1', 'x2'], ["x1"]))
coll.append(ot.SymbolicFunction(['x1', 'x2'], ["x2"]))
basis = ot.Basis(coll)
algo = ot.LinearModelStepwiseAlgorithm(X, basis, Y, i_min)
algo.run()
result = algo.getResult()
print(result)

> class=LinearModelResult beta=class=Point name=Unnamed dimension=3 values=[-nan,-nan,-nan] formula=[[x1,x2]->[0],[x1,x2]->[x2],[x1,x2]->[x1]]
```

Issue #2094, opened 2022-07-13
getSampleAtVertex() is not robust


mbaudin47:
It seems to me that the method `getSampleAtVertex(nodeIndex)` does not check its input. Hence, any integer `nodeIndex` can be used, which may produce memory errors.
With OT1.19, the following script:
```python
import openturns as ot
scale = [1.0]
amplitude = [1.0]
covarianceModel = ot.SquaredExponential(scale, amplitude)
grid_min = 0.0
grid_max = 2.0
grid_size = 2
grid_step = (grid_max - grid_min) / (grid_size - 1)
mesh = ot.RegularGrid(0.0, grid_step, grid_size)
gaussianProcess = ot.GaussianProcess(covarianceModel, mesh)
sample_size = 10
sample = gaussianProcess.getSample(sample_size)
print("getSampleAtVertex(2) = ")
print(sample.getSampleAtVertex(2))
```
produce:
```
getSampleAtVertex(2) = 
0 : [ 4.65653e-310 ]
1 : [ 1.03978e-312 ]
2 : [ 1.58101e-322 ]
3 : [ 4.65653e-310 ]
4 : [ 8.37404e+242 ]
5 : [ 4.65653e-310 ]
6 : [ 1.58101e-322 ]
7 : [ 6.93594e-310 ]
8 : [ 1.58101e-322 ]
9 : [ 1.58101e-322 ]
```
But this cannot be correct, since there are only two nodes, with indices 0 and 1. Hence, there is no node with index equal to 2. According to the code:

https://github.com/openturns/openturns/blob/8584beb36c7f3282147b10e83acb327da234c3e5/lib/src/Base/Stat/ProcessSampleImplementation.cxx#L518

there does not seem to be a check that the node index is valid, i.e. between 0 and `verticesNumber - 1` where:
```
const UnsignedInteger verticesNumber = mesh_.getVerticesNumber();
```

regislebrun:
This method was primarily developed as a private method, used in a controlled context by other methods. We forgot to add the argument check when its status changed...

PR #2112, opened 2022-07-28
misc

jschueller:
- #2094
- #2098
- #2101
- #2129 
- Add help doc to write gallery examples
- support for discrete scipy distributions


lib/src/Base/Graph/Pie.cxx
```diff
@@ -232,6 +236,29 @@ void Pie::checkData(const Point & data) const
...
+  const Scalar epsilon = ResourceMap::GetAsScalar("Pie-SmallValue");
+  for (UnsignedInteger i = 0; i < size; ++ i)
+    if (data[i] > epsilon)
+    {
+      filteredData.add(data[i]);
+      if (labels_.getSize())
+        filteredLabels.add(labels_[i]);
+    }
+    else
+      LOGWARN(OSS() << "removed the value " << data[i] << " from pie chart as it is too small");
```
regislebrun, line 256:
```
"... with label " << labels_[i]
```
We could also aggregate all the labels associated to a total value less than epsilon and put a unique label made of all these labels, in order to show on the piechart that they contribute to nothing. Otherwise, the message will be silently ignored in a notebook session...

regislebrun (commented):
Removing the small contributors from the piechart looks too radical to me, see my comment

lib/src/Base/Graph/Pie.cxx
jschueller: ok, dropped that commit, I'll open an issue instead

regislebrun (approved):

python/src/Distribution.i
```diff
@@ -171,9 +179,15 @@ class SciPyDistribution(PythonDistribution):
...
-        # since 1.9, pdf for continuous, pmf for discrete
-        pdf = self._dist.pdf(X[0])
+        x0 = float(X[0])
+        if self.isContinuous():
+            pdf = self._dist.pdf(x0)
+        else:
+            if x0.is_integer():
+                pdf = self._dist.pmf(x0)
+            else:
+                # newer scipy pmf is a continuous function
```
sofianehaddad, line 189: hmm suppose you have an equivalent of `UserDefined` with support points of type `(n+1)/2`, `n integer` (you are supposing discrete distribution + support is some `integer`...) 

Why don't we use `ppf` to get the support ? Seems that it is more robust than `pmf`

jschueller: I am usin ppf to get the support, see code above:
lb = dist.ppf(cdf_epsilon)
I also just saw that there's a "support" method that can be called instead of ppf, using that for the first call

For discrete but non-integral concern there is no api other than that, so currently I assume discrete means integral for now.
Browsing into scipy distributions I did not see any non-integral, but I may be wrong.

Anyway I added some comments for these aspects

jschueller:
update: added fix for #2152

PR #2133, opened 2022-09-14
Fixed ConditionedGaussianProcess

regislebrun:
The management of vertices included into the conditioning set was partly wrong.


jschueller (approved):
now it matches the comment :)

sofianehaddad (approved):

PR #2134, opened 2022-09-14
Fixed Triangular

### LINKED ISSUE #2095: Triangular::computeCharacteristicfunction() produces NaNs
When m==a or m==b, there is a division by 0 in this method:
```
import openturns as ot

d1 = ot.Triangular(-2.0, -2.0, 4.0)
print(d1.computeCharacteristicFunction(1.0))
d2 = ot.Triangular(-2.0, 4.0, 4.0)
print(d2.computeCharacteristicFunction(1.0))
```
produces:
```
(nan+nanj)
(nan+nanj)
```

regislebrun:
It closes #2095.

The characteristic function produced NaNs in degenerate cases where two breakpoints were equal.

### DIFF
--- a/lib/test/t_Trapezoidal_std.cxx
+++ b/lib/test/t_Trapezoidal_std.cxx
@@ -43,140 +43,159 @@ int main(int, char *[])
     // Test basic functionnalities
     checkClassWithClassName<TestObject>();
 
+    Collection< Trapezoidal > coll;
+    coll.add(Trapezoidal(1.0, 1.2, 3.0, 14.0));   // a<>b<>c<>d
+    coll.add(Trapezoidal(1.0, 1.0, 3.0, 14.0));   // a==b<>c<>d
+    coll.add(Trapezoidal(1.0, 1.2, 1.2, 14.0));   // a<>b==c<>d
+    coll.add(Trapezoidal(1.0, 1.0, 1.0, 14.0));   // a==b==c<>d
+    coll.add(Trapezoidal(1.0, 1.2, 14.0, 14.0));  // a<>b<>c==d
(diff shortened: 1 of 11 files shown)


jschueller (commented):
could you add the testcase ?

jschueller (approved):

lib/test/t_Trapezoidal_std.cxx
```diff
@@ -43,140 +43,159 @@ int main(int, char *[])
...
     {
-      fullprint << "Kolmogorov test for the generator, sample size=" << size << " is " << (FittingTest::Kolmogorov(distribution.getSample(size), distribution).getBinaryQualityMeasure() ? "accepted" : "rejected") << std::endl;
-      size *= 10;
+      // Instantiate one distribution object
+      Trapezoidal distribution(coll[nTrapezoidal]);
+      fullprint << "Distribution " << distribution << std::endl;
+      std::cout << "Distribution " << distribution << std::endl;
+
+      // Is this distribution elliptical ?
+      fullprint << "Elliptical = " << (distribution.isElliptical() ? "true" : "false") << std::endl;
```
josephmure, line 64: This sort of check could be made using assertions.

regislebrun: As discussed during our last technical meeting, you are right, it could be done using assertions, but as there are two to four such tests in each of the 80 tests related to distributions, I don't see the need to cherry-pick some parts of some tests to convert them to assertions, and unfortunately I don't have the time to convert this test to the new format as a whole...

josephmure (approved):
Never mind my comment about unit tests, it is not critical.

sofianehaddad (approved):

PR #2143, opened 2022-09-26
Update plot_estimate_conditional_quantile.py

adutfoy:
The method computeConditionalQuantile computes the quantile of the last component given all the other variables: if X=(X_1, ..., X__d), then computeQuantile computes the quantile of X_d given that (X_1, .., X_{d-1})=(x_1, ..., x_{d-1}). The example was illustrating the quantile of X_1 given X_2.

### DIFF
--- a/python/doc/examples/data_analysis/distribution_fitting/plot_estimate_conditional_quantile.py
+++ b/python/doc/examples/data_analysis/distribution_fitting/plot_estimate_conditional_quantile.py
@@ -4,7 +4,7 @@ Estimate a conditional quantile
 """
 # sphinx_gallery_thumbnail_number = 8
 # %%
-# From a multivariate data sample we estimate its distribution with kernel smoothing. Here we present a bivariate distribution  X= (X1, X2). We use the `computeConditionalQuantile` method to estimate the 90% quantile :math:`Q_1` of the conditional variable :math:`X1|X2` :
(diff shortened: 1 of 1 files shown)


python/doc/examples/data_analysis/distribution_fitting/plot_estimate_conditional_quantile.py
```diff
@@ -7,9 +7,9 @@
 # From a multivariate data sample we estimate its distribution with kernel smoothing. Here we present a bivariate distribution  X= (X1, X2). We use the `computeConditionalQuantile` method to estimate the 90% quantile :math:`Q_1` of the conditional variable :math:`X1|X2` :
```
jschueller, line 7: X2|X1 here too then (at end of line) ?

python/doc/examples/data_analysis/distribution_fitting/plot_estimate_conditional_quantile.py
```diff
@@ -191,7 +191,7 @@
                  '$x_2$', '$Q_1(x_2)$', True, '')
```
josephmure, line 191: I think the title of the last graph is also wrong: it should be X2 | X1=x1

python/doc/examples/data_analysis/distribution_fitting/plot_estimate_conditional_quantile.py
adutfoy: Yes... it has been included in the proposition of Joseph.

python/doc/examples/data_analysis/distribution_fitting/plot_estimate_conditional_quantile.py
adutfoy: Ok. I change it.

josephmure (approved):

jschueller:
moved to #2112

PR #2148, opened 2022-09-30
Update use_case_wingweight.rst

m-balesdent:
Correction of Latex typos on wingweight.

### DIFF
--- a/python/doc/usecases/use_case_wingweight.rst
+++ b/python/doc/usecases/use_case_wingweight.rst
@@ -24,11 +24,11 @@ with:
 
 - :math:`S_w \sim\mathcal{U}(150, 200)`, the wing area (ft^2)
 
-- :math:`W_fw \sim\mathcal{U}(220, 300)`,  the weight of fuel in the wing (lb)
+- :math:`W_{fw} \sim\mathcal{U}(220, 300)`,  the weight of fuel in the wing (lb)
 		 
 - :math:`A : \sim\mathcal{U}(6, 10)`, the aspect ratio (-)
 		 
-- :math:`Lambda : \sim\mathcal{U}(-10, 10)`, the quarter-chord sweep angle (deg)
+- :math:`\Lambda : \sim\mathcal{U}(-10, 10)`, the quarter-chord sweep angle (deg)
 		 
 - :math:`q : \sim\mathcal{U}(16, 45)`, the dynamic pressure at cruise (lb/ft^2)
 


josephmure (approved):

regislebrun:
I used this use-case in a presentation on sensitivity analysis with OT. I had some questions on it:
* would it be possible to use \ell instead of l in the formula? IMO It makes it easier to read.
* I don't see how the conclusions regarding the effect of the input variables on the weight can be drawn from the cross-cuts. For example, the (l, \Lambda) plot shows an evolution of the weight wrt both variables
* the computation of the Sobol indices based on a chaos expansion shows all the ugly guts of the chaos. Maybe a first run using the user-friendly interface:
```
sizePCE = 800
inputDesignPCE = m.distributionX.getSample(sizePCE)
outputDesignPCE = m.model(inputDesignPCE)

algo = ot.FunctionalChaosAlgorithm(
    inputDesignPCE, outputDesignPCE, m.distributionX)

algo.run()
result = algo.getResult()
print(result.getResiduals())
print(result.getRelativeErrors())
```
would convince the user that it is easy to use, then the detailed interface is used in order to show that, with some tuning of the parameters, the same results are obtained with 200 runs instead of 800 runs.
* an additional comment could be made after the chaos expansion: as the most important variables acts only through decoupled first degree contributions, the hypothesis of a linear dependence between the input variables and the weight is legitimate, which explains why both SRC^2 and Taylor give the exact same results even if the first one is based on a L^2 linear approximation and the second one is based on a linear expansion around the mean value of the input variables.

regislebrun (changes requested):
See my remarks in the main thread.

m-balesdent:
Hello,

>     * I don't see how the conclusions regarding the effect of the input variables on the weight can be drawn from the cross-cuts. For example, the (l, \Lambda) plot shows an evolution of the weight wrt both variables

This graph is just used to preliminary understand the behavior of this dimensional function with respect to the inputs.

>     * the computation of the Sobol indices based on a chaos expansion shows all the ugly guts of the chaos. Maybe a first run using the user-friendly interface:

You're right. The default tuning of chaos expansion  works well on this case.

>     * an additional comment could be made after the chaos expansion: as the most important variables acts only through decoupled first degree contributions, the hypothesis of a linear dependence between the input variables and the weight is legitimate, which explains why both SRC^2 and Taylor give the exact same results even if the first one is based on a L^2 linear approximation and the second one is based on a linear expansion around the mean value of the input variables.

This is now added in the example.

The PR #2150 implements your remarks.

jschueller:
could you put everything in the same PR ? either that one or the other, and close one

PR #2150, opened 2022-10-03
Wingweight

m-balesdent:
Changes of wingweight example to answer  @regislebrun remarks

### DIFF
--- a/python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_wingweight.py
+++ b/python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_wingweight.py
@@ -331,7 +331,7 @@ view = otv.View(graph)
 
 # %%
 # Now, we estimate the Sobol' indices using Polynomial Chaos Expansion.
-# We create a Functional Chaos Expansion using basic tunings.
+# We create a Functional Chaos Expansion.
 sizePCE = 800
 inputDesignPCE = m.distributionX.getSample(sizePCE)
 outputDesignPCE = m.model(inputDesignPCE)
@@ -369,9 +369,10 @@ view = otv.View(graph)
 # ------------
 
 # %%
-# We then estimate the HSIC indices using a data-driven approach by exploiting the design of experiments used to train the PCE model.
-inputDesignHSIC = inputDesignPCE
-outputDesignHSIC = outputDesignPCE
+# We then estimate the HSIC indices using a data-driven approach.
+sizeHSIC = 500
+inputDesignHSIC = m.distributionX.getSample(sizeHSIC)
+outputDesignHSIC = m.model(inputDesignHSIC)
 
 covarianceModelCollection = []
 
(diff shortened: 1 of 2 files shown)


jschueller:
dont you want to include the changes from #2148 here ?

m-balesdent:
> seems the example is way too slow now and exceeds the timeout (100s) <10s is preferred

The example is a tutorial to compare a lot of sensitivity analysis methods (SRC, SRRC, Taylor, Pearson, Sobol with Saltelli and Sobol with polynomial chaos, HSIC). So it can take much time than classical example ;)

Before, the polynomial chaos was done in "expert mode" in which all the basis, truncation, input distribution were defined by the user. To follow Regis' remark, this part has been changed to a "push-button" mode in which only the input and output samples are given. Consequently, the chaos polynomial algorithm has to estimate the distribution, basis, etc. that may cause this increase of duration.

jschueller:
If you have the distribution you can still pass it to the simplified constructor.

jschueller:
it's the hsic computation at the end that take forever, try lowering the doe size

m-balesdent:
> it's the hsic computation at the end that take forever, try lowering the doe size

Thanks. I've changed that point and it now takes 50s on my laptop.

jschueller:
50s is still a bit high, can you reduce again ? let's say at most 20s

m-balesdent:
> 50s is still a bit high, can you reduce again ? let's say at most 20s

Now it takes 17s. Is it OK for you?

jschueller:
great, thanks

python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_wingweight.py
```diff
@@ -370,17 +360,19 @@
 # %%
 #
 # The Sobol' indices confirm the previous analyses, in terms of ranking of the most influent variables. We also see that five variables have a quasi null total Sobol' indices, that indicates almost no influence on the wing weight.
-# There is no discrepancy between first order and total Sobol' indices, that indicates no or very low interaction between the variables in the variance of the output.
+# There is no discrepancy between first order and total Sobol' indices, that indicates no or very low interaction between the variables in the variance of the output. 
+# As the most important variables act only through decoupled first degree contributions, the hypothesis of a linear dependence between the input variables and the weight is legitimate, which explains why both squared SRC and Taylor give the exact same results even if the first one is based on a :math:`\mathcal{L}^2` linear approximation and the second one is based on a linear expansion around the mean value of the input variables.
```
jschueller, line 364: these lines are way too long and make the review difficult

jschueller (approved):

python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_wingweight.py
m-balesdent: > these lines are way too long and make the review difficult

The line has been splitted.

sofianehaddad (approved):

PR #2160, opened 2022-10-13
Explicit link with atomic

josephmure:
Needed for Debian:
https://sources.debian.org/src/openturns/1.19-3/debian/patches/linker_flags_for_armel.patch/

### DIFF
--- a/lib/src/CMakeLists.txt
+++ b/lib/src/CMakeLists.txt
@@ -100,6 +100,8 @@ endif ()
 target_link_libraries (OT PUBLIC ${OPENTURNS_LIBRARIES})
 target_link_libraries (OT PRIVATE ${OPENTURNS_PRIVATE_LIBRARIES})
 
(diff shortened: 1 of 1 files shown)


jschueller:
https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/CheckAtomic.cmake

josephmure:
@pgrt Do you have any idea about this?

pgrt:
Hi  @jschueller and @josephmure ,

I guess, Julien, that your idea is to call try_compile without the addition of the proposed flags and to add them in case the compile attempts fails, so that the flags are not included by default?
I trust this makes sense here, indeed.

jschueller:
see #2187

josephmure:
Thanks Julien!

pgrt:
Yes, thanks a lot for lending an helping hand and implementing this!

Cheers,
Pierre

PR #2170, opened 2022-10-17
Update plot_sensitivity_wingweight.py

m-balesdent:
Modification of example to have shaded surfaces instead of contours to ease the interpretation of the influence of the variables following past technical committee recomandations.


jschueller:
you probably want to change target branch to 1.20

m-balesdent:
Actually it's a patch for current RC1

jschueller:
yes, so it should go to the 1.20 branch to land in the final v1.20 tag

m-balesdent:
transfer to 1.20

PR #2173, opened 2022-10-18
Update GaussLegendre_doc.i.in

adutfoy:
Fixed typo in the doc


jschueller:
moved to #2163 to include it in 1.20 branch instead of master

PR #2180, opened 2022-11-02
Deprecate BoxCoxFactory::build(..., BasisCollection)

sofianehaddad:
As `GLM(BasisCollection)` is already deprecated, we do the same here for `BoxCoxFactory::build(...,BasisCollection)`.

`BoxCoxFactory::build(...Basis)` was relying on the previous method thus we reimplement this method to not rely anymore on the deprecated method and to avoid useless `deprecate warning msg`

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Transformation/ProcessTransformation/BoxCoxFactory.cxx
+++ b/lib/src/Uncertainty/Algorithm/Transformation/ProcessTransformation/BoxCoxFactory.cxx
@@ -83,7 +83,7 @@ public:
   Point computeLogLikelihood(const Point & lambda) const
   {
     const UnsignedInteger size = sample_.getSize();
-    // Define BoxCox trannsformation for sample
+    // Define BoxCox transformation for sample
     BoxCoxEvaluation myBoxFunction(Point(1, lambda[0]));
     // compute the mean of the transformed sample using the Box-Cox function
     const Sample outSample(myBoxFunction(sample_));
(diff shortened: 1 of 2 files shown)


lib/src/Uncertainty/Algorithm/Transformation/ProcessTransformation/BoxCoxFactory.cxx
```diff
@@ -381,6 +484,7 @@ BoxCoxTransform BoxCoxFactory::build(const Sample & inputSample,
                                      const Point & shift,
                                      GeneralLinearModelResult & result)
 {
+  LOGWARN(OSS() << "BoxCoxFactory::build(inputSample, outpuSample, covarianceModel, basisCollection, shift) is deprecated");
```
jschueller, line 487: can you update the changelog ?
maybe you could also send the PR that removes the method in master

sofianehaddad: 1st point ok (already done)
For the second point I will have a specific PR for both GLM/KA & BoxCoxFactory

jschueller (approved):

TODO
```diff
@@ -15,3 +15,4 @@ Drop deprecated Drawable.draw|clean & Graph.draw|clean|getRCommand|makeR*|GetExt
 Drop deprecated GaussianProcess.GIBBS
 Remove FunctionalChaosSobolIndices::summary
 Remove FunctionalChaosResult Function ctor
+Drop deprecated BoxCoxFactory::build methods for 1.21
```
jschueller, line 18: I wont be able to merge this, better drop it

sofianehaddad: Done

regislebrun (approved):

PR #2203, opened 2022-12-08
Added SpecFunc::AccurateSum

regislebrun:
This method allows to compute the sum of a sequence of floats with an extended precision, using either MPFR or Kahan's method.

### DIFF
--- a/lib/src/Base/Func/SpecFunc/SpecFunc.cxx
+++ b/lib/src/Base/Func/SpecFunc/SpecFunc.cxx
@@ -874,6 +874,7 @@ Scalar SpecFunc::HyperGeom_1_1(const Scalar p1,
 #endif
 }
 
+
 // Complex hypergeometric function of type (1,1): HyperGeom_1_1(p1, q1, x) = \sum_{n=0}^{\infty} [\prod_{k=0}^{n-1} (p1 + k) / (q1 + k)] * x^n / n!
 Complex SpecFunc::HyperGeom_1_1(const Scalar p1,
                                 const Scalar q1,
@@ -1346,5 +1347,26 @@ UnsignedInteger SpecFunc::BinomialCoefficient(const UnsignedInteger n,
   return value;
 }
 
+Scalar SpecFunc::AccurateSum(const Point & v)
+{
+#ifdef OPENTURNS_HAVE_MPFR
+  boost::multiprecision::mpfr_float_500 sum(0.0);
+  for (UnsignedInteger i = 0; i < v.getSize(); ++i)
+    sum += v[i];
+  return sum.convert_to<Scalar>();
+#else
+  // Here we use Kahan's compensated summation
+  Scalar sum = 0.0;
+  Scalar error = 0.0;
+  for (UnsignedInteger i = 0; i < v.getSize(); ++i)
+    {
+      const Scalar y = v[i] - error;
+      const Scalar t = sum + y;
+      error = (t - sum) - y;
+      sum = t;
+    }
+  return sum;
+#endif
+}
 
 END_NAMESPACE_OPENTURNS
--- a/lib/src/Uncertainty/Algorithm/Analytical/SystemFORM.cxx
+++ b/lib/src/Uncertainty/Algorithm/Analytical/SystemFORM.cxx
@@ -25,6 +25,7 @@
 #include "openturns/IntersectionEvent.hxx"
 #include "openturns/UnionEvent.hxx"
 #include "openturns/Normal.hxx"
+#include "openturns/SpecFunc.hxx"
 
 BEGIN_NAMESPACE_OPENTURNS
 
@@ -176,7 +177,9 @@ void SystemFORM::run()
     numberOfPoincareTerms = 2 * numberOfPoincareTerms + 1;
   }
 
-  Scalar eventProbability = 0.0;
+  // We store all the terms of the Poincare formula in order to use an extra-precision summation for the final result
+  Point signedProbabilities(numberOfPoincareTerms);
+  Scalar sign = 1.0;
   for (UnsignedInteger k = 0; k < numberOfPoincareTerms; ++ k)
   {
     // retrieve region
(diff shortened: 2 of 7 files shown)


lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -1007,9 +1008,10 @@ Scalar DistributionImplementation::computeProbabilityGeneral(const Interval & in
         }
       } // j
       const Scalar cdf = computeCDF(c);
-      probability += (evenLower ? cdf : -cdf);
+      probabilities[i] = (evenLower ? cdf : -cdf);
     } // i
-
+    std::cerr << "probabilities=" << probabilities << std::endl;
```
jschueller, line 1013: leftover from debug?

python/test/t_SpecFunc_std.py
```diff
@@ -38,3 +38,7 @@
     "ibeta(0.0, 1.0, 0.95, tail)=",
     ot.SpecFunc.RegularizedIncompleteBeta(0.0, 1.0, 0.95, True),
 )
+
+x = [1.0, 2.0**54, -2.0**54]
+print("sum(x)=", sum(x))
+print("accurate sum(x)=", ot.SpecFunc.AccurateSum(x))
```
josephmure, line 44:
```suggestion
assert sum(x) == 0.0, "sum(x) nonzero"
assert ot.SpecFunc.AccurateSum(x) == 1.0, "accurate sum(x) not 1"
```

python/test/t_SpecFunc_std.expout
```diff
@@ -60,3 +60,5 @@ ibeta(2.0, 0.0, 0.95)= 0.0
 ibeta(0.0, 1.0, 0.95)= 1.0
 ibeta(2.0, 0.0, 0.95, tail)= 1.0
 ibeta(0.0, 1.0, 0.95, tail)= 0.0
+sum(x)= 0.0
+accurate sum(x)= 1.0
```
josephmure, line 64:
```suggestion
```

josephmure (changes requested):

lib/src/Uncertainty/Model/DistributionImplementation.cxx
regislebrun: Yep. Fixed.

lib/src/Uncertainty/Algorithm/Analytical/SystemFORM.cxx
```diff
@@ -214,17 +217,15 @@ void SystemFORM::run()
...
-    const Scalar poincarreProbability = normal.computeCDF(-1.0 * regionBeta);
+    signedProbabilities[k] = sign * normal.computeCDF(-1.0 * regionBeta);
+    sign = -sign;
 
-    LOGINFO(OSS() << "SystemFORM: poincarre probability [" << k << "]=" << poincarreProbability);
-
-    // trick to update the poincarre sum
-    // we invert the sign of each consecutive term to follow the order of our recursive expansion of the Poincarre formula
-    eventProbability = -eventProbability + poincarreProbability;
+    LOGINFO(OSS() << "SystemFORM: poincarre probability [" << k << "]=" << signedProbabilities[k]);
```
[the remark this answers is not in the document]
regislebrun: @josephmure This one is from @jschueller and is fixed in another PR

josephmure (approved):

jschueller (approved):

jschueller:
you can fix the conflicts now

jschueller:
the $ms compiler is not happy :[

regislebrun:
@jschueller I saw it. It must be linked with a stupid optimization simplifying Kahan's algorithm as if float arithmetic was associative and commutative. Maybe a pragma such as
```
#pragma optimize( "", off )
```
would do the job ([details](https://learn.microsoft.com/en-us/cpp/preprocessor/optimize?view=msvc-170)), as suggested [here](https://stackoverflow.com/questions/32718989/disable-optimization-on-specific-msvc-version)

regislebrun:
@jschueller After inspection, it is not the fault of the MS compiler, the test was too stringent for Kahan's method. It was masked on Linux platform due to the use of the alternative MPFR implementation. Any hope to have it on Windows?

josephmure:
@regislebrun If the test is too stringent, it should be relaxed IMO. Its purpose is to test that Kahan's method is well implemented so it should account for possible platform-related differences.

regislebrun:
@josephmure You are right, and it is exactly what it does now. The Kahan method gives the exact same results on all the platforms and hopefully will continue to give the same results in the future ;-)

jschueller:
on windows we distribute mingw binaries and mpfr is available there, so I'll probably wont bother testing mpfr with msvc

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -812,6 +812,13 @@ Scalar SpecFunc::Stirlerr(const UnsignedInteger n)
 }
 
 // Hypergeometric function of type (1,1): HyperGeom_1_1(p1, q1, x) = \sum_{n=0}^{\infty} [\prod_{k=0}^{n-1} (p1 + k) / (q1 + k)] * x^n / n!
+// To prevent Microsoft compiler for abusive optimization
+#ifdef _WIN32
+#ifdef _MSC_VER
+#pragma optimize( "", off )
```
jschueller, line 818: so optimization can be re-enabled now ?
note that _MSC_VER implies _WIN32

jschueller (approved):

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -812,6 +812,7 @@ Scalar SpecFunc::Stirlerr(const UnsignedInteger n)
 }
 
 // Hypergeometric function of type (1,1): HyperGeom_1_1(p1, q1, x) = \sum_{n=0}^{\infty} [\prod_{k=0}^{n-1} (p1 + k) / (q1 + k)] * x^n / n!
+// To prevent Microsoft compiler for abusive optimization
```
jschueller, line 815: drop this one

regislebrun: done

PR #2212, opened 2023-01-09
Kriging update

sofianehaddad:
Now we do not use anymore a collection of `Basis` but only a `Basis`. This last one should be `finite` (`FiniteBasis`) and has a collection of function $f_k: R^d \rightarrow R^q$ if we consider a `q-output` function (and thus kriging)

For example, if d=2, q=3, and we want linear functions, a way to build such a `Basis` is given hereafter:
```python
d = 2
q = 3
linear_basis = ot.LinearBasisFactory(d).build()

# Get the constant function
constant_function = linear_basis.build(0)
# Get the affine function
affine_function = linear_basis.build(1)

# Extent the function to a multivariate basis
phi_0 = ot.AggregatedFunction([constant_function] * q)
phi_1 = ot.AggregatedFunction([affine_function] * q)
basis = ot.Basis([phi_0, phi_1])
```

Of course we can have different marginals for each function $\phi_k$.  As an example, suppose we want a `Legendre` basis for the first output marginal, `Hermite` for the second one and still a 2-d input example, we can have the following script :

```python
input_dim = 2

legendre_builder = ot.OrthogonalProductPolynomialFactory([ot.LegendreFactory()] * input_dim)
hermite_builder = ot.OrthogonalProductPolynomialFactory([ot.HermiteFactory()] * input_dim)

# Generate the multivariate function (quadratic)
basis_size = ot.LinearEnumerateFunction(input_dim).getStrataCumulatedCardinal(2)
basis = []
for i in range(basis_size):
    phi_i = ot.AggregatedFunction([legendre_builder.build(i), hermite_builder.build(i)])
    basis.append(phi_i)

basis = ot.Basis(basis)
```

We benefit from this PR to fix some "hidden" issues (related to the multi-output kriging when a trend is required & differ from constant basis), clean a bit the code in order to prepare the future `GaussianProcessRegression` classes


lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/KrigingEvaluation.hxx
```diff
@@ -84,7 +84,7 @@ public:
 protected:
 
   // BasisCollection ==> Persistent for save attribut
-  BasisPersistentCollection basis_;
+  Basis basis_;
```
jschueller, line 87: we need backward-compatible save/load methods

sofianehaddad: Any example on how to implement it?

jschueller: yes, see OrthogonalUniVariatePolynomial:
```
void OrthogonalUniVariatePolynomial::load(Advocate & adv)
{
  UniVariatePolynomialImplementation::load(adv);
  // recurrenceCoefficients_ changed type from PersistentCollection<Coefficients> to Sample in 1.19
  // without backward compatibility, see https://github.com/openturns/openturns/pull/1961
  if (adv.getStudyVersion() >= 102000)
    adv.loadAttribute("recurrenceCoefficients_", recurrenceCoefficients_);
  else
  {
    PersistentCollection<Coefficients> coefficientsColl;
    adv.loadAttribute("recurrenceCoefficients_", coefficientsColl);
    const UnsignedInteger size = coefficientsColl.getSize();
    recurrenceCoefficients_ = PersistentCollection<Scalar>(3 * size);
    for (UnsignedInteger i = 0; i < size; ++ i)
      for (UnsignedInteger j = 0; j < 3; ++j)
        recurrenceCoefficients_[3 * i + j] = coefficientsColl[i][j];
  }
```

sofianehaddad: Ok I see the point, thanks
There will be however some "lost" because the oldest implementation allowed basis collection to be wih different sizes (and it was bogus)

jschueller (approved):

jschueller:
can we try another optimization algo, or disable optimization ?

jschueller:
yes, the random generator definetely has to be reset for each example as all are run in the same process,
and so depending on the examples enabled it can change the output, this is explained in the dev guide

sofianehaddad:
While fixing covariance param, I am expecting to have **all** model parameters equal
However, trend coefficients seems to be different between `artifact` and `laptop`

On `artifact`, we get `Trend coeff=  [8.13771e-05,0.0416623,0.00258901,-2.77803e-10,9.54839e-07,2.18707e-07,3.43745e-07,1.01771,0.0126075,1.24023e-06,0.0288537,-0.00590047,0.000128351,1.70837,0.162497,-0.000148037,0.0588951,0.00393705,-0.000617047,0.073682,0.00375045,0.000209744,2.85549,0.153265,0.00563198,0.127405,0.00776773,6.70055e-05,0.0498137,0.00302664]#30`

On `my laptop`, `Trend coeff=  [0.000111517,0.433554,0.00314669,-1.92729e-10,9.60517e-07,2.64898e-07,8.98426e-07,1.01923,0.0128945,9.46653e-07,0.00931923,-0.00694004,7.82142e-05,0.73435,0.175136,-0.00042437,-3.57125,0.00652978,-0.000597502,-1.77771,0.00398987,0.000183413,2.39676,0.163105,0.00556484,32.0938,0.00823919,7.94159e-05,0.13876,0.00391612]#30`

As a consequence, $q_2$ are lower in `artifcat` and models are less accurate on tail. Trend coeff are given as $\beta = F^t C^{-1} F y$

Any idea on why such differences ?

jschueller:
hard to say, could be due to threading, maybe try to disable tbb ? on your mac there is also probably openblas threading

python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_multioutput_firesatellite.py
```diff
@@ -53,7 +54,8 @@
 # We choose to use a constant trend.
 
 # %%
-basis = ot.ConstantBasisFactory(m.dim).build()
+linear_basis = ot.LinearBasisFactory(m.dim).build()
+basis = ot.Basis([ot.AggregatedFunction([linear_basis.build(k)] * 3) for k in range(linear_basis.getSize())])
```
josephmure, line 58: This is not really related to this PR, but `build` is a strange name for a method that really gets the element of the basis with a given index. Shouldn't the name rather be `get`, or even better, couldn't we allow indexation, like this: ``linear_basis[k]`` ?

python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_multioutput_firesatellite.py
```diff
@@ -53,7 +54,8 @@
 # We choose to use a constant trend.
```
josephmure, line 54: The trend is no longer constant with your changes.

josephmure (commented):
Good idea, I have a few questions/remarks.

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
```diff
@@ -55,8 +55,8 @@ KrigingEvaluation::KrigingEvaluation (const BasisCollection & basis,
 {
   if (basis.getSize() > 0)
   {
-    if (basis.getSize() != covarianceModel.getOutputDimension())
-      throw InvalidArgumentException(HERE) << "In KrigingEvaluation::KrigingEvaluation, output sample dimension (" << covarianceModel.getOutputDimension()  << ") does not match multi-basis dimension (" << basis_.getSize() << ")";
+    if (basis[0].getOutputDimension() != covarianceModel.getOutputDimension())
```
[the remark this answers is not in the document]
sofianehaddad: Agreed. However we don't use directly this class but through `KrigingAlgorithm/GeneralLinearModelAlgorithm`, that do the job correctly

python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_multioutput_firesatellite.py
sofianehaddad: exact, need to update it. As I am struggling with other issues, I missed this one, thanks

python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_multioutput_firesatellite.py
sofianehaddad: The `Basis` class is independent from `kriging`. 

One of the ability of such classes is to `build on the fly` a function (polynomial for example)

linear_basis[k] yields a `TypeError` even if operator exists because it is masked in python

Thus no choice

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx
```diff
@@ -422,7 +393,7 @@ void GeneralLinearModelAlgorithm::run()
   CovarianceModel reducedCovarianceModelCopy(reducedCovarianceModel_);
   reducedCovarianceModelCopy.setActiveParameter(covarianceModel_.getActiveParameter());
 
-  result_ = GeneralLinearModelResult(inputSample_, outputSample_, metaModel, residuals, relativeErrors, basisCollection_, trendCoefficients, reducedCovarianceModelCopy, optimalLogLikelihood);
+  result_ = GeneralLinearModelResult(inputSample_, outputSample_, metaModel, residuals, relativeErrors, basis_, beta_, reducedCovarianceModelCopy, optimalLogLikelihood);
```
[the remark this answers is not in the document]
sofianehaddad: maybe it was more explicit but moving to `beta_` purpose is to have the same behavior / naming as in `KrigingAlgorithm` class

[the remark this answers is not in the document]
josephmure: For this purpose, you could as well rename `beta_` in the KrigingAlgorithm class.

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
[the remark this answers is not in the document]
josephmure: I am not sure I agree with this idea: if your logic is sound, then this check is not needed in the first place. But can we be sure C++ users (who have direct access to this class unless I am mistaken) will never do anything with it and potentially need the check?

python/doc/examples/meta_modeling/kriging_metamodel/plot_kriging_multioutput_firesatellite.py
josephmure: I don't understand: do you mean that a Basis object can build a linear combination of its elements? Where is this ability used?

Also, couldn't we unmask the [] operator in Python?

sofianehaddad: For the second point, you might ask @jschueller There is probably a good reason for masking it

For the first point, I mean that we might generate functions on a "factory mode". If you define or embed a `LegendreFactory` into yourr `Basis`, you can ask for a `legendre` function up to a degree ... in fact no limitation as we consider infinite basis. This is why we have a `build` method here (factory mode)

sofianehaddad:
should be ok for merging

jschueller (approved):

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx
```diff
@@ -354,48 +334,39 @@ void GeneralLinearModelAlgorithm::run()
...
   {
     // Care ! collection should be non empty
-    metaModel = DualLinearCombinationFunction(allFunctionsCollection, trendCoefficientsSample);
+    metaModel = AggregatedFunction(marginalFunctions);
   }
   else
   {
-    // If no basis ==> zero function
+      // If no basis ==> zero function
 #ifdef OPENTURNS_HAVE_ANALYTICAL_PARSER
```
regislebrun, line 370: This may be removed as OT now embeds Exprtk

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
```diff
@@ -187,16 +193,16 @@ Point KrigingEvaluation::operator()(const Point & inP) const
...
-    const Basis localBasis(basis_[i]);
-    const Point betaBasis(beta_[i]);
-    const UnsignedInteger basisSize = localBasis.getSize();
-    for (UnsignedInteger j = 0; j < basisSize; ++j)
-      value[i] += localBasis[j](inP)[0] * betaBasis[j];
+    const Point phi_i(basis_[i](inP));
+    for (UnsignedInteger outputMarginal = 0; outputMarginal < getOutputDimension(); ++outputMarginal)
+    {
+      value[outputMarginal] += phi_i[outputMarginal] * beta_[index];
+      index += 1;
```
regislebrun, line 204: `++index;`

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/KrigingEvaluation.cxx
```diff
@@ -287,26 +293,17 @@ Sample KrigingEvaluation::operator()(const Sample & inS) const
...
-    const Basis localBasis(basis_[i]);
-    const Point betaBasis(beta_[i]);
-    const UnsignedInteger basisSize = localBasis.getSize();
-    // For the i-th Basis (marginal), take into account the trend
-    Sample fi(size, 1);
-
-    for (UnsignedInteger j = 0; j < basisSize; ++j)
+    // Evaluate basis_[i] on the full sample & benefit from the parallelism
+    // of the underlying functions
+    const Sample phi_i = basis_[i](inS);
```
regislebrun, line 301: `cont Sample phi_i(basis_[i](inS))`

regislebrun (approved):
Only minor remarks. `Basis` should definitely provide the `getInputDimension()` and `getOutputDimension()` methods

lib/src/Base/Stat/TensorizedCovarianceModel.cxx
```diff
@@ -200,7 +200,8 @@ Matrix TensorizedCovarianceModel::partialGradient(const Point & s,
 /* Parameters accessor */
 void TensorizedCovarianceModel::setFullParameter(const Point & parameter)
 {
-  const UnsignedInteger parameterDimension = getParameter().getDimension();
+  const UnsignedInteger parameterDimension = getFullParameter().getSize();
+
```
jschueller, line 204: great! can you add mathias python testcase ?

sofianehaddad: ok but a simplified one (without optim)

PR #2238, opened 2023-02-14
Fixed KernelSmoothing

adutfoy:
When the bounding option is activated and the bound is given, the algorithhm may consider that the sample has constant values due to a wrong initialization of variables.


regislebrun:
The transformation method could be implemented too, but it is a separate issue. Here there is clearly a bug: you cannot fix 0.0 as the given lower bound and apply no upper boundary correction without throwing a spurious exception.

jschueller (approved):

sofianehaddad (approved):

PR #2242, opened 2023-02-16
Update ExprTk to 0.0.2

ArashPartow:


jschueller:
thanks for the update, it looks like an "override" keyword is missing, it generates a warning promoted to an error on the macos build log:
```
exprtk.hpp:5852:25: error: 'vector_size' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
            std::size_t vector_size() const
```
https://github.com/ArashPartow/exprtk/pull/40

ArashPartow:
yeah agreed that seems to be the case (-Werror,-Winconsistent-missing-override).

I'm upgrading the library to C++17, i'll fix these issues as part of that,.  Though it's interesting that the linux build does not have the same error.

PR #2256, opened 2023-03-28
Update plot_chaos_cantilever_beam_integration.py

adutfoy:
Fixed typo.

### DIFF
--- a/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cantilever_beam_integration.py
+++ b/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cantilever_beam_integration.py
@@ -32,7 +32,7 @@ cb = cantilever_beam.CantileverBeam()
 dist_E = cb.E
 dist_F = cb.F
 dist_L = cb.L
-dist_I = cb.II
+dist_I = cb.I
 distribution = cb.independentDistribution
 
 # %%


python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cantilever_beam_integration.py
```diff
@@ -32,7 +32,7 @@
 dist_E = cb.E
 dist_F = cb.F
 dist_L = cb.L
-dist_I = cb.II
+dist_I = cb.I
```
mbaudin47, line 35: This is wrong, it really is `II` according to [the code](https://github.com/openturns/openturns/blob/71d3b4e9905649bca18279da3135a5b6d2a49829/python/src/usecases/cantilever_beam.py#L72). May be to avoid a confusion of the candidate `I` name with the `L` letter, already used as the length of the beam?

mbaudin47 (changes requested):
I suggest to cancel this PR because it fails to produce a working script.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cantilever_beam_integration.py
jschueller: I modified it to follow flake8 recommendations

adutfoy: I agree: we can delete this PR.

Issue #2046, opened 2022-05-14
[Debian] DistFunc_binomial issues in tests on arm64 and ppc64el


pgrt:
Hello,

During the Debian packaging of version 1.18, I noticed the following test failure on the architectures arm64 and ppc64el :

343/492 Test #343: cppcheck_DistFunc_binomial .......................................***Failed    0.15 sec
--- /<<PKGBUILDDIR>>/lib/test/t_DistFunc_binomial.expout	2021-11-10 07:42:45.000000000 +0000
+++ /<<PKGBUILDDIR>>/build-python3.10/lib/test/t_DistFunc_binomial.out	2022-04-11 22:20:54.053398612 +0000
@@ -70,7 +70,7 @@
 dBinomial(11, 0.1, 6)=0.000273
 dBinomial(11, 0.1, 7)=2.17e-05
 dBinomial(11, 0.1, 8)=1.2e-06
-dBinomial(11, 0.1, 9)=4.46e-08
+dBinomial(11, 0.1, 9)=4.45e-08
 dBinomial(11, 0.1, 10)=9.9e-10
 dBinomial(11, 0.1, 11)=1e-11
 dBinomial(12, 0.1, 0)=0.282
@@ -193,11 +193,11 @@
 dBinomial(5, 0.5, 4)=0.156
 dBinomial(5, 0.5, 5)=0.0312
 dBinomial(6, 0.5, 0)=0.0156
-dBinomial(6, 0.5, 1)=0.0938
+dBinomial(6, 0.5, 1)=0.0937
 dBinomial(6, 0.5, 2)=0.234
 dBinomial(6, 0.5, 3)=0.312
 dBinomial(6, 0.5, 4)=0.234
-dBinomial(6, 0.5, 5)=0.0938
+dBinomial(6, 0.5, 5)=0.0937
 dBinomial(6, 0.5, 6)=0.0156
 dBinomial(7, 0.5, 0)=0.00781
 dBinomial(7, 0.5, 1)=0.0547

I guess one only has to decrease the number of digits in the output to correct this issue. Maybe you would not like to do it, in which case I would make it a Debian-specific change.

Best,

Pierre

jschueller:
I found out that this is due to the floating-point expression contraction optimization.
One can work-around it by setting "-ffp-contract=off" in CXXFLAGS.
Not an incorrect result per say, but our reference is x86_64 where such optimizations are not available.

mbaudin47:
[...]

PR #2263, opened 2023-03-30
Test how OT reacts to MaxScalar = infinity

josephmure:
EDIT (20/11/2023) : 

- Main change: a new compilation Flag `OPENTURNS_MAXSCALAR_IS_INF` can be used to set `SpecFunc::MaxScalar` to infinity. False by default (so the current behavior where it has the value `1.79769e+308` remains the default).
- A new `SpecFunc::ActualMaxScalar` value is set to `1.79769e+308`. This is needed for some algorithms which legitimately need the maximum finite scalar. When the `OPENTURNS_MAXSCALAR_IS_INF` compilation flag is False, both `SpecFunc::ActualMaxScalar` and `SpecFunc::MaxScalar` have the value `1.79769e+308`.
- Optimization algorithms are tweaked as a result of these changes.
- Unit tests have been tweaked to work regardless of the value of the `OPENTURNS_MAXSCALAR_IS_INF` compilation option.
- A new series of unit tests has been added to `t_DistFunc_binomial` C++ and Python tests to test fringe behavior of the Binomial distribution.


josephmure:
Hi, on the macOS CI machine, I get an error that I don't understand:
```
/Users/runner/work/openturns/openturns/lib/test/t_KPermutationsDistribution_std.cxx:71:12: error: unused variable 'LPDF' [-Werror,-Wunused-variable]
    Scalar LPDF = distribution.computeLogPDF( point );
```
However the variable `LPDF`  **is** used at line 72:

https://github.com/openturns/openturns/blob/4a9cc56d8e9736a68724becdf30ce165cec5b9ac/lib/test/t_KPermutationsDistribution_std.cxx#L72

Does anybody have an idea? @jschueller @sofianehaddad @regislebrun @mbaudin47

jschueller:
the assert is probably ignored in release build

josephmure:
So how can I solve this?

jschueller:
use assert_equal, else the assert will be ignored and the test also

mbaudin47:
> the assert is probably ignored in release build

Very true: see the [stackoverflow](https://stackoverflow.com/questions/5354314/how-can-i-completely-disable-calls-to-assert) message.

lib/src/Base/Optim/TNC.cxx
```diff
@@ -380,7 +380,7 @@ int TNC::ComputeObjectiveAndGradient(double *x, double *f, double *g, void *stat
 
     outP = problem.getObjective().operator()(inP);
 
-    if (!SpecFunc::IsNormal(outP[0]))
+    if (SpecFunc::IsNaN(outP[0]))
```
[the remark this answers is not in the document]
josephmure: I am not sure: having infinity as output is one thing, having infinity as input is another: that would imply that the optimization domain contains infinity, which is not possible anyway, I think.

Anyway, my approach in this PR is to try to modify the smallest possible amount of code. The reason I modified line 383 is that it was breaking several tests and examples, but line 378 does not. If you agree, I would therefore rather keep line 378 as is.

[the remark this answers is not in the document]
mbaudin47: ok

lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx
```diff
@@ -1635,6 +1636,7 @@ protected:
     Point operator() (const Point & point) const override
     {
       const Scalar logPDF = p_distribution_->computeLogPDF(point);
+      if (logPDF == SpecFunc::LowestScalar) return Point(1, 0.0);
```
mbaudin47, line 1639: I cannot understand why this code is necessary. If std::exp() handles -INF, then -exp(-INF) should be -0 and multiplied by a negative number (since logPDF must be negative here), this should give a +0. So your code suggests that std::exp(-INF) does not produce +0: what does it produce then?

josephmure: The problem is the line below: https://github.com/openturns/openturns/blob/a18c583c8a35239d8db361dd77178b9c87f1836d/lib/src/Uncertainty/Model/openturns/DistributionImplementation.hxx#L1640

Without the check, we would get the indeterminate form:
```
-std::exp(-INF) * (-INF) = (-0) * (-INF) = NAN
```
Of course, we know that the limit of x log(x) is 0 when x goes to 0, but the arithmetic of floats does not.

mbaudin47 (approved):

mbaudin47:
@josephmure : I suggest to finalize the code and the review so that it can make its way into 1.22.

regislebrun:
IMO at least a test similar to t_Distribution_quantile.py should be added to test computeCDF(), computeComplementaryCDF(), computeSurvivalFunction(), computePDF() in the corner cases (e.g. at the bounds of the range)

mbaudin47:
I suggest to consider the CSV files available at #2046 that I created for Scilab and available at https://github.com/openturns/openturns/issues/2046#issuecomment-1337563967. These tests are focused on corner cases and might reveal some difficulties.

josephmure:
@mbaudin47 @regislebrun I have added the tests proposed by @mbaudin47 we discussed during the technical committee meeting.

josephmure:
This PR could IMO make it to 1.22 if the necessary reviews come in. @jschueller @regislebrun

regislebrun (approved):
Nice work. IMO it should be merged and activated by default during the continuous integration for at least a few weeks, just to check if everything is ok on the various platforms.

josephmure:
For some reason, although infinities are applied now (OPENTURNS_MAXSCALAR_IS_INF is ON), when I run the command 
```python
print(ot.SpecFunc.MaxScalar)
```
I get 1.7976931348623157e+308. And yet the infinities are clearly applied elsewhere:

```python
print(ot.SpecFunc.MaxScalar)
print(ot.Uniform().computeLogPDF([3.0]))
```

produces:

```
1.7976931348623157e+308
-inf
```

In 1.22rc1, it produces:

```
1.7976931348623157e+308
-1.7976931348623157e+308
```

Any idea why that happens? This phenomenon explains the test failures.

josephmure:
@jschueller I have fixed the GEV issue, so I believe this PR can finally be merged.

josephmure:
@jschueller I have fixed the GEV issue, so I believe this PR can finally be merged.

validation/src/distributions/binom.r
```diff
@@ -0,0 +1,102 @@
+
+// Copyright (C) 2012 - Prateek Papriwal
+//
+// This file must be used under the terms of the CeCILL.
+// This source file is licensed as described in the file COPYING, which
+// you should have received as part of this distribution.  The terms
+// are also available at
+// http://www.cecill.info/licences/Licence_CeCILL_V2-en.txt
```
jschueller, line 8: I did not see that before, but the license must be added (eg COPYING.CeCILL_V2) and the files using this specific license must be mentioned in the LICENSE file

josephmure: I have added the license file to the folder containing the script. I don't think we need to put it at the forefront of the repo since, as you said, it applies to a stand-alone file unrelated to OpenTURNS code.

jschueller: but its also used in python/test/t_binomial_dataset.csv

josephmure: Not that it matters (as you said, this file is independent from OT), but CeCILL is actually compatible with the LGPL (and even the GPL for that matter): http://www.cecill.info/faq.fr.html#compatible

jschueller: you cannot include cecill code into an lgpl project, but the other way around is ok

josephmure: Ah yes indeed, you are right. I think I will simply duplicate the COPYING.CECILL_V2 file in the test directories because putting it in the main folder might give people the impression parts of OpenTURNS are licensed under CECILL.

josephmure: Yes, but all of them refer to parts of the library. The file is not very large anyway.

TODO
```diff
@@ -0,0 +1 @@
+When the OPENTURNS_MAXSCALAR_IS_INF compilation option is removed, some checks marked TODO JM need to be removed as well.
```
jschueller, line 1: maybe we could also rename MaxScalar to Infinity and ActualMaxScalar to MaxScalar when that option disappears

jschueller:
we added a new test in t_distribution_std that uses the value MaxScalar, I believe that replacing that to ActualMaxScalar should make the CI happy again, but I wonder why it only succeeds on linux & mingw

josephmure:
But why did CI work on https://github.com/openturns/openturns/pull/2263/commits/622a35e50885f48ef48ab80419084caf1399bc1b ? Nothing has changed since that commit (except for the licensing stuff, but that should be irrelevant to tests).

josephmure:
It seems that CI is running the test in master rather than the test in this branch? Since when does it do that?

jschueller:
that would explain why it still succeeds on circleci

josephmure:
Anyway I have rebased in order to get the same error on Linux and I have fixed the problematic computeCDFs methods.

josephmure:
> if github does something different, that would explain why it still succeeds on circleci

So what does GitHub do? Does it run the tests on master rather than the PR branch (that would be super bad!) or does it merge the PR branch in master before running the tests (that would actually be really good since if would prevent surprises at merge time).

TODO
josephmure: I agree, added to TODO.

PR #2273, opened 2023-04-18
Update NAIS.cxx

m-balesdent:
Modification of iteration number to reflect that an initial sampling has been performed. Allows to compute the correct value of OuterSampling.

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
@@ -162,7 +162,7 @@ void NAIS::run()
     auxiliaryDistribution = computeAuxiliaryDistribution(auxiliaryInputSample, weights);
   }
 
-  UnsignedInteger iterationNumber  = 0;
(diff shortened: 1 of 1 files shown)


m-balesdent:
> that's not consistent with other algos, ie subset where the number of iterations does not take into account the initial doe



But the outer sampling is note correct in this case. Would it be correct to keep iteration number but to replace the outersampling by getMaximumOuterSampling() * (iterationNumber+1) ?

jschueller:
ah, you mean for the result ? yes indeed, this makes sense to count the initial DOE for the evaluation cost

m-balesdent:
> ah, you mean for the result ? yes indeed, this makes sense to count the initial DOE for the evaluation cost

Yes. The same occurs in Cross Entropy.

jschueller:
in subset too then

jschueller (commented):
I checked and this is what is done in subset too, so +1

jschueller (approved):

sofianehaddad (approved):

josephmure (approved):

PR #2277, opened 2023-04-20
Modification of NAIS API

m-balesdent:
Hello,

In the PR[ #2272](https://github.com/openturns/openturns/pull/2272)  about Importance Sampling by Cross Entropy, it was agreed that the name of the attribute "rhoQuantile" is not meaningful for the user. So, it has been changed by "quantileLevel". 

This PR proposes to propagate this change for the NAIS algorithm, as NAIS and Cross Entropy Importance Sampling have similar API.

Furthermore, the sphinx gallery thumbnail picture of  the NAIS example has been changed to set a graph that is more representative of the algorithm behavior.


m-balesdent:
Mistake during merging process

PR #2279, opened 2023-04-20
Improved IntervalMesher

regislebrun:
Now it works in any dimension. The diamond flag is still limited to dimensions 1 and 2.

This capability is mandatory for many applications in higher dimension, such as the definition of distributions over complex domains, see PR #2276.


jschueller (approved):
good to go
probably need to rebase for CI to pass

KieranDelamotte:
Do we still use the expout mechanism in tests?

regislebrun:
I don't know how to check a mesh using assertions, otherwise I would have use it. Do you have any idea on how to do that efficiently?

jschueller:
there is the method checkValidity, otherwise you can probably also test some properties, like the range via getMin/getMax ...

regislebrun:
Using assertion imposes to write the reference value in the test. Even for a very simple 5D mesh (a unique cube split into simplices), you get 32 points in dimension 5 and 120 indices of size 6, quite a lot of things to write into the test script. And it remains to do the same for the other dimensions... The most error-prone part is the connections between vertices, not the vertices coordinates, so a simple range computation will not be very useful.
I will try to embed the reference mesh in the script and see if it is still manageable.

regislebrun:
BTW I forgot to remove the test on the dimension in LevelSetMesher. I will push it soon.

regislebrun:
After a second though, I don't see the point of copying the reference mesh into the script instead of having it in an external file. And it is essentially what we do with the expout file. Your opinion on it?

jschueller:
as I said its also fine to continue using the expout for now

jschueller:
@sofianehaddad ?

sofianehaddad (approved):

PR #2288, opened 2023-04-27
Added 3 examples on regression analysis

mbaudin47:
This PR creates 3 new examples on regression analysis.
- "_Compute confidence interval of a univariate noisy function._" This example shows how to evaluate a point wise confidence interval of a regression model. It shows how to evaluate the point wise C.I. of the mean estimator and the observation. This is a toy model to present the method. [Link to the latest doc build](https://output.circle-artifacts.com/output/job/4f01c985-b16c-4141-adea-51643e673988/artifacts/0/html/auto_meta_modeling/general_purpose_metamodels/plot_regression_sinus.html#sphx-glr-auto-meta-modeling-general-purpose-metamodels-plot-regression-sinus-py)
- "_Compute confidence interval of a regression model from data._" This example is the same as before, but uses a given multivariate data set. This can be used for real world studies. [Link to the latest doc build](https://output.circle-artifacts.com/output/job/4f01c985-b16c-4141-adea-51643e673988/artifacts/0/html/auto_numerical_methods/general_methods/plot_regression_interval.html#sphx-glr-auto-numerical-methods-general-methods-plot-regression-interval-py)
- "_Compute leave-one-out error of a polynomial chaos expansion._" This example shows how to create the `DesignProxy` and use it to compute the LOO Q2 error. This is the first example of this class in the library. [Link to the latest doc build](https://output.circle-artifacts.com/output/job/4f01c985-b16c-4141-adea-51643e673988/artifacts/0/html/auto_meta_modeling/polynomial_chaos_metamodel/plot_pce_design.html#sphx-glr-auto-meta-modeling-polynomial-chaos-metamodel-plot-pce-design-py)

It also creates a new `LinthurstDataset` usecase.

image

**Figure 1.** The 3 new examples in the PR.

image

**Figure 2.** Pointwise confidence intervals of a linear least squares polynomial order 4 fit on a noisy function.


python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_sinus.py
```diff
@@ -0,0 +1,435 @@
...
+# ------------
+#
+# In this example, we compute the pointwise confidence interval of the
+# estimator of the conditional expectation given an input.
+# We consider noisy observations of the sine function.
+# Then we perform linear least squares regression to fit an order 4
+# polynomial.
+# For a given point x, the code computes the confidence interval
+# of the prediction y.
+# This it the confidence interval of the conditional expectation
```
josephmure, line 18:
```suggestion
# This is the confidence interval of the conditional expectation
```

python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_sinus.py
```diff
@@ -0,0 +1,435 @@
...
+    return curve
+
+
+graph = ot.Graph("Polynomial curve fitting", "x", "y", True, "topright")
+# The "unknown" function
+graph.add(plotFunction(g, palette[0]))
+view = otv.View(graph)
+
+# %%
+# This seems a nice, smooth function to approximate with polynomials.
```
josephmure, line 64:
```suggestion
# This is a nice, smooth function to approximate with polynomials.
```

python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_sinus.py
```diff
@@ -0,0 +1,435 @@
...
+graph.add(plotFunction(g, palette[0]))
+# Training set
+graph.add(plotData(xTrain, yTrain, palette[1]))
+view = otv.View(graph)
+
+# %%
+# We see that the noisy observations of the function are relatively
+# large compared to the function values.
+# It may not be obvious that a regression model can fit well to that
+# data.
```
josephmure, line 124:
```suggestion
# It may not be obvious that a regression model can fit that data well.
```

python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_sinus.py
```diff
@@ -0,0 +1,435 @@
...
+# training sample.
+
+designSampleTrain = basisFunction(xTrain)
+print(designSampleTrain[:5])
+
+
+# %%
+# Convert the design sample into a design matrix and create
+# an instance of the :class:`openturns.QRMethod` class.
+# This class has the :meth:`openturns.QRMethod.getGramInverse` method that
```
josephmure, line 161:
```suggestion
# an instance of the :class:`~openturns.QRMethod` class.
# This class has the :meth:`~openturns.QRMethod.getGramInverse` method that
```

python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_sinus.py
```diff
@@ -0,0 +1,435 @@
...
+# Compute the confidence interval.
+
+# %%
+alpha = 0.95
+confidenceIntervalMean = computeRegressionConfidenceInterval(
+    lsqMethod, betaHat, sigma2Hat, designSampleTest, alpha=alpha
+)
+
+# %%
+# On output, the `confidenceIntervalMean` variable is a `Sample`
```
josephmure, line 328:
```suggestion
# On output, the `confidenceIntervalMean` variable is a :class:`~openturns.Sample`
```

python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_sinus.py
```diff
@@ -0,0 +1,435 @@
...
+        label="Mean %.0f%%" % (100.0 * alpha),
+    )
+)
+view = otv.View(graph)
+
+# %%
+# We see that the pointwise confidence interval contains the true
+# model for most points.
+# For a small set of points, there are points which are not within
+# the bounds, but there are not too far away.
```
josephmure, line 379:
```suggestion
# the bounds, but are not too far away.
```

python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_sinus.py
```diff
@@ -0,0 +1,435 @@
...
+    alpha=alpha,
+    mean=False,
+)
+
+
+# %%
+# Then we plot the function, its least squares approximation, the
+# C.I. of the mean and the C.I. of the observations.
+
+# %%
```
josephmure, line 402:
```suggestion
```
sphinx_gallery_thumbnail_number should not appear as plain text but as code.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
josephmure, line 1: Is there no relevant graph this example could produce? Like a plot of the residuals? (This can be done in a later PR if you don't have time.)

mbaudin47: Actually, I do not think that a plot would increase the value of this particular example, but I might be wrong. My intention was to provide the code that shows the `DesignProxy` in action. This is a design matrix: no particular graph can show this matrix, which is essentially an algebraic object. Gatien Chopard currently works on plots which show the leave-one-out method in action: this may provide an excellent example, but this is off topic in this particular example.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
```diff
@@ -0,0 +1,152 @@
...
+adaptiveStrategy = ot.FixedStrategy(multivariateBasis, basisSize)
+chaosAlgorithm = ot.FunctionalChaosAlgorithm(
+    xTrain, yTrain, im.distributionX, adaptiveStrategy, projectionStrategy
+)
+chaosAlgorithm.run()
+chaosResult = chaosAlgorithm.getResult()
+
+# %%
+# The :class:`~openturns.DesignProxy` class provides methods used to solve
+# least squares problem.
```
mbaudin47, line 57:
```suggestion
# the least squares problem.
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
```diff
@@ -0,0 +1,152 @@
...
+print("Basis size = ", basisSize)
+adaptiveStrategy = ot.FixedStrategy(multivariateBasis, basisSize)
+chaosAlgorithm = ot.FunctionalChaosAlgorithm(
+    xTrain, yTrain, im.distributionX, adaptiveStrategy, projectionStrategy
+)
+chaosAlgorithm.run()
+chaosResult = chaosAlgorithm.getResult()
+
+# %%
+# The :class:`~openturns.DesignProxy` class provides methods used to solve
```
mbaudin47, line 56:
```suggestion
# The :class:`~openturns.DesignProxy` class provides methods used create the objects necessary to solve
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
```diff
@@ -0,0 +1,152 @@
...
+# least squares problem.
+# More precisely, it provides the :meth:`~openturns.DesignProxy.computeDesign`
+# method that we need.
```
mbaudin47, line 59:
```suggestion
# method that we need to evaluate the design matrix.
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
```diff
@@ -0,0 +1,152 @@
...
+chaosResult = chaosAlgorithm.getResult()
+
+# %%
+# The :class:`~openturns.DesignProxy` class provides methods used to solve
+# the least squares problem.
+# More precisely, it provides the :meth:`~openturns.DesignProxy.computeDesign`
+# method that we need.
+# In many cases we do not need that matrix, but the Gram matrix (or its inverse).
+# The :class:`~openturns.DesignProxy` class is needed by a least squares solver,
+# e.g. :class:`~openturns.QRMethod` that knows how to actually compute the coefficients.
```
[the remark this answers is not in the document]
mbaudin47: That might be confusing indeed. I would point to [linear_regression.html](http://openturns.github.io/openturns/latest/theory/data_analysis/linear_regression.html), but that particular page does not present the relevant topics. Let $n$ be the number of observations and $m$ be the number of coefficients of the linear model. Let $D$ be the [design matrix](https://en.wikipedia.org/wiki/Design_matrix), i.e. the matrix that produces the predictions of the linear regression model from the coefficients:

$$
\hat{y} = D \boldsymbol{a}
$$

where $\boldsymbol{a} \in \mathbb{R}^m$ is the vector of coefficients, $D \in \mathbb{R}^{n \times m}$ and $\hat{y} \in \mathbb{R}^n$ is the vector of predictions. The linear least squares problem is:

$$
\arg \min_{\boldsymbol{a} \in \mathbb{R}^m} \left\|\| D \boldsymbol{a} - \boldsymbol{y} \right\|\|.
$$

The solution is given by the normal equations, i.e. the vector of coefficients is the solution of the following linear system of equations:

$$
D^T D \boldsymbol{a} = D^T \boldsymbol{y}
$$

The point is: to solve a linear least squares problem, we need the design matrix $D$, which is the primary goal of the `DesignProxy` class. When we use the QR decomposition, we actually do not need to evaluate it in our script: the `QRMethod` class knows how to compute the solution without evaluating the Gram matrix $D^T D$. We may need the inverse Gram matrix $(D^T D)^{-1}$ sometimes, for example when we want to create a D-[optimal design](https://en.wikipedia.org/wiki/Optimal_design). Finally, when we want to compute the analytical leave-one-out error, we need to compute the diagonal of the  [projection matrix](https://en.wikipedia.org/wiki/Projection_matrix), that is $D (D^T D)^{-1} D^T$. For all these purposes, the `DesignProxy` is _the_ tool.

Perhaps I could just write this as an introduction to the page. What do you think about it?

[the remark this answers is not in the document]
mbaudin47: Fixed at cad72481d, at the price of many more (hopefully interesting) details.

mbaudin47:
@Gatienn: Please comment this PR if you feel to.

mbaudin47:
@m-balesdent : Please review this PR if you are interested by regression analysis.

python/doc/examples/meta_modeling/general_purpose_metamodels/plot_regression_interval.py
```diff
@@ -0,0 +1,267 @@
...
+#
+# In this example, we compute the pointwise confidence interval of the
+# estimator of the conditional expectation given an input.
+# More precisely, we compute confidence intervals of the output of
+# a regression model.
+# The linear regression model is an order 1 multivariate polynomial.
+# This model is built from a dataset.
+
+# %%
+
```
mbaudin47, line 19: fdsqfdsqfdsqfdsqfdsq

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
```diff
@@ -0,0 +1,429 @@
...
+# - When we use the QR decomposition, we actually do not need to evaluate it in
+#   our script: the :class:`~openturns.QRMethod` class knows how to compute the
+#   solution without evaluating the Gram matrix `D^T D`.
+# - We may need the inverse Gram matrix
+#   :math:`(D^T D)^{-1}` sometimes, for example when we want to create
+#   a D-optimal design.
+# - Finally, when we want to compute the analytical leave-one-out error,
+#   we need to compute the diagonal of the  projection matrix :math:`H`.
+#
+# For all these purposes, the `DesignProxy` is _the_ tool.
```
mbaudin47, line 83:
```suggestion
# For all these purposes, the `DesignProxy` is *the* tool.
```

sofianehaddad (commented):
I have a couple of remarks about the examples here :

 - First in many examples, we use `low level classes` such as `DesignProxy`, `QRMethod`. As we can not tag examples and make distinction between examples for `advance users` & `newbies`, it is maybe risky. Renaming the examples may help ? Or putting examples using low level functions in other section ?
 - If we have a look at the examples, for instance `plot_regression_sinus` and `plot_regression_interval`, we are implementing 
 the same `computeRegressionConfidenceInterval` function in both. 
 I think implementing functions in the examples should be done in few specific cases or for demonstrating comparable features outside the library (ex other transformations than BoxCox...) In the current applicaion, it is rather a missing feature in the API and I will implement it soon in `LinearModelAnalysis`
 - If we focus on the graph aspect, I see that very often we use the `Graph` & `Drawable` objects, which is cumbersome.
 I prefer using directly `matplotlib` to make people aware of the interoperability between `openturns` & the python `ecosystem`.
 As user, I never use `Graph` objects to plot `lines` or a `cloud` but I rely on `Class.drawXXX` to expose the services. 

In sum, we probably have to "calibrate" the complexity of our examples and make sure we are addressing the right needs for the right users.

mbaudin47:
> In sum, we probably have to "calibrate" the complexity of our examples and make sure we are addressing the right needs for the right users.

mbaudin47:
Thank you for this review. Let me try to explain the choices here:

>     * First in many examples, we use `low level classes` such as `DesignProxy`, `QRMethod`.
There is currently no "Advanced" help section, but I can introduce the example as an "advanced" example.

>     * If we have a look at the examples, for instance `plot_regression_sinus` and `plot_regression_interval`, we are implementing the same `computeRegressionConfidenceInterval` function in both.

My goal was not to point that there is a missing feature, but that, given the design matrix, lots of derived calculation can be done. Hence, this example will be as a starting point for the reader. A new feature would be very helpful of course: please use the current example as a bases for the C++ implementation. But, event when the feature will be available, this may not completely remove the need for an example.

>     * If we focus on the graph aspect, I see that very often we use the `Graph` & `Drawable` objects, which is cumbersome.

The "cumbersome" adjective is somewhat arbitrary. 1) I consistently use OpenTURNS API in the examples using advices from @regislebrun in all places where the current API does the job. Using another library in the help pages makes sense only if the content is not relevant or too clumsy. 2) If you can point specific segments of the example that may be improved using Matplotlib, I can adapt the script. 3) This provides OpenTURNS examples of the graphics classes, which is the whole point of help pages isn'it?

mbaudin47:
@josephmure @adutfoy : Please review this example if you are interested on this topic.

mbaudin47:
I have a branch to implement the Q2 LOO at : https://github.com/mbaudin47/openturns/tree/chaosvalidation

mbaudin47:
@sofianehaddad : Ok, let's make a summary of the issues you found here and try to see what can be done.
- The `computeRegressionConfidenceInterval()` function is duplicated in two examples ("Compute confidence intervals of a regression model from data" and "Compute confidence intervals of a univariate noisy function"). Still, each example is specific: "Compute confidence intervals of a regression model from data" show how to use the `DesignProxy` on a multivariate case and the second directly builds the design matrix on a univariate case where a plot of the function can be made. My suggestion is to keep these examples as is. When the feature will be available in the `LinearModelAnalysis`, then I suggest to update the "Compute confidence intervals of a univariate noisy function" example and use the feature. The interesting part of the first example is to show how to use `DesignProxy`, so that questions such as the one appearing in [a Discourse topic](https://openturns.discourse.group/t/pce-design-matrix-and-coefficient-fit/265/5) can find an answer.
- On the "advanced" property of these examples, I updated the content to mention this. When we will have more than 2 examples, then a "Advanced" section will make sense, but that is not currently the case.
- On the fact that the examples uses the OpenTURNS graphics API, I think that not integrating the examples for this reason would make no sense. If you have, however, specific comments on the scripts that can be simplified, then we can discuss on a case by case basis. I suggest to make a detailed review if that seems relevant to you.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
```diff
@@ -0,0 +1,447 @@
...
+#     \vect{y} = (y^{(1)}, \ldots, y^{(n)})^T.
+#
+#
+# Consider the following set of inputs, let aside the :math:`i`-th input:
+#
+# .. math::
+#
+#     \cD_{-i} := \left\{\vect{x}^{(1)}, \ldots, \vect{x}^{(i - 1)}, \vect{x}^{(i + 1)}, \ldots, \vect{x}^{(n)}\right\}
+#
+# for :math:`i \in \{1, ..., n\}`.
```
mbaudin47, line 146: The letter $i$ should be changed for $j$, so that the notations can be kept consistent across pages.

sofianehaddad:
> @sofianehaddad : Ok, let's make a summary of the issues you found here and try to see what can be done.
> 
>     * The `computeRegressionConfidenceInterval()` function is duplicated in two examples ("Compute confidence intervals of a regression model from data" and "Compute confidence intervals of a univariate noisy function"). Still, each example is specific: "Compute confidence intervals of a regression model from data" show how to use the `DesignProxy` on a multivariate case and the second directly builds the design matrix on a univariate case where a plot of the function can be made. My suggestion is to keep these examples as is. When the feature will be available in the `LinearModelAnalysis`, then I suggest to update the "Compute confidence intervals of a univariate noisy function" example and use the feature. The interesting part of the first example is to show how to use `DesignProxy`, so that questions such as the one appearing in [a Discourse topic](https://openturns.discourse.group/t/pce-design-matrix-and-coefficient-fit/265/5) can find an answer.

In that case, please mention in the title and first descriptions that purpose is to highlight the usage of low level classes. It is important to alert that the heart is the usage of `DesignProxy` rather than `how to compute confidence interval for linear models with OpenTURNS?`  
> 
>     * On the "advanced" property of these examples, I updated the content to mention this. When we will have more than 2 examples, then a "Advanced" section will make sense, but that is not currently the case.
> 
I had in mind [that section](https://openturns.github.io/openturns/latest/auto_numerical_methods/index.html
Advanced
>     * On the fact that the examples uses the OpenTURNS graphics API, I think that not integrating the examples for this reason would make no sense. If you have, however, specific comments on the scripts that can be simplified, then we can discuss on a case by case basis. I suggest to make a detailed review if that seems relevant to you.

For me it is "low level". I want to make sure that new users won't be scared or held back by the "heavy" use of the openturns Graph API and they won't realize that matplotlib can be used directly

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_pce_design.py
mbaudin47: Fixed at b6fa07671d806cb4189983595d4f50028b5d664a

python/src/usecases/linthurst_dataset.py
```diff
@@ -0,0 +1,79 @@
+"""
+Use case : Linthurst data set
+=============================
+"""
+import openturns as ot
+
+
+class LinthurstDataset:
```
jschueller, line 8: maybe the module & class names could be simplified:
"linthurst_dataset.LinthurstDataset().data" -> "linthurst.Linthurst.data"
like we have for "coles.Coles.wavesurge"

python/doc/examples/numerical_methods/general_methods/plot_regression_sinus.py
```diff
@@ -0,0 +1,436 @@
...
+#
+
+# %%
+# In order to approximate the function with polynomials up to degree 4,
+# we create a list of strings containing the associated monomials.
+# We perform the loop from 0 up to `totalDegree` (but the `range`
+# function takes `totalDegree + 1` as its second input argument).
+
+totalDegree = 4
+polynomialCollection = ["x^%d" % (degree) for degree in range(0, totalDegree + 1)]
```
jschueller, line 140: prefer fstrings: f"x^{degree}"

python/doc/examples/numerical_methods/general_methods/plot_regression_sinus.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+# %%
+# Then we plot the true function, its noisy observations and the least
+# squares model of degree 4.
+
+
+def plotPredictions(xTest, yHatTest, totalDegree, color):
+    curve = ot.Curve(xTest, yHatTest)
+    curve.setLegend("L.S. degree %d" % (totalDegree))
```
jschueller, line 212: here too

ChangeLog
```diff
@@ -37,6 +37,7 @@
  * LeastSquaresExpansion, IntegrationExpansion (openturns.experimental)
  * UniformOverMesh (openturns.experimental)
  * GeneralizedExtremeValueValidation (openturns.experimental)
+ * LinthurstDataset
```
jschueller, line 40: maybe specify its an use-case module (I also forgot to add Coles, would you mind adding it here as well ?)

jschueller (approved):
Its a good compromise to have the low-level examples in numerical_methods instead
because I agree this could be confusing to promote very low-level classes

python/doc/examples/numerical_methods/general_methods/plot_regression_sinus.py
mbaudin47: Ok. Fixed at d603557b2.

python/doc/examples/numerical_methods/general_methods/plot_regression_sinus.py
mbaudin47: Ok. Fixed at d603557b2.

ChangeLog
mbaudin47: Ok. Fixed at 0df9c238e

sofianehaddad (approved):

mbaudin47:
The code was actually merged into #2336. This is why I delete this branch.

PR #2313, opened 2023-05-19
Pretty-print of DualLinearCombinationFunction

mbaudin47:
## The problem
This PR is related to the objects which are produced by the polynomial chaos expansion from `FunctionalChaosAlgorithm` and available in `FunctionalChaosResult`. The issue was detected in the discussion during PR [2236](https://github.com/openturns/openturns/pull/2236).
The pretty-print of the composed metamodel returned by `FunctionalChaosResult.getComposedMetamodel()` is not satisfactory. We see that the expression is very long. 1) It uses the raw exponentiation operator `^`. 2) The indices of the inputs are shown without subscripts, e.g. `x0` instead of $x_0$. 3) Some extra parentheses are shown, which produce e.g. `(x0)` instead of the simpler `x0`. The next figure is presented in the example [plot-functional-chaos-exploitation-py](http://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_exploitation.html#sphx-glr-auto-meta-modeling-polynomial-chaos-metamodel-plot-functional-chaos-exploitation-py).

PR Pretty Print polynômes - 3

**Figure 1.** The composed metamodel returned by `FunctionalChaosResult.getComposedMetamodel()` in OpenTURNS 1.20 from the example `plot_functional_chaos_exploitation.py`.

## Analysis

- Within the polynomial chaos, the composed metamodel is a `DualLinearCombinationFunction` and is created [there](https://github.com/openturns/openturns/blob/457622ecf5264389f3739316e9c8cdc92aaea083/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx#L73). The current code returns a Function, but it is easy to change this to get the `DualLinearCombinationFunction`.  

- The top-level object is the `DualLinearCombinationFunction` class. It somewhat hides the internal function [here](https://github.com/openturns/openturns/blob/457622ecf5264389f3739316e9c8cdc92aaea083/lib/src/Base/Func/DualLinearCombinationFunction.cxx#L43). It does not seem easy to access it from Python's potential pretty-printer. Hence, it is much easier to produce the pretty-print string right from the C++, then to raise it up to the Python layer. 

- The C++ almost-pretty-print is at [UniVariatePolynomialImplementation.cxx](https://github.com/openturns/openturns/blob/457622ecf5264389f3739316e9c8cdc92aaea083/lib/src/Base/Func/UniVariatePolynomialImplementation.cxx#L75). One potential method to print this would be via HTML.

## Result

image

## Differences between str and repr_html

The `ProductPolynomialEvaluation` class implements the product of univariate polynomials. The `UniVariatePolynomialImplementation` class implements the univariate polynomials involved in the product.

There are several differences between `__repr_html__` and `__str__`.
- The exponentiation operator is expressed as the `^` character:
```
# __str__
(1.73205 * x0) * (0.612372 - 1.22474 * x1^2 + 0.204124 * x1^4)
```
We can use the HTML `sup` operator :
```
# __repr_html__
(1.73205 * x0) * (0.612372 - 1.22474 * x1<sup>2</sup> + 0.204124 * x1<sup>4</sup>)
```
which is rendered as :

PR 2313 pretty-print polynômes - figure - 1

- The `*` operator is used as the mathematical operator:
```
# __str__
(1.73205 * x0) * (0.612372 - 1.22474 * x1^2 + 0.204124 * x1^4)
```
The correct HTML expression uses the unicode character $\times$ as `<span>&#215;</span>`:
```
# __repr_html__
(1.73205 <span>&#215;</span> x0) <span>&#215;</span> (0.612372 - 1.22474 <span>&#215;</span> x1<sup>2</sup> + 0.204124 <span>&#215;</span> x1<sup>4</sup>)
```
which is rendered as:

PR 2313 pretty-print polynômes - figure - 2

- Sometimes there is a single variable within curly braces :
```
# __str__
(1.125 - 11.25 * x0^2 + 13.125 * x0^4) * (x1)  # Example A
(x0) × (x1)  # Example B
(1.73205 * x0) * (-1.22474 * x1 + 0.408248 * x1^3)  # Example C
```
This appears for example when a polynomial is the power of a marginal variable (like in example A or B). More generally, it appears when all coefficients are zero, except the leading one (like in example C).
The HTML is:
```__repr_html__
(1.125 - 11.25 <span>&#215;</span> x0<sup>2</sup> + 13.125 <span>&#215;</span> x0<sup>4</sup>) <span>&#215;</span> x1  # Example A
```
which is rendered as:

PR 2313 pretty-print polynômes - figure - 3


jschueller:
there are no "_repr_html_" methods in the c++ lib to properly cross the tree of function hierarchy, instead it is defined in Object.i to call repr by default and a few other objects override this method in their .i

mbaudin47:
> there are no "_repr_html_" methods in the c++ lib to properly cross the tree of function hierarchy
> instead it is defined for Object to call repr by default and a few other objects override this method in their .i

So how to evaluate DualLinearCombinationEvaluation.__repr_html__()?

lib/src/Base/Common/openturns/PersistentObject.hxx
```diff
@@ -151,6 +151,13 @@ public:
     return OSS() << offset << __repr__();
   }
 
+  /* String converter */
+  inline
```
mbaudin47, line 155:
```suggestion
```

lib/src/Base/Common/openturns/PersistentObject.hxx
```diff
@@ -151,6 +151,13 @@ public:
...
+  String __repr_html__() const
```
mbaudin47, line 156: It should be virtual.

python/src/DualLinearCombinationFunction.i
```diff
@@ -8,3 +8,11 @@
 
 %include openturns/DualLinearCombinationFunction.hxx
 namespace OT { %extend DualLinearCombinationFunction { DualLinearCombinationFunction(const DualLinearCombinationFunction & other) { return new OT::DualLinearCombinationFunction(other); } } }
+
+%pythoncode %{
+def __DualLinearCombinationFunction_repr_html(self):
+    """Get HTML representation."""
+    return self.__repr_html__()
```
mbaudin47, line 15: Once the code is evaluated, please write the HTML code inside the SWIG/Python layer.

lib/src/Base/Common/openturns/PersistentObject.hxx
mbaudin47: I do not understand this suggestion. The `__repr__()`is `inline`:
https://github.com/openturns/openturns/blob/31fc1f66f383304a354eafd6bb674ce0dc1c9260/lib/src/Base/Common/openturns/PersistentObject.hxx#L141
So why should'nt be `__repr_html__` be inline too?

lib/src/Base/Common/openturns/PersistentObject.hxx
mbaudin47: Why should this particular method be `virtual`, and not the other ones?

lib/src/Base/Func/DualLinearCombinationEvaluation.cxx
```diff
@@ -118,37 +121,58 @@ String DualLinearCombinationEvaluation::__str__(const String & ) const
         if (first) oss << value;
         else if (value > 0.0) oss << " + " << value;
         else oss << " - " << -value;
-        first = false;
```
mbaudin47, line 121: This removal is because the case where the expression is first is already taken into account in the `if (mustPrint)` segment at line 133 below. This is why this code can be removed without any change in the produced output.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx
```diff
@@ -266,7 +266,7 @@ FunctionalChaosResult::FunctionCollection FunctionalChaosResult::getReducedBasis
 }
 
 /* Composed meta model accessor */
-Function FunctionalChaosResult::getComposedMetaModel() const
+DualLinearCombinationFunction FunctionalChaosResult::getComposedMetaModel() const
```
mbaudin47, line 269: This code allows to get the pretty-print, which is attached to the `DualLinearCombinationFunction` class.

mbaudin47:
TODO:
- [ ] `result.getMetaModel()` should benefit from the pretty-print e.g. when there is an independent copula.

See:

https://github.com/openturns/openturns/blob/95f0e0e09ecad661f865c6dfbcc46a3fbe518cb9/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx#L74

jschueller:
mark the PR as ready when done else I probably wont review

mbaudin47:
> mark the PR as ready when done else I probably wont review

Done. I fixed all issues. I could provide a pretty-print to `ComposedFunction`, but this was not my initial goal, so I rather stop the PR at this point. Providing a pretty-print to `ComposedFunction` can be done afterwards.

mbaudin47:
@regislebrun . You asked why the composed metamodel was pretty-printed and not the metamodel in the [PCE exploitation](http://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_exploitation.html) example. This is because this particular example uses a standard normal input distribution. Hence the distribution of the standardized Z is equal to the distribution of the physical X. Your point was that the identity transformation was used in the metamodel in this specific example. Here is a simplified script:
```Python
import openturns as ot

dimension = 2
input_names = ["x1", "x2"]
formulas = ["cos(x1 + x2)", "(x2 + 1) * exp(x1 - 2 * x2)"]
model = ot.SymbolicFunction(input_names, formulas)
distribution = ot.Normal(dimension)
x = distribution.getSample(30)
y = model(x)
algo = ot.FunctionalChaosAlgorithm(x, y, distribution)
algo.run()
result = algo.getResult()
mm = result.getMetaModel()
print("getMetaModel - type", type(mm))
print("print = ", mm)
cm = result.getComposedMetaModel()
print("getComposedMetaModel - type", type(cm))
print("print = ", cm)
```
I added two intermediate prints:
- at https://github.com/openturns/openturns/blob/a4e2515ce8cce8d6a203dcb56ae470cd60f1bc93/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosAlgorithm.cxx#L275 to print the variable `noTransformation` ;
- at https://github.com/openturns/openturns/blob/a4e2515ce8cce8d6a203dcb56ae470cd60f1bc93/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx#L73 to print the value of `transformation.getEvaluation().getImplementation()->getClassName()`.

Here is the output of the script:
```bash
devel@OT-VM:~/GIT/openturns-printpoly/build (printpoly)$ python3 ~/Documents/plot_functional_chaos_exploitation_simplifie.py 
DBG - FunctionalChaosAlgorithm::run
DBG - noTransformation = false
DBG - FunctionalChaosResult()
DBG - transformation - classname = ComposedEvaluation
getMetaModel - type <class 'openturns.func.Function'>
print =  ([0.00304403,1.45374] + [...])o((| y0 = [x0]->[x0]
| y1 = [x1]->[x1]
)o(class=LinearEvaluation name=Unnamed center=[0,0] constant=[0,0] linear=[[ 1 0 ]
 [ 0 1 ]]))
getComposedMetaModel - type <class 'openturns.func.DualLinearCombinationFunction'>
print =  [0.00304403,1.45374] + [...]
```
We see that:
- the distribution is not equal to the standardized distribution (and this is not expected) ;
- the name of the transformation is "ComposedEvaluation" and not "IdentityEvaluation".

lib/src/Base/Common/openturns/PersistentObject.hxx
mbaudin47: Fixed.

lib/src/Base/Common/openturns/PersistentObject.hxx
```diff
@@ -151,6 +151,13 @@ public:
     return OSS() << offset << __repr__();
   }
 
+  /* String converter */
+  inline
+  String __repr_html__() const override
+  {
+    return OSS() << __repr_html__();
```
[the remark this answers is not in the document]
mbaudin47: Fixed

lib/src/Base/Common/openturns/PersistentObject.hxx
mbaudin47: This code was removed.

jschueller:
you can try rebase if the comparison bug is still there

mbaudin47:
> you can try rebase if the comparison bug is still there

I just checked that: the same behaviour is obtained.

regislebrun:
The ComposedDistribution::equal() method is wrong when the two distributions have an independent copula. The following implementation of equal() and operator== solve the issue:
```
/* Comparison operator */
Bool ComposedDistribution::operator ==(const ComposedDistribution & other) const
{
  if (this == &other) return true;
  // The copula...
  // Store the result of hasIndependentCopula() as it may be costly.
  const Bool hasIndependent = hasIndependentCopula();
  // One of the copulas is the independent one, not the other
  if (!hasIndependent == other.hasIndependentCopula()) return false;
  // None of the copulas is the independent one, and the two copulas are different
  if (!hasIndependent && !(copula_ == other.getCopula())) return false;
  // Then the marginals
  for (UnsignedInteger i = 0; i < dimension_; ++i)
  {
    const Distribution left(distributionCollection_[i]);
    const Distribution right(other.distributionCollection_[i]);
    if (!(left == right)) return false;
  }
  return true;
}

Bool ComposedDistribution::equals(const DistributionImplementation & other) const
{
  // First, test the dimension
  if (!(dimension_ == other.getDimension())) return false;
  // Second, check with cast
  const ComposedDistribution* p_other = dynamic_cast<const ComposedDistribution*>(&other);
  if (p_other != 0) return (*this == *p_other);
  // Third, check by properties
  // We coud go there eg. when comparing a ComposedDistribution([Normal()]*2) with a Normal(2)
  // The copula...
  // Store the result of hasIndependentCopula() as it may be costly.
  const Bool hasIndependent = hasIndependentCopula();
  // One of the copulas is the independent one, not the other
  if (!hasIndependent == other.hasIndependentCopula()) return false;
  // None of the copulas is the independent one, and the two copulas are different
  if (!hasIndependent && !(copula_ == other.getCopula())) return false;
  // Then the marginals
  for (UnsignedInteger i = 0; i < dimension_; ++i)
  {
    const Distribution left(distributionCollection_[i]);
    const Distribution right(other.getMarginal(i));
    if (!(left == right)) return false;
  }
  return true;
}
```
Feel free to introduce it into this PR or another

mbaudin47:
Thank you for the bug fix. I will include it within this PR. Can you suggest a unit test?

jschueller:
@regislebrun you meant the bug happens with a dependent copula ?

lib/src/Base/Func/DualLinearCombinationFunction.cxx
```diff
@@ -64,4 +64,9 @@ String DualLinearCombinationFunction::__str__(const String & offset) const
   return getImplementation()->__str__(offset);
 }
 
+String DualLinearCombinationFunction::__repr_html__() const
+{
+  return getImplementation()->__repr_html__();
+}
+
```
jschueller, line 71: this should not be needed as its already defined in the parent class (Function)

lib/src/Base/Func/SymbolicFunction.cxx
```diff
@@ -141,6 +141,14 @@ String SymbolicFunction::__str__(const String & offset) const
   return getImplementation()->__str__(offset);
 }
 
+/* String converter */
+String SymbolicFunction::__repr_html__() const
+{
+  OSS oss;
+  oss << getImplementation()->__repr_html__();
+  return oss;
+}
```
jschueller, line 150: that one too

lib/src/Base/Func/openturns/Function.hxx
```diff
@@ -86,6 +86,9 @@ public:
   /** String converter */
   String __str__(const String & offset = "") const override;
 
+  /** String converter */
+  String __repr_html__() const override;
```
jschueller, line 90: html converter

regislebrun:
No. The case of two independent copulas was not properly detected. There was an early detection of the case where one of the copula is independent and not the other, using the hasIndependentCopula() method, then the two copulas were compared either if they were both independent or both dependent. The problem is that you can have an independent copula due to a specific choice of parameter values (e.g. a zero correlation for a normal distribution) even if the copula is not the independent one in general. In this case it is better to use the hasIndependentCopula() method instead of a dirct comparison between copulas.

lib/src/Base/Func/openturns/SymbolicFunction.hxx
```diff
@@ -58,6 +58,7 @@ public:
   /** String converter */
   String __repr__() const override;
   String __str__(const String & offset = "") const override;
+  String __repr_html__() const override;
```
jschueller, line 61: unneeded

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/FunctionalChaosResult.hxx
```diff
@@ -132,7 +132,7 @@ private:
   FunctionPersistentCollection Psi_k_;
 
   /** Composed meta model */
-  Function composedMetaModel_;
+  DualLinearCombinationFunction composedMetaModel_;
```
jschueller, line 135: here too, revert it to Function

python/src/DualLinearCombinationFunction.i
jschueller: yes, you had to overload it from c++, so this can be dropped

jschueller:
> No. The case of two independent copulas was not properly detected. There was an early detection of the case where one of the copula is independent and not the other, using the hasIndependentCopula() method, then the two copulas were compared either if they were both independent or both dependent. The problem is that you can have an independent copula due to a specific choice of parameter values (e.g. a zero correlation for a normal distribution) even if the copula is not the independent one in general. In this case it is better to use the hasIndependentCopula() method instead of a dirct comparison between copulas.

ok, but the current code seems ok as it gives True for:
`ot.Normal(2) == ot.ComposedDistribution([ot.Normal()]*2)`

regislebrun:
@jschueller I only tested the C++ code, and with the current master, 
```
ComposedDistribution(Collection<Distribution>(2, Normal())) == Normal(2)
```
is false.
I don't understand how two distributions with different dimensions can be equal at the Python level, as the first test in EllipticalDistribution::equals() and ComposedDistribution::equals() is to check if the dimensions are equal or not!

jschueller:
I added a fix and test in misc2

lib/src/Base/Func/DualLinearCombinationFunction.cxx
josephmure, line 71:
```suggestion
```

lib/src/Base/Func/openturns/DualLinearCombinationFunction.hxx
```diff
@@ -52,6 +52,7 @@ public:
   /** String converter */
   String __repr__() const override;
   String __str__(const String & offset = "") const override;
+  String __repr_html__() const override;
```
josephmure, line 55:
```suggestion
```

python/src/DualLinearCombinationFunction.i
```diff
@@ -8,3 +8,11 @@
...
+
+DualLinearCombinationFunction._repr_html_ = __DualLinearCombinationFunction_repr_html
+%}
```
josephmure, line 18:
```suggestion
```

lib/src/Base/Func/SymbolicFunction.cxx
```diff
@@ -141,6 +141,14 @@ String SymbolicFunction::__str__(const String & offset) const
...
+
```
josephmure, line 151:
```suggestion
```

lib/src/Base/Func/SymbolicFunction.cxx
mbaudin47, line 150:
```suggestion
```

lib/src/Base/Func/openturns/Function.hxx
```diff
@@ -86,6 +86,9 @@ public:
   /** String converter */
   String __str__(const String & offset = "") const override;
 
+  /** String converter */
```
josephmure, line 89:
```suggestion
  /** HTML converter */
```

lib/src/Base/Func/openturns/SymbolicFunction.hxx
mbaudin47, line 61:
```suggestion
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/FunctionalChaosResult.hxx
josephmure, line 135:
```suggestion
  Function composedMetaModel_;
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx
mbaudin47: That was just for the record, not as a suggestion for code improvement.

lib/src/Base/Func/DualLinearCombinationEvaluation.cxx
mbaudin47: That comment is for the record and this is why I resolve it.

lib/src/Base/Func/openturns/Function.hxx
```diff
@@ -86,6 +86,9 @@ public:
   /** String converter */
   String __str__(const String & offset = "") const override;
 
+  /** HTML converter */
```
mbaudin47, line 89:
```suggestion
  /** html converter */
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx
mbaudin47, line 269:
```suggestion
Function FunctionalChaosResult::getComposedMetaModel() const
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/FunctionalChaosResult.hxx
```diff
@@ -98,7 +98,7 @@ public:
   virtual FunctionCollection getReducedBasis() const;
 
   /** Composed meta model accessor */
-  virtual Function getComposedMetaModel() const;
+  virtual DualLinearCombinationFunction getComposedMetaModel() const;
```
mbaudin47, line 101:
```suggestion
  virtual Function getComposedMetaModel() const;
```

lib/src/Base/Func/SymbolicFunction.cxx
```diff
@@ -141,6 +141,7 @@ String SymbolicFunction::__str__(const String & offset) const
   return getImplementation()->__str__(offset);
 }
 
+
```
mbaudin47, line 144:
```suggestion
```

mbaudin47:
The code is now tight, but the pretty-print feature is broken after 01b07108bda4747afbbdf513f57b4ecc6fe429d3. This is why I move it to draft.

python/src/Function.i
```diff
@@ -435,4 +435,6 @@ class PythonFunction(Function):
             instance.isVariableLinear = variablesLinearity
                           
         return Function(instance)
+
+Function._repr_html_ = Function.__repr_html__
```
jschueller, line 439: oh, so the right python name is actually _repr_html_
it should be renamed in the cxx side so this alias can be dropped then (else it wont be available for other objects)

mbaudin47: I thought of this way, but it breaks the consistency of the C++ API naming:
```cpp
// Option 1 : rename in C++ so that Python can recognize the naming
virtual String __repr__() const;
virtual String _repr_html_() const;  // or __repr_html__ ?
virtual String __str__(const String & offset = "") const;
```
In other words, the C++ shows doubled underscores, except for the `html` repr. The advantage of this naming is the corresponding Python classes are automatically pretty-printed, without any need for the `.i` trick.

The option 2 is as currently implemented in the PR. I agree with you: this means that only classes that have the `_repr_html_` trick in their `.i` will get pretty-printed. One drawback is that we choose the C++ name `_ repr_html_` not based on an internal C++ standard way of programming, but based on the naming imposed on the Python side.

I did not find any C++ standard on this topic: this way of programming seems to be Python-specific.

I do not known if you think of these details when posting your comment. Which option should I pick finally?

jschueller: Use `_repr_html_`

mbaudin47: Ok. Fixed at 1b4003f67c39301cfeb121c6446786b08b032249.

python/src/DualLinearCombinationFunction.i
mbaudin47: Ok. This is done after 270217be782ff9c0cf28dbb5cb0d519137ede00e.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/FunctionalChaosResult.hxx
```diff
@@ -30,7 +30,7 @@
 #include "openturns/Function.hxx"
 #include "openturns/Distribution.hxx"
 #include "openturns/OrthogonalBasis.hxx"
-
+#include "openturns/DualLinearCombinationFunction.hxx"
```
jschueller, line 33:
```suggestion
```

mbaudin47:
@jschueller, @regislebrun , @sofianehaddad, @josephmure : This seems OK for merging according to me, provided that a final round of reviews confirms this.

jschueller:
good for me

mbaudin47:
I am not making any progress on this PR. Any review will be highly appreciated.

jschueller (approved):

josephmure (approved):

mbaudin47:
@regislebrun @sofianehaddad : This PR seems to be ready, but a final round of review may help before the release.

mbaudin47:
@jschueller : As far as I can see, we cannot hope to get a review from the Airbus team who seems busy. Given that 2 approvals are given, the content and the amount of review, I think that this PR can be merged so that it can make its way to the 1.22 release.

Issue #2323, opened 2023-05-25
Minor wording issues in the cross entropy importance sampling example


JPelamatti:
Some minor wording issues / typos I noticed in this [example](http://openturns.github.io/openturns/master/auto_reliability_sensitivity/reliability/plot_crossentropy.html)

- margin(s) should be marginal distribution(s) or marginal(s) 
- standard space is missing an article 'the' at a few locations in the text
- For more advanced use -> For a more advanced usage

I might have missed other stuff.
Cheers

jschueller:
cc @m-balesdent @l-brevault

PR #2326, opened 2023-05-31
Reset auxiliary distribution of Cross Entropy Importance Sampling before use

m-balesdent:
As discussed in the previous meeting, after a run, current versions of Cross Entropy algorithms (StandardSpace and PhysicalSpace) keep in memory the auxiliary density as default auxiliary distribution. This may have impact in case of multiple runs. Furthermore, this behavior is not the same as Subset Sampling. 

This PR proposes to automatically reset the auxiliary distribution with default values in case of multiple runs of StandardSpaceCrossEntropyImportanceSampling and PhysicalSpaceCrossEntropyImportanceSampling. 

To do so, a new method  `resetAuxiliaryDistribution` has been created in order to reinitialize the auxiliary distribution to the default one in case of multiple runs of Importance Sampling.

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Simulation/CrossEntropyImportanceSampling.cxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/CrossEntropyImportanceSampling.cxx
@@ -92,9 +92,19 @@ Point CrossEntropyImportanceSampling::optimizeAuxiliaryDistributionParameters(co
 }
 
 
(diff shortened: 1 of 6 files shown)


regislebrun (approved):

jschueller:
would you mind also fixing the typos reported in #2323 ?

jschueller (approved):

PR #2329, opened 2023-06-01
Update plot_crossentropy.py

l-brevault:
This PR solves the issue [#2323](https://github.com/openturns/openturns/issues/2323) by correcting several typos.

### DIFF
--- a/python/doc/examples/reliability_sensitivity/reliability/plot_crossentropy.py
+++ b/python/doc/examples/reliability_sensitivity/reliability/plot_crossentropy.py
@@ -5,7 +5,7 @@ Cross Entropy Importance Sampling
 # %%
 #
 # The objective is to evaluate a failure probability using Cross Entropy Importance Sampling.
-# Two versions working in standard or physical spaces are implemented.
+# Two versions working in the standard or physical spaces are implemented.
 # See :class:`~openturns.experimental.StandardSpaceCrossEntropyImportanceSampling` and :class:`~openturns.experimental.PhysicalSpaceCrossEntropyImportanceSampling`.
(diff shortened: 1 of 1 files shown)


python/doc/examples/reliability_sensitivity/reliability/plot_crossentropy.py
```diff
@@ -5,7 +5,7 @@
 # %%
 #
 # The objective is to evaluate a failure probability using Cross Entropy Importance Sampling.
-# Two versions working in standard or physical spaces are implemented.
+# Two versions working in the standard or physical spaces are implemented.
```
mbaudin47, line 8: I think that the subject "version" cannot be associated with the verb "work": a version cannot work. So I suggest to remove the verb, and the result is:
```suggestion
# Two versions in the standard or physical spaces are implemented.
```

python/doc/examples/reliability_sensitivity/reliability/plot_crossentropy.py
```diff
@@ -356,4 +356,4 @@
 
 
 # %%
-# The optimized auxiliary distribution with the dependency between the two margins allows one to better fit the failure domain resulting a lower coefficient of variation.
+# The optimized auxiliary distribution with the dependency between the two marginals allows one to better fit the failure domain resulting a lower coefficient of variation.
```
mbaudin47, line 359:
```suggestion
# The optimized auxiliary distribution with a dependency between the two marginals allows one to better fit the failure domain, resulting in a lower coefficient of variation.
```

mbaudin47 (changes requested):

sofianehaddad (approved):

jschueller:
could you update the  Changelog to add it to the list of bugs fixed ?

PR #2336, opened 2023-06-19
Doc: Added 3 examples on regression analysis

jschueller:
Create a new Linthurst use case for regression


mbaudin47:
This is a follow-up to #2288.

PR #2346, opened 2023-06-24
Update plot_chaos_sobol_confidence.py

mbaudin47:
Define an interesting plot as the default, instead of the first figure in the script. The goal of the example is to plot the confidence interval, so that the figure should show that.

image

**Figure 1.** The figure of the example "Compute Sobol’ indices confidence intervals", in the example section. **Before** the PR.

image

**Figure 2.** The figure of the example "Compute Sobol’ indices confidence intervals", in the example section. **After** the PR.

### DIFF
--- a/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_sobol_confidence.py
+++ b/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_sobol_confidence.py
@@ -57,6 +57,7 @@ Y = g(X)
 
 # %%
 
+
 def computeSparseLeastSquaresChaos(X, Y, basis, total_degree, distribution):
     """
     Create a sparse polynomial chaos with least squares.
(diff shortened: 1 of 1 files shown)


python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_sobol_confidence.py
```diff
@@ -57,7 +57,6 @@
 
 # %%
 
-
 def computeSparseLeastSquaresChaos(X, Y, basis, total_degree, distribution):
```
mbaudin47, line 60:
```suggestion

def computeSparseLeastSquaresChaos(X, Y, basis, total_degree, distribution):
```

JPelamatti (approved):

mbaudin47:
@sofianehaddad : This is a kind reminder of the request of a review on this topic.

l-brevault (approved):

Issue #2347, opened 2023-06-26
Build error of `/usr/include/tbb/machine/gcc_generic.h:39:20: error: operator '||' has no left operand` in s390x on Fedora


topazus:
```bash

[...]

PR #2349, opened 2023-06-27
Misc

jschueller:
- #2347
- Fix distribution operator==
- Add issue templates
- Deprecate Pagmo.setGenerationNumber for Pagmo.setMaximumIterationNumber
- Fix missing docstrings by completing doc or hiding unneeded methods in various classes
- Allow to stop Log redirection
- Fix windows compilation of Ceres when built with full features (windows.h conflict)


.github/ISSUE_TEMPLATE/0_bug.yml
```diff
@@ -0,0 +1,68 @@
...
+    attributes:
+      label: What happened?
+      description: What should have happened instead? Please provide as many details as possible.
+    validations:
+      required: true
+  - type: textarea
+    id: info
+    attributes:
+      label: How to reproduce the issue?
+      description: Please add a minimal script that helps replicating the issue.
```
mbaudin47, line 17:
```suggestion
      description: Please add a minimal Python script that helps replicating the issue.
```

python/doc/examples/reliability_sensitivity/design_of_experiments/plot_plot_design.py
```diff
@@ -1,112 +0,0 @@
-"""
```
mbaudin47, line 1: I do not agree with this change, that is, removing the `PlotDesign` method. This is because `PlotDesign` shows how a `Sample` fills the space. This is an essential property of some space filling designs, such as LHS and low discrepancy sequences. See [^Santner2003] chapter 5 page 121 for different methods on this topic. Hence, providing this feature in the library is very different from having X/YTicks (which is a good thing). It provides not only a graphical value, but a scientific value for the user. 

[^Santner2003]: Santner, T. J., Williams, B. J., Notz, W. I., & Williams, B. J. (2003). The design and analysis of computer experiments (Vol. 1). New York: Springer.

python/doc/pyplots/LHSExperiment.py
```diff
@@ -8,5 +8,15 @@
 bounds = distribution.getRange()
 lhs = ot.LHSExperiment(distribution, size)
 sample = lhs.generate()
-subdivisions = [size] * dim
```
mbaudin47, line 11: This line shows that the number of subdivisions is equal to the sample size, which is the essential property of a LHS. Replacing this code by a calculation obscures the code and hides the mathematical feature of this particular design of experiment.

python/src/viewer.py
```diff
@@ -886,6 +891,8 @@ def PlotDesign(
     >>> subdivisions = [size]*dim
     >>> fig = PlotDesign(sample, bounds, subdivisions)
     """
+    warnings.warn('PlotDesign is deprecated in favor of DrawPairs', DeprecationWarning)
```
mbaudin47, line 894: No, `PlotDesign` has a completely different purpose and cannot be replaced with `DrawPairs`.

mbaudin47 (changes requested):
This PR also creates new get/set X/YTicks on `Graph` which is useful.

Removing PlotDesign does not seem relevant to me, and this is why I request changes on this specific part of the PR.

.github/ISSUE_TEMPLATE/0_bug.yml
jschueller: it can be c++ or Python, lets say code

sofianehaddad:
Why are you hiding `TruncatedDistribution::getSimplified`? Otherwise fine with me

jschueller:
its i not meant to be used on its own: it is only used in randommixture

sofianehaddad (approved):

jschueller:
hi @regislebrun do you mind having a look at this one ?

python/src/TruncatedDistribution.i
```diff
@@ -6,5 +6,7 @@
 
 %include TruncatedDistribution_doc.i
 
+%ignore OT::TruncatedDistribution::getSimplifiedVersion;
```
regislebrun, line 9: Why do you need to hide this method? If I remember well, Anne was using it in her studies some time ago.

regislebrun (approved):
Excepted the point on hiding TruncatedDistribution::getSimplifiedVersion(), it is good to go for me.

PR #2353, opened 2023-07-11
New use case : two-degree-of-freedom oscillator

m-balesdent:
This PR proposes a new use case presented by JM Bourinet at the users day of 2023. This use case deals with a two-degree-of-freedom non linear oscillator which involves a highly non linear limit state function. This use case presents 8 uncertainties modeled according to Log-Normal distributions. The failure probability is very low (3.78 10-7).

To illustrate this use case, an example which deals with FORM / SORM is also included in this PR. This example allows to show the linear approximation of FORM and the refinement proposed by SORM on cross-cuts graphs near the design point. Eventually, post analytical Importance Sampling is performed in order to correct the FORM approximation.


python/src/usecases/oscillator.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class Oscillator:
+    """
+    Data class for the oscillator example.
+
+    Attributes
+    ----------
+
+    dim : Constant
```
jschueller, line 15: I expected a type here (like int, float), not Constant

python/src/usecases/oscillator.py
```diff
@@ -0,0 +1,217 @@
...
+          dim=8, dimension of the problem
+
+    model : `SymbolicFunction`
+            The limit state function
+
+    muMp : Constant
```
jschueller, line 21: idem, float

jschueller (approved):
great

python/src/usecases/oscillator.py
```diff
@@ -0,0 +1,217 @@
...
+    distributionKp : `LogNormalMuSigmaOverMu` distribution of the spring stiffness of the primary system
+                      ot.LogNormalMuSigmaOverMu(muKp, sigmaOverMuKp).getDistribution()
+
+    muKs : Constant
+          muKs = 0.01, mean of the spring stiffness of the secondary system
+
+    sigmaOverMuKs : Constant
+          sigmaOverMuKs = 0.2, coefficient of variation of the spring stiffness of the secondary system
+
+    distributionKs : `LogNormalMuSigmaOverMu` distribution of the spring stiffness of the secondary system
```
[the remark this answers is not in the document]
jschueller: nevertheless, this attribute is of type "LogNormalMu", not "LogNormalMuSigmaOverMu"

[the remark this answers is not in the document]
m-balesdent: understood.

m-balesdent:
@vchabri and @mbaudin47, maybe you could be interested by reviewing this PR ?

python/doc/_static/oscillator.png
mbaudin47, line 1: I guess that you have the vector .svg figure. I suggest to write the .svg file in the same directory. This may allow to increase / decrease bitmap resolution if needed, or switch to a vector version when needed.

python/src/usecases/oscillator.py
```diff
@@ -0,0 +1,217 @@
...
+        # Random variable : Fs
+        self.muFs = 27.5
+        self.sigmaOverMuFs = 0.1
+
+        # Random variable : S0
+        self.muS0 = 100.0
+        self.sigmaOverMuS0 = 0.1
+
+        # create the limit state function model
+        formula = "Fs-3*ks*sqrt("
```
mbaudin47, line 139: Please separate arithmetic operators with blank spaces so that the code can be read more easily.

python/src/usecases/oscillator.py
```diff
@@ -0,0 +1,217 @@
...
+
+        # Intensity of white noise excitation
+        self.distributionS0 = ot.LogNormalMuSigmaOverMu(
+            self.muS0, self.sigmaOverMuS0
+        ).getDistribution()
+        self.distributionS0.setName("Intensity of white noise excitation")
+        self.distributionS0.setDescription(["S0"])
+
+        # Joint distribution of the input parameters
+        self.distribution = ot.ComposedDistribution(
```
mbaudin47, line 206: I suggest to create a new attribute with the reference probability with the largest possible number of significant digits. This will allow to create benchmark scripts when needed. Also, please write in the comments of this function how this reference probability was estimated. This may allow to improve the reference when possible.

python/doc/usecases/use_case_oscillator.rst
```diff
@@ -0,0 +1,97 @@
...
+- :math:`\zeta_p \sim \text{LogNormalSigmaOverMu}(  \mu_{\zeta_p} = 0.05, \delta_{\zeta_p} = 0.4)`
+- :math:`\zeta_s \sim \text{LogNormalSigmaOverMu}(  \mu_{\zeta_s} = 0.02,  \delta_{\zeta_s} = 0.5)`
+- :math:`F_s \sim \text{LogNormalSigmaOverMu}(  \mu_{F_s} = 27.5,  \delta_{F_s} = 0.1)`
+- :math:`S_0 \sim \text{LogNormalSigmaOverMu}( \mu_{S_0} = 100,  \delta_{S_0} = 0.1)`
+
+
+
+The failure probability is:
+
+.. math:: P_f = \text{Prob}(G(m_p,m_s,k_p,k_z,\zeta_p,\zeta_s,F_s,Z_0) \leq 0).
```
mbaudin47, line 72: Please use the `\Prob` macro from https://github.com/openturns/openturns/blob/master/python/doc/math_notations.sty

python/test/t_oscillator.py
```diff
@@ -0,0 +1,12 @@
+#! /usr/bin/env python
+
+from openturns.testing import assert_almost_equal
+from openturns.usecases import oscillator
+
+"""
+Test the WingWeightModel data class on a reference point.
```
mbaudin47, line 7: You mean "Oscillator", not WingWeight I guess.

python/test/t_oscillator.py
```diff
@@ -0,0 +1,12 @@
...
+"""
+osc = oscillator.Oscillator()
+
+
+assert_almost_equal(osc.model(osc.distribution.getMean()), [23.1897], 1e-2)
```
mbaudin47, line 12: This unit test is almost empty. And, indeed, testing such a difficult benchmark function is not that easy (otherwise we would not create it). I feel, however, that any bug in the code may become unnoticed.

Please consider using a more thorough testing of the attributes, such as the fact that the most advanced FORM-IS method provides a rough approximation of the reference probability without being too costly (say, within 1 second or so). This will make maintenance much more easy. You may use the unit tests from otbenchmark as a reference : https://github.com/mbaudin47/otbenchmark/blob/master/tests/test_RelabilityProblem14.py

mbaudin47 (changes requested):
Thank you for this highly requested use case!
I feel that there is room for improvement here, especially on the testing side. 

As a side topic, may you share your thoughts on which rules we may use to select which of the OT library of the Otbenchmark module should contain future use cases like this one?

mbaudin47:
The most spectacular figure is that one, where the design point is in the center.

image

python/doc/examples/reliability_sensitivity/reliability/plot_estimate_probability_form_oscillator.py
```diff
@@ -0,0 +1,391 @@
...
+event = ot.ThresholdEvent(G, ot.Less(), 0.0)
+event.setName("failure")
+
+# %%
+# Cross cuts in the physical space
+# ---------------------------------
+
+# %%
+# Let’s have a look on 2D cross cuts of the limit state function. For each 2D cross cut, the other variables are fixed to the input distribution mean values.
+# This graph allows one to have a first idea of the variations of the function in pair of dimensions. The colors of each contour plot are comparable.
```
JPelamatti, line 49:
```suggestion
# This graph allows one to have a first idea of the variations of the function in pairs of dimensions. The colors of each contour plot are comparable.
```

python/doc/examples/reliability_sensitivity/reliability/plot_estimate_probability_form_oscillator.py
```diff
@@ -0,0 +1,391 @@
...
+
+# %%
+# To get an accurate estimate of this probability, a billion of simulations are necessary.
+
+# %%
+# FORM Analysis
+# -------------
+
+# %%
+# To get a first idea of the failure probability with reduced simulation budget, one can use the First Order Realibily Method.
```
JPelamatti, line 125:
```suggestion
# To get a first idea of the failure probability with reduced simulation budget, one can use the First Order Realiability Method (FORM).
```

python/doc/usecases/use_case_oscillator.rst
```diff
@@ -0,0 +1,97 @@
...
+The mass ratio is equal to:
+
+.. math:: \gamma = \frac{m_s}{m_p}
+
+The tuning parameter of the system is equal to:
+
+.. math:: \theta = \frac{\omega_p - \omega_s}{\omega_a}
+
+Eight uncertainties are considered in the system:
+- on the masses of the primary and secondary systems (:math:`m_p` and :math:`m_s`),
```
JPelamatti, line 50: The bullet points here are not rendered correctly in the artefact

python/doc/examples/reliability_sensitivity/reliability/plot_estimate_probability_form_oscillator.py
mbaudin47:
```suggestion
# To get a first idea of the failure probability with reduced simulation budget, one can use the First Order Reliability Method.
```

python/src/usecases/oscillator.py
m-balesdent: Actually, black is not able to modify strings formula.

python/src/usecases/oscillator.py
m-balesdent: Thanks for your remark. Due to the magnitude order of this probability, estimating with accuracy this probability requires 10^10 or 10^11 samples. So I couldn't do that on my laptop and I prefered to indicate the value of the reference paper.

python/test/t_oscillator.py
m-balesdent: Thanks for your remark. I added a test using FORM.

python/doc/usecases/use_case_oscillator.rst
m-balesdent: fixed

python/doc/usecases/use_case_oscillator.rst
```diff
@@ -0,0 +1,97 @@
...
+- on the spring stiffeness of the primary and secondary oscillators (:math:`k_p` and :math:`k_s`),
+- on the damping ratios of the primary and secondary systems (:math:`\zeta_p` and :math:`\zeta_s`),
+- on the force capacity of the secondary spring (:math:`F_s`),
```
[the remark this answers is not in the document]
m-balesdent: fixed

python/doc/usecases/use_case_oscillator.rst
m-balesdent: fixed

python/doc/_static/oscillator.png
m-balesdent: added

m-balesdent:
> Thank you for this highly requested use case! I feel that there is room for improvement here, especially on the testing side.
> 
> As a side topic, may you share your thoughts on which rules we may use to select which of the OT library of the Otbenchmark module should contain future use cases like this one?

I think we have to choose examples that are quite challenging or that are illustrative to describe several features of the reliability algorithms. For this specific example, it has a significant  curvature and a very low probability of failure. Moreover, it is a commonly used test-case to assess the performance of relability algorithms, especially the ones involving surrogate models.

python/src/usecases/oscillator.py
mbaudin47: I cannot be sure until I try, but my guess is that an importance sampling algorithm with a SORM approximation may perform correctly, isn'it? This is especially true with a symbolic function, since its derivatives are exact in this case. With that combination of algorithms and implementation, a large sample may not be required to acheive a good accuracy. Am I right?

python/src/usecases/oscillator.py
JPelamatti: Oh, nevermind then :)

JPelamatti (approved):

python/doc/examples/reliability_sensitivity/reliability/plot_estimate_probability_form_oscillator.py
```diff
@@ -0,0 +1,391 @@
+"""
+Use the FORM - SORM algorithms on non linear function
```
vchabri: "Non linear" is more British English. In US English, one should better use "nonlinear".

python/src/usecases/oscillator.py
mbaudin47: This can be improved afterwards anyway.

m-balesdent:
> > The most spectacular figure is that one, where the design point is in the center.
> > image
> 
> This image is wonderful. However, I would like to point out two things:
> 
>     * (1) If this plot is generated in the Gaussian standard space, I would recommend to add a point (or a square) for the center of the space (i.e., u=(0,0)) in order to better identify the ``rareness'' of the failure event. With such an information, one would better understand how far is the limit-state surface from the origin;
> 
>     * (2) A last effort could be done in order to add a legend to this plot (maybe in the right-upper corner of the plot, since it will be always a lower-triangular matrix plot). This legend could mention the following items: origin (u=(0,0)), MPFP (u*), FORM LSS approximation (dashed line), ''true'' LSS (dark blue line), green area (Safe domain), orange area (Failure domain).
> 
> 
> Again, this plot is beautiful and inspiring. This is just a minor suggestion in order to enhance it.

Thanks for your proposals. It is now added in the graphs.

jschueller:
do you use the same matplotlib version locally ? continuous integration uses the latest (3.7.2)

python/doc/examples/reliability_sensitivity/reliability/plot_estimate_probability_form_oscillator.py
```diff
@@ -0,0 +1,414 @@
...
+
+# %%
+# To get an accurate estimate of this probability, a billion of simulations are necessary.
+
+# %%
+# FORM Analysis
+# -------------
+
+# %%
+# To get a first idea of the failure probability with reduced simulation budget, one can use the First Order Realiability Method (FORM).
```
mbaudin47, line 126:
```suggestion
# To get a first idea of the failure probability with reduced simulation budget, one can use the First Order Reliability Method (FORM).
```

mbaudin47 (changes requested):

m-balesdent:
> do you use the same matplotlib version locally ? continuous integration uses the latest (3.7.2)

Thanks for the help!

mbaudin47 (approved):
Adding a new attribute should be done for all use cases, I agree.

mbaudin47:
Suggestion: use `\Prob`:

https://github.com/openturns/openturns/blob/b0234d2710d729ffffc86cd5ddde47a1c6c949f6/python/doc/math_notations.sty#L121

m-balesdent:
> `\Prob`:



> Suggestion: use `\Prob`:
> 
> https://github.com/openturns/openturns/blob/b0234d2710d729ffffc86cd5ddde47a1c6c949f6/python/doc/math_notations.sty#L121

this is now fixed

PR #2373, opened 2023-08-23
Convert a FunctionalChaosResult into a LinearModelResult

mbaudin47:
The goal of this PR is to convert a PCE into a linear model. 

## The problem
In the current implementation, when we create a polynomial chaos expansion based on linear regression, we get a `FunctionalChaosResult`. Mathematically, this is the result of a linear regression. However, the Python object is not the `LinearModelResult` that is needed to perform the `LinearModelAnalysis` that actually implements the regression analysis in the library. The goal of this PR is to be able to create a `LinearModelResult` from a `FunctionalChaosResult`.

## Features
- `LinearModelAnalysis`:
  - Creates a new `LinearModelAnalysis::getResidualsStandardError()` method.
  - New HTML Python pretty-print.
- Creates a new `FunctionalChaosResult::getLinearModelResult()` method.
- `LinearModelResult`
  - Creates new `LinearModelResult.getDesign()` and `LinearModelResult.getSigma2()` methods. These are attributes of the class which had no corresponding `get` methods.
  - Creates a new `LinearModelResult::__repr_markdown__()` method.

## Doc
- A new example of a `PythonRandomVector` is created: `plot_distribution_linear_regression.py`. It plots the distribution of the estimators of the variance and standard deviation of linear regression. This example uses the `LinearFunction` class.
- A new theory help page of linear regression is created.
- The help page of the `LinearFunction` class is improved.

## Fixed bugs
- In the ASCII-art pretty-print, the standard deviation is estimated using a code which is only available in the pretty-print source code ([see this code](https://github.com/openturns/openturns/blob/1d861ea09d429a01ae3d946ad99c2baac11b1495/lib/src/Uncertainty/Algorithm/MetaModel/LinearModel/LinearModelAnalysis.cxx#L137)). This is wrong by design, because it prevents from implementing a unit test. Implement the new `LinearModelAnalysis.getResidualsStandardError()` method instead.
- In the normality test for the residuals, the Kolmogorov-Smirnov uses a biased estimator of the standard deviation ([see this code](https://github.com/openturns/openturns/blob/1d861ea09d429a01ae3d946ad99c2baac11b1495/lib/src/Uncertainty/Algorithm/MetaModel/LinearModel/LinearModelAnalysis.cxx#L325)). Uses the new `LinearModelAnalysis.getResidualsStandardError()` method instead.

## Questions
- [ ] The standard deviation estimator looks suspicious to me: [see this code](https://github.com/openturns/openturns/blob/1d861ea09d429a01ae3d946ad99c2baac11b1495/lib/src/Uncertainty/Algorithm/MetaModel/LinearModel/LinearModelAnalysis.cxx#L137). I cannot find a reference for this estimator. I cannot understand why we should not use the square root of the variance. I admit, however, that the distribution of this estimator is correct, as shown in `plot_distribution_linear_regression.py`.
- [ ] Wait for the merge of #2330 for the `isRegression_` attribute of `FunctionalChaosResult`. Then, if regression is false, generate an exception.

### DIFF
--- a/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosResult.cxx
@@ -27,6 +27,9 @@
 #include "openturns/DualLinearCombinationFunction.hxx"
 #include "openturns/Curve.hxx"
 #include <unordered_map>
+#include "openturns/DesignProxy.hxx"
+#include "openturns/LeastSquaresMethod.hxx"
+#include "openturns/SpecFunc.hxx"
 
 BEGIN_NAMESPACE_OPENTURNS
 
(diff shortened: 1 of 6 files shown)


sofianehaddad:
I am not sure it makes sense to make such an analysis with `LinearModelAnalysis`. Even if we solve numerically similar problems (LS) the underlying assumptions differ. I might be wrong but as there is no gaussian assumption on chaos residuals (in comparison with `LinearModel`), you can provide bad guidance to users when using this post-processing.

On the other hand we can not segregate chaos results obtained using `least squares` or `integration`. Suppose you get a chaos using an `integration`method, with the routine you are computing new coefficients (via `least squares`) , the model/result you provide to `LinearModelResult` is not the one you intitially get & you will make probably bad assumptions...

mbaudin47:
> I might be wrong but as there is no gaussian assumption on chaos residuals (in comparison with `LinearModel`), you can provide bad guidance to users when using this post-processing.

I try to elaborate on this.
- In the classical polynomial chaos expansion, the output observations are exact values of the function: there is no noise. Hence, analyzing the distribution of the residuals does not have the same meaning as in classical regression analysis. 
- Notice, however, that the classical regression does not know the source of the discrepancy between observations and predictions. In this case, we only observe some residuals and analyze them. If may happen that these residuals have the Gaussian distribution, and some conclusions can then be drawn from this, but this is not guaranteed _beforehand_. 
- Anyway, this is exactly the same as in the `LinearModelResult`: in fact, you do not know where the observations come from. Now suppose I use the `LinearModelAlgorithm` class to compute the PCE coefficients: how would you know this from the point of view of the `LinearModelAlgorithm`? In fact, we cannot make any difference between these two cases, and this is why this PR makes a step into that direction.
- In a future PR, I plan to implement the forward stagewise, LASSO, Ridge and ElasticNet regression on the `LinearModel` level, and then to use these new features to provide new methods to compute the coefficients. Indeed, there is, after all, no reason why LARS should be the only method to compute the PCE coefficients using regression. This will provide the forward stepwise, backard stepwise, both, forward stagewise, Ridge and ElasticNet as new computation methods for the coefficients of the PCE.

I do not understand your second point: can you please give more details on this use-case?

mbaudin47:
Please split this PR into 2 sub-PR:
- one part which can be integrated (the doc and other improvements),
- one part which focuses on the conversion (which sensitive).

regislebrun:
I disagree with the following sentence in the description of this PR:
*In the current implementation, when we create a polynomial chaos expansion based on linear regression, we get a FunctionalChaosResult. Mathematically, this is the result of a linear regression.*
A functional chaos decomposition is **not** a linear regression, even if one uses a least squares method to compute the decomposition. In linear regression, the mathematical model is
```
Y=\sum_{k=0}^n\alpha_k\phi_k(X)+\varepsilon
```
where `X` is a random vector taking values in `R^n`, `\varphi_k` are given functions from `R^n` into `R^p` and `\varepsilon` is a random vector taking values in `R^p`, independent from `Y`. Note that the functions don't have to be orthonormal wrt the distribution of `X`.

In functional chaos expansion, the mathematical model is, up to a diffeomorphism:
```
Y=\sum_{k=0}^\infty}\alpha_k\phi_k(X)
```
where the functions `\phi_k` are orthonormal wrt the distribution of `X`. If the sum is truncated to `k\leq n`, the associated residual `sum_{k>n}\alpha_k\phi_k(X)` is **not** independent from `X` so analyzing this residual the same way as in regression is wrong.

mbaudin47:
This PR will never be merged because there is a methodogical issue with it. This is why i close it.

Issue #2354, opened 2023-07-17
IterativeThresholdExceedance


mdelozzo:
Hello,
I'm estimating probabilities with `IterativeThresholdExceedance` and I have question: would it be possible to add an option, e.g. `includeThreshold`, to count the number of times the threshold is reached or exceeded (and not only exceeded)? In order to estimate $\mathbb{P}[X\geq t]$, and not only $\mathbb{P}[X > t]$. Or is there another technique I haven't seen?

jschueller:
hi @mdelozzo, indeed it computes $\mathbb{P}[X > t]$
maybe a ComparisonOperator could be used in this class to get the desired behavior
but maybe you can work around with a well chosen $\mathbb{P}[X > t - \epsilon]$ ?

mdelozzo:
yes, clearly, I'm doing very well with this workaround at the moment.

Another question: would it be possible to add a class `IterativeProbability ` instead of letting the user calculate it by hand with `getThresholdExceedance` and `getIterationNumber`?

jschueller:
It can be a new method in that same class

PR #2387, opened 2023-09-12
Misc

jschueller:
Mainly optimization fixes this time:
- Fix OptimizationResult always accepting first point regardless of constraint
- Disable NLopt PRAXIS/AGS/STOGO (nlopt bugs) & run with deterministic seed (SetSeed => setSeed)
- Fix for DLib maximization with version<19.19
- enchanced MultiStart to take constraint violation under account

Also
- Allow stop redirecting Log to file
- #2354 : add a Comparison operator to the class instead hardwiring (X>T)
- Drop matrix/tensor implementation classes from SWIG
- Fix for Sample parsing spaces in csv labels


sofianehaddad (approved):

jschueller:
@regislebrun did you have a look at this one ? I'd really like to merge it before starting another branch

lib/src/Base/Optim/Dlib.cxx
```diff
@@ -645,28 +640,30 @@ void Dlib::run()
...
+//                                         ub);
+//       else
+//         dlib::find_max (searchStrategy,
+//                         stopStrategy,
+//                         objectiveDlibFunction,
+//                         objectiveDlibGradient,
+//                         optimPoint,
+//                         SpecFunc::MaxScalar);
+//     }
+
```
regislebrun, line 661: Dead code?

lib/src/Base/Optim/openturns/DlibFunctions.hxx
```diff
@@ -43,15 +43,16 @@ class DlibGradient
 {
 public:
   // Constructor
-  explicit DlibGradient(const Gradient & gradient);
+  DlibGradient(const Gradient & gradient, const Bool minimization = true);
```
regislebrun, line 46: Why did you removed the explicit specifier?

lib/src/Base/Optim/openturns/DlibFunctions.hxx
```diff
@@ -60,10 +61,11 @@ class DlibHessian
 {
 public:
   // Constructor
-  explicit DlibHessian(const Hessian & hessian);
+  DlibHessian(const Hessian & hessian);
```
regislebrun, line 64: same here

regislebrun (approved):
Only one question for curiosity ;-)

lib/src/Base/Optim/openturns/DlibFunctions.hxx
jschueller: seems like a mistake

lib/src/Base/Optim/Dlib.cxx
jschueller: yes, replace that with an explanation of why we dont use find_max

PR #2411, opened 2023-10-02
Update plot_chaos_ishigami.py

mbaudin47:
Now the example prints the coefficients _and_ the part of variance. The original sentence was wrong: the code prints the part of variance, but the text says that it prints the coefficients.

### DIFF
--- a/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_ishigami.py
+++ b/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_ishigami.py
@@ -269,16 +269,17 @@ for i in range(im.dim):
 # %%
 partOfVariance = chaosSI.getPartOfVariance()
 chaosResult = chaosSI.getFunctionalChaosResult()
+coefficients = chaosResult.getCoefficients()
 orthogonalBasis = chaosResult.getOrthogonalBasis()
 enumerateFunction = orthogonalBasis.getEnumerateFunction()
 indices = chaosResult.getIndices()
 basisSize = indices.getSize()
-print("Index, global index, multi-index, coefficient")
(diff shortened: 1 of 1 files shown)


python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_ishigami.py
```diff
@@ -269,16 +269,17 @@ def plotXvsY(sampleX, sampleY):
...
 indices = chaosResult.getIndices()
 basisSize = indices.getSize()
-print("Index, global index, multi-index, coefficient")
+print("Index, global index, multi-index, coefficient, part of variance")
 for i in range(basisSize):
     globalIndex = indices[i]
     multiIndex = enumerateFunction(globalIndex)
     if partOfVariance[i] > 1.0e-3:
-        print("%d, %d, %s, %.4f" % (i, globalIndex, multiIndex, partOfVariance[i]))
+        print("%d, %d, %s, %.4f, %.4f" % (i, globalIndex, multiIndex, coefficients[i], partOfVariance[i]))
```
mbaudin47, line 282:
```suggestion
        print("%d, %d, %s, %.4f, %.4f" % (i, globalIndex, multiIndex, coefficients[i, 0], partOfVariance[i]))
```

mbaudin47:
> why do we recode the summary we developed ?
The example is interesting because it shows how the different methods to manages multi-indices and coefficients work together. This is fundamental when we want to go beyond existing services. However, we should make clear that there is a built-in feature which does exactly this. I will rewrite this more clearly.

sofianehaddad (approved):

sofianehaddad:
> why do we recode the summary we developed ? regardless it looks good
Good point :-)

PR #2415, opened 2023-10-04
Format Markdown table in example

mbaudin47:
The goal of this PR is to show how a table containing results from a UQ study from OpenTURNS can be easily formatted using the Markdown format.

For example, the code:
```python
from IPython.display import Markdown
markdown = ""
# Itemized list
markdown += "- item\n"
markdown += "- item\n"
markdown += "\n"
# Table
markdown += "| Index | Value |\n"
markdown += "|-------|-------|\n"
n = 5
for i in range(n):
    markdown += "| %5s | %5s |\n" % (i, 1 + i)
markdown += "\n"
Markdown(markdown)
```
prints within a Jupyter Notebook:

image

But inside the Sphinx doc produced by OpenTURNS, this is not rendered:

image

Another solution is to format the table using HTML. But this is cumbersome because of the HTML format itself. Using Pandas for this purpose is overkill. 

I have two questions:
- How to solve this problem nicely?
- In the continuous integration machine, the example produces the error:
```
ModuleNotFoundError: No module named 'IPython'
```
Isn't IPython a default package?

### DIFF
--- a/python/doc/examples/data_analysis/manage_data_and_samples/plot_use_markdown_format.py
+++ b/python/doc/examples/data_analysis/manage_data_and_samples/plot_use_markdown_format.py
@@ -1,0 +1,169 @@
+"""
+Use Markdown to format results
+==============================
+"""
+# %%
+#
+# In this example, we use the Markdown format to pretty-print
+# results from an uncertainty quantification study.
+# We first show how the Markdown function from the IPython module
+# works.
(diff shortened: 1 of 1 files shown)


jschueller:
you probably cannot do that ; IPython is not a default package anywhere, except the full anaconda distro maybe

mbaudin47:
Ok, so let us forget this PR.

Issue #2423, opened 2023-10-11
mutable OptimizationAlgorithm


jschueller:
### What happened?

I noticed in GPD factory the solver was mutable, this not clean, better just make a copy, this can be improved in some other classes too:

### How to reproduce the issue?

```shell
$ grep -nr 'mutable OptimizationAlgorithm' ../lib/
../lib/src/Uncertainty/Process/openturns/WhittleFactory.hxx:166:  mutable OptimizationAlgorithm  solver_;
../lib/src/Uncertainty/Process/openturns/ARMALikelihoodFactory.hxx:194:  mutable OptimizationAlgorithm  solver_;
../lib/src/Uncertainty/Distribution/openturns/OrderStatisticsMarginalChecker.hxx:62:  mutable OptimizationAlgorithm  solver_;
../lib/src/Uncertainty/Distribution/openturns/MeixnerDistribution.hxx:148:  mutable OptimizationAlgorithm  solver_;
../lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GeneralLinearModelAlgorithm.hxx:214:  mutable OptimizationAlgorithm solver_;
```

### Version

master

### Operating System

unknown

### Installation media

unknown

Issue #2406, opened 2023-09-29
Student does not give access to all its parameters


adutfoy:
### What happened?

When the Student is created from the (nu, mu, sigma, R) API: 

- the getMu() method is implemented only when the dimension is 1.
- the setMu() method is implemented only when the dimension is 1.

It should be implemented for any dimensions. 

### How to reproduce the issue?

```shell
from openturns import *
s = Student(1.5, [1.0, 2.0], [3.0, 4.0], CorrelationMatrix(2))
s.getMu()
s.setMu([3.0, 4.0])
```
```


### Version

1.22dev

### Operating System

Linux

### Installation media

from source

### Additional Context

_No response_

jschueller:
it returns a Scalar, so we cannot change this without breaking api

adutfoy:
Yes I saw that it was returning a scalar. But I don't understand why we did not implement the accessor of the parameter in the other dimensions.

jschueller:
indeed, that's a problem

jschueller:
there is also getMean which gives what you want

mbaudin47:
The `getMu()` method should be changed and return a `Point`, whatever the dimension. The doc should be changed accordingly.

PR #2426, opened 2023-10-12
api changes: setVerbose, Elliptical

jschueller:
- use the global Log::Debug channel instead of setVerbose methods

- drop some mutable attributes for #2423 (except for classes that are being worked out in other branches: GPD, GEV)

- EllipticalDistribution API changes for #2429 (& #2406)

- #2439 

- Stable changes from #2417 + drawMeanResidualLifePlot

- Deprecated Point.clean (used in a single test! assert_almost_equal must be used instead)
- Deprecated Gamma.setKLambda (move to protected like the other distributions)

- #2466


sofianehaddad (approved):

jschueller:
TODO move computeDensityGenerator doc after #2405 is merged

PR #2433, opened 2023-10-18
Update plot_chaos_cleaning_strategy.py

adutfoy:
A PC approximation is not a regression meta model. There is no epsilon here!

### DIFF
--- a/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
+++ b/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
@@ -11,30 +11,26 @@ Create a sparse chaos by integration
 # Polynomial chaos expansion
 # --------------------------
 #
-# Let :math:`g : \mathcal{X} \rightarrow \mathbb{R}` be a function
-# where :math:`\mathcal{X} \subseteq \mathbb{R}^p` is the domain of :math:`g`.
-# Let :math:`f` be a probability density function on :math:`\mathcal{X}`.
-# Let :math:`T` be the iso-probabilistic transformation from the physical
-# space :math:`\mathcal{X}` to the standard space :math:`\mathcal{\bar{X}}`:
+# Let :math:`g : \mathcal{D} \rightarrow \mathbb{R}` be a function
+# where :math:`\mathcal{D} \subseteq \mathbb{R}^p` is the domain of :math:`g`.
+# We consider :math:`\vect{X}` a random vector in :math:`\mathcal{D}` which
+# probability density function is denoted by :math:`f`.
+# We assume that :math:`g(\vect{X})` has a finite second order moment.
+# Let :math:`T` be an isoprobabilistic transformation such that :math:`\vect{Z} = T(\vect{X})`
+# follows an unisolvent distribution.
+# We consider :math:`h` the function defined by:
 #
 # .. math::
 #
-#   \xi = T(\boldsymbol{x}) \in \mathcal{\bar{X}}
+#   h = g \circ T^{-1}.
 #
-# for any :math:`\boldsymbol{x} \in \mathcal{X}`.
-# Let :math:`h` be the function defined by the equation :
-#
-# .. math::
-#
-#   h(\boldsymbol{\xi}) = \left(g \circ T^{-1}\right)(\boldsymbol{\xi})
-#
-# for any :math:`\boldsymbol{\xi} \in \mathcal{\bar{X}}`.
-# The polynomial chaos decomposition of :math:`h` is ([blatman2009]_ page 73) :
+# The polynomial chaos decomposition of :math:`h` with respect to the measure of
+# :math:`\vect{Z}` is ([blatman2009]_ page 73) :
 #
 # .. math::
 #
-#   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathbb{N}^p}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
+#   h(\boldsymbol{z}) = \sum_{\boldsymbol{\alpha} \in \mathbb{N}^p}
+#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{z})
 #
 # where :math:`\boldsymbol{\alpha} = (\alpha_1, ..., \alpha_p) \in \mathbb{N}^p`
 # is a multiindex, :math:`a_{\boldsymbol{\alpha}} \in \mathbb{R}` is the
@@ -70,8 +66,8 @@ Create a sparse chaos by integration
 #
 # .. math::
 #
-#   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}^{d}}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
+#   \tidle{h}(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}^{d}}
+#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}).
 #
 # In order to ensure a low error, we may choose a large value of the
 # parameter :math:`P`. This, however, leads to a large number of
@@ -106,15 +102,15 @@ Create a sparse chaos by integration
 #
 #   \mathcal{A}^{d,j} = \left\{ \boldsymbol{\alpha} \in \mathbb{N}^p
 #   \; | \; \|\boldsymbol{\alpha}\|_1 \leq d, \;
-#   \; \|\boldsymbol{\alpha}\|_0 \leq j\right\}
+#   \; \|\boldsymbol{\alpha}\|_0 \leq j\right\}.
 #
 # Therefore, the rank-`j` polynomial chaos expansion is:
 #
 # .. math::
 #
-#   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in
+#   \tidle{h}(\boldsymbol{z}) = \sum_{\boldsymbol{\alpha} \in
 #   \mathcal{A}^{d,j}} a_{\boldsymbol{\alpha}}
-#   \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
+#   \psi_{\boldsymbol{\alpha}}(\boldsymbol{z}).
 #
 # The rank is now a hyperparameter of the model: [blatman2009]_ suggests
 # to use :math:`j = 2, 3, 4`. An example of low-rank PCE for the G-Sobol'
@@ -129,7 +125,7 @@ Create a sparse chaos by integration
 # If :math:`\textrm{card}\left(\mathcal{A}^{d}\right)` is large, many coefficients
 # may be poorly estimated, which may reduce the quality of the metamodel. We may
 # want to select a subset of the coefficients which best predict the output.
-# In other words, we may compute a subset
+# In other words, we may compute a subset :
 #
 # .. math::
 #
@@ -139,8 +135,8 @@ Create a sparse chaos by integration
 #
 # .. math::
 #
-#   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
+#   \tidle{h}(\boldsymbol{z}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}}
+#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{z})
 #
 # An enumeration rule is a function from the set of integers :math:`k` to
 # the corresponding set of multiindices :math:`\boldsymbol{\alpha}`. More
--- a/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
+++ b/python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
@@ -34,12 +34,12 @@ Create a sparse chaos by integration
 # .. math::
 #
 #   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathbb{N}^p}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}) + \epsilon
+#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
 #
 # where :math:`\boldsymbol{\alpha} = (\alpha_1, ..., \alpha_p) \in \mathbb{N}^p`
 # is a multiindex, :math:`a_{\boldsymbol{\alpha}} \in \mathbb{R}` is the
 # coefficient,  :math:`\psi_{\boldsymbol{\alpha}} : \mathcal{\bar{X}} \rightarrow \mathbb{R}`
-# is a multivariate polynomial and :math:`\epsilon` is a random variable.
+# is a multivariate polynomial.
 
 # %%
 #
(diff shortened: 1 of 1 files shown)


regislebrun (approved):

jschueller (approved):

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -71,7 +71,7 @@
 # .. math::
 #
 #   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}^{d}}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}) + \epsilon.
+#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
```
mbaudin47, line 74: I think that the suggested correction is wrong because the expansion is truncated up to the degree $d$ (as explained on lines 53-69). The consequence is that we cannot have equality between the expansion and the function. 

This is why the error was _not_ to write $+\epsilon$. The error was to write that $\epsilon$ is a random variable (which was written earlier), which is not.

I suggest to write:

$$
\widetilde{h}(\boldsymbol{\xi}) 
= \sum_{\boldsymbol{\alpha} \in \mathcal{A}^{d}}
a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
$$

where $\widetilde{h}$ is an approximation of the $h$ function.

PS
One of the reasons I got confused at this stage is the equation 2.54 page 34 in [^LeMaître2010]. But the significant difference is that the left hand side of [^LeMaître2010] is a random variable. One of the potential sources of confusions is that the greek letter $\xi$ does not make clear the distinction with its random counterpart, which should write $\Xi$.

[^LeMaître2010]: Le Maître, O., & Knio, O. M. (2010). _Spectral methods for uncertainty quantification: with applications to computational fluid dynamics._ Springer Science & Business Media.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -34,12 +34,12 @@
 # .. math::
 #
 #   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathbb{N}^p}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}) + \epsilon
+#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
```
mbaudin47, line 37: Very true. Indeed (assuming the orthogonal basis of polynomials is complete), the expansion converges to the function.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -106,15 +106,15 @@
 #
 #   \mathcal{A}^{d,j} = \left\{ \boldsymbol{\alpha} \in \mathbb{N}^p
 #   \; | \; \|\boldsymbol{\alpha}\|_1 \leq d, \;
-#   \; \|\boldsymbol{\alpha}\|_0 \leq j\right\}.
+#   \; \|\boldsymbol{\alpha}\|_0 \leq j\right\}
```
mbaudin47, line 109:
```suggestion
#   \; \|\boldsymbol{\alpha}\|_0 \leq j\right\}.
```

This is a wrong suggestion because the point ends the sentence. Please consider for example eq. 1 page 6 in [^Knuth]:

image


[^Knuth]: Donald E. Knuth, Tracy Larrabee, and Paul M. Roberts. Mathematical Writing. Available at [this link](https://jmlr.csail.mit.edu/reviewing-papers/knuth_mathematical_writing.pdf).

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -106,15 +106,15 @@
...
 #
 # Therefore, the rank-`j` polynomial chaos expansion is:
 #
 # .. math::
 #
 #   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in
 #   \mathcal{A}^{d,j}} a_{\boldsymbol{\alpha}}
-#   \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}) + \epsilon.
+#   \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
```
mbaudin47, line 117: This suggestion is not OK. As before, the truncation up to rank $j$ produces an expansion which is an approximation $\widetilde{h}$ of the function $h$.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -141,7 +141,6 @@
 #
 #   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}}
 #   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi})
-#   + \epsilon.
```
mbaudin47, line 144: The same suggestion applies as before.

mbaudin47 (changes requested):
Thank you for pointing that error. The suggestions, however, are not OK neither. Your sentence "A PC approximation is not a regression meta model." is true, but this was not the intention of this particular page. The goal was to show different ways to truncate the expansion.

Since we are at updating this page, I suggest to update the notations and to switch from the greek letter $\xi$ to the roman letter $z$. The main advantage of this choice is that we can then make a difference with the random counterpart $Z$. This is not as easy with the letter $\Xi$, which may suggest a set. In this case, the domain of the observation $z$ can be denoted $\mathcal{Z}$. If you will, I can finish this.

jschueller:
I think there is a confusion here:
- Anne thinks the epsilon means the normal regression noise 
- Michael wants to explicit the difference between the real model and the metamodel
Dont we usually omit this epsilon when it means "approximation" and just use the "\hat" notation ?
Please resolve this

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
josephmure: Shouldn't the fact the orthogonal basis of polynomials must be complete be added to the text?

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -34,12 +34,12 @@
 # .. math::
```
josephmure, line 34: $\xi$ should be $\boldsymbol{\xi}$ at line 22 (use the \vect macro)

josephmure (changes requested):
I agree with @mbaudin47 's remarks, and the solution proposed by @jschueller.

jschueller:
hello @adutfoy, do you think this will be ready for 1.22 ? or later ?

adutfoy:
I took into account all your remarks. Tell me if it is ok!

@josephmure , @mbaudin47 : this PR needs your validation!

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -11,35 +11,31 @@
...
-# Let :math:`T` be the iso-probabilistic transformation from the physical
-# space :math:`\mathcal{X}` to the standard space :math:`\mathcal{\bar{X}}`:
+# Let :math:`g : \mathcal{D} \rightarrow \mathbb{R}` be a function
+# where :math:`\mathcal{D} \subseteq \mathbb{R}^p` is the domain of :math:`g`.
+# We consider :math:`\vect{X}` a random vector in :math:`\mathcal{D}` which
+# probability density function is denoted by :math:`f`.
+# We assume that :math:`g(\vect{X})` has a finite second order moment.
+# Let :math:`T` be an isoprobabilistic transformation such that :math:`\vect{Z} = T(\vect{X})`
+# follows a distribution uniquely defined by all its moments.
+# We consider :math:`h` the function defined by:
```
mbaudin47, line 21:
```suggestion
# Let :math:`h` be the function defined by:
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -90,31 +86,31 @@
...
+#   \; \|\vect{\alpha}\|_0 \leq j\right\}.
 #
 # Therefore, the rank-`j` polynomial chaos expansion is:
 #
 # .. math::
 #
-#   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in
-#   \mathcal{A}^{d,j}} a_{\boldsymbol{\alpha}}
-#   \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}) + \epsilon.
+#   \tilde{h}(\vect{z}) = \sum_{\vect{\alpha} \in
```
mbaudin47, line 111: widetilde is larger and cover the symbol better:
```suggestion
#   \widetilde{h}(\vect{z}) = \sum_{\vect{\alpha} \in
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -129,7 +125,7 @@
 # If :math:`\textrm{card}\left(\mathcal{A}^{d}\right)` is large, many coefficients
 # may be poorly estimated, which may reduce the quality of the metamodel. We may
 # want to select a subset of the coefficients which best predict the output.
-# In other words, we may compute a subset
+# In other words, we may compute a subset :
```
mbaudin47, line 128:
```suggestion
# In other words, we may compute a subset:
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -56,26 +52,26 @@
...
 #
-# is the 1-norm of the multi-index :math:`\boldsymbol{\alpha}`.
+# is the 1-norm of the multi-index :math:`\vect{\alpha}`.
 # Therefore, the truncated polynomial chaos expansion is:
 #
 # .. math::
 #
-#   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}^{d}}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}) + \epsilon.
+#   \tilde{h}(\vect{z}) = \sum_{\vect{\alpha} \in \mathcal{A}^{d}}
```
mbaudin47, line 69:
```suggestion
#   \widetilde{h}(\vect{z}) = \sum_{\vect{\alpha} \in \mathcal{A}^{d}}
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -56,26 +52,26 @@
...
+#   a_{\vect{\alpha}} \psi_{\vect{\alpha}}(\vect{z}).
 #
 # In order to ensure a low error, we may choose a large value of the
 # parameter :math:`P`. This, however, leads to a large number of
-# coefficients :math:`\boldsymbol{\alpha} \in \mathcal{A}^{d}` to
+# coefficients :math:`\vect{\alpha} \in \mathcal{A}^{d}` to
 # estimate. More precisely, the number of coefficients to estimate
 # is ([blatman2009]_ page 73) :
```
mbaudin47, line 76:
```suggestion
# is (see [blatman2009]_ page 73) :
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -56,26 +52,26 @@
...
+# is the 1-norm of the multi-index :math:`\vect{\alpha}`.
 # Therefore, the truncated polynomial chaos expansion is:
 #
 # .. math::
 #
-#   h(\boldsymbol{\xi}) = \sum_{\boldsymbol{\alpha} \in \mathcal{A}^{d}}
-#   a_{\boldsymbol{\alpha}} \psi_{\boldsymbol{\alpha}}(\boldsymbol{\xi}) + \epsilon.
+#   \tilde{h}(\vect{z}) = \sum_{\vect{\alpha} \in \mathcal{A}^{d}}
+#   a_{\vect{\alpha}} \psi_{\vect{\alpha}}(\vect{z}).
 #
```
mbaudin47, line 71:
```suggestion
#
# where :math:`\widetilde{h}` is an approximation of :math:`h`.
#
```

mbaudin47 (changes requested):
This PR greatly clarifies these equations. Only a few suggestions seems required to me to finalize the PR.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
adutfoy: ok

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -11,35 +11,31 @@
 # Polynomial chaos expansion
 # --------------------------
 #
-# Let :math:`g : \mathcal{X} \rightarrow \mathbb{R}` be a function
-# where :math:`\mathcal{X} \subseteq \mathbb{R}^p` is the domain of :math:`g`.
-# Let :math:`f` be a probability density function on :math:`\mathcal{X}`.
-# Let :math:`T` be the iso-probabilistic transformation from the physical
-# space :math:`\mathcal{X}` to the standard space :math:`\mathcal{\bar{X}}`:
+# Let :math:`g : \mathcal{D} \rightarrow \mathbb{R}` be a function
```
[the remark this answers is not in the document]
adutfoy: Ok. I removed the mention of the support. There is no interest to mention it here.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
adutfoy: ok

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
adutfoy: ok

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
adutfoy: ok

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
```diff
@@ -11,35 +11,31 @@
...
 #
 # .. math::
 #
-#   \xi = T(\boldsymbol{x}) \in \mathcal{\bar{X}}
+#   h = g \circ T^{-1}.
 #
-# for any :math:`\boldsymbol{x} \in \mathcal{X}`.
-# Let :math:`h` be the function defined by the equation :
+# The polynomial chaos decomposition of :math:`h` with respect to the measure of
+# :math:`\vect{Z}` is ([blatman2009]_ page 73) :
```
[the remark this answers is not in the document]
adutfoy: ok

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cleaning_strategy.py
adutfoy: ok

mbaudin47 (approved):
Absolutely great. Thank you for this!

josephmure (approved):

PR #2368, opened 2023-08-12
Fixes doc issues

mbaudin47:
This PR fixes two bugs:
- the first kriging example mentioned by @mdelozzo in #2364
- the bugs mentioned in the API doc of  `KrigingAlgorithm` by @mdelozzo in #2361 
- the bugs mentioned on `Sample` example in #2359

This PR also fixes other doc bugs.
- This PR also fixes bugs in the Bayesian calibration doc:

image

- This PR provides more detailed explanations on the `ParametricFunction` class.
- It provides an improved example to set a column or a row of a `Sample`.
- It removes unnecessary casts (i.e. class conversions) in some examples.

To do this, I used `docfast.py` with the following parameters:
```bash
$ git add . && cd build && make -j8 && make install && cd .. && python3 utils/docfast.py build-sphinxonly response_surface.rst kriging.rst KrigingAlgorithm.rst GeneralLinearModelAlgorithm.rst glm.rst plot_kriging_hyperparameters_optimization.py
```
When I updated the example with `docfast`, I also added the `KrigingAlgorithm` so that I could made the links active (otherwise, they are just in bold face). But the kriging is in the "Response surface" section of the API help page, so I had to add this section.

PR #2447, opened 2023-10-30
Fixes the pretty-print of ComposedDistribution

mbaudin47:
The goal of this PR is to improve the pretty-print feature of the `ComposedDistribution`.

## The problem

Currently, when we create a `ComposedDistribution`, this distribution can be cast into an ordinary `Distribution`. This is what is done in [MetaModelAlgorithm.cxx#L145](https://github.com/openturns/openturns/blob/0a592bd5397c6625809ed4b7bc4c535d53fae4e8/lib/src/Uncertainty/Algorithm/MetaModel/MetaModelAlgorithm.cxx#L145), where the `MetaModelAlgorithm.BuildDistribution` method returns a `Distribution` which is based on a `ComposedDistribution`. One of the consequences of this is that the pretty-print of the `ComposedDistribution` is not triggered, as shown in the next figure.

PR2447 - Before
**Figure 1.** Printing a `Distribution` containing a `ComposedDistribution` before the PR.

The goal of this PR is to get a pretty-print as shown in the next figure.

PR2447 - After
**Figure 2.** Printing a `Distribution` containing a `ComposedDistribution` after the PR.

## Analysis

The fact that a `Distribution` which actually contains a `ComposedDistribution` does not get a pretty-print can be surprising, since PR #2282 introduced a pretty-print for a `ComposedDistribution`. The reason for this behaviour is that the pretty-print is implemented in Python. Since the object is a `Distribution`, the Python pretty-print for this object is triggered, which prints a basic distribution instead of a Composed one. This was found in PR #2368.

The conclusion is that the pretty-print feature of a `ComposedDistribution` must be implemented in C++ so that object-oriented programming can take place.

## Implementation

The PR includes the content of commit 74586396af5684c438765d1523f12bc52abb76be, with a C++ implementation.

## Questions

- [x] Should we rename `__repr_markdown__` into `_repr_markdown_`? Renamed, the method cannot be, said Yoda.


jschueller:
you can keep __repr_markdown__ as this is not used directly by Python

python/src/DistributionImplementation.i
```diff
@@ -92,3 +92,11 @@ Distribution __rmul__(Scalar s)
 
 
 } }
+
+%pythoncode %{
+def __DistributionImplementation_repr_markdown(self):
+    """Get Markdown representation."""
+    return self.__repr_markdown__()
+
+DistributionImplementation._repr_markdown_ = __DistributionImplementation_repr_markdown
```
jschueller, line 101: why rename it ?

python/test/t_MetaModelAlgorithm_std.py
```diff
@@ -26,5 +26,9 @@
...
     [[0], [142.857], [285.714], [428.571], [571.429], [714.286], [857.143], [1000.0]]
 )
-d = ot.MetaModelAlgorithm.BuildDistribution(sample)
-print(d)
+distribution = ot.MetaModelAlgorithm.BuildDistribution(sample)
+print(distribution)
+print("distribution (HTML)")
+print(distribution._repr_html_())
+print("distribution (Markdown)")
+print(distribution._repr_markdown_())
```
jschueller, line 34: make up your mind

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
```diff
@@ -36,14 +41,42 @@
...
 distribution = ot.FunctionalChaosAlgorithm.BuildDistribution(inputSample)
 
+# %%
+# Let us explore the distribution with its fitted parameters.
+
+# %%
+distribution
+
+# %%
+# We can also analyse its properties in more details.
```
josephmure, line 78:
```suggestion
# We can also analyse its properties in more detail.
```

See https://www.linguee.fr/francais-anglais/traduction/de+mani%C3%A8re+plus+d%C3%A9taill%C3%A9e.html

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
```diff
@@ -52,11 +85,16 @@
 distribution.getCopula()
 
 # %%
-# The previous call to `BuildDistribution` is what is done internally by the following constructor of `FunctionalChaosAlgorithm`.
+# The previous call to `BuildDistribution` is what is done internally by the
```
josephmure, line 88:
```suggestion
# The previous call to
# :meth:`~openturns.FunctionalChaosAlgorithm.BuildDistribution`
# is what is done internally by the
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
```diff
@@ -52,11 +85,16 @@
...
 # %%
 algo = ot.FunctionalChaosAlgorithm(inputSample, outputSample)
 
+# %%
+# The previous constructor is the main topic of the example
+# :doc:`Create a polynomial chaos metamodel from a data set </auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos>`.
+
 # %%
 # Reset default settings
 ot.ResourceMap.Reload()
```
josephmure, line 100:
```suggestion
ot.ResourceMap.Reload() # reset default settings
```
This looks odd otherwise:

image

josephmure (changes requested):

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
```diff
@@ -4,9 +4,14 @@
 """
 # %%
 #
-# In this example we show how to use the `BuildDistribution` function to fit a distribution to an input sample.
-# This function is used by the `FunctionalChaosAlgorithm` class when we want to create a polynomial chaos metamodel and we have a design of experiments which have been computed beforehand.
-# In this case, we have to identify the distributions which best fit to the input sample in order to define the input probabilistic model.
+# In this example we show how to use the
+# :class:`~openturns.BuildDistribution` function to fit a distribution to an
```
mbaudin47, line 8:
```suggestion
# :meth:`~openturns.MetaModelAlgorithm.BuildDistribution` function to fit a distribution to an
```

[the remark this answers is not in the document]
mbaudin47: Thank you for pointing this. But, after looking at the code, I realized that this static method actually comes from `MetaModelAlgorithm`.

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
mbaudin47, line 88:
```suggestion
# The previous call to :meth:`~openturns.MetaModelAlgorithm.BuildDistribution` is what is done internally by the
```

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
```diff
@@ -36,14 +41,42 @@
...
+
+# %%
+# In order to speed this example, let us reduce this value.
 
 # %%
 ot.ResourceMap.SetAsUnsignedInteger("FittingTest-LillieforsMaximumSamplingSize", 100)
 
 # %%
+# Then we fit the distribution.
 distribution = ot.FunctionalChaosAlgorithm.BuildDistribution(inputSample)
```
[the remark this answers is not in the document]
mbaudin47: Indeed, that would make sense. This should be the flagship of the Step B.

python/src/DistributionImplementation.i
mbaudin47: This is for consistency with _repr_html_ from Python's point of view, in order to make it consistent with other `_repr_` functions in [IPython.display.html](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html)

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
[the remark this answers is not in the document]
mbaudin47: Fixed at 07ee567bb7d86da4455183cb2bc92c78dafa10cc

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
```diff
@@ -36,14 +41,42 @@
 
 # %%
 # Create a functional chaos model.
-# First, we need to fit a distribution on the input sample. We can do this automatically with the Lilliefors test.
+# The algorithm used by :class:`~openturns.BuildDistribution`
```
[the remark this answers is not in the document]
mbaudin47: Fixed at 07ee567bb7d86da4455183cb2bc92c78dafa10cc

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
mbaudin47: Fixed at 07ee567bb7d86da4455183cb2bc92c78dafa10cc

python/src/DistributionImplementation.i
jschueller: whatever, use the same name everywhere so renames like this can be dropped

mbaudin47: Ok. I removed the unnecessary renames at 25a1b7f8d.

python/test/t_MetaModelAlgorithm_std.py
mbaudin47: I do not understand: Shouldn't we test them both?

josephmure (approved):

jschueller:
a rebase is needed

python/doc/examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_build_distribution.py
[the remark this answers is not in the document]
jschueller: no, its metamodel-oriented to yield the simplest parametric models

regislebrun (approved):
Nice work! Notebook users (e.g. my students) will love it!

lib/src/Base/Common/openturns/Object.hxx
```diff
@@ -113,6 +113,17 @@ public:
...
+  /**
+   * Html string converter
+   *
+   * This method shows Html information on the
+   * internal state of an object. It is used for printing the object
+   * and should only show relevant information.
+   *
+   * @return A Html string
+   */
+  virtual String _repr_html_() const;
```
jschueller, line 125: Now its declared 2 times from the other PR

python/test/t_MetaModelAlgorithm_std.py
jschueller: it was about the naming

jschueller: nevermind, its okay now

PR #2448, opened 2023-10-31
Update plot_pce_design.py

mbaudin47:
Fixed row index of the design matrix.


jschueller:
maybe you could group these small doc PRs to shorten the list next time ?

mbaudin47:
> please group these small doc PRs to shorten the list

You are right: the content is moved to #2450, that will gather several doc improvements.

PR #2491, opened 2023-11-24
Use default colors when adding a graph to another graph

josephmure:
# Goal

Build upon PR #2422 by expanding the use of default colors : 

- PR #2422 applies a default color to a Drawable when it is added to a Graph (provided its color has not been previously explicitly set)
- This PR does the same  to all Drawables of a Graph when that Graph is added to another.

# Implementation

This PR adds a protected method to DrawableImplementation: `setColor(String color, Bool isExplicitlySet)`

This protected method is used by the friend class GraphImplementation to set the color and the isExplicitlySet_ attribute at the same time. Allows is to unexplicitly set colors when it applies the default palette.


mbaudin47:
To make the review easier, I write below a list of the examples which are changed by the PR:
- [plot_estimate_non_parametric_distribution.html](https://openturns.github.io/openturns/latest/auto_data_analysis/distribution_fitting/plot_estimate_non_parametric_distribution.html)
- [plot_overfitting_model_selection.html](https://openturns.github.io/openturns/latest/auto_meta_modeling/general_purpose_metamodels/plot_overfitting_model_selection.html)

mbaudin47 (changes requested):
Thank you for that PR: it is an excellent addition.
I feel, however, that one of the examples you updated can be made even better. To improve on this, I created a PR on your repo:

https://github.com/josephmure/openturns/pull/12

Please consider including it and see how it improves the examples from your point of view.

mbaudin47 (approved):

josephmure:
@jschueller Could you please review?

lib/src/Base/Graph/GraphImplementation.cxx
```diff
@@ -160,7 +160,7 @@ void GraphImplementation::add(const Drawable & aDrawable)
   Drawable&drawable = drawablesCollection_[drawableCount];
   /* TODO: It would be more efficient to have a method to obtain the ith color from a palette */
   if (!drawable.getImplementation()->isColorExplicitlySet())
-    drawable.setColor(DrawableImplementation::BuildDefaultPalette(drawableCount + 1)[drawableCount]);
+    drawable.getImplementation()->setColor(DrawableImplementation::BuildDefaultPalette(drawableCount + 1)[drawableCount], false);
```
jschueller, line 163: this will bypass the copyonwrite and modify the original as well

lib/src/Base/Graph/GraphImplementation.cxx
```diff
@@ -196,7 +196,7 @@ void GraphImplementation::setDrawables(const DrawableCollection & drawableCollec
   {
     Drawable & drawable = drawablesCollection_[i];
     if (!drawable.getImplementation()->isColorExplicitlySet())
-      drawable.setColor(palette[i]);
+      drawable.getImplementation()->setColor(palette[i], false);
```
jschueller, line 199: here too

PR #2497, opened 2023-11-29
CovarianceModel: Add nuggetFactor to optimizable parameters

### LINKED ISSUE #1330: NuggetFactor cannot be made active
As discussed with @regislebrun and @mbaudin47 , there is a nugget factor-related  bug in covariance models: the nugget factor cannot be made active.

This bug exists even in version 1.14.

josephmure:
# Description

This PR adds the `nuggetFactor` to the list of parameters of a `CovarianceModel` so it can be optimized by `GeneralLinearModelAlgorithm`.

It changes the parameter order (size in parentheses) from:

1. scale (inputDimension usually)
2. amplitude (outputDimension)
3. output correlation (outputDimension * (outputDimension-1) / 2)
4. extra parameters (?)

to:
1. scale (inputDimension usually)
2. **nuggetFactor** (1)
3. amplitude (outputDimension)
4. output correlation (outputDimension * (outputDimension-1) / 2)
5. extra parameters (?)

Furthermore, in `GeneralLinearModelAlgorithm`, the default optimization lower bound for the `nuggetFactor` is set to `0.0`. It has the same upper bound as the other parameters (`1e2`).

EDIT: in addition, this PR implements new methods to make activation and deactivation of standard parameters easier: `activateScale`, `activateNuggetFactor` and `activateAmplitude`. Tests and an example are provided.

# Motivation

Although the creation of an algebra of covariance models is planned, which would be a smarter way to handle the nuggetFactor (which could be viewed as the addition of a `DiracCovarianceModel`), @vchabri has an EDF usecase which requires OpenTURNS to estimate nugget factors.

Closes #1330

### DIFF
--- a/lib/src/Base/Stat/CovarianceModelImplementation.cxx
+++ b/lib/src/Base/Stat/CovarianceModelImplementation.cxx
@@ -54,9 +54,10 @@ CovarianceModelImplementation::CovarianceModelImplementation(const UnsignedInteg
   , isDiagonal_(true)
   , isStationary_(false)
   , nuggetFactor_(ResourceMap::GetAsScalar("CovarianceModel-DefaultNuggetFactor"))
-  , activeParameter_(inputDimension_ + outputDimension_)
+  , activeParameter_(0)
 {
-  activeParameter_.fill();
+  for (UnsignedInteger i=0; i < inputDimension_ + 1 + outputDimension_; ++i)
+    if(i != inputDimension_) activeParameter_.add(i);
   updateOutputCovariance();
 }
 
@@ -74,11 +75,12 @@ CovarianceModelImplementation::CovarianceModelImplementation(const Point & scale
   , isDiagonal_(true)
   , isStationary_(false)
   , nuggetFactor_(ResourceMap::GetAsScalar("CovarianceModel-DefaultNuggetFactor"))
-  , activeParameter_(inputDimension_ + outputDimension_)
+  , activeParameter_(0)
 {
   setAmplitude(amplitude);
   setScale(scale);
-  activeParameter_.fill();
+  for (UnsignedInteger i=0; i < inputDimension_ + 1 + outputDimension_; ++i)
+    if(i != inputDimension_) activeParameter_.add(i);
 }
 
 /* Standard constructor with scale, amplitude and spatial correlation parameter parameter */
@@ -96,11 +98,12 @@ CovarianceModelImplementation::CovarianceModelImplementation(const Point & scale
   , isDiagonal_(true)
   , isStationary_(false)
   , nuggetFactor_(ResourceMap::GetAsScalar("CovarianceModel-DefaultNuggetFactor"))
-  , activeParameter_(inputDimension_ + outputDimension_)
+  , activeParameter_(0)
 {
   setAmplitude(amplitude);
   setScale(scale);
-  activeParameter_.fill();
+  for (UnsignedInteger i=0; i < inputDimension_ + 1 + outputDimension_; ++i)
+    if(i != inputDimension_) activeParameter_.add(i);
   setOutputCorrelation(spatialCorrelation);
 }
 
@@ -118,7 +121,7 @@ CovarianceModelImplementation::CovarianceModelImplementation(const Point & scale
   , isDiagonal_(spatialCovariance.isDiagonal())
   , isStationary_(false)
   , nuggetFactor_(ResourceMap::GetAsScalar("CovarianceModel-DefaultNuggetFactor"))
-  , activeParameter_(inputDimension_ + outputDimension_)
+  , activeParameter_(0)
 {
   setScale(scale);
   for (UnsignedInteger i = 0; i < outputDimension_; ++i)
@@ -136,7 +139,8 @@ CovarianceModelImplementation::CovarianceModelImplementation(const Point & scale
       for (UnsignedInteger i = j + 1; i < outputDimension_; ++i)
         outputCorrelation_(i, j) = spatialCovariance(i, j) / (amplitude_[i] * amplitude_[j]);
   } // !isDiagonal
-  activeParameter_.fill();
+  for (UnsignedInteger i=0; i < inputDimension_ + 1 + outputDimension_; ++i)
+    if(i != inputDimension_) activeParameter_.add(i);
 }
 
 /* Virtual constructor */
@@ -947,7 +951,12 @@ void CovarianceModelImplementation::setFullParameter(const Point & parameter)
     scale_[i] = parameter[index];
     ++ index;
   }
-  // Second the amplitude parameter
+  // Second the nugget factor
+  if (!(parameter[index] >= 0.0))
+      throw InvalidArgumentException(HERE) << "In CovarianceModelImplementation::setFullParameter, the component " << index << " of nuggetFactor is negative";
+  nuggetFactor_ = parameter[index];
+  ++ index;
+  // Third the amplitude parameter
   for (UnsignedInteger i = 0; i < outputDimension_; ++ i)
   {
     if (!(parameter[index] > 0.0))
@@ -956,7 +965,7 @@ void CovarianceModelImplementation::setFullParameter(const Point & parameter)
     ++ index;
   }
   CorrelationMatrix outputCorrelation(outputDimension_);
-  // Third the output correlation parameter, only the lower triangle
+  // Fourth the output correlation parameter, only the lower triangle
   for (UnsignedInteger i = 0; i < outputDimension_; ++ i)
     for (UnsignedInteger j = 0; j < i; ++ j)
     {
@@ -972,9 +981,11 @@ Point CovarianceModelImplementation::getFullParameter() const
   // Here we manage only the generic parameters
   // First the scale parameter
   Point parameter(getScale());
-  // Second the amplitude parameter
+  // Second the nugget factor
+  parameter.add(nuggetFactor_);
+  // Third the amplitude parameter
   parameter.add(getAmplitude());
-  // Third the spatial correation parameter, only the lower triangle
+  // Fourth the spatial correation parameter, only the lower triangle
   if (isDiagonal_)
   {
     if (outputDimension_ > 1)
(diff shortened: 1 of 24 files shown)


lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -54,9 +54,10 @@ CovarianceModelImplementation::CovarianceModelImplementation(const UnsignedInteg
   , isDiagonal_(true)
   , isStationary_(false)
   , nuggetFactor_(ResourceMap::GetAsScalar("CovarianceModel-DefaultNuggetFactor"))
-  , activeParameter_(inputDimension_ + outputDimension_)
+  , activeParameter_(0)
 {
-  activeParameter_.fill();
+  for (UnsignedInteger i=0; i < inputDimension_ + 1 + outputDimension_; ++i)
+    if(i != inputDimension_) activeParameter_.add(i);
```
sofianehaddad, line 60: what about `i` for  `i >= inputDimension_` ? don't you consider `amplitude` to be active?

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -96,11 +98,12 @@ CovarianceModelImplementation::CovarianceModelImplementation(const Point & scale
   , isDiagonal_(true)
   , isStationary_(false)
   , nuggetFactor_(ResourceMap::GetAsScalar("CovarianceModel-DefaultNuggetFactor"))
-  , activeParameter_(inputDimension_ + outputDimension_)
+  , activeParameter_(0)
 {
   setAmplitude(amplitude);
   setScale(scale);
-  activeParameter_.fill();
+  for (UnsignedInteger i=0; i < inputDimension_ + 1 + outputDimension_; ++i)
```
sofianehaddad, line 105: same here

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -136,7 +139,8 @@ CovarianceModelImplementation::CovarianceModelImplementation(const Point & scale
       for (UnsignedInteger i = j + 1; i < outputDimension_; ++i)
         outputCorrelation_(i, j) = spatialCovariance(i, j) / (amplitude_[i] * amplitude_[j]);
   } // !isDiagonal
-  activeParameter_.fill();
+  for (UnsignedInteger i=0; i < inputDimension_ + 1 + outputDimension_; ++i)
+    if(i != inputDimension_) activeParameter_.add(i);
```
sofianehaddad, line 143: same here

lib/src/Base/Stat/CovarianceModelImplementation.cxx
```diff
@@ -947,7 +951,12 @@ void CovarianceModelImplementation::setFullParameter(const Point & parameter)
     scale_[i] = parameter[index];
     ++ index;
   }
-  // Second the amplitude parameter
+  // Second the nugget factor
```
sofianehaddad, line 954: in order to not break the current API, I will rather sugget to put this block after the specific one for `amplitude`

lib/src/Base/Stat/IsotropicCovarianceModel.cxx
```diff
@@ -146,16 +146,18 @@ Description IsotropicCovarianceModel::getFullParameterDescription() const
...
+  if (!(scale[0] > 0.0))
+    throw InvalidArgumentException(HERE) << "In IsotropicCovarianceModel::setScale: the scale is << " << scale[0] << " but should be positive";
   kernel_.setScale(scale);
   scale_ = scale;
 }
 
 /* Nugget factor accessor */
 void IsotropicCovarianceModel::setNuggetFactor(const Scalar nuggetFactor)
 {
+  if (!(nuggetFactor >= 0.0))
```
sofianehaddad, line 159: I think this check is already done by `kernel_.setNuggetFactor`

sofianehaddad (commented):
If we discard that nugget factor shall be removed soon, the way to include the `nugget` in the optimization process should be a bit different in order to avoid breaking API at least and do things efficiently!

you also need to consider the nugget factor in the optimization process - for the moment there is no update in that part and results might be wrong!

image

josephmure:
> If we discard that nugget factor shall be removed soon, the way to include the `nugget` in the optimization process should be a bit different in order to avoid breaking API at least and do things efficiently!
> 
> you also need to consider the nugget factor in the optimization process - for the moment there is no update in that part and results might be wrong!
> 
> image

I don't think OpenTURNS uses the parametrization $\sigma^2 \boldsymbol{R} + \tau^2 \boldsymbol{I}_n$, but rather

$$
\sigma^2 \left( \boldsymbol{R} + \eta \boldsymbol{I}_n \right)
$$

See for example:

https://github.com/openturns/openturns/blob/588873084cdd62b47e395542d3fc9f306ae057b2/lib/src/Base/Stat/MaternModel.cxx#L98

As we are optimizing $\sigma$ and $\eta$ rather than $\sigma$ and $\tau$, I don't think the optimization process as it is currently is wrong.

Concerning your point about efficiency, I am not sure what you mean?

lib/src/Base/Stat/CovarianceModelImplementation.cxx
josephmure: What these lines are supposed to do is: activate the scale and amplitude, but not the nuggetFactor, which is between the scale and the amplitude in the parameter order.

This is done to preserve the current behavior as the default (scale and amplitude are active, but not the nuggetFactor).

lib/src/Base/Stat/CovarianceModelImplementation.cxx
josephmure: You know, this is actually what I did at first, but I decided against it in the end, because when you read the expression of the covariance matrix:

$$
\sigma² (\boldsymbol{R}_\theta + \eta \boldsymbol{I}_n)
$$

you want to : 
1. determine the input correlation matrix $\boldsymbol{R}_\theta$ (parametrized by the scale parameter), then 
2. the $\eta$ scalar which is added to its diagonal elements (the nuggetFactor), 
3. and then the amplitude $\sigma$ whose square is applied to the resulting augmented "correlation" matrix $\boldsymbol{R}_\theta + \eta \boldsymbol{I}_n$.

Furthermore, note that `CorrelationMatrixImplementation` actually has additional parameters after the scale and the amplitude: the nondiagonal elements of the **output** correlation matrix. So, IMO, it is more logical to progress from the parameters which affect the **input** correlation matrix (scale and nuggetFactor) to the parameters which affect the output covariance matrix (the amplitude values and the coefficients of the output correlation matrix).

lib/src/Base/Stat/IsotropicCovarianceModel.cxx
josephmure: True, but doing it here allows to catch the exception earlier and produce an error message which might be a little more clear. However, if you think this is overkill, I can remove it.

sofianehaddad:
it is not a book! Here is the [link](https://www.jstatsoft.org/index.php/jss/article/view/v051i01/641)



> > If we discard that nugget factor shall be removed soon, the way to include the `nugget` in the optimization process should be a bit different in order to avoid breaking API at least and do things efficiently!
> > you also need to consider the nugget factor in the optimization process - for the moment there is no update in that part and results might be wrong!
> > image
> 
> I don't think OpenTURNS uses the parametrization σ2R+τ2In, but rather
> 
> σ2(R+ηIn)
> 
> See for example:
> 
> https://github.com/openturns/openturns/blob/588873084cdd62b47e395542d3fc9f306ae057b2/lib/src/Base/Stat/MaternModel.cxx#L98
> 
> As we are optimizing σ and η rather than σ and τ, I don't think the optimization process as it is currently is wrong.
> 
> Concerning your point about efficiency, I am not sure what you mean?
> 
> PS: what is the book you took a screenshot of?

I am aware that we use the convention $\sigma^2  (R +\tau I_n)$. For a fixed $\tau$ value, it makes sense
If $\tau$ has to be calibrated, we should consider a domain of variation that we don't master. Using the convention proposed for the specifc caase where $\tau$ is unknow allows to fix a ratio $\alpha$ in ]0, 1[

josephmure:
I don't understand why mingw has so many failures, and on tests which do not seem to be related to this PR: pyinstallcheck_Study_std or pyinstallcheck_Sample_std or pyinstallcheck_Mesh_std for example.

lib/src/Base/Stat/CovarianceModelImplementation.cxx
josephmure: I think this ordering point is less important now this PR includes easy-to-use methods to activate/deactivate parameters: `activateScale`, `activateNuggetFactor`, `activateAmplitude`.

mbaudin47:
@vchabri : Please review this PR, which may or may not suit your needs.

jschueller (approved):

sofianehaddad (commented):
Nice work, only minor comment : could you add more "tests" with `nugget` factor active?
The idea is to "check" the feature is working accurately

We can define a use-case in 3d using data + perturbation
Also,I would suggest a comparison (for example) between 
 - `AbsoluteExponential([1] * 3)`  
 - `ProductCovarianceModel([AbsoluteExponential(1)]*3)`

This allows at the same time to "validate" the feature in `ProductCovarianceModel`

vchabri:
@mbaudin47 : I don't know whether this PR suits or do not suit "my needs". Not sure that I'm the only one that needs an estimation strategy for the nugget factor in GP regression. It seems more to be a "general" need than a personal one. I thank Joseph (and others contributors/reviewers too) for the time he spent on this. I will try to use the new version in the 'oticscream' module I develop in order to see whether it improves the Q². Despite that, I'm sure it will help several Kriging users as a preliminary step towards the next one (with the improvement of the arithmetic of covariance functions).

vchabri (approved):
Nice work. This paves the way towards a really efficient Kriging with nugget factor.

josephmure:
> Nice work, only minor comment : could you add more "tests" with `nugget` factor active? The idea is to "check" the feature is working accurately
> 
> We can define a use-case in 3d using data + perturbation Also,I would suggest a comparison (for example) between
> 
>     * `AbsoluteExponential([1] * 3)`
> 
>     * `ProductCovarianceModel([AbsoluteExponential(1)]*3)`
> 
> 
> This allows at the same time to "validate" the feature in `ProductCovarianceModel`

Unfortunately I did not have time to do this before the PR was merged. I will try to do it in a different PR.

sofianehaddad:
ok thanks, I think it was merged too quickly

PR #2505, opened 2023-12-11
Distribution: Efficient computeProbability

jschueller:
- for 1d-continuous distributions with analytical CDF override computeProbability to call computeProbabilityGeneral1D
benchmark: with Burr 1e5 computeProbability calls in 0.5s instead of 7s before
- for 1d-continuous distributions where GaussKronrod is used (eg Meixner), make sure we dont integrate outside the range
- for discrete distributions, make sure getSupport(interval) does not return points outside range, else computeProbability can be very slow and inexact (was done for most but not all)
- student.computeCDF returns nan with big values (#2506)

/cc @regislebrun

### DIFF
--- a/lib/src/Uncertainty/Distribution/StudentFunctions.cxx
+++ b/lib/src/Uncertainty/Distribution/StudentFunctions.cxx
@@ -51,12 +51,35 @@ Scalar StudentCDF(const Scalar nu,
   if (x == 0.0) return 0.5;
   if (nu == 1.0) return (tail ? 0.5 - (std::atan(x) * M_1_PI) : 0.5 + (std::atan(x) * M_1_PI));
   const Scalar x2 = x * x;
-  if (nu == 2.0) return (tail ? 0.5 - 0.5 * (x / std::sqrt(2.0 + x2)) : 0.5 + 0.5 * (x / std::sqrt(2.0 + x2)));
+  if (nu == 2.0)
+  {
+    if (std::abs(x) < 1.0) return (tail ? 0.5 - 0.5 * (x / std::sqrt(2.0 + x2)) : 0.5 + 0.5 * (x / std::sqrt(2.0 + x2)));
+    // tail xor (x < 0.0)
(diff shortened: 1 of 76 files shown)


lib/src/Uncertainty/Distribution/StudentFunctions.cxx
```diff
@@ -49,6 +49,9 @@ Scalar StudentCDF(const Scalar nu,
 {
   if (!(nu > 0.0)) throw InvalidArgumentException(HERE) << "Error: nu must be positive, here nu=" << nu;
   if (x == 0.0) return 0.5;
+#ifdef OPENTURNS_HAVE_BOOST
+  return (tail ? boost::math::cdf(boost::math::complement(boost::math::students_t(nu), x)) : boost::math::cdf(boost::math::students_t(nu), x)) ;
+#else
```
regislebrun, line 54: If Boost is not available, we still have the bug. The problem is that the formulas are correct but should be rewritten for large |x| by dividing both the numerator and the denominator by their leading terms $|x|^{\nu/2}$. Be careful, the sign of x will show up in the numerator, but as the numerator is an odd polynomial function, you just have to select the same formula when (tail != (x<0.0)). For example, when $\nu=2$, we should have:
```
if (nu == 2.0)
{
  if (std::abs(x) < 1.0) return (tail ? 0.5 - 0.5 * (x / std::sqrt(2.0 + x2)) : 0.5 + 0.5 * (x / std::sqrt(2.0 + x2)));
  // tail xor (x < 0.0)
  return (tail != (x < 0.0) ? 0.5 - 0.5 / std::sqrt(2.0 / x2 + 1.0) : 0.5 + 0.5 / std::sqrt(2.0 / x2 + 1.0));
}
```

regislebrun (changes requested):
only a partial review

lib/src/Uncertainty/Distribution/StudentFunctions.cxx
jschueller: thanks for the hint, I rewrote all the formulas

regislebrun (approved):

PR #2528, opened 2024-01-29
Modification of input distribution for firesatellite use case

### LINKED ISSUE #2527: Strange probabilistic model for the fire satellite use-case
### What happened?

The probabilistic model of the fire satellite use-case (https://openturns.github.io/openturns/latest/usecases/use_case_fireSatellite.html) is made of independent normal distributions. It allows to produce negative values for quantities which are physically constrained to be positive. Many of these normal distributions have a standard deviation much lower than their mean value, resulting into a numerical truncation to positive values, but others like the reflectance factor or the drag coefficient can take negative values with significant probabilities.

```shell
Read the use-case description
```

### Version

1.23dev

### Operating System

Linux

### Installation media

from source

### Additional Context

_No response_

m-balesdent:
Hello,

Closes  #2527

This PR proposed a modification of the probability distribution of input parameters of the firesatellite use case, to avoid non realistic samples (negative for positive quantities). 

Truncated  Normal distribution is used instead of Normal distribution.

### DIFF
--- a/python/doc/usecases/use_case_fireSatellite.rst
+++ b/python/doc/usecases/use_case_fireSatellite.rst
@@ -22,7 +22,7 @@ The multidisciplinary analysis that is used to compute the converged values of t
 
 
 This test case is composed of nine random variables.
-These variables are distributed according to truncated Normal distribution :math:`\mathcal{TN}(\mu,\sigma,a,b).
+These variables are distributed according to truncated Normal distribution :math:`\mathcal{TN}(\mu,\sigma,a,b)`.
 The hyperparameters are :math:`\mu` the mean, :math:`\sigma` the standard deviation, :math:`a` the lower bound and :math:`b` the upper bound.
 
 - :math:`H \sim \mathcal{TN}(18e6, 1e6, 15e6, 21e6)`: the altitude (m)
--- a/python/doc/usecases/use_case_fireSatellite.rst
+++ b/python/doc/usecases/use_case_fireSatellite.rst
@@ -21,7 +21,8 @@ The multidisciplinary analysis that is used to compute the converged values of t
     :width: 50%
 
 
-This test case is composed of nine random variables distributed according to truncated Normal distributions :math:`\mathcal{TN}(\mu,\sigma,a,b)` with :math:`\mu` the mean, :math:`\sigma` the standard deviation, :math:`a` the lower bound and :math:`b` the upper bound:
+This test case is composed of nine random variables.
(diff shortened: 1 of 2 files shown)


jschueller (commented):
its not always necessary to truncate when sigma is small wrt to the mean
(when the numerical range does not contain negative values)
but it doesnt hurt either

note that the body of the commit should contain "Closes #xxx" (instead of the commit title)
+1

python/doc/usecases/use_case_fireSatellite.rst
```diff
@@ -21,25 +21,25 @@ The multidisciplinary analysis that is used to compute the converged values of t
     :width: 50%
 
 
-This test case is composed of nine random variables:
+This test case is composed of nine random variables distributed according to truncated Normal distributions :math:`\mathcal{TN}(\mu,\sigma,a,b)` with :math:`\mu` the mean, :math:`\sigma` the standard deviation, :math:`a` the lower bound and :math:`b` the upper bound:
```
jschueller, line 24: skip lines, its a bit long

m-balesdent:
> its not always necessary to truncate when sigma is small wrt to the mean (when the numerical range does not contain negative values) but it doesnt hurt either
> 
> note that the body of the commit msg should contain "Closes #xxx" (instead of the commit title) +1

The commit msg has been changed.

Concerning the truncated distribution, it could be more consistent to keep the same distribution for all the variables.

josephmure (approved):

python/src/usecases/fireSatellite_function.py
```diff
@@ -156,31 +156,31 @@ def __init__(self):
...
-        self.RD = ot.Normal(5.0, 1.0)
+        self.RD = ot.TruncatedNormal(5.0, 1.0, 5.0 - 3 * 1.0, 5.0 + 3 * 1.0)
 
         # Moment arm for aerodynamic torque
-        self.Lalpha = ot.Normal(2.0, 0.4)
+        self.Lalpha = ot.TruncatedNormal(2.0, 0.4, 2.0 - 3 * 0.4, 2.0 + 3 * 0.4)
 
         # Drag coefficient
-        self.Cd = ot.Normal(1.0, 0.3)
+        self.Cd = ot.TruncatedNormal(1.0, 0.3, 1.0 - 3 * 0.3, 1.0 + 3 * 0.3)
```
regislebrun, line 183: positive, but it was close ;-)

regislebrun (approved):

PR #2529, opened 2024-01-29
Update importance_form.rst

mbaudin47:
Define what exactly is SO(n), which is not so well known in linear algebra.

### DIFF
--- a/python/doc/theory/reliability_sensitivity/importance_form.rst
+++ b/python/doc/theory/reliability_sensitivity/importance_form.rst
@@ -23,6 +23,8 @@ input components with respect to the realization of the event. They are often se
 of the impact of modeling the
 input components as random variables rather than fixed values. FORM importance factors are defined as follows.
 
+Let :math:`\mat{R}\in{\cS\cO}_n(\Rset)` be the special orthogonal group
+of square orthogonal matrices of dimension :math:`n` with determinant 1.
 The :ref:`isoprobabilistic transformation <isoprobabilistic_transformations>` :math:`T` used in the FORM and
 SORM approximation is a diffeomorphism from
 :math:`\supp{\vect{X}}` into :math:`\Rset^n`, such that the


python/doc/theory/reliability_sensitivity/importance_form.rst
```diff
@@ -23,6 +23,8 @@ input components with respect to the realization of the event. They are often se
 of the impact of modeling the
 input components as random variables rather than fixed values. FORM importance factors are defined as follows.
 
+Let :math:`\mat{R}\in{\cS\cO}_n(\Rset)` be the special orthogonal group
+of square orthogonal matrices of dimension :math:`n` with determinant 1.
```
adutfoy, line 27: The sentence is redundant. You should name the space and precise its definition after: 

Let :math:`\mat{R}\in{\cS\cO}_n(\Rset)` be the special orthogonal group which is the set of square matrices :math:`\mat{M} in \cM_n(\Rset)`  such that :math:`\mat{M} \Tr{\mat{M}} = I_n` and with dertminant equal to 1.

mbaudin47, line 27:
```suggestion
Let :math:`\mat{R} \in {\cS\cO}_n(\Rset)` be the special orthogonal group which is the set of square matrices
:math:`\mat{M} \in \cM_n(\Rset)` such that :math:`\mat{M} \Tr{\mat{M}} = \mat{I}_n` and with
determinant equal to 1.
```

mbaudin47: Fixed at 267b7b8f5ec2dd5cd636181c1969bad3d85f5c94. Please tell me if that suits your requirement.

adutfoy: ok!

adutfoy (approved):

josephmure (approved):

Issue #2557, opened 2024-02-27
Shipping openturns and poissoninv GPL license


mwestphal:
### What is the idea?

Reading the license of OpenTURNS and the GPL exception for its non optional component poissoninv, it looks like shipping openturns in another software would ship poinssoninv under the GPL license hence contaminating openturns and any software linked with it.

In my case, I'm interested to ship OpenTURNS in ParaView binary release.

I see two possible solutions.

1. Ask Mike Giles for an exception to the GPL for not only openturns but for ParaView as well.
2. Find a way to disable poissoninv in OpenTURNS, if that even makes sense.

Please let me know what you think before I start reaching out to Mike or putting my hand into OpenTURNS code.

Is 2. something that you may be interested in ?

jschueller:
see COPYING.poissinv

mwestphal:
LGPL is fine, but I don't see anything related to using open turns API in mike emails. He talks about shipping poisoning in open turns binary and with open turns source file. I don't see a specific exception related to using poisininv through open turns API.

jschueller:
I thought that was implied

mwestphal:
So I've been reading carefully the different GPL exception in openturns:
 1. https://github.com/openturns/openturns/blob/master/COPYING.hmat
 2. https://github.com/openturns/openturns/blob/master/COPYING.KolmogorovSmirnovDist
 3. https://github.com/openturns/openturns/blob/master/COPYING.poissinv
 
 For 1., it is clear that the exception is "in the limited from of the developement of the Open Turns software, you are allowed to used the “Hmatrix library”, recently referenced in Open Source with the GPL v3 License [instead of using the GPLv2]." which does not constitute that much of an exception, because anyone shipping OpenTurns with hmat would end up with a LGPLv3 + GPLv2 software which is not compatible. That being said, hmat is optional, so I'm not impacted.
 
 For 2. Your email initially refer to an exception the accord project got to use the code as LGPL as part of their code. Looking at accord code I found this: https://github.com/Azure/Accord-NET/blob/master/Sources/Accord.Statistics/Testing/KolmogorovSmirnovTest.cs, where there is no exception, just a straight up (re) licensing of the code as LGPL.
 So regarding the exception, Richard states (translated):
 
 ```
 You can use our program for the Kolmogorov-Smirnov under the LGPL license
 ```
 
[...]

Issue #2552, opened 2024-02-21
ComputeQuantile for discrete variables


adutfoy:
### What happened?

We have a discrete variable, and we want to calculate its quantile of order p.
The two following methods should be identical: 
computeQuantile(p) and computeQuantile(1-p, True).
But the results obtained differ....

### How to reproduce the issue?

```shell
from openturns import *

n=16
alpha = 0.75
bin = Binomial(n, alpha)

proba = 0.95
q1 = bin.computeQuantile(proba)[0]
q2 = bin.computeQuantile(1-proba, True)[0]
print('q1, q2 = ', q1, q2)

# Test q1: 
print(bin.computeCDF(q1))
# 0.9899774042423815 > proba ==> ok!
# Test q2: 
print(bin.computeCDF(q2))
# 0.9365235602017492< proba ==> problem!
```

### Version

1.23dev

### Operating System

Linux

### Installation media

from source

jschueller:
so it's a doc-only problem ?

mbaudin47:
As surprising as it seems, I encountered exactly the same bug at the same time :). This is because we are working on the confidence intervals of a quantile, known as "Wilks's method". In this method, we want to compute indices $k_1$ and $k_2$ such that $\mathbb{P}(K \leq k_1)\leq \beta_1$ and $\mathbb{P}(K \geq k_2) \geq \beta_2$ where $K$ is a binomial distribution. This amounts to compute quantiles of the binomial distribution. But this only motivates the topic and is not relevant to solve the bug.

It is not a doc problem, but an issue in the implementation of `computeQuantile(False)`. Consider the following code, which compute the inverse of the complementary CDF function at point $\alpha = 0.9$:
```python
n = 3
p = 0.5
n_points = 1000
distribution = ot.Binomial(n, p)
x = distribution.computeQuantile(0.9, True)
```
prints:
```
x =  [-1]
```
This is a bug, since the quantile of a Binomial distribution is in the set $\set{0, ..., n}$. Hence, $x=-1$ cannot be a quantile.

To make this clearer, I compared OpenTURNS' [Binomial.html](http://openturns.github.io/openturns/latest/user_manual/_generated/openturns.Binomial.html) with SciPy's [binom.html](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binom.html).

```python
# %%
import openturns as ot
import openturns.viewer as otv
import matplotlib.pyplot as plt
import numpy as np
[...]

Issue #2541, opened 2024-02-07
The computeQuantile() method of a Mixture can fail


mbaudin47:
### What happened?

I created a `Mixture` and computed its quantile of level 0.5. 
- The script produces the value 1.5.
- The correct value is 1.

![image](https://github.com/openturns/openturns/assets/31351465/a922796b-a279-40a2-a349-14af08dadad6)

This is a trap: the distribution is specifically chosen so that its CDF is flat on the nonempty interval [1,2].

### How to reproduce the issue?

The script:

```python
distribution = ot.Mixture([ot.Uniform(0.0, 1.0), ot.Uniform(2.0, 3.0)], [0.5, 0.5])
alpha = 0.5
z = distribution.computeQuantile(alpha)[0]
print("z = ", z)
```

prints:
```
z = 1.5
```

### Version

1.21

### Operating System

Linux

### Installation media

conda

jschueller:
isnt any value in [1,2] correct ?

mbaudin47:
Here is the PDF:

![output](https://github.com/openturns/openturns/assets/31351465/a7af9271-346d-4a0d-9cd5-72fea4b57f97)

It is true, indeed, that the value x = 1.5 is so that f(0) = 0. But why should this be related to the definition of the quantile?

By definition, we have:

$$
x_\alpha = \textrm{inf} \set{x \in \mathbb{R}  |  \alpha \leq F(x)}
$$

Taking $x = 0.5$, we get:

$$
x_{0.5} = \textrm{inf} \set{x \in \mathbb{R}  | 0.5 \leq F(x)}
$$

Since $F(x) = 0.5$ for any $x in [1,2]$, it seems to me that the solution is:

$$
x_{0.5} = \textrm{inf} \set{[1,2]} = 1.
$$

jschueller:
indeed sometimes its not defined with the inf, the english wp page states thats its more a convention:
https://en.wikipedia.org/wiki/Quantile_function

but the french page has the inf:
https://fr.wikipedia.org/wiki/Fonction_quantile

mbaudin47:
You are right. This is done at:

https://github.com/openturns/openturns/blob/4fc2024d0cbccb92b4c5f70df8fd36598c903eb5/lib/src/Uncertainty/Model/DistributionImplementation.cxx#L2307

[...]

Issue #2525, opened 2024-01-25
Geometric range starts at 1


jschueller:
### What happened?

our Geometric distribution range starts at 1 but computeRange sets 0
its an easy fix but then the ChiSquared test fails because it wants to check if the sample has values outside the support of the distribution, see key "FittingTest-ChiSquaredCheckSample"

### How to reproduce the issue?

```shell
ot.Geometric(0.5).getRange()
```

### Version

master

### Operating System

unknown

### Installation media

unknown

PR #2564, opened 2024-03-06
Misc 5

jschueller:
- #2525 
- #2552: fix binomial quantile
- #2541:

Régis: "At this point you can add a 1st order correction. You solved $F(x_p)=p-\epsilon$, and you are looking for $\eta$ such that $F(x_p+\eta)=p$. But, up to the first order, $F(x_p+\eta)=F(x_p)+\eta F'(x_p)$, so $p=p-\epsilon+\eta F'(x_p)$, resulting into $\eta=\epsilon/F'(x_p)$. It works only if $F'(x_p)\neq 0$, which is the case for a continuous mixture but not for a discrete distribution or a mix of discrete and continuous distributions."
- replace poissinv code by boost.math (see #2557)
- split some keepIntact methods (#2218)
- fix EGO stopping criterion
- #2571
- fix inversewishart variance
- #2524 (& NormalGamma)


lib/src/Uncertainty/Distribution/BayesDistribution.cxx
```diff
@@ -403,29 +400,28 @@ void BayesDistribution::computeCovariance() const
...
   else integrator = GaussLegendre(Indices(2, static_cast<UnsignedInteger>(std::ceil(std::sqrt(1.0 * integrationNodesNumber_)))));
+
+  // The conditioning covariance can be reused as is
+  const UnsignedInteger conditioningDimension = conditioningDistribution_.getDimension();
+  const CovarianceMatrix conditioningCovariance(conditioningDistribution_.getCovariance());
+  for (UnsignedInteger i = 0; i < conditioningDimension; ++ i)
+    for (UnsignedInteger j = 0; j <= i; ++ j)
+      covariance_(i, j) = conditioningCovariance(i, j);
+
   // Integrate over the interval (-inf, y] of the conditioning Gamma distribution
```
regislebrun, line 416: Strange comment...

regislebrun (approved):
Good job. I checked only the C++ part, good to go for me.

lib/src/Uncertainty/Distribution/BayesDistribution.cxx
jschueller: I didnt get either

jschueller:
hi @sofianehaddad can we merge this one ?

sofianehaddad (approved):

PR #2573, opened 2024-03-19
Added processsample and field sum and difference operators

JPelamatti:
This PR implements sum and difference operators for Field and ProcessSample objects, as well as equality checks for testing meshes, fields and proccessamples


lib/src/Base/Stat/openturns/FieldImplementation.hxx
```diff
@@ -145,6 +145,18 @@ public:
...
+  FieldImplementation & operator += (const Sample & translation);
+  FieldImplementation & operator -= (const Sample & translation);
+  FieldImplementation & operator += (const Point & translation);
+  FieldImplementation & operator -= (const Point & translation);
+
+  /** Translate realizations */
+  FieldImplementation operator + (const Sample & translation);
+  FieldImplementation operator - (const Sample & translation);
+  FieldImplementation operator + (const Point & translation);
+  FieldImplementation operator - (const Point & translation);
```
jschueller, line 159: const

lib/src/Base/Stat/openturns/ProcessSampleImplementation.hxx
```diff
@@ -179,7 +179,19 @@ public:
...
+  ProcessSampleImplementation & operator += (const ProcessSampleImplementation & translation);
+  ProcessSampleImplementation & operator -= (const ProcessSampleImplementation & translation);
+
+  /** Translate realizations */
+  ProcessSampleImplementation operator + (const Sample & translation);
+  ProcessSampleImplementation operator - (const Sample & translation);
+  ProcessSampleImplementation operator + (const Point & translation);
+  ProcessSampleImplementation operator - (const Point & translation);
+  ProcessSampleImplementation operator + (const ProcessSampleImplementation & translation);
+  ProcessSampleImplementation operator - (const ProcessSampleImplementation & translation);
```
jschueller, line 193: const

jschueller (approved):

sofianehaddad (approved):

JPelamatti:
With the addition of python tests, this PR should be ready to be merged

python/test/CMakeLists.txt
```diff
@@ -237,6 +237,7 @@ ot_pyinstallcheck_test (NonStationaryCovarianceModelFactory_std)
 ot_pyinstallcheck_test (ProductCovarianceModel_std)
 ot_pyinstallcheck_test (StationaryFunctionalCovarianceModel_std IGNOREOUT)
 ot_pyinstallcheck_test (LatentVariableModel_std IGNOREOUT)
+ot_pyinstallcheck_test (ProcessAndFieldOperators_std)
```
jschueller, line 240: we should prefer assert-based tests over expout files
 (its ok to test only one component to simplify testing as much as possible)

regislebrun (approved):

python/test/CMakeLists.txt
JPelamatti: This choice was due to the impossibility of asserting equality between objects other than samples & points, and I chose  a more thorough testing over non-unitarty size & multi-dimensional fields.
I am open to simplify the tests to assert based ones if necessary

mbaudin47: Perhaps you could introduce a new assert_almost_equal_field() function? It would check that a computed Field is almost equal to a reference Field. This would be the building block of your test (and perhaps other unit tests as well). You would need to implement a new assert_almost_equal_mesh(). These codes would locate into ottest. They could be based on assert on Points and Sample, which already exist.

JPelamatti: I added assert operators for meshes, fields and processsamples for future testing

python/test/CMakeLists.txt
```diff
@@ -237,6 +237,7 @@ ot_pyinstallcheck_test (NonStationaryCovarianceModelFactory_std)
 ot_pyinstallcheck_test (ProductCovarianceModel_std)
 ot_pyinstallcheck_test (StationaryFunctionalCovarianceModel_std IGNOREOUT)
 ot_pyinstallcheck_test (LatentVariableModel_std IGNOREOUT)
+ot_pyinstallcheck_test (ProcessAndFieldOperators_std IGNOREOUT)
```
jschueller, line 240: should be named "ProcessSample_Field_operators" (without std suffix)
or even better could be split accross 2 files: ProcessSample_operators & Field_operators

JPelamatti:
Please note that in the new assert_equal functions for fields and processsamples, the same tolerances are used to check the field values as well as the vertices values.

python/test/CMakeLists.txt
jschueller: thanks a lot

PR #2594, opened 2024-03-29
Correct PosteriorDistribution equation

josephmure:


jschueller (approved):

adutfoy:
Ok. Just a minor correction.
You wrote "ie the marginal distribution of Y". I propose : "ie the distribution of Y".

PR #2597, opened 2024-03-31
New inversion methods for a Matrix

mbaudin47:
- The new `SquareMatrix.inverse()` and `SymmetricMatrix.inverse()` methods are created.

Example:
```python
sqMatrix = ot.SquareMatrix([[1.0, 2.0, 3.0], [3.0, 2.0, 1.0], [2.0, 1.0, 3.0]])
sqInverse = sqMatrix.inverse()
```

This is a part of PR #2330.

### DIFF
--- a/lib/src/Base/Type/MatrixImplementation.cxx
+++ b/lib/src/Base/Type/MatrixImplementation.cxx
@@ -32,7 +32,6 @@
 #include "openturns/Os.hxx"
 #include "openturns/Exception.hxx"
 #include "openturns/Sample.hxx"
-#include "openturns/IdentityMatrix.hxx"
 
 BEGIN_NAMESPACE_OPENTURNS
 
--- a/lib/src/Base/Type/MatrixImplementation.cxx
+++ b/lib/src/Base/Type/MatrixImplementation.cxx
@@ -896,37 +896,27 @@ MatrixImplementation MatrixImplementation::solveLinearSystemSquare(const MatrixI
 }
 
 /* Square matrix inverse */
-MatrixImplementation MatrixImplementation::inverseSquareInPlace()
+MatrixImplementation MatrixImplementation::inverseSquare() const
 {
   if (nbColumns_ != nbRows_ ) throw InvalidDimensionException(HERE) << "The matrix has " << nbRows_ << " and " << nbColumns_ << ", expected a square matrix.";
   MatrixImplementation identity(nbRows_, nbColumns_);
   for(UnsignedInteger i = 0; i < nbRows_; ++i)
     identity(i, i) = 1.0;
-  const MatrixImplementation inverseMatrix(solveLinearSystemSquareInPlace(identity));
-  return inverseMatrix;
-}
-
-MatrixImplementation MatrixImplementation::inverseSquare() const
-{
   MatrixImplementation A(*this);
-  return A.inverseSquareInPlace();
+  const MatrixImplementation inverseMatrix(A.solveLinearSystemSquareInPlace(identity));
+  return inverseMatrix;
 }
 
 /* Symmetric matrix inverse */
-MatrixImplementation MatrixImplementation::inverseSymInPlace()
+MatrixImplementation MatrixImplementation::inverseSym() const
 {
   if (nbColumns_ != nbRows_ ) throw InvalidDimensionException(HERE) << "The matrix has " << nbRows_ << " and " << nbColumns_ << ", expected a square matrix.";
   MatrixImplementation identity(nbRows_, nbColumns_);
   for(UnsignedInteger i = 0; i < nbRows_; ++i)
     identity(i, i) = 1.0;
-  const MatrixImplementation inverseMatrix(solveLinearSystemSymInPlace(identity));
-  return inverseMatrix;
-}
-
-MatrixImplementation MatrixImplementation::inverseSym() const
-{
   MatrixImplementation A(*this);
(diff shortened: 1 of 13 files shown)


lib/src/Base/Type/MatrixImplementation.cxx
```diff
@@ -895,6 +896,40 @@ MatrixImplementation MatrixImplementation::solveLinearSystemSquare(const MatrixI
...
 
+/* Square matrix inverse */
+MatrixImplementation MatrixImplementation::inverseSquareInPlace()
+{
+  if (nbColumns_ != nbRows_ ) throw InvalidDimensionException(HERE) << "The matrix has " << nbRows_ << " and " << nbColumns_ << ", expected a square matrix.";
+  MatrixImplementation identity(nbRows_, nbColumns_);
+  for(UnsignedInteger i = 0; i < nbRows_; ++i)
+    identity(i, i) = 1.0;
+  const MatrixImplementation inverseMatrix(solveLinearSystemSquareInPlace(identity));
+  return inverseMatrix;
```
mbaudin47, line 907: Comment from @regislebrun : "The method is supposed to work in-place, no?"

lib/src/Base/Type/MatrixImplementation.cxx
```diff
@@ -895,6 +896,40 @@ MatrixImplementation MatrixImplementation::solveLinearSystemSquare(const MatrixI
...
+}
+
+/* Symmetric matrix inverse */
+MatrixImplementation MatrixImplementation::inverseSymInPlace()
+{
+  if (nbColumns_ != nbRows_ ) throw InvalidDimensionException(HERE) << "The matrix has " << nbRows_ << " and " << nbColumns_ << ", expected a square matrix.";
+  MatrixImplementation identity(nbRows_, nbColumns_);
+  for(UnsignedInteger i = 0; i < nbRows_; ++i)
+    identity(i, i) = 1.0;
+  const MatrixImplementation inverseMatrix(solveLinearSystemSymInPlace(identity));
```
mbaudin47, line 923: Another comment from @regislebrun : "same here"

lib/src/Base/Type/MatrixImplementation.cxx
```diff
@@ -32,6 +32,7 @@
 #include "openturns/Os.hxx"
 #include "openturns/Exception.hxx"
 #include "openturns/Sample.hxx"
+#include "openturns/IdentityMatrix.hxx"
```
regislebrun, line 35: Why do you include this file? It is not used anywhere

mbaudin47: This is true. Fixed at 9c332e51a

lib/src/Base/Type/MatrixImplementation.cxx
mbaudin47: Fixed at 149e87072

jschueller:
Maybe we dont want to do that: we actually rarely need to compute the inverse of a matrix but only solve the linear system

mbaudin47:
In most cases, you are right. But WE sometimes really need the inverse, for example when we want to compute the hat matrix of a linear least squares problem. The leverages are on the diagonal.

mbaudin47:
@jschueller , @regislebrun : Can you please review this PR?

KieranDelamotte:
@mbaudin47 
Maybe the pseudo-inverse can help ?

mbaudin47:
> @mbaudin47 
> Maybe the pseudo-inverse can help ?

The QR or SVD decompositions can help computing the projection matrix of a linear least squares model: i derived the equations and tested that yesterday. But this cannot be done at the matrix level. It can be implemented in the LinearModel class, which can access the decomposition. Whatever the implementation, this is not a topic of the Matrix class (this PR) or the cross validation (the target PR). The implementation can be improved afterwards once the interface is defined: there is currently no LinearModelResult.getHatMatrix() available.

lib/src/Base/Type/MatrixImplementation.cxx
```diff
@@ -895,6 +895,30 @@ MatrixImplementation MatrixImplementation::solveLinearSystemSquare(const MatrixI
...
+/* Square matrix inverse */
+MatrixImplementation MatrixImplementation::inverseSquare() const
+{
+  if (nbColumns_ != nbRows_ ) throw InvalidDimensionException(HERE) << "The matrix has " << nbRows_ << " and " << nbColumns_ << ", expected a square matrix.";
+  MatrixImplementation identity(nbRows_, nbColumns_);
+  for(UnsignedInteger i = 0; i < nbRows_; ++i)
+    identity(i, i) = 1.0;
+  MatrixImplementation A(*this);
+  const MatrixImplementation inverseMatrix(A.solveLinearSystemSquareInPlace(identity));
+  return inverseMatrix;
```
mbaudin47, line 907: Comment from @regislebrun :

The method is supposed to work in-place, no?

mbaudin47: I removed the method `MatrixImplementation::inverseSquareInPlace()`.

mbaudin47:
@jschueller , @regislebrun : Can you please review this PR?

jschueller:
I still dont think we should add inverse methods

jschueller (approved):

l-brevault (approved):

regislebrun (approved):

Issue #2366, opened 2023-08-12
The maths notations in the help pages are inconsistent


mbaudin47:
The mathematical notations in the help pages are inconsistent, which makes the understanding of the methods and API more difficult than necessary. 
- Consider for example the creation of a surrogate model: two candidates in the library are PCE and kriging. But the [KrigingResult](https://openturns.github.io/openturns/latest/user_manual/response_surface/_generated/openturns.KrigingResult.html) name it $\tilde{\mathcal{M}}$ while the [PCE](https://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_exploitation.html#sphx-glr-auto-meta-modeling-polynomial-chaos-metamodel-plot-functional-chaos-exploitation-py) names it $\hat{g}$. 
- The functions in the basis are name $\Psi_k$ in the [PCE](https://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_exploitation.html#sphx-glr-auto-meta-modeling-polynomial-chaos-metamodel-plot-functional-chaos-exploitation-py) and $\varphi_\ell$ in the [kriging](https://openturns.github.io/openturns/latest/user_manual/response_surface/_generated/openturns.KrigingResult.html).
- The physical model is named $f$ in [Function](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.Function.html) and $g$ in the [PCE](https://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_exploitation.html#sphx-glr-auto-meta-modeling-polynomial-chaos-metamodel-plot-functional-chaos-exploitation-py).
- The output dimension of the physical model is $d'$ in the [Function](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.Function.html) but $p$ in [KrigingResult](https://openturns.github.io/openturns/latest/user_manual/response_surface/_generated/openturns.KrigingResult.html).

Most of these changes in the maths notations are unnecessary. 

One of the methods I plan to use is to create a list of macros in math_notations.sty and use them consistently. I already began with \model, and I plan to continue. This will fix many bugs with a minimum effort. Here is a list of suggestions :
- \physicalmodel;
- \metamodel;
- \probabilitydensityfunction;
- \inputdimension;
- \inputvector;
- \outputdimension;
- \ouputvector;
- \basisdimension;
[...]

PR #2574, opened 2024-03-19
Hackaton 2024: resolved some doc issues

adutfoy:
I modified the following files: 

- Theory / Probabilistic modelling / Copulas : Closes Issues #2558
- Theory / Probabilistic modelling / Parametric models
- Theory / Meta modelling / Taylor expansion: pb in titles
- Theory / Reliability and sensitivity:  Closes  #2332: The doc of Sobol' indices has issues doc

I also introduced (with the help of Michaël! :-) ) math_notations within all the files .rst of the directories:

-  "Reliability Sensivity"
- "Probabilistic Modelling'
- "Data analysis/ Fitting tests": I largely modified the text
- "Data analysis/ Comparison of two samples": I modified the text

PR #2598, opened 2024-04-01
Improve doc of polynomial chaos expansion

mbaudin47:
This PR improves some help pages.

I introduced a set of macro in `math_notations.sty` to solve #2366. The goal is to consistently use them across the help pages. This is because I notice that the `FunctionalChaosResult` and `FunctionalChaosAlgorithm` help pages do not use the same notations consistently. Indeed, we should use consistent notations not only within a single page, but also across pages on the same topic. Moreover, because some help pages are inherited from mother classes, several help pages must be updated accordingly. For example, `FunctionalChaosResult` inherits from `MetaModelResult`. Hence, both pages must be updated accordingly. Otherwise, the sample size, for example, is denoted $n$ in some pages, but $N$ in the inherited page. The only way to consistently use the same notation is to introduce a macro and use it everywhere it is needed.

This is the doc part of the PR #2372 which is unrelated to other developments.

### DIFF
--- a/python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_sobol.py
+++ b/python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_sobol.py
@@ -4,38 +4,50 @@ Estimate Sobol' indices for the Ishigami function by a sampling method: a quick 
 """
 # %%
 #
-# In this example, we estimate the Sobol' indices for the :ref:`Ishigami function <use-case-ishigami>` by sampling methods.
+# In this example, we estimate the Sobol' indices for the
+# :ref:`Ishigami function <use-case-ishigami>` by sampling methods.
 #
 
 # %%
 # Introduction
 # ------------
 #
-# In this example we are going to quantify the correlation between the input variables and the output variable of a model thanks to Sobol indices.
+# In this example we are going to quantify the correlation between the input
+# variables and the output variable of a model thanks to Sobol indices.
 #
-# Sobol indices are designed to evaluate the importance of a single variable or a specific set of variables.
-# Here the Sobol indices are estimated by sampling from the distributions of the input variables and propagating uncertainty through a function.
+# Sobol indices are designed to evaluate the importance of a single variable
+# or a specific set of variables.
+# Here the Sobol indices are estimated by sampling from the distributions of
+# the input variables and propagating uncertainty through a function.
 #
-# In theory, Sobol indices range from 0 to 1; the closer an index value is to 1, the better the associated input variable explains the function output.
+# In theory, Sobol indices range from 0 to 1; the closer an index value is
+# to 1, the better the associated input variable explains the function output.
 #
 # Let us denote by :math:`d` the input dimension of the model.
 #
 # Sobol' indices can be computed at different orders.
 #
-# * First order indices evaluate the importance of one input variable at a time.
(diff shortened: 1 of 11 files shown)


mbaudin47:
@josephmure  : This PR seems to be ready from my point of view. Can you please review it?

python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_sobol.py
```diff
@@ -147,11 +190,15 @@ def plotXvsY(sampleX, sampleY, figsize=(15, 3)):
...
 #   alone produce almost 30% (0.6 - 0.3) of the total variance.
 # - The variable :math:`X_2` has the highest first order index: approximately 0.4.
-#   However, it has little interaction with other variables since its total order indice is close to its first order index.
+#   However, it has little interaction with other variables since its total
+#   order index is close to its first order index.
 # - The variable :math:`X_3` has a first order index close to zero.
-#   However, it has an impact to the total variance thanks to its interactions with :math:`X_1`.
-# - We see that the variability of the estimates is quite high even with a relatively large sample size.
-#   Moreover, since the exact first order Sobol' index for :math:`X_3` is zero, its estimate has a 50% chance of being nonpositive.
+#   However, it has an impact to the total variance thanks to its interactions
```
josephmure, line 196:
```suggestion
#   However, it has an impact on the total variance thanks to its interactions
```

josephmure (approved):
Good to go after a small fix!

mbaudin47:
@adutfoy : Thank you for the review. I will take this into account ASAP.

>     * Your notation seems a little strange... [...]

Actually, there is a limitations in the current notations in OT's doc. I personnally use the notations presented in [Notations.md](https://github.com/mbaudin47/otsupgalilee-eleve/blob/master/0-Deroulement/Notations.md).

- You suggest to use${\boldsymbol{x}^j∈R^{n_X},j=1…n}$ rather than ${\boldsymbol{x}^j∈R^{n_X}}_{j=1…n}$. Ok: I will do that.
- You write "_The notation_ $\boldsymbol{x}^{(i)}$ _is confusing: there is no order statistics here! the () are unnecessary IMO._" I do not agree because 1) we cannot write $\boldsymbol{x}^{i}$ since this would mean to exponentiate a vector, which has no meaning. 2) You imply that there are order statistics involved, but an order statistics would write $x_{(i)}$ so that there are two differences here: a) this is an index, and not an exponent. b) there is no particular meaning of sorting a vector sample because there is no obvious order in dimension greater than 2.

All in all, I think that this notation is appropriate.

mbaudin47:
>     * You should quote the  exact condition of the closure when the basis is polynomial:  the distribution μZ must be fully characterized by all its moments. (unisolvante in french...)

Do you have a specific bibliographic reference that can be cited on this topic?

mbaudin47:
>     * You should also add that if the distribution has only a finite number of finite moments, then the orthonormal polynomial family has a finite number of members. For example: the Cauchy distribution has a first order moment (expectation) which is infinite: the orthonormal polynomial family is reduced to constant polynomials.

Do you have a specific bibliographic reference that can be cited on this topic?

adutfoy (approved):

mbaudin47:
@jschueller : I think that this PR is good for merging.

jschueller:
@mbaudin47 can you rebase to make sure it builds with #2574 merged ?

PR #2606, opened 2024-04-05
Sum covariance model

JPelamatti:
This is a draft PR to initiate the discussion concerning how we could create a sum of covariance models in the actual set-up


sofianehaddad:
Question : do we target a basic sum or a weighted sum ?

JPelamatti:
> Question : do we target a basic sum or a weighted sum ?

In the proposed implementation, we would handle the individual amplitudes of all covariance models, it is therefore equivalent to a weighted sum

lib/src/Base/Stat/SumCovarianceModel.cxx
```diff
@@ -0,0 +1,508 @@
...
+      const Description localFullParameterDescription(collection_[i].getFullParameterDescription());
+      for (UnsignedInteger k = 0; k < extraParameterNumber_[i]; ++ k)
+        description.add(OSS() << localFullParameterDescription[collection_[i].getScale().getSize() + 2 + k] << "_" << i);
+    }
+  }
+
+  return description;
+}
+
+void SumCovarianceModel::setScale(const Point & scale)
```
mbaudin47, line 429: Isn't `getScale()` to be implemented?

mbaudin47:
- Please provide a review of software implementing arithmetic of covariance models: GPytorch, GPFlow, Gpy, KerGP.
- A generic class which provides services related to a collection of covariance models may help to manage this class and the existing TensorizedCovarianceModel and ProductCovarianceModel classes. This class may implement methods based on dictionnaries.

Issue #2602, opened 2024-04-03
The pretty-print of a ParametricFunction does not show the name of the parameters


mbaudin47:
### What happened?

The pretty-print of a ParametricFunction does not show the name of the parameters.

The code below prints a parametric function where *x* and *y* are parameters. The pretty-print, attached to the variable *g*, prints:
- Input dimension = 1
- Input description = [z]
- Output dimension = 2
- Output description = [S,H]
- Parameter = class=Point name=Unnamed dimension=2 values=[2,3]

Notice that the parameters have the names *x* and *y* cannot be seen here: only the positions are shown.

But when we use `print(g)`, we see the parameters:
```
ParametricEvaluation([x,y,z]->[x + y,x * z + y], parameters positions=[0,1], parameters=[x : 2, y : 3], input positions=[2])
```

So the names *x* and *y* of the parameters:
- cannot be seen with the pretty-print,
- can be seen with the raw `print`.

This is a bug from my point of view, because the purpose of the pretty-print is to show the relevant information.

### How to reproduce the issue?

```shell
# %%
import openturns as ot

# %%
f = ot.SymbolicFunction(["x", "y", "z"], ["x + y", "x * z + y"])
f.setOutputDescription(["S", "H"])
print(f)

# %%
g = ot.ParametricFunction(f, [0, 1], [2.0, 3.0])
g

# %%
print(g)
```

### Version

1.22

### Operating System

unknown

### Installation media

unknown

Issue #2596, opened 2024-03-30
The legends of the drawPDF() graph have a wrong order


mbaudin47:
### What happened?

The [plot_estimate_non_parametric_copula.html](http://openturns.github.io/openturns/latest/auto_data_analysis/estimate_dependency_and_copulas/plot_estimate_non_parametric_copula.html) plots the contours of a 2 dimensions distribution. The `drawPDF()` method plots:

![image](https://github.com/openturns/openturns/assets/31351465/41faf26e-22f5-4d79-b552-54116f9210eb)

We see that the order of the legends in the plot do not correspond to the order of the values: the values should be sorted in increasing order, but the legends are sorted so that the lowest value is the last, instead of being the first. In other words, the order of the legends is correct, except for the first one.

### How to reproduce the issue?

```python
import openturns as ot
import openturns.viewer as viewer
from matplotlib import pylab as plt

R = ot.CorrelationMatrix(2)
R[1, 0] = 0.4
copula = ot.NormalCopula(R)
sample = copula.getSample(30)
distribution = ot.BernsteinCopulaFactory().build(sample)
graph = distribution.drawPDF()
view = viewer.View(graph)
```

### Version

1.22

### Operating System

unknown

### Installation media

unknown

### Additional Context

@ariste-sarl : Did you notice that?
_No response_

jschueller:
I dont think so

mbaudin47:
Can you elaborate on this?

jschueller:
its about the order of the legend, not the z-axis order

Issue #2545, opened 2024-02-09
Formatting Issues on LatentVariableModel


m-balesdent:
### What happened?

Hello,

The documentation page of [LatentVariableModel ](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.experimental.LatentVariableModel.html#openturns.experimental.LatentVariableModel)presents several formatting issues.

Some links are not active : 

![image](https://github.com/openturns/openturns/assets/26813843/2af7c839-16c0-454f-a3dc-d8231cf2c1c2)

Documentation is missing for `getLevelNumber` method : 

![image](https://github.com/openturns/openturns/assets/26813843/b1b99e55-6e63-4306-abc2-679e7b26de94)

The note of computeCrossCovariance is not consistent : 

![image](https://github.com/openturns/openturns/assets/26813843/20059fdc-dd5e-4f1c-95d0-777090dd52b0)

### How to reproduce the issue?

```shell
see https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.experimental.LatentVariableModel.html#openturns.experimental.LatentVariableModel
```

### Version

doc

### Operating System

all

### Installation media

unknown

Issue #2517, opened 2024-01-16
Cannot build OpenTURNS with doc if optional dependencies bison/flex are unavailable


josephmure:
### What happened?

When compiling OpenTURNS on a machine which does not have all optional dependencies, I am unable to build the doc:

```python
Warning, treated as error:
/home/GIT/openturns/build/python/src/sphinx_build/examples/data_analysis/distribution_fitting/plot_estimate_gev_venice.py failed to execute correctly: Traceback (most recent call last):
  File "/home/GIT/openturns/build/python/src/sphinx_build/examples/data_analysis/distribution_fitting/plot_estimate_gev_venice.py", line 28, in <module>
    data = coles.Coles().venice
           ^^^^^^^^^^^^^
  File "/home/GIT/openturns/build/python/src/site-packages/openturns/usecases/coles.py", line 53, in __init__
    dataset = ot.Sample.ImportFromCSVFile(fn, ",")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/GIT/openturns/build/python/src/site-packages/openturns/typ.py", line 7840, in ImportFromCSVFile
    return _typ.Sample_ImportFromCSVFile(*args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: NotYetImplementedException : OpenTURNS has been compiled without bison/flex support

make[2]: *** [python/src/CMakeFiles/sphinx_html.dir/build.make:781 : python/src/CMakeFiles/sphinx_html] Erreur 2
make[1]: *** [CMakeFiles/Makefile2:19626 : python/src/CMakeFiles/sphinx_html.dir/all] Erreur 2
make: *** [Makefile:166 : all] Erreur 2
```

Note that this occurs even when I pass the ` --compile-no-warning-as-error` option to CMAKE.

### How to reproduce the issue?

```shell
Build OpenTURNS with -DUSE_SPHINX=ON and with bison/flex missing.
```

### Version

master at 61f0de3491571a933207c252b006a4ff23ea046b

### Operating System

Linux

### Installation media

from source

PR #2608, opened 2024-04-10
Misc 6

jschueller:
- fix Pagmo MINLP input variables reordering
- add Function.setStopCallback method: the use-case is the interruption of Salome/YACS async eval API in Persalys
- Fix viewer z-order (again), else points appear behind grid
- #2596 
- Use kolmogorov distribution from scipy https://github.com/openturns/openturns/issues/2557: translate some code from Python and also include some of the needed C code
- Disable HMat by default (because of license)
- #2621
- #2602
- #2545 
- #2627 
- #2628 
- #2517
- Drop Tensor/Matrix.clean method call from str methods & tests
- Add nanoflann (from #2566)
- Add Matrix.frobeniusNorm (from #2566)


sofianehaddad:
Maybe specific `kolmogorov` files (including `dd_*`) might be in a specific folder ? (

jschueller:
> Maybe specific `kolmogorov` files (including `dd_*`) might be in a specific folder ? (

good idea, lets move it into /cephes

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -173,25 +174,37 @@ TruncatedDistribution * TruncatedDistribution::clone() const
...
   Distribution localDistribution(distribution_);
   String kind(localDistribution.getImplementation()->getClassName());
   // Delve into the antecedents until we get something which is not a truncated distribution
+  UnsignedInteger level = 1;
   while (kind == "TruncatedDistribution")
   {
     const TruncatedDistribution * truncatedDistribution = dynamic_cast<const TruncatedDistribution *>(localDistribution.getImplementation().get());
     localDistribution = truncatedDistribution->getDistribution();
     kind = localDistribution.getImplementation()->getClassName();
+    ++ level;
```
regislebrun, line 200: Maybe you could add a comment on the fact that the nested truncation intervals have already been intersected during the nested range computations.

lib/src/Uncertainty/Distribution/cephes/dd_real.h
```diff
@@ -0,0 +1,143 @@
+/*
```
regislebrun, line 1: So with this, we embed an extended precision library in addition to the optional mpfr/mpc multiprecision libraries used in SpecFunc. At one point we should coordinate this, in particular we could make it more visible troughout the whole library eg. to implement roundoff sensitive algorithms (orthonormalization, dot products...)

regislebrun (commented):
Only minor remarks. Approved if it does not change by 100+ lines of codes since this review.

lib/src/Uncertainty/Distribution/cephes/dd_real.h
jschueller: yes, I did not want to rewrite that part

lib/src/Base/Func/SpecFunc/openturns/SpecFunc.hxx
```diff
@@ -31,6 +31,7 @@ BEGIN_NAMESPACE_OPENTURNS
 namespace SpecFunc
 {
 
+#ifndef SWIG
```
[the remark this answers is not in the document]
jschueller: ok I can drop it but imho they are not necessary

PR #2615, opened 2024-04-18
Firesat test case correction

l-brevault:
Hello,

This is a modification of the firesat test case due to an incompatibility between the documentation and the code.

Sincerely,

### DIFF
--- a/python/src/usecases/fireSatellite_function.py
+++ b/python/src/usecases/fireSatellite_function.py
@@ -351,7 +351,7 @@ class FireSatelliteModel:
         Ptot = P_ACS + P_other
 
         # power production capability at beginning of life
-        P_BOL = eta * Fs * Id * m.cos(theta)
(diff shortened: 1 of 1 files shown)


python/src/usecases/fireSatellite_function.py
```diff
@@ -351,7 +351,7 @@ def power(self, inputs):
         Ptot = P_ACS + P_other
 
         # power production capability at beginning of life
-        P_BOL = eta * Fs * Id * m.cos(theta)
+        P_BOL = eta * Fs * Id * m.cos(self.i)
```
jschueller, line 354: theta is not used anymore and linter fails

l-brevault: Yes, the modification has been made to remove theta

sofianehaddad (approved):

PR #2654, opened 2024-05-12
Improve low discrepancy sequences

### LINKED ISSUE #2653: The Faure sequence is wrong
### What happened?

The Faure sequence generates wrong points in any dimension.

Here are the proofs of this.
- In dimension 2, we see that a set of 4 points does not correctly fill the 4 base-2 elementary volumes.
- In dimension 3, we see that 9 first points generated are wrong. The basis is obviously equal to 2 instead of 3. The third dimension is a copy of the first dimension.
- The estimated value of the mean of the GSobol' function does not converge to its true value.

Many thanks to Félix Husson (EDF & Institut Galilée) for pointing that bug.

One of the reasons of the bug is undetected is the way the unit test [t_FaureSequence_std.cxx](https://github.com/openturns/openturns/blob/master/lib/test/t_FaureSequence_std.cxx) is implemented. Since the algorithm produces a set of points, the simplest and most efficient way to implement the unit test is to check the coordinates of the set of points against a pre-defined reference `Sample`. This reference sample can be computed from another library, e.g. Matlab, Scilab, Mathematica, etc. Several dimensions e.g. 2, 3, and 4 should be checked. Instead, the first part of the current unit test checks the library against itself, which prevents from detecting any bug. The second part of the unit test estimates the number $\pi$. Unfortunately, the algorithm seems to converge with relative accuracy equal to $10^{-4}$. Actually, this unit test is still of poor value. Indeed, the fact that the estimate converges is a weak proof that the method is OK. Indeed, we may expect that the _rate_ of convergence is OK (which is not). Finally, the sample size is equal to 1000, instead of being equal to a power of the basis, i.e. 2 in this particular case.

#### Proof #1: In dimension 2

```python
import openturns as ot
import openturns.usecases.ishigami_function as ishigami
import openturns.viewer as viewer
import matplotlib.pyplot as plt

basis = 2
dimension = 2
taille = basis**dimension
distribution = ot.ComposedDistribution([ot.Uniform(0.0, 1.0)] * dimension)
sequence = ot.FaureSequence()
experiment_QMC = ot.LowDiscrepancyExperiment(sequence, distribution, taille, False)
sample = experiment_QMC.generate()
fig = viewer.PlotDesign(
    sample,
    subdivisions=[basis] * dimension,
    bounds=ot.Interval([0.0] * dimension, [1.0] * dimension),
)
plt.title(f"n = {taille}")
fig.set_figwidth(3.0)
fig.set_figheight(3.0)
```

image
**Figure 1.** The first 4 points of the Faure sequence in dimension 2. There should be one single point for each elementary volume, but the upper left cell contain 2 points.

#### Proof #2: In dimension 3

The script below defines the first 9 points of the Faure sequence in dimension 3. This is from [^Baudin2013].
```python
# First 9 points of Faure sequence in dimension 3
expected = ot.Sample([
    [1 / 3, 1 / 3, 1 / 3],
    [2 / 3, 2 / 3, 2 / 3],
    [1 / 9, 4 / 9, 7 / 9],
    [4 / 9, 7 / 9, 1 / 9],
    [7 / 9, 1 / 9, 4 / 9],
    [2 / 9, 8 / 9, 5 / 9],
    [5 / 9, 2 / 9, 8 / 9],
    [8 / 9, 5 / 9, 2 / 9],
])
```

Now compare to the result from OpenTURNS.
```
dimension = 3
taille = 9
distribution = ot.ComposedDistribution([ot.Uniform(0.0, 1.0)] * dimension)
sequence = ot.FaureSequence()
experiment_QMC = ot.LowDiscrepancyExperiment(sequence, distribution, taille, False)
sample = experiment_QMC.generate()
sample
```
which prints:

<table>
  <tr><td></td><th>y0</th><th>y1</th><th>y2</th></tr>
  <tr><th>0</th><td>0.5</td><td>0.5</td><td>0.5</td></tr>
  <tr><th>1</th><td>0.25</td><td>0.75</td><td>0.25</td></tr>
  <tr><th>2</th><td>0.75</td><td>0.25</td><td>0.75</td></tr>
  <tr><th>3</th><td>0.125</td><td>0.625</td><td>0.125</td></tr>
  <tr><th>4</th><td>0.625</td><td>0.125</td><td>0.625</td></tr>
  <tr><th>5</th><td>0.375</td><td>0.375</td><td>0.375</td></tr>
  <tr><th>6</th><td>0.875</td><td>0.875</td><td>0.875</td></tr>
  <tr><th>7</th><td>0.0625</td><td>0.9375</td><td>0.0625</td></tr>
  <tr><th>8</th><td>0.5625</td><td>0.4375</td><td>0.5625</td></tr>
</table>

We see that the basis is equal to 2 instead of 3. We see that the third dimension is a copy of the first one.

[^Baudin2013]: "Low Discrepancy Toolbox Manual", Michael Baudin. Version 0.1. May 2013. https://gitlab.com/scilab/forge/lowdisc

#### Proof #3: Convergence on GSobol'

Notice that the lack of convergence of the faulty Faure sequence is not so obvious on the Ishigami function, because the third input variable is no so influential. Only the speed of convergence is affected here.

We consider the GSobol' test function and compute its mean.

```python
a=[0.0, 9.0, 99.0]
def GSobolModel(X):
    X = ot.Point(X)
    d = X.getDimension()
    Y = 1.0
    for i in range(d):
        Y *= (abs(4.0 * X[i] - 2.0) + a[i]) / (1.0 + a[i])
    return ot.Point([Y])

gSobolFunction = ot.PythonFunction(dimension, 1, GSobolModel)
gSobolFunction.setOutputDescription(["Y"])

# Define the distribution
distributionList = [ot.Uniform(0.0, 1.0) for i in range(dimension)]
distribution = ot.ComposedDistribution(distributionList)

name = "GSobol"

# Compute Mean
gSobolMean = 1.0


dimension = 3
basis = 3  # ie the smallest prime number greater or equal to the dimension
sampleSize = basis ** 7  # This is significant!
sequence = ot.FaureSequence()
experiment_QMC = ot.LowDiscrepancyExperiment(sequence, im.distributionX, sampleSize, False)
inputSample = experiment_QMC.generate()
outputSample = gSobolFunction(inputSample)
print(f"n = {sampleSize} , estimate = {outputSample.computeMean()[0]}")
print(f"Expected = {gSobolMean}")
```
We get:
```
n = 2187 , estimate = 10.722134899604974
Expected = 1.0
```

### Version

1.22

### Operating System

unknown

### Installation media

unknown

### Additional Context

_No response_

mbaudin47:
- Improve unit test of Halton sequence
- Improve unit test of Faure sequence
- Improve unit test of Sobol' sequence
- Closes #2653


lib/test/t_FaureSequence_std.cxx
```diff
@@ -31,31 +31,74 @@ int main(int, char *[])
...
     Scalar probabilityEstimate = 1.0 * pointInsideCircle / sampleSize;
     Scalar probability = M_PI / 4.0;
-    Scalar relativeError = std::abs(probability - probabilityEstimate) / probability;
     fullprint << "sample size=" << sampleSize << std::endl;
-    fullprint << "relative error to Pi=" << relativeError << std::endl;
+    fullprint << "computed probability =" << probabilityEstimate << std::endl;
+    fullprint << "expected probability =" << probability << std::endl;
+    Scalar rtol = 10.0 / sampleSize;
+    assert_almost_equal(probabilityEstimate, probability, rtol);
   }
```
mbaudin47, line 102: Please translate the unit test from GSobol'

lib/src/Base/Stat/LowDiscrepancySequenceImplementation.cxx
```diff
@@ -381,9 +381,9 @@ Unsigned64BitsInteger LowDiscrepancySequenceImplementation::GetNextPrimeNumber(c
 #ifdef OPENTURNS_HAVE_PRIMESIEVE
   primesieve::iterator it;
 #if PRIMESIEVE_VERSION_MAJOR >= 11
-  it.jump_to(n - 1);
+  it.jump_to(n);
```
jschueller, line 384: Could you change the commit title ? the "closes #..." should be in the message body
A rebase is needed anyway

mbaudin47: Ok. Fixed and rebased.

ChangeLog
```diff
@@ -27,6 +27,8 @@
 
 === Miscellaneous ===
 
+=== Bug fixes (total 45) ===
+ * #2653 (The Faure sequence is wrong)
```
jschueller, line 31: please drop these lines, I will update the 1.23 section when tagging

jschueller (approved):

ChangeLog
mbaudin47, line 31:
```suggestion
```

lib/src/Base/Stat/LowDiscrepancySequenceImplementation.cxx
regislebrun, line 384: Line 380, the test `if (n == 0) return 2;` should be replaced by `if (n <= 2) return 2;`

lib/test/t_FaureSequence_std.cxx
```diff
@@ -31,31 +31,74 @@ int main(int, char *[])
...
+    Sample expected3D(8, 3);
+    expected3D[0] = Point({1.0 / 3.0, 1.0 / 3.0, 1.0 / 3.0});
+    expected3D[1] = Point({2.0 / 3.0, 2.0 / 3.0, 2.0 / 3.0});
+    expected3D[2] = Point({1.0 / 9.0, 4.0 / 9.0, 7.0 / 9.0});
+    expected3D[3] = Point({4.0 / 9.0, 7.0 / 9.0, 1.0 / 9.0});
+    expected3D[4] = Point({7.0 / 9.0, 1.0 / 9.0, 4.0 / 9.0});
+    expected3D[5] = Point({2.0 / 9.0, 8.0 / 9.0, 5.0 / 9.0});
+    expected3D[6] = Point({5.0 / 9.0, 2.0 / 9.0, 8.0 / 9.0});
+    expected3D[7] = Point({8.0 / 9.0, 5.0 / 9.0, 2.0 / 9.0});
+    FaureSequence sequence3D(3);
```
regislebrun, line 74: Have you checked the values for larger samples, even if these values don't have to be checked in the test? This is just the beginning of the sequence and surprises can pop up for larger values...

regislebrun (approved):
Very nice work! Thanks for digging into this not-so-easy piece of code. I would be reassured if you checked the values for larger sequences.

mbaudin47:
- @regislebrun states that the dimension 2 works perfectly, which is the contrary of what @mbaudin47 . Burkardt (C++ and Matlab) and Dimitry Shvorob (Mathworks.com, 15373, "Generate a Faure sequence"). 
- TODO : extend the sample size in the unit tests up to size TODO

jschueller:
good to go after the first Regis comment is addressed

josephmure (approved):

PR #2672, opened 2024-06-02
Generic implementation of the partial estimation of a distribution when a subset of its parameter values is known

regislebrun:
I moved the mechanism already proposed in LeastSquaresDistributionFactory, MaximumLikelihoodFactory, QuantileMatchingFactory, MethodOfMomentsFactory at the DistributionFactoryImplementation level in order to allow the user to fix a part of the parameter values before to estimate the other values.

It is done by using the MaximumLikelihoodFactory estimation as a second pass after a full estimation based on the current buildAsXXX() methods. This first estimation gives the starting point of the subsequent restricted MLE estimation.

Many files are impacted, but for most of them it is only a one-liner added before to return the current result, such as [here](https://github.com/regislebrun/openturns/blob/7d289b76dbcdfc169be6a22f6dc52e0b29816187/lib/src/Uncertainty/Distribution/AliMikhailHaqCopulaFactory.cxx#L112)

For some other factories already based on MLE, I simply added the known parameter values and indices to the estimation.

Systematic tests are needed, maybe through a generic test, but I don't see how to propose validated reference values for these many configurations.


lib/src/Uncertainty/Model/DistributionFactoryImplementation.cxx
```diff
@@ -199,18 +200,51 @@ void DistributionFactoryImplementation::setBootstrapSize(const UnsignedInteger b
...
+  adv.saveAttribute("knownParameterValues_", knownParameterValues_);
+  adv.saveAttribute("knownParameterIndices_", knownParameterIndices_);
 }
 
 /* Method load() reloads the object from the StorageManager */
 void DistributionFactoryImplementation::load(Advocate & adv)
 {
   PersistentObject::load(adv);
   adv.loadAttribute("bootstrapSize_", bootstrapSize_);
+  adv.loadAttribute("knownParameterValues_", knownParameterValues_);
```
jschueller, line 246: if (hasattribute(''knownParameterValues_")) ...

lib/src/Uncertainty/Distribution/BetaFactory.cxx
```diff
@@ -75,7 +75,9 @@ Beta BetaFactory::buildAsBeta(const Sample & sample) const
 
   const Scalar mu = sample.computeMean()[0];
   const Scalar sigma = sample.computeStandardDeviation()[0];
-  return buildAsBeta(BetaMuSigma(mu, sigma, a, b).evaluate());
+  Beta result(buildAsBeta(BetaMuSigma(mu, sigma, a, b).evaluate()));
+  adaptToKnownParameter(sample, &result);
```
jschueller, line 79: it looks setDescription is missing here unlike the others

jschueller (approved):
good idea, only minor remarks

lib/src/Uncertainty/Distribution/BetaFactory.cxx
regislebrun: done

lib/src/Uncertainty/Model/DistributionFactoryImplementation.cxx
regislebrun: done

jschueller:
TODO: tests/examples

mbaudin47:
Please add a new unit test, e.g. based on [t_Distribution_std.py](https://github.com/openturns/openturns/blob/master/python/test/t_Distribution_std.py).

mbaudin47:
> TODO: tests/examples

Please add a new unit test, e.g. based on [t_Distribution_std.py](https://github.com/openturns/openturns/blob/master/python/test/t_Distribution_std.py).

regislebrun:
@jschueller @mbaudin47 I added a generic test for the new feature, see python/test/t_DistributionFactory_std.py for the details. Some adjustments had to be made in order to check all the factories, it may be link to unknown bugs as the coverage is now much improved wrt the factories

python/test/t_DistributionFactory_std.py
```diff
@@ -46,9 +47,77 @@
...
+            rEps = 1.0
+        if factoryName == "TriangularFactory":
+            aEps = 0.025
+            rEps = 0.025
+        if factoryName == "VonMisesFactory":
+            aEps = 0.02
+            rEps = 0.02
+        if factoryName == "NegativeBinomialFactory":
+            aEps = 0.05
+            rEps = 0.05
```
jschueller, line 100: it would be simpler to write this as a dictionnary, you dont even need to store both values since they're equal

lib/src/Uncertainty/Model/DistributionFactoryImplementation.cxx
```diff
@@ -211,6 +241,10 @@ void DistributionFactoryImplementation::load(Advocate & adv)
 {
   PersistentObject::load(adv);
   adv.loadAttribute("bootstrapSize_", bootstrapSize_);
+  if (adv.hasAttribute("knownParameterValues_"))
+    adv.loadAttribute("knownParameterValues_", knownParameterValues_);
+  if (adv.hasAttribute("knownParameterIndices_"))
+    adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
```
jschueller, line 247: these sould appear in the save method too

lib/src/Uncertainty/Distribution/QuantileMatchingFactory.cxx
```diff
@@ -386,9 +370,11 @@ void QuantileMatchingFactory::load(Advocate & adv)
   DistributionFactoryImplementation::load(adv);
   adv.loadAttribute("distribution_", distribution_);
   adv.loadAttribute("probabilities_", probabilities_);
-  adv.loadAttribute("knownParameterValues_", knownParameterValues_);
-  adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
   adv.loadAttribute("optimizationBounds_", optimizationBounds_);
+  if (adv.hasAttribute("knownParameterValues_"))
+    adv.loadAttribute("knownParameterValues_", knownParameterValues_);
+  if (adv.hasAttribute("knownParameterIndices_"))
+    adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
```
jschueller, line 377: now they're loaded twice

lib/src/Uncertainty/Distribution/MethodOfMomentsFactory.cxx
```diff
@@ -412,9 +399,11 @@ void MethodOfMomentsFactory::load(Advocate & adv)
   DistributionFactoryImplementation::load(adv);
   adv.loadAttribute("distribution_", distribution_);
   adv.loadAttribute("momentOrders_", momentOrders_);
-  adv.loadAttribute("knownParameterValues_", knownParameterValues_);
-  adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
   adv.loadAttribute("optimizationBounds_", optimizationBounds_);
+  if (adv.hasAttribute("knownParameterValues_"))
+    adv.loadAttribute("knownParameterValues_", knownParameterValues_);
+  if (adv.hasAttribute("knownParameterIndices_"))
+    adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
```
jschueller, line 406: here too

lib/src/Uncertainty/Distribution/MaximumLikelihoodFactory.cxx
```diff
@@ -431,10 +413,12 @@ void MaximumLikelihoodFactory::save(Advocate & adv) const
...
 {
   DistributionFactoryImplementation::load(adv);
-  adv.loadAttribute("knownParameterValues_", knownParameterValues_);
-  adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
   adv.loadAttribute("optimizationBounds_", optimizationBounds_);
   adv.loadAttribute("optimizationInequalityConstraint_", optimizationInequalityConstraint_);
+  if (adv.hasAttribute("knownParameterValues_"))
+    adv.loadAttribute("knownParameterValues_", knownParameterValues_);
+  if (adv.hasAttribute("knownParameterIndices_"))
+    adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
```
jschueller, line 421: and here

lib/src/Uncertainty/Distribution/LeastSquaresDistributionFactory.cxx
```diff
@@ -330,10 +309,12 @@ void LeastSquaresDistributionFactory::save(Advocate & adv) const
...
 {
   DistributionFactoryImplementation::load(adv);
-  adv.loadAttribute("knownParameterValues_", knownParameterValues_);
-  adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
   adv.loadAttribute("optimizationBounds_", optimizationBounds_);
   adv.loadAttribute("optimizationInequalityConstraint_", optimizationInequalityConstraint_);
+  if (adv.hasAttribute("knownParameterValues_"))
+    adv.loadAttribute("knownParameterValues_", knownParameterValues_);
+  if (adv.hasAttribute("knownParameterIndices_"))
+    adv.loadAttribute("knownParameterIndices_", knownParameterIndices_);
```
jschueller, line 317: and here

mbaudin47:
- Can you please execute the unit test and report the list of Factories for which the test does not pass, i.e. the list of factories for which the MaximumLikelihoodEstimator provides an estimation different from the default one? After this, we can create a specific issue for the guilty factories.
- Can these issues be generated by the use of the MaximumLikelihoodEstimator in some cases, e.g. estimating bounds of distributions? This is because this estimator can sometimes fail or have a different behavior from the `SpecificFactory.build()` method, e.g. `GumbelFactory()` use the method of moments or `LogNormalFactory()`, which has several mixed strategies.

mbaudin47:
Here is an example of a Python which is available before this PR:
```python
import openturns as ot
ot.RandomGenerator.SetSeed(0)
sample = ot.LogNormal(1.5, 2.5, -1.5).getSample(1000)
factory = ot.MaximumLikelihoodFactory(ot.LogNormal())
factory.setKnownParameter([2], -1.5)  # Set gamma to a known value
estimated = factory.build(sample)  # Estimate mu and sigma only
```

Here is an example of a Python which is available after this PR:
```python
import openturns as ot
ot.RandomGenerator.SetSeed(0)
sample = ot.LogNormal(1.5, 2.5, -1.5).getSample(1000)
factory = ot.LogNormalFactory()  # This is the log-normal factory
factory.setKnownParameter([2], -1.5)  # Set gamma to a known value
estimated = factory.build(sample)  # Estimate mu and sigma only
```

adutfoy:
We have not fullly understood what had been implemented.
Regis explained the strategy behind: "It is done by using the MaximumLikelihoodFactory estimation as a second pass after a full estimation based on the current buildAsXXX() methods. This first estimation gives the starting point of the subsequent restricted MLE estimation."
Doing that, we use all the strategies developped for each distribution to get a good full point: this point is reduced to the coordinates that have to be estimated. This point is used to create the starting point for the partial MLH estimator which converges quickly due to the good starting point. Moreover, the potential constraints for admissible parameters values are naturallly taken into account is the starting point is admissible.
Conclusion: event of the lines have not changed a lot as Michaël showed it, the new developpements have a lot of advantages...

python/test/t_DistributionFactory_std.py
```diff
@@ -46,9 +47,56 @@
...
+}
+
+for factory in factories:
+    print("factory=", factory)
+    try:
+        factoryName = factory.getImplementation().getClassName()
+        if factoryName in params:
+            eps = params[factoryName]
+        else:
+            eps = 0.01e100
```
adutfoy, line 81: is it 0.01 ?? :-)

adutfoy:
We answer here the previous questions: 

- "Why removing HistogramFactory based on the fact that it is nonparametric, but keep KernelSmoothing?"
KernelSmoothing is not in the list returned by  GetContinuousUniVariateFactories() and GetContinuousMultiVariateFactories().
HistogramFactory has been removed because it is not possible to recover an histogram with the same number of parameters (number of bins, width of each bin, ...) than the one which generated the data.

- "The fact that the test succeeds for LogNormalFactory is especially strange, since this it has several estimation methods: does it really succeds is this is just luck?"
No it is not luck... :-)

- "In the case where we evaluate LogNormalFactory.buildAsLogNormal() and set the method as the method of moments. It seems that the adaptToKnownParameter() ignores that choice and use the maximum likelihoood method. In this case, how can be use the method of moments for this specific factory?"
The unsconstrained estimator is calculated with the method chosen by the User. Only the last step is done with MLE which starting point is the previous  unsconstrained estimator.

- "Can these issues be generated by the use of the MaximumLikelihoodEstimator in some cases, e.g. estimating bounds of distributions? This is because this estimator can sometimes fail or have a different behavior from the SpecificFactory.build() method, e.g. GumbelFactory() use the method of moments or LogNormalFactory(), which has several mixed strategies."
It is true that even if the starting point is very good, the laste step using the MLE might not define a proper estimator. This has to be inspected case by case. In these case,s, it is straightforward to swith from the MLE to another method in the code (methods of moments, ...).

- "Can you please execute the unit test and report the list of Factories for which the test does not pass, i.e. the list of factories for which the MaximumLikelihoodEstimator provides an estimation different from the default one? After this, we can create a specific issue for the guilty factories."
CARE!!! there is a typo in the test... the eps should be 0.01 which means that **the given list is excatly the list of pathologic distributions**...... and not the contrary....
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

python/test/t_DistributionFactory_std.py
regislebrun: Oups! It was supposed to be 0.01. Fixed

python/test/t_DistributionFactory_std.py
```diff
@@ -46,9 +47,56 @@
...
+    "VonMisesFactory": 0.02,
+    "NegativeBinomialFactory": 0.05,
+}
+
+for factory in factories:
+    print("factory=", factory)
+    try:
+        factoryName = factory.getImplementation().getClassName()
+        if factoryName in params:
+            eps = params[factoryName]
```
jschueller, line 79: you could use `dict.get(key, default_value)` which is much shorter than the if/else

lib/src/Uncertainty/Model/DistributionFactoryImplementation.cxx
regislebrun: done

lib/src/Uncertainty/Distribution/MethodOfMomentsFactory.cxx
regislebrun: done

python/test/t_DistributionFactory_std.py
regislebrun: done, thks

python/test/t_DistributionFactory_std.py
```diff
@@ -46,9 +47,53 @@
...
+                # Generate all the combinations of known parameter
+                combinations = ot.Combinations(k, n).generate()
+                for combination in combinations:
+                    indices = [int(x) for x in combination]
+                    factory.setKnownParameter([refParameter[i] for i in indices], indices)
+                    res = factory.build(sample)
+                    resParameter = res.getParameter()
+                    ott.assert_almost_equal(resParameter, refParameter, eps, eps)
+    except Exception:
+        print("Skip", factory, "no getClassName()")
```
jschueller, line 98: I dont see why getClassName would fail here
perhaps the block fails because of another error ? or maybe the try/except could be removed ?

regislebrun: @jschueller SmoothedUniformFactory has no getClassName()...

jschueller: indeed, one can also use the distribution name:
factory.build().getImplementation().getClassName()
=> SmoothedUniform

jschueller: ok for me then, just needs a rebase (because of the renames)

jschueller:
@sofianehaddad @mbaudin47 @josephmure @JPelamatti can someone please review ?

jschueller:
hello Regis,
sorry but the save/load methods of the following classes are still not ok:
MethodOfMomentsFactory, LeastSquaresDistributionFactory, MaximumLikelihoodFactory, QuantileMatchingFactory
the known parameter attributes are being moved to the parent class so they should not be saved/loaded there

regislebrun:
@jschueller It should be good now...

jschueller:
great, we just need another review @sofianehaddad @mbaudin47

mbaudin47 (approved):
This new feature will be a very convenient one. But there are missing unit tests: the new methods `getKnownParameterIndices()` and `getKnownParameterValues()` are untested. This is one of the reasons CodeCov is not as happy as possible.

image

I suggest to create either a unit test or an example, so that these methods are tested.

mbaudin47 (changes requested):
I suggest to create a unit test to the new methods `getKnownParameterIndices()` and `getKnownParameterValues()`.

PR #2702, opened 2024-06-25
Update FilonQuadrature_doc.i.in

adutfoy:
Wrong description


adutfoy:
I moved these modifications into the PR#2699

PR #2703, opened 2024-06-25
Update CubaIntegration_doc.i.in

adutfoy:
Replaced optimization by integration in the doc


adutfoy:
I moved these modifications into the PR#2699

PR #2705, opened 2024-06-25
Update plot_randomize_sample_lines.py

mbaudin47:
Simplified the use of KPermutationsDistribution.


jschueller:
could you move it to #2650 ?

jschueller:
moved to #2760

PR #2724, opened 2024-07-23
Update DistributionImplementation_doc.i.in

mbaudin47:
Fixes the doc of Distribution.computePDFGradient()


jschueller:
moved to #2760

PR #2757, opened 2024-09-06
modification of kernel to speed up EGO example and converge

### LINKED ISSUE #1302: EGO Example should show convergence
i tried to run the code embeded in http://openturns.github.io/openturns/latest/user_manual/_generated/openturns.EfficientGlobalOptimization.html?highlight=ego

the code runs just well but the optimum reached by OT is incorrect and result.getIterationNumber() returns 0

i work with the version 1.13 on a windows 10 computer

regards,

m-balesdent:
This PR closes #1302.

The EGO example has been modified to enhance convergence of the algorithm without increasing the number of iteration. To do so, the kernel has been changed to Matern which is mored adapted than SquareExponential in order to deal with Ackley function.

### DIFF
--- a/python/doc/bibliography.rst
+++ b/python/doc/bibliography.rst
@@ -270,8 +270,8 @@ Bibliography
     *Spectral methods for uncertainty quantification: with applications to computational fluid dynamics*. Springer Science & Business Media.
 .. [lemieux2009] Lemieux, C. (2009). *Monte Carlo and Quasi-Monte Carlo Sampling*.
     Springer. Springer Series in Statistics.
-.. [leriche2021] Le Riche, R., & Picheny, V. (2021). *ERevisiting Bayesian optimization in the
-    light of the COCO benchmark.* Structural and Multidisciplinary 
+.. [leriche2021] Le Riche, R., & Picheny, V. (2021). *Revisiting Bayesian optimization in the
+    light of the COCO benchmark.* Structural and Multidisciplinary
     Optimization, 64, 3063-3087.
 .. [liu2006] Liu, R., & Owen, A. B. (2006). *Estimating mean dimensionality of
     analysis of variance decompositions.* Journal of the American Statistical
--- a/python/doc/bibliography.rst
+++ b/python/doc/bibliography.rst
@@ -270,6 +270,9 @@ Bibliography
     *Spectral methods for uncertainty quantification: with applications to computational fluid dynamics*. Springer Science & Business Media.
 .. [lemieux2009] Lemieux, C. (2009). *Monte Carlo and Quasi-Monte Carlo Sampling*.
     Springer. Springer Series in Statistics.
+.. [leriche2021] Le Riche, R., & Picheny, V. (2021). *ERevisiting Bayesian optimization in the
+    light of the COCO benchmark.* Structural and Multidisciplinary 
+    Optimization, 64, 3063-3087.
 .. [liu2006] Liu, R., & Owen, A. B. (2006). *Estimating mean dimensionality of
     analysis of variance decompositions.* Journal of the American Statistical
     Association, 101 (474), 712-721.
--- a/python/doc/examples/numerical_methods/optimization/plot_ego.py
+++ b/python/doc/examples/numerical_methods/optimization/plot_ego.py
@@ -102,7 +102,7 @@ view = viewer.View(graph)
 # We use default settings (1.0) for the scale parameters of the covariance model, but configure the amplitude to 0.1, which better corresponds to the properties of the Ackley function.
 
 # %%
-covarianceModel = ot.SquaredExponential([1.0] * dim, [0.5])
+covarianceModel = ot.MaternModel([1.0] * dim, [0.5], 1.5)
 basis = ot.ConstantBasisFactory(dim).build()
 kriging = ot.KrigingAlgorithm(inputSample, outputSample, covarianceModel, basis)
 kriging.run()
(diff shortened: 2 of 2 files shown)


jschueller (approved):

m-balesdent:
> Hi @m-balesdent thanks for the PR
> 
> I made some experiments, `Matern 5/2` seems a bit better (at least on the `Ackley` use-case) and keep the same regularity as the `SquaredExponential`
> 
> With @mbaudin47 we were trying to improve these examples during the hackathon in term of speed up and precision. With `maximumEvaluationNumber` set to `50`, the new covariance model and by setting `Cobyla` instead of `TNC` for the Kriging purposes, results seem a bit better as illustrated here after
> 
> ```
> Kriging time build :  0.005391120910644531
> EGO time run :  2.2223830223083496
> ```
> 
> image

Thanks for your experiments. I suggest to add this PR to the agenda of the next technical committee.

sofianehaddad (approved):

mbaudin47:
@m-balesdent : Don't you think that the Ackley function is a bit too difficult for the kriging to find its minimum using a starting sample size equal to 50? This observation is because, counting from my eyes, I found approximately 80 local minimum in this function. So, if we were to know exactly the locations of these minimum _beforehand_, it would require at least 80 points to get a very approximate view of the trend of the function.

image

**Figure 1.** The Ackley has many local minimums.

If we look at the Branin function instead, there are only 3 local minimums. Therefore, there are much better odds to make a good kriging starting from an initial sample made of 50 points.

image

**Figure 2.** The Branin has three local minimums.

So, I would use the `MetamodelValidation` in order to see what is the minimum sample size of the kriging which makes a satisfactory kriging: I bet that the Branin function is much easier to approximate. Don't you think?

Therefore, the EGO algorithm should perform better on this test case. As a side note, it might be interesting to show how the `OptimalLHSExperiment` behaves in this case, because there is no actual use of this case _in kriging context_.

m-balesdent:
> @m-balesdent : Don't you think that the Ackley function is a bit too difficult for the kriging to find its minimum using a starting sample size equal to 50? This observation is because, counting from my eyes, I found approximately 80 local minimum in this function. So, if we were to know exactly the locations of these minimum _beforehand_, it would require at least 80 points to get a very approximate view of the trend of the function.
> 
> image
> 
> **Figure 1.** The Ackley function has many local minimums.
> 
> If we look at the Branin function instead, there are only 3 local minimums. Therefore, there are much better odds to make a good kriging starting from an initial sample made of 50 points.
> 
> image
> 
> **Figure 2.** The Branin function has three local minimums.
> 
> So, I use `MetamodelValidation` in order to see what is the minimum sample size of the kriging which makes a satisfactory kriging: I bet that the Branin function is much easier to approximate.
> 
> image
> 
> **Figure 3.** Simple validation on the Ackley test function, with a test sample size equal to 50 made of observations different from the training sample. The Q2=0.36 shows that the kriging metamodel is poor.
> 
> image
> 
> **Figure 4.** Simple validation on the Branin test function, with a test sample size equal to 50 made of observations different from the training sample. The Q2=0.98 shows that the kriging metamodel is quite accurate.
> 
> In the two next figure, we look at the Q2 score depending on the size of the training sample. We see that the Ackley test function requires more than 130 observations to get a Q2 consistently greater than 0.8. For the Branin test function, we see that 50 observations lead to a Q2 generally greater than 0.9. Hence, the Ackley test function requires a larger sample size to be correctly approximated by the kriging metamodel.
> 
> image
> 
> **Figure 5.** The Q2 score depending on the size of the training sample for the Ackley test function.
> 
> image
> 
> **Figure 6.** The Q2 score depending on the size of the training sample for the Branin test function.
> 
> Therefore, the EGO algorithm should perform better on the Branin test case than on the Ackley test case. As a side note, it might be interesting to show how the `OptimalLHSExperiment` behaves in this case, because there is no actual use of this optimal LHS class used to build the experimental design _in kriging context_ (the example at [plot_kriging_hyperparameters_optimization.html](http://openturns.github.io/openturns/latest/auto_meta_modeling/kriging_metamodel/plot_kriging_hyperparameters_optimization.html) uses it to show how to enhance the estimation of the hyperparameters).

Thanks for you experiments. We have to keep in mind that kriging for optimization purpose is not the same as kriging for metamodeling. Papers from literature recommand to start with 3 to 10 times the dimension of the input space as initial DoE (see for example https://link.springer.com/article/10.1007/s00158-021-02977-1

Indeed, a lot of people prefer to let the EGO algorithm discovers the regions instead of giving before hand a large DoE.

In the Ackley example, it is not necessary to explore all the local minima if the kriging model is confident about the values at the center of these minima. The EGO algorithm will place a point at one of these minima if, taken into account the uncertainty, the potential value of these minima is better than the current optimum.

For example, I tried to solve the Ackley problem with an initial DoE of 20 (so 10 times the problem dimension) and with 40 iterations of EGO (leading to the same computational budget of the current example). You will find below the convergence curve : 

image

So we could stop the algorithm at iteration 15 (so with only 35 evaluated points) to get an accurate optimum.

Here is the map of evaluated points : 
image

At the end of the optimization process with EGO, we do not obtain a kriging model that is performant to model the function everywhere. The kriging model is just a tool to find the global optimum. 
For example, if I build the metamodel at the end of EGO, I find : 

image

This metamodel is just used to "filter" the optima and concentrates on the global one.

m-balesdent:
I modified in depth the example to be more in line with current recommandations on EGO community : 
 https://link.springer.com/article/10.1007/s00158-021-02977-1

"Our results validate most common practices: the Matern 5/2 kernel, no data transformation and no non-stationary GPs by default, and a small initial DoE."

I put less point in the initial DoE and let the algorithm converge with more iterations. As a result, it reached the optima for both functions. For that reason, I propose to remove the part of the example dealing with BFGS because it is not required anymore.

mbaudin47 (approved):

l-brevault (approved):

jschueller (approved):
example now runs in 2.36s, great

Issue #1179, opened 2019-06-23
The SobolSimulationAlgorithm class has doc issues.


mbaudin47:
The `SobolSimulationAlgorithm` class has at least two documentation issues.
* The purpose of the batchsize parameter is unclear, so that the user cannot understand why he would modify it.
* There is not mention of the SobolSimulationAlgorithm-DefaultBatchSize and SobolSimulationAlgorithm-DefaultBlockSize resource map keys so that the default values of the parameters are unknown.

Moreover, this class has no example in the examples section. One example could be based on [this example](https://github.com/mbaudin47/otsupgalilee-eleve/blob/master/4-Sensibilite/Exercice-Sensibilite-Sobol-cantilever-beam.ipynb).

## The batchsize parameter

I this section, I provide an explanation of the batchsize parameter.

The purpose of the batchsize parameter is to send feedback to the user more often. Two feedback systems are available: the progress callback allows to send information from the algorithm to the user in order, e.g. to update a progress bar; the stop callback allows to send information back from the user to the algorithm so that we can stop the algorithm when required, e.g. with a stop button in a graphical user interface. 

Evaluating the design of experiments associated with just one single block requires blocksize*(nx+2) evaluations, where nx is the number of input variables. If blocksize or nx are large, then evaluating a block may require a certain computer time. However, there are particular uses of the OpenTURNS library where we want to get a regular feedback from the algorithm. Consider, for example, a graphical user interface where a progress bar informs the user of the progression of the algorithm.  If the blocksize parameter was the only one available to the user, the user would get a feedback only each blocksize*(nx+2) evaluations. Instead, everytime a new batch has been evaluated, the progress and stop callbacks are evaluated so that the graphical user interface can be updated.

mbaudin47:
This issue is managed in a contract.

PR #2760, opened 2024-09-12
Misc

jschueller:
- #1179 
- #2761 
- #2773
- Move CrossEntropy, StudentCopula, TruncatedOverMesh, GEVValidation out of experimental
- Workaround a matplotlib 3.6 issue with single contours labels in viewer
- Implement JointDistribution.getSupport
- Account for drawing range in 2-d discrete PDF graphs and dont write support over values, change point style
- in TruncatedDistribution we can use the simplified version for most services, not just sampling
- Fix for FieldFCA algos segfaulting with 1 eigen value, and missing some modes


lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -4047,8 +4048,8 @@ Graph DistributionImplementation::drawLogPDF(const Point & xMin,
   }
   if (isDiscrete())
   {
-    const Sample support(getSupport());
-    const Point probabilities(getProbabilities());
+    const Sample support(getSupport(Interval(xMin, xMax)));
```
regislebrun, line 4051: good catch

regislebrun (approved):
A small bug to fix, otherwise ok

m-balesdent (approved):
I'm OK with the changes made on CrossEntropy and NAIS

PR #2775, opened 2024-10-07
Replaced assert with assert_equal in C++ unit tests

mbaudin47:
When `assert` is used in C++ unit tests, it does not check anything. Using this may prevent from seeing actual bugs. This PR replaces `assert` with `assert_equal` in C++ unit tests

### DIFF
--- a/lib/test/t_ExperimentIntegration_std.cxx
+++ b/lib/test/t_ExperimentIntegration_std.cxx
@@ -62,14 +62,14 @@ int main(int, char *[])
 
   fullprint << "Test integrate()" << std::endl;
   const Point ishigamiMeanComputed(integration.integrate(ishigamiFunction));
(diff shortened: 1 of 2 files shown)


lib/test/t_ExperimentIntegration_std.cxx
```diff
@@ -62,14 +62,14 @@ int main(int, char *[])
 
   fullprint << "Test integrate()" << std::endl;
   const Point ishigamiMeanComputed(integration.integrate(ishigamiFunction));
-  assert(ishigamiMeanComputed.getDimension() == 1);
+  assert_equal(ishigamiMeanComputed.getDimension(), (UnsignedInteger) 1);
```
jschueller, line 65: use 1L instead of (UnsignedInteger) 1

jschueller:
moved to #2774

PR #2776, opened 2024-10-09
Improved JointByConditioningDistribution

regislebrun:
Now, the compute[Sequential]Conditional[PDF|CDF[Quantile] methods have a dedicated implementation, resulting into a huge performance boost.

The main goal is to allow the use of JointByConditioningDistribution in FORM/SORM algorithms.


jschueller (approved):

jschueller:
great, it seems truncated, parametrized did not override these either (done in misc)

sofianehaddad (approved):

PR #2781, opened 2024-10-16
Change default colormap to Viridis

### LINKED ISSUE #2626: Non-cyclic contour hue
### What happened?

An idea mentioned by Regis to improve readability of contour plots was to avoid purple colors

contour2
contour1

But this clashes with #2587

```shell
diff --git a/lib/src/Base/Func/EvaluationImplementation.cxx b/lib/src/Base/Func/EvaluationImplementation.cxx
index a56d7f823..5d6437c9b 100644
--- a/lib/src/Base/Func/EvaluationImplementation.cxx
+++ b/lib/src/Base/Func/EvaluationImplementation.cxx
@@ -447,7 +447,7 @@ Graph EvaluationImplementation::draw(const UnsignedInteger firstInputMarginal,
     current.setLabels(Description(1, labels[i]));
     current.setDrawLabels(false);
     current.setLegend(labels[i]);
-    current.setColor(Contour::ConvertFromHSV((360.0 * i / levels.getDimension()), 1.0, 1.0));
+    current.setColor(Contour::ConvertFromHSV((270.0 * i / levels.getDimension()), 1.0, 1.0));
     graph.add(current);
   }
   return graph;
diff --git a/lib/src/Base/Stat/CovarianceModelImplementation.cxx b/lib/src/Base/Stat/CovarianceModelImplementation.cxx
index 96f26e59c..f153e4c8f 100644
--- a/lib/src/Base/Stat/CovarianceModelImplementation.cxx
+++ b/lib/src/Base/Stat/CovarianceModelImplementation.cxx
@@ -1277,7 +1277,7 @@ Graph CovarianceModelImplementation::draw(const UnsignedInteger rowIndex,
     current.setLabels(Description(1, labels[i]));
     current.setDrawLabels(false);
     current.setLegend(labels[i]);
-    current.setColor(Contour::ConvertFromHSV((360.0 * i / levels.getDimension()), 1.0, 1.0));
+    current.setColor(Contour::ConvertFromHSV((270.0 * i / levels.getDimension()), 1.0, 1.0));
     graph.add(current);
   }
   return graph;
```

josephmure:
Closes #2626


jschueller:
was that the conclusion of the ending discussion of last CT ?

these become obsolete:
```
../python/doc/examples/calibration/bayesian_calibration/plot_hierarchical_calibration_fission_gas.py:423:ot.ResourceMap.SetAsString("Contour-DefaultColorMap", "viridis")
../python/doc/examples/graphs/plot_graphs_loglikelihood_contour.py:16:ot.ResourceMap.SetAsString("Contour-DefaultColorMap", "viridis")
```

josephmure:
@jschueller I noted a change in the master w.r.t. the 1.23 version : 

- master : https://openturns.github.io/openturns/master/auto_probabilistic_modeling/distributions/plot_bayes_distribution.html 
image

- 1.23 : https://openturns.github.io/openturns/1.23/auto_probabilistic_modeling/distributions/plot_bayes_distribution.html
image

jschueller:
> @jschueller I noted a change in the master w.r.t. the 1.23 version :
yes that's on purpose

josephmure:
> was that the conclusion of the ending discussion of last CT ?

We reached 2 conclusions and left 1 question open:
- there is no longer a point in trying to fix the HSV cycling inside OT since we now rely entirely on matplotlib. In fact, the 
- viridis can be used as default as long as we check it does not break the doc (I have, and if I missed something we can still fix it in the RC)
- @regislebrun thinks we should switch the default color map norm from "linear" to "rank", but we could not reach a consensus on this subject.

As a side note, it seems that `Drawable::BuildRainbowPalette` is not used anywhere in our codebase (except its own documentation). Maybe we should remove it?

jschueller (approved):
ok, I guess there's no problem to use viridis then, but we should rediscuss the norm

mbaudin47 (approved):
This is a small code with great improvement: thank you!

Issue #2780, opened 2024-10-16
Student from (nu, mu, Covariance) ?


jschueller:
### What is the idea?

For the point conditional distribution we saw that the specific formula operates on the covariance matrix,
but unlike the Normal we dont have a ctor for that.

Issue #1417, opened 2020-03-20
The example of the BarPlot graph is too complicated


mbaudin47:
The example of the BarPlot graph is too complicated: 

http://openturns.github.io/openturns/master/user_manual/_generated/openturns.BarPlot.html

Below is a simpler example.

In this example, we create a `BarPlot` with 3 bars
* starting at 0.0, 
* the first bar has a width equal to 1.0, a height equal to 2.0, 
* the second bar has a width equal to 3.0, a height equal to 4.0, 
* the first bar has a width equal to 2.0, a height equal to 3.0.

```python
import openturns.viewer as otv
import openturns as ot
graph = ot.Graph('BarPlot', 'X', 
                 'Y', True, 'topright')
origin = 0.0
data = [[1.0, 2.0], [3.0, 4.0], [2.0, 3.0]]
barPlot = ot.BarPlot(data, origin, 'orange', 
                     'shaded', 'dashed', 'label')
graph.add(barPlot)
otv.View(graph)
```

The previous script produces the following figure.

![Screenshot from 2020-03-20 22-17-51](https://user-images.githubusercontent.com/31351465/77207054-b4719e00-6af8-11ea-8ff7-0767d9e514c1.png)

PR #2784, opened 2024-10-17
Misc

jschueller:
- #2780 
- #1417 
- MultiStart fixes for overflow in reported number of calls & history @sofianehaddad 
 see http://openturns.github.io/openturns/latest/auto_numerical_methods/optimization/plot_optimization_rastrigin.html
- Fixes for undocumented Normal/Student dimension ctors
- Renamed DistributionChecker to DistributionValidation for consistency with KL/LM/FC/...Validation etc @mbaudin47 
- A workaround for another matplotlib 3.6.x contour bug in the colorbar throwing IndexError


sofianehaddad (approved):
Thanks for the `MultiStart`
Ok also for renaming

lib/src/Uncertainty/Distribution/Normal.cxx
```diff
@@ -102,7 +101,6 @@ Normal::Normal(const Point & mean,
                            , CorrelationMatrix(mean.getDimension())
                            , 1.0)
   , logNormalizationFactor_((-1.0 * mean.getDimension()) * SpecFunc::LOGSQRT2PI)
-  , hasIndependentCopula_(false)
```
josephmure, line 105: Shouldn't you use checkIndependentCopula in this constructor?

josephmure (approved):
Let's get this into 1.24, but can you please first check my comment above?

jschueller:
I added two other commits to simplify the barplot example and workaround a matplotlib bug in viewer

josephmure:
Maybe update the ChangeLog with DistributionValidation (which can be useful to check a PythonDistribution).

jschueller:
> Maybe update the ChangeLog with DistributionValidation (which can be useful to check a PythonDistribution).

will do, a global update of Changelog is needed

Issue #2794, opened 2024-10-29
The drawPDF() method of JointDistribution can take ages to produce the graph with no reason


regislebrun:
### What happened?

The default implementation of drawPDF() doesn't take into account the tensorized nature of the sample on which the PDF is evaluated to produce the graph, nor the factorization on the PDF. As a result, the marginal PDF and CDF are evaluated N^2 times instead of N times. If the marginal PDF/CDF are costly to evaluate (e.g large sample kernel smoothing, composite distribution, random mixture) it takes so much time that it becomes unusable.

### How to reproduce the issue?

```shell
import openturns as ot
import openturns.viewer as otv
import time as time

# Build some costly marginal distributions
marginal0 = ot.Normal() - ot.Normal().sin()
marginal1 = ot.Normal().exp() + ot.Normal()
# And a cheap core
core = ot.GumbelCopula(3.5)
joint = ot.JointDistribution([marginal0, marginal1], core)

# Now, draw the joint PDF bypassing the automatic range computation
a = [marginal0.computeScalarQuantile(0.1), marginal1.computeScalarQuantile(0.1)]
b = [marginal0.computeScalarQuantile(0.9), marginal1.computeScalarQuantile(0.9)]
N = 41

t0 = time.time()
g = joint.drawPDF(a, b, [N]*2)
t1 = time.time()
print("t=", t1 - t0, "s")
view = otv.View(g)
view.save("jointPDF.png")

produces
```
t= 131.87123012542725 s
```

With the following code:
```python
import openturns as ot
import openturns.viewer as otv
import time as time

# This function allows to draw the PDF of a JointDistribution much faster than
# the default algorithm by taking into account the tensorized nature of the grid
# and the partial factorization of the distribution
def drawJointPDF(joint, xMin, xMax, discretization):
    m0 = joint.getMarginal(0)
    m1 = joint.getMarginal(1)
    core = joint.getCore()
    N0 = discretization[0]
    N1 = discretization[1]
    listX0 = [[xMin[0] + (xMax[0] - xMin[0]) * i / (N0 - 1)] for i in range(N0)]
    listX1 = [[xMin[1] + (xMax[1] - xMin[1]) * i / (N1 - 1)] for i in range(N1)]
    pdf0 = [m0.computePDF(x) for x in listX0]
    cdf0 = [m0.computeCDF(x) for x in listX0]
    pdf1 = [m1.computePDF(x) for x in listX1]
[...]

Issue #2786, opened 2024-10-18
The colors of some examples are broken


mbaudin47:
### What happened?

It seems that some examples now have broken colors.

For example, [plot_functional_chaos_graphs.html](https://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.html) was as follows in latest.

![image](https://github.com/user-attachments/assets/5f79f074-3e07-4c77-82b8-a945db92b8c3)

**Figure 1.** The Laguerre polynomials in [latest/plot_functional_chaos_graphs.html](https://openturns.github.io/openturns/latest/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.html).

![image](https://github.com/user-attachments/assets/61a5a8e1-e0ab-48e9-a4e7-ca22ed38958b)

**Figure 2.** Laguerre polynomials in [master/plot_functional_chaos_graphs](https://openturns.github.io/openturns/master/auto_meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos_graphs.html).

It seems to me that the faulty PR is [commit 2b85507e1c63e62e4aaf3d94e63d48709d393928](https://github.com/openturns/openturns/commit/2b85507e1c63e62e4aaf3d94e63d48709d393928#diff-84a7321b8c1468bbb9fada83b9dc00c86a91db28c4e5aa882c7b2c34bea8b2d7).

The same is true for [master/plot_subset_sampling.html](https://openturns.github.io/openturns/master/auto_reliability_sensitivity/reliability/plot_subset_sampling.html).

![image](https://github.com/user-attachments/assets/046f204a-e53a-4478-a59d-8702354a2389)
**Figure 3.** Thresholds in subset sampling in [latest/plot_subset_sampling.html](https://openturns.github.io/openturns/latest/auto_reliability_sensitivity/reliability/plot_subset_sampling.html).

![image](https://github.com/user-attachments/assets/d72af067-d299-418c-8186-6e4fb81e5fbe)
**Figure 4.** Thresholds in subset sampling in [master/plot_subset_sampling.html](https://openturns.github.io/openturns/master/auto_reliability_sensitivity/reliability/plot_subset_sampling.html).

### How to reproduce the issue?

```shell
Read the doc
```

### Version

1.23

### Operating System

unknown

### Installation media

unknown

mbaudin47:
[...]

Issue #2778, opened 2024-10-11
I suggest to create a unit test to the new methods `getKnownParameterIndices()` and `getKnownParameterValues()`.


josephmure:
I suggest to create a unit test to the new methods `getKnownParameterIndices()` and `getKnownParameterValues()`.

_Originally posted by @mbaudin47 in https://github.com/openturns/openturns/pull/2672#pullrequestreview-2349823832_

Issue #2723, opened 2024-07-22
FunctionalChaosAlgorithm.run() can fail


mbaudin47:
### What happened?

In the next script, I create a full polynomial chaos expansion and estimate its coefficients using least squares. This produces:
```
File c:\\Users\\c61372\\AppData\\Local\\miniforge3\\envs\\myEnvironment\\Lib\\site-packages\\openturns\\metamodel.py:4075, in FunctionalChaosAlgorithm.run(self)
   4067 def run(self):
   4068     r\"\"\"
   4069     Compute the metamodel.
   4070 
   (...)
   4073     Evaluates the metamodel and stores all the results in a result structure.
   4074     \"\"\"
-> 4075     return _metamodel.FunctionalChaosAlgorithm_run(self)

RuntimeError: Error in function boost::math::gamma_q_inv<double>(double, double): Overflow Error"
```

### How to reproduce the issue?

```python
# %%
import openturns as ot
import numpy as np
import openturns.viewer as otv

print("OT Version = ", ot.__version__)

# %%
totalDegree = 1
nTrain = 25  # size of the training sample
nTest = 1000

# %%
dist_E = ot.Beta(0.9, 3.1, 2.8e7, 4.8e7)
dist_E.setDescription(["E"])
f_para = ot.LogNormalMuSigma(3.0e4, 9.0e3, 15.0e3)  # in N
dist_F = ot.ParametrizedDistribution(f_para)
dist_F.setDescription(["F"])
dist_L = ot.Uniform(250.0, 260.0)  # in cm
dist_L.setDescription(["L"])
dist_I = ot.Beta(2.5, 4, 310.0, 450.0)  # in cm^4
dist_I.setDescription(["I"])

trueDistribution = ot.ComposedDistribution([dist_E, dist_F, dist_L, dist_I])

dim_input = 4  # dimension of the input
dim_output = 1  # dimension of the output


def function_beam(X):
    E, F, L, I = X
    Y = F * (L ** 3) / (3 * E * I)
    return [Y]


g = ot.PythonFunction(dim_input, dim_output, function_beam)

# %%
ot.RandomGenerator.SetSeed(0)

# %%
inputTrain = trueDistribution.getSample(nTrain)
outputTrain = g(inputTrain)

# %%
#inputDistribution = ot.FunctionalChaosAlgorithm.BuildDistribution(inputTrain)
[...]

PR #2795, opened 2024-10-30
Misc

jschueller:
- #2778
- #2786
- #2794
- #2796
- #2797
- #2798
- #2805 
- #2810
- #2811
- #2812
- #2813
- #2830
- #2832
- #2723
- Return inf in RegularizedIncompleteGammaInverse
- Fix wrong DeconditionedDistribution dimension
- Fix for PosteriorDistribution not taking link function into account
- Fix for UniformOverMesh/TruncatedOverMesh being super slow within a Truncated
- Fix for TruncatedOverMesh mean
- Fix for BlockIndependentDistribution::computeConditional PDF/CDF/Quantile
- Fix for Normal::computeConditionalQuantile 
- DistributionValidation: Check isotransfo + condtional PDF/CDF/quantile


lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -3832,10 +3852,17 @@ Graph DistributionImplementation::drawPDF(const Point & xMin,
   const GraphImplementation::LogScale scale = static_cast<GraphImplementation::LogScale>((logScaleX ? 1 : 0) + (logScaleY ? 2 : 0));
   if (isContinuous())
   {
-    Graph graph(PDFWrapper(this).draw(xMin, xMax, pointNumber, scale));
-    graph.setXTitle(description_[0]);
-    graph.setYTitle(description_[1]);
-    graph.setTitle(String(OSS() << getDescription() << " iso-PDF"));
+    // retrieve x/y discretization from a dummy graph
+    const Drawable dummyDrawable(LinearFunction(Point(2), Point(1), Matrix(1, 2, Collection<Scalar>({1.0, 1.0}))).draw(xMin, xMax, pointNumber, scale).getDrawable(0));
```
regislebrun, line 3856: Why not using `SymbolicFunction(["x", "y"],["x+y"])`? It is both faster (x6.64 on my laptop) and more readable. And if you go for `SymbolicFunction(["x", "y"],["0"])` you get a 10% boost!

regislebrun (approved):
Note that the improvement done for drawPDF() can also be done for drawLogPDF() and drawCDF(), using the same trick.

lib/src/Uncertainty/Distribution/JointDistribution.cxx
```diff
@@ -514,6 +514,112 @@ Scalar JointDistribution::computePDF(const Point & point) const
...
+  const UnsignedInteger nX = x.getSize();
+  const UnsignedInteger nY = y.getSize();
+  const Distribution mX(getMarginal(0));
+  const Distribution mY(getMarginal(1));
+  Sample pdfX(LogPDFWrapper(mX.getImplementation().get())(x));
+  Sample pdfY(LogPDFWrapper(mY.getImplementation().get())(y));
+  const Sample uSample(computeCDFMarginalGrid2D(x, y));
+  const Sample corePDF(core_.computeLogPDF(uSample));
+  UnsignedInteger index = 0;
+  Sample jointPDF(nX * nY, 1);
```
regislebrun, line 577: jointLogPDF

lib/src/Uncertainty/Distribution/JointDistribution.cxx
```diff
@@ -514,6 +514,112 @@ Scalar JointDistribution::computePDF(const Point & point) const
...
+Sample JointDistribution::computeLogPDFGrid2D(const Sample & x, const Sample & y) const
+{
+  const UnsignedInteger nX = x.getSize();
+  const UnsignedInteger nY = y.getSize();
+  const Distribution mX(getMarginal(0));
+  const Distribution mY(getMarginal(1));
+  Sample pdfX(LogPDFWrapper(mX.getImplementation().get())(x));
+  Sample pdfY(LogPDFWrapper(mY.getImplementation().get())(y));
+  const Sample uSample(computeCDFMarginalGrid2D(x, y));
+  const Sample corePDF(core_.computeLogPDF(uSample));
```
regislebrun, line 575: coreLogPDF

regislebrun (approved):

lib/src/Base/Func/SymbolicParserExprTk.cxx
```diff
@@ -228,15 +228,19 @@ SymbolicParserExprTk::ExpressionCollection SymbolicParserExprTk::allocateExpress
   symbol_table.add_function("besselY1", y1);
   for (UnsignedInteger inputIndex = 0; inputIndex < inputDimension; ++ inputIndex)
   {
+    const String varName(inputVariablesNames_[inputIndex]);
+    if (symbol_table.symbol_exists(varName))
+      throw InvalidArgumentException(HERE) << "Variable name '" << varName << "' clobbers an existing symbol";
     if (!symbol_table.add_variable(inputVariablesNames_[inputIndex], stack[inputIndex]))
```
ArashPartow, line 234: @jschueller 

Minor nit: Could `varName `be used here instead of `inputVariablesNames_[inputIndex]`

regislebrun:
I have put additional developments in my branch misc since you merged it into your branch (I haven't noticed the PR was merged). As all these new developments are merged into a unique commit, I don't know how to extract the part posterior to the merge of the PR in order to propose a new PR. Some important new features are proposed, in particular a rework of the update() method allowing for a much faster construction of PointConditionalDistribution instances when a simplified version exists. Could you please help me on this point?

jschueller:
@regislebrun done

lib/src/Uncertainty/Distribution/PointConditionalDistribution.cxx
```diff
@@ -331,8 +331,8 @@ void PointConditionalDistribution::update()
     // find a feasible starting point
     SobolSequence sequence(dimension);
     Point start;
-    const UnsignedInteger size = ResourceMap::GetAsUnsignedInteger("PointConditionalDistribution-RatioUniformCandidateNumber");
-    for (UnsignedInteger i = 0; i < size; ++ i)
+    const UnsignedInteger candidateNumber = ResourceMap::GetAsUnsignedInteger("PointConditionalDistribution-RatioUniformCandidateNumber");
+    for (UnsignedInteger k = 0; k < candidateNumber; ++ k)
```
regislebrun, line 335: And what happen if no candidate with postive PDF is found after the fixed number of tries? No exception? It is why I opted for a while loop. It all depends on if we want to insure that the RNG is initialized, even if it takes time, or if we accept to throw because the initialization is too costly. But if we don't throw the candidate is 100% sure to be bad after the loop!

lib/src/Uncertainty/Distribution/PointConditionalDistribution.cxx
```diff
@@ -368,24 +368,42 @@ void PointConditionalDistribution::update()
       problemVI.setMinimization(false);
       if (ub[i] > 0.0)
       {
+        // find a feasible starting point in [0, ub]
+        for (UnsignedInteger k = 0; k < candidateNumber; ++ k)
+        {
+          Point candidate(sequence.generate(1)[0]);
```
regislebrun, line 374: ??? Why not `sequence.generate()`?

lib/src/Uncertainty/Distribution/PointConditionalDistribution.cxx
```diff
@@ -368,24 +368,42 @@ void PointConditionalDistribution::update()
...
         algo.run();
         supV_[i] = std::exp(algo.getResult().getOptimalValue()[0]);
         LOGDEBUG(OSS() << "supV_[" << i << "]=" << supV_[i] << " v*=" << algo.getResult().getOptimalPoint());
       }
       if (lb[i] < 0.0)
       {
+        // find a feasible starting point in [lb, 0]
+        for (UnsignedInteger k = 0; k < candidateNumber; ++ k)
+        {
+          Point candidate(sequence.generate(1)[0]);
```
regislebrun, line 395: same here

regislebrun (changes requested):

lib/src/Uncertainty/Distribution/PointConditionalDistribution.cxx
```diff
@@ -222,19 +234,74 @@ class PointConditionalDistributionVBoundEvaluation : public EvaluationImplementa
...
+      endBlock += distributions[blockCounter].getDimension();
+    }
+    conditioningIndicesBlocks[blockCounter].add(conditioningIndex - beginBlock);
+    conditioningValuesBlocks[blockCounter].add(sortedConditioningPairs[conditioningCounter].second);
+  } // for conditioningCounter
+  // Now, decide what to do for each block
+  Collection<Distribution> newBlocks;
+  for (UnsignedInteger i = 0; i < distributions.getSize(); ++i)
+  {
+    // If the block is fully conditioned, skip it
```
mbaudin47, line 270: You can also code the opposite case: if there is no conditioning, you can directly set the block.

lib/src/Uncertainty/Distribution/UniformOverMesh.cxx
```diff
@@ -194,33 +194,32 @@ Scalar UniformOverMesh::computePDF(const Point & point) const
...
-  Scalar cdf = 0.0;
-  const Interval quadrant(Point(getDimension(), -SpecFunc::Infinity), point);
-  const Interval intersection(quadrant.intersect(getRange()));
+  if (interval.getDimension() != getDimension()) throw InvalidArgumentException(HERE) << "UniformOverMesh interval must have dimension " << getDimension() << ", got " << interval.getDimension();
+  Scalar probability = 0.0;
+  const Interval intersection(interval.intersect(getRange()));
   if (intersection.isEmpty())
-    cdf = 0.0;
+    probability = 0.0;
   else if (intersection == getRange())
```
regislebrun, line 205: This test will fail for distributions with unbounded range. The variable quadrant has been built using two points, so its bounds are finite. As a result intersection has finite bounds, and if range has infinite bounds the test will fail even if the numerical bounds are equal. Either quadrant is built using the finite flag of the lower bound, or the test is based on bounds only (my preference):
```
else if (intersection.getLowerBound() == getRange().getLowerBound() && intersection.getUpperBound() == getRange().getUpperBound())
```

regislebrun (approved):
Only a minor improvement suggested, not mandatory.

lib/src/Uncertainty/Distribution/UniformOverMesh.cxx
jschueller: assuming you're referring to TruncatedOverMesh, this is ok since the range is also computed as an intersection with the mesh range, so a finite interval too, I just checked with the Normal from the test
for UniformOverMesh the range is just the mesh range of course

lib/src/Uncertainty/Distribution/PosteriorDistribution.cxx
```diff
@@ -107,7 +107,7 @@ Point PosteriorDistribution::computeLikelihood(const Point & theta) const
 Scalar PosteriorDistribution::computeLogLikelihood(const Point & theta) const
```
regislebrun, line 107: Here theta should be renamed y to match the notations of the reference document. Theta is the image of y by the link function, and is also the name of the parameter of the conditioned distribution.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -590,8 +590,10 @@ Distribution TruncatedDistribution::getMarginal(const Indices & indices) const
   if (useSimplifiedVersion_)
     return simplifiedVersion_.getMarginal(indices);
 
-  Interval marginalBounds(bounds_.getMarginal(indices));
-  return new TruncatedDistribution(distribution_.getMarginal(indices), marginalBounds);
+  if (distribution_.hasIndependentCopula())
```
regislebrun, line 593: This case is already taken into account in the simplification mechanism

lib/src/Uncertainty/Distribution/PosteriorDistribution.cxx
regislebrun: Other point: in computePDF(), we compute the log-pdf before to take its exponential. It would be more logical to implement computeLogPDF() then to call it from computePDF()

regislebrun (approved):
Nice to have added the fix for the TruncatedDistribution. Only minor remarks, nothing critical.

lib/src/Uncertainty/Distribution/PosteriorDistribution.cxx
jschueller: done, theta=f(y)

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: there is no simplification for general n-d distribution

jschueller:
@regislebrun I added a few commits for TruncatedNormal and NormalCopula, I hope I dont discover anymore bugs :]

lib/src/Uncertainty/Distribution/Normal.cxx
```diff
@@ -653,7 +653,7 @@ Point Normal::computeSequentialConditionalQuantile(const Point & q) const
       result[i] = mean_[i] + sigma_[i] * DistFunc::qNormal(q[i]);
     return result;
   }
-  return mean_ + inverseCholesky_.solveLinearSystem(DistFunc::qNormal(q));
+  return EllipticalDistribution::computeSequentialConditionalQuantile(q);
```
regislebrun, line 656: I don't understand what has motivated this change. Using the following script:
```python
import openturns as ot

# Build a multivariate Normal distribution
R = ot.CorrelationMatrix(3, [1.0, 0.5, 0.2, 0.5, 1.0, 0.3, 0.2, 0.3, 1.0])
mu = [1.0, 2.0, 3.0]
sigma = [2.0, 1.5, 3.0]
distribution = ot.Normal(mu, sigma, R)
# Extract its iso-proba tranformation and its inverse
T = distribution.getIsoProbabilisticTransformation()
invT = distribution.getInverseIsoProbabilisticTransformation()
# Build the corresponding Rosenblatt transformations
Ros = ot.Function(ot.RosenblattEvaluation(distribution))
invRos = ot.Function(ot.InverseRosenblattEvaluation(distribution))
# Check if the composition is the identity
# First, the Nataf version
sample = distribution.getSample(100)
res = invT(T(sample))
print("err Nataf=", (res - sample).computeRawMoment(2))
# Second, the Rosenblatt version
res = invRos(Ros(sample))
print("err Ros=", (res - sample).computeRawMoment(2))
# Third, the Rosenblatt/Nataf mix
res = invRos(T(sample))
print("err Ros/Nat=", (res - sample).computeRawMoment(2))
# Fourth, the Nataf/Rosenblatt mix
res = invT(Ros(sample))
print("err Nat/Ros=", (res - sample).computeRawMoment(2))
```
I gat with the current master:
```
err Nataf= [0,2.66241e-32,1.72563e-32]
err Ros= [2.55036e-30,8.69621e-30,3.15051e-30]
err Ros/Nat= [2.5103e-30,8.67994e-30,3.04993e-30]
err Nat/Ros= [2.5103e-30,8.67402e-30,3.1091e-30]
```
as expected, and with your branch:
```
err Nataf= [0,2.66241e-32,1.72563e-32]
err Ros= [2.55036e-30,0.00983886,0.125701]
err Ros/Nat= [2.5103e-30,0.00983886,0.125701]
err Nat/Ros= [2.5103e-30,8.67402e-30,3.1091e-30]
```
which shows a problem...

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -172,17 +172,52 @@ TruncatedDistribution * TruncatedDistribution::clone() const
...
+Distribution TruncatedDistribution::dispatchTruncation(const Collection<Distribution> & distributions) const
+{
+  const UnsignedInteger size = distributions.getSize();
+  Collection<Distribution> newBlocks(size);
+  UnsignedInteger startIndex = 0;
+  for (UnsignedInteger i = 0; i < size; ++ i)
+  {
+    Indices blockIndices(distributions[i].getDimension());
+    blockIndices.fill(startIndex);
+    newBlocks[i] = TruncatedDistribution(distributions[i], bounds_.getMarginal(blockIndices));
```
regislebrun, line 184: Here we could store the simplified version of the marginal as we do in PointConditionalDistribution

regislebrun (changes requested):
There is something wrong with your fix or with my understanding of the problem. IMO we should take the time to fix it, not in a hurry.

lib/src/Uncertainty/Algorithm/Transformation/NatafEllipticalCopula/NatafEllipticalCopulaEvaluation.cxx
```diff
@@ -112,29 +111,32 @@ Matrix NatafEllipticalCopulaEvaluation::parameterGradient(const Point & ) const
...
 /* Method load() reloads the object from the StorageManager */
 void NatafEllipticalCopulaEvaluation::load(Advocate & adv)
 {
   EvaluationImplementation::load(adv);
   adv.loadAttribute( "standardDistribution_", standardDistribution_ );
-  adv.loadAttribute( "cholesky_", cholesky_ );
+  if (adv.hasAttribute("inverseCholesky_"))
+    adv.loadAttribute("inverseCholesky_", inverseCholesky_);
+  else
+    adv.loadAttribute("cholesky_", inverseCholesky_); // OT < 1.24
```
regislebrun, line 139: Shouldn't we inverse the cholesky factor stored in the old versions?

jschueller: only the attribute name was wrong: it was the inverse all along!
see NormalCopula::getIsoProbabilisticTransformation

Issue #2839, opened 2024-11-29
Unable to build without MUPARSER and without EXPRTK 


mwestphal:
### What happened?

Building with `-DUSE_MUPARSER=OFF -DUSE_EXPRTK=OFF` result in a failing build:

```
/home/glow/dev/paraview/deps/openturns/src/lib/src/Base/Func/SymbolicFunction.cxx: In member function ‘OT::Description OT::SymbolicFun
/home/glow/dev/paraview/deps/openturns/src/lib/src/Base/Func/SymbolicFunction.cxx:147:23: error: ‘SymbolicEvaluation’ does not name a 
  147 |   return dynamic_cast<SymbolicEvaluation*>(getEvaluation().getImplementation().get())->getFormulas();
      |                       ^~~~~~~~~~~~~~~~~~
      |                       SymbolicFunction
/home/glow/dev/paraview/deps/openturns/src/lib/src/Base/Func/SymbolicFunction.cxx:147:41: error: expected ‘>’ before ‘*’ token
  147 |   return dynamic_cast<SymbolicEvaluation*>(getEvaluation().getImplementation().get())->getFormulas();
      |                                         ^
/home/glow/dev/paraview/deps/openturns/src/lib/src/Base/Func/SymbolicFunction.cxx:147:41: error: expected ‘(’ before ‘*’ token
  147 |   return dynamic_cast<SymbolicEvaluation*>(getEvaluation().getImplementation().get())->getFormulas();
      |                                         ^
      |                                         (
/home/glow/dev/paraview/deps/openturns/src/lib/src/Base/Func/SymbolicFunction.cxx:147:42: error: expected primary-expression before ‘>
  147 |   return dynamic_cast<SymbolicEvaluation*>(getEvaluation().getImplementation().get())->getFormulas();
      |                                          ^
/home/glow/dev/paraview/deps/openturns/src/lib/src/Base/Func/SymbolicFunction.cxx:147:88: error: ‘class OT::EvaluationImplementation’ 
  147 |   return dynamic_cast<SymbolicEvaluation*>(getEvaluation().getImplementation().get())->getFormulas();
      |                                                                                        ^~~~~~~~~~~
/home/glow/dev/paraview/deps/openturns/src/lib/src/Base/Func/SymbolicFunction.cxx:147:101: error: expected ‘)’ before ‘;’ token
  147 |   return dynamic_cast<SymbolicEvaluation*>(getEvaluation().getImplementation().get())->getFormulas();
[...]

PR #2840, opened 2024-11-29
CMake: Fail if there is no analytical parser

mwestphal:
Fix https://github.com/openturns/openturns/issues/2839


CMakeLists.txt
```diff
@@ -219,6 +219,9 @@ if (USE_EXPRTK)
   set (OPENTURNS_HAVE_ANALYTICAL_PARSER TRUE)
   set (SYMBOLICPARSER_DEFAULT_BACKEND ExprTk)
 endif ()
+if (NOT OPENTURNS_HAVE_ANALYTICAL_PARSER)
+  message(FATAL_ERROR "Analytical paser required, please enable either USE_MUPARSER or USE_EXPRTK")
```
mwestphal, line 223:
```suggestion
  message(FATAL_ERROR "Analytical parser required, please enable either USE_MUPARSER or USE_EXPRTK")
```

jschueller (approved):

regislebrun (approved):

regislebrun:
@jschueller I was pretty sure that ExprTk was always available (as we include its unique source file in our source tree) and that MuParser was an additional capability. As I am not very fluent in CMake, could you please tell me if it is possible to have both parsers or if we have to choose at compile time which parser will be used?

mwestphal:
> @jschueller I was pretty sure that ExprTk was always available (as we include its unique source file in our source tree) and that MuParser was an additional capability. As I am not very fluent in CMake, could you please tell me if it is possible to have both parsers or if we have to choose at compile time which parser will be used?

That's definitely possible to implement such logic. I could use muparser when `USE_MUPARSER=ON` and fallback to exprtk when `USE_MUPARSER=OFF`.

regislebrun:
> > @jschueller I was pretty sure that ExprTk was always available (as we include its unique source file in our source tree) and that MuParser was an additional capability. As I am not very fluent in CMake, could you please tell me if it is possible to have both parsers or if we have to choose at compile time which parser will be used?
> 
> That's definitely possible to implement such logic. I could use muparser when `USE_MUPARSER=ON` and fallback to exprtk when `USE_MUPARSER=OFF`.

It would be a perfect solution. BTW I assumed that at least one parser was available in several places in the code, e.g in PosteriorDistribution when I compute the moments by integration. It would be a pity to only have a conditional access to these computations despite the inclusion of ExprTk source code!

PR #2854, opened 2024-12-20
Update plot_sensitivity_sobol_multivariate.py

### LINKED ISSUE #2569: Plotting graphic missing in Example file
### What happened?

Hello Openturn !

I find a missing one or two lines which is not a big bug, don't worry.

I am implementing a Sobol analysis for my project, and when looking at the example  [here](https://openturns.github.io/openturns/latest/auto_reliability_sensitivity/sensitivity_analysis/plot_sensitivity_sobol_multivariate.html), I found that running the script  (.py one) provided the in page doesn't yield the graph as in the webpage. By looking around I found that you just need to add two lines, in the correct place.
`
from matplotlib import pylab as plt
plt.show()
`

So nothing alarming, but I guess if one want to reproduce directly the result he would be surprised. (But may be it is intended I don't know)

```shell
Donwload the file and run it, one would not see the final graph appear.
```

### Version

1.22

### Operating System

Windows

### Installation media

pip

### Additional Context

_No response_

m-balesdent:
This PR closes #2569 by adding `viewer.View.ShowAll()`

### DIFF
--- a/python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_sobol_multivariate.py
+++ b/python/doc/examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_sobol_multivariate.py
@@ -102,6 +102,7 @@ print("Agg. total order indices: ", agg_total_order)
 # %%
 graph = sensitivityAnalysis.draw()
 view = viewer.View(graph)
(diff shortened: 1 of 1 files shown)


jschueller:
this is not enough to close #2569 at many other examples are affected

Issue #2886, opened 2025-02-18
Changing the indices of inputs in a python function changes the result of `LinearLeastSquaresCalibration`


bmazurie:
### What happened?

hi,
I was comparing different calibration methods when I encountered a behavior that I could not understand. 

I was using the Chaboche model to compare calibration exactely as it is done in [OpenTURNS documentation](http://openturns.github.io/openturns/latest/auto_calibration/least_squares_and_gaussian_calibration/plot_calibration_chaboche.html?highlight=chaboche). However little twist, for my use case I am using Python function, thus I coded the [Chaboche mechanical model](http://openturns.github.io/openturns/latest/usecases/use_case_chaboche.html#use-case-chaboche) inside an `ot.PythonFunction`.

The issue I had : 
I defined two python functions, the first one (let's call it `exec_ot_sorted`) with the same order for the inputs as it is in the documentation : `Strain,R,C,Gamma`. In the second function (let's call it `exec_ot_unsorted`) I changed the names and the order : `C,R,gamma,strain_epsilon`. 
I used the jupyter-notebook from the documentation. For the first function, I kept `calibratedIndices = [1, 2, 3]`. For the second function, I changed to `calibratedIndices = [1, 0, 2]`. 

I then run `mycf = ot.ParametricFunction(exec_ot, calibratedIndices, thetaPrior)` with `exec_ot` defined above, and `algo = ot.LinearLeastSquaresCalibration(mycf, observedStrain, observedStress, thetaPrior, "SVD")` for each case.

The result of `LinearLeastSquaresCalibration` for `exec_ot_sorted` function is the same as the one from the documentation.:
![Image](https://github.com/user-attachments/assets/0f928993-6e1b-49de-94ca-72652222b02d)

 But the results are different for `exec_ot_unsorted`:

![Image](https://github.com/user-attachments/assets/9e14a85c-6fa6-4c1f-994c-723b9473cfdc)

Please see the jupyter notebooks attached (one for sorted and one for unsorted, function `exec_ot` for Chaboche model is defined at the beginning)

[plot_calibration_chaboche_ChabochePython.zip](https://github.com/user-attachments/files/18848341/plot_calibration_chaboche_ChabochePython.zip)

**Did I forget to change something in the notebook for the unsorted function?** 

**Is there any mathematical reason for this behavior?** 

### How to reproduce the issue?

```shell
[...]

Issue #2758, opened 2024-09-08
RatioDistribution is undocumented, untested and unavailable in Python


josephmure:
### What happened?

Apparently it is supposed to be the distribution of the random variable $XY$ when $X$ and $Y$ are independent scalar random variables following potentially different distributions.

### How to reproduce the issue?

```shell
Everything is in the title.
```

### Version

1.24dev

### Operating System

all

### Installation media

unknown

jschueller:
you mean the distribution of X/Y
actually its unused in the c++ side too as Distribution::operator/ uses the ProductDistribution of X * (1/Y) with 1/Y as a CompositeFunction with the inverse function
lets ask the author if it can de dropped

josephmure:
> you mean the distribution of X/Y

No actually: see the code

https://github.com/openturns/openturns/blob/05589f5dd80820e1849efb3280a3469c60f215cb/lib/src/Uncertainty/Distribution/RatioDistribution.cxx#L123

PR #2887, opened 2025-02-19
Misc

jschueller:
- #2886 
- #2758 
- #2905
- Check LOLA input sample components are not constant
- Fixes to make some examples reproducible


lib/src/Uncertainty/Distribution/Poisson.cxx
```diff
@@ -140,19 +140,17 @@ Scalar Poisson::computeLogPDF(const Point & point) const
 Scalar Poisson::computeCDF(const Point & point) const
 {
   if (point.getDimension() != 1) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=1, here dimension=" << point.getDimension();
-
   const Scalar k = point[0];
   if (k < -supportEpsilon_) return 0.0;
-  return DistFunc::pGamma(floor(k) + 1.0, lambda_, true);
+  return DistFunc::pPoisson(lambda_, k);
```
mbaudin47, line 145: This may reveal that pGamma has a bug.

mbaudin47 (changes requested):
The bugfix does not help to understand why the bug appeared.

lib/src/Uncertainty/Distribution/Poisson.cxx
jschueller: moving to #2912

sofianehaddad (approved):

PR #2881, opened 2025-02-05
Update sensitivity_sobol_from_pce.rst

mbaudin47:
Fixed typos.

PR #2888, opened 2025-02-21
Update HSICEstimatorConditionalSensitivity_doc.i.in

mbaudin47:
Typo.


jschueller:
could you move this #2881 for example ?

Issue #2890, opened 2025-03-03
calibration MAP outside CI


jschueller:
### What happened?

MAP is outside CI for this gaussian/nonlinear calibration with 2 outputs, no booststrap

### How to reproduce the issue?

```shell
import math as m
import openturns as ot


def _exec(X):
    wab, wcd, t, P, L, rho, g, a, b = X
    wab = wab / 1000
    wcd = wcd / 1000
    t = t / 1000
    L = L / 1000
    P = P * 1000
    MB = 1 / 3 * (P * L) + (rho * g * wcd * t * (L**a)) / 18
    Masse = rho * t * L * (2 / (3 * m.sin(b)) * wab + wcd)
    return [MB, Masse]


prior = ot.JointDistribution([ot.Normal(0.1, 0.01)] * 3)
model = ot.PythonFunction(9, 2, _exec)


# model = ot.SymbolicFunction(["wab", "wcd", "t", "P", "L", "rho", "g", "a", "b"], ["1 / 3 * (P/1000 * L/1000) + (rho * g * wcd/1000 * t/1000 * ((L/1000)^a)) / 18", "rho * t/1000 * L/1000 * (2 / (3 * sin(b)) * wab/1000 + wcd/1000)"])
model_p = ot.ParametricFunction(model, [6, 7, 8], prior.getMean())

obs_sample = ot.Sample.ImportFromCSVFile("t_Calibration_console.csv", ",")
obs_in = obs_sample.getMarginal([0, 1, 2, 3, 4, 5])
obs_out = obs_sample.getMarginal([10, 11])
err_cov = ot.CovarianceMatrix(2)
for i in range(2):
    err_cov[i, i] = 0.1**2
p0cov = prior.getCovariance()
p0 = prior.getMean()
algo = ot.GaussianNonLinearCalibration(model_p, obs_in, obs_out, p0, p0cov, err_cov)
# algo = ot.NonLinearLeastSquaresCalibration(model_p, obs_in, obs_out, p0)

problem = ot.LeastSquaresProblem()
names = ot.OptimizationAlgorithm.Build(problem)

names = []
if ot.PlatformInfo.HasFeature("cminpack"):
    names.append("CMinpack")
if ot.PlatformInfo.HasFeature("ceres"):
    names.append("LEVENBERG_MARQUARDT")

for name in names:
    solver = ot.OptimizationAlgorithm.GetByName(name)
    algo.setOptimizationAlgorithm(solver)
[...]

PR #2892, opened 2025-03-04
Fix nonlinear leastsquares and Gaussian calibration

regislebrun:
The posterior distribution was not centered at the MAP when using the Laplace approximation. It was due to the fact that a new linear least squares or linear Gaussian calibration was triggered and the posterior distribution was built using this new MAP.

See #2890 for the details.

### DIFF
--- a/lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
+++ b/lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
@@ -364,10 +364,63 @@ void GaussianNonLinearCalibration::run()
     parameterPosterior = KernelSmoothing().build(thetaSample);
   }
   else
-  {
-    GaussianLinearCalibration algo(model_, inputObservations_, outputObservations_, thetaStar, getParameterPrior().getCovariance(), error.getCovariance());
-    algo.run();
-    parameterPosterior = algo.getResult().getParameterPosterior();
+    {
+      // Compute the linearization
+      Function parametrizedModel(model_);
+      parametrizedModel.setParameter(thetaStar);
+      // Flatten everything related to the model evaluations over the input observations
+      const UnsignedInteger parameterDimension = getParameterPrior().getDimension();
+      const UnsignedInteger outputDimension = getOutputObservations().getDimension();
+      const UnsignedInteger size = getOutputObservations().getSize();
+      Matrix gradientObservations = MatrixImplementation(parameterDimension, size * outputDimension);
+      UnsignedInteger shift = 0;
+      UnsignedInteger skip = parameterDimension * outputDimension;
+      for (UnsignedInteger i = 0; i < size; ++i)
+	{
+	  const Matrix parameterGradient(parametrizedModel.parameterGradient(getInputObservations()[i]));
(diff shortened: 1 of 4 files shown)


jschueller:
we should add the script from #2890

lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
```diff
@@ -364,10 +364,63 @@ void GaussianNonLinearCalibration::run()
...
   else
-  {
-    GaussianLinearCalibration algo(model_, inputObservations_, outputObservations_, thetaStar, getParameterPrior().getCovariance(), error.getCovariance());
-    algo.run();
-    parameterPosterior = algo.getResult().getParameterPosterior();
+    {
+      // Compute the linearization
+      Function parametrizedModel(model_);
+      parametrizedModel.setParameter(thetaStar);
+      // Flatten everything related to the model evaluations over the input observations
```
mbaudin47, line 371: Is this a copy of this code from lines 105 to 173?https://github.com/openturns/openturns/blob/07a9b2174d4cfbf8382df26ed663e3402558299e/lib/src/Uncertainty/Bayesian/GaussianLinearCalibration.cxx#L105
If yes, then it is mathematically correct, but the code can be factored. Otherwise I do not exactly understand what is done.

mbaudin47 (changes requested):

lib/src/Uncertainty/Bayesian/NonLinearLeastSquaresCalibration.cxx
```diff
@@ -287,9 +287,30 @@ void NonLinearLeastSquaresCalibration::run()
...
   else
   {
-    LinearLeastSquaresCalibration linearAlgo(model_, inputObservations_, outputObservations_, thetaStar);
-    linearAlgo.run();
-    parameterPosterior = linearAlgo.getResult().getParameterPosterior();
+    // Prepare the inverse Gram
+    // Compute the linearization
+    Function parametrizedModel(model_);
+    parametrizedModel.setParameter(thetaStar);
+    // Flatten everything related to the model evaluations over the input observations
```
mbaudin47, line 294: Isn't this a copy of the next code from lines 92 to 121 (with the exception of the diagonal regularization)?
https://github.com/openturns/openturns/blob/07a9b2174d4cfbf8382df26ed663e3402558299e/lib/src/Uncertainty/Bayesian/LinearLeastSquaresCalibration.cxx#L92
If yes, then the code can be factored, including the regularization.

mbaudin47 (changes requested):
The code seems to be mathematically correct, but the implementation could be improved by factoring the calculation of the covariance matrices in the linear classes, and used in the nonlinear classes.

lib/src/Uncertainty/Bayesian/GaussianNonLinearCalibration.cxx
regislebrun: yes it is. For now, it is more a quick and dirty fix, only to check if it is what we expect from a mathematical point of view and if the implementation is mathematically correct. I postpone the refactoring of the code to after our next technical committee. The idea is to factor things between GaussianNonLinearCalibration and GaussianLinearCalibration the way I did it between NonLinearLeastSquaresCalibration and LinearLestSquaresCalibration, but as it is more involved I prefer to discuss it before to go for a mess.

jschueller:
moved to #2912

PR #2895, opened 2025-03-18
Modification of DrawPairsMarginal

### LINKED ISSUE #2894: DrawPairsMarginals should use the same axes on all projections
### What is the idea?

The DrawPairsMarginals plots the marginal distribution on the diagonal and the (i, j)-projection on the extra-diagonals of a Grid. But the X axes limits should be made consistent on each column, so that the plots match.

Consider for example a distribution made of 3 independent uniform distributions. 

```python
import openturns as ot

dimension = 3
multi_uniform = ot.JointDistribution([ot.Uniform() for i in range(dimension)])
sample = multi_uniform.getSample(100)
grid = ot.VisualTest.DrawPairsMarginals(sample , multi_uniform)
view = otv.View(grid, figure_kw={"figsize": (6.0, 6.0)})
```

The previous script produces the plot:

Image

We see that the X interval of each projection is not consistent with the X marginal interval. There can be two different cases:
- the X range of the marginal distribution can be smaller than the X rang of the marginal projection,
- the X range of the marginal distribution can be larger than the X rang of the marginal projection.

Whatever the case, the X ranges on a single column should match, so that the distributions and the clouds can correspond.

 

### Why is this needed?

_No response_

### Additional Context

_No response_

m-balesdent:
This PR closes #2894 

A little modification has been added in `DrawPairsMarginal` of the `VisualTest` class to ensure consistency on the x-axes between the `drawPDF()`  graphs in the diagonal  and the pairplots outside the diagonal.

Here is a before / after example with : 
```python
import openturns as ot

ot.RandomGenerator.SetSeed(1)
multi_uniform = ot.JointDistribution([ot.Normal(),ot.Triangular(0.2,0.3,0.5), ot.Triangular(),ot.Uniform()])
sample = multi_uniform.getSample(500)
grid = ot.VisualTest.DrawPairsMarginals(sample , multi_uniform)
view = otv.View(grid, figure_kw={"figsize": (6.0, 6.0)})
```

Before : 

image

After : 

image


jschueller (approved):
nice

regislebrun (approved):

lib/src/Uncertainty/StatTests/VisualTest.cxx
```diff
@@ -343,7 +343,7 @@ GridLayout VisualTest::DrawPairsMarginals(const Sample & sample, const Distribut
   const Description description(sample.getDescription());
   for (UnsignedInteger i = 0; i < dimension; ++ i)
   {
-    Graph pdfGraph(distribution.getMarginal(i).drawPDF());
+    Graph pdfGraph(distribution.getMarginal(i).drawPDF(sample.getMarginal(i).getMin()[0], sample.getMarginal(i).getMax()[0]));
```
mbaudin47, line 346: I would rather use a small margin to ensure that the data is safely shown.

mbaudin47 (changes requested):
The code might be improved to prevent the edges from being hidden.

lib/src/Uncertainty/StatTests/VisualTest.cxx
regislebrun: or maybe the other way: as drawPDF() produces a quite nice graph, we could use it to present the clouds.

jschueller: I think the clouds would be quite narrow if you look at the image from the issue

m-balesdent: Thanks for your comments. I tried to implement your solution in C++.

This is the result

image

lib/src/Uncertainty/StatTests/VisualTest.cxx
```diff
@@ -353,6 +364,19 @@ GridLayout VisualTest::DrawPairsMarginals(const Sample & sample, const Distribut
       const Indices indices = {j, i};
       const Cloud cloud(sample.getMarginal(indices), "blue", "fsquare", "");
       Graph graph("", i == dimension - 1 ? description[j] : "", j == 0 ? description[i] : "", true, "topright");
+      
+      Point minRange(2,0.0);
+      minRange[0] = axesMin[j];
+      minRange[1] = axesMin[i];
```
jschueller, line 370: I suggest `const Point minRange = {axesMin[j], axesMin[i]};`

lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -1706,7 +1706,8 @@ void ResourceMap::loadDefaultConfiguration()
   addAsUnsignedInteger("VisualTest-CloudMediumSize", 1000);
   addAsUnsignedInteger("VisualTest-KendallPlot-MonteCarloSize", 100);
   addAsScalar("VisualTest-DependenceConfidenceLevel", 0.95);
-
+  addAsScalar("VisualTest-DrawPairsMarginals-AxesMargin", 0.10);
```
jschueller, line 1709: should be added in openturns.conf.in too

lib/src/Uncertainty/StatTests/VisualTest.cxx
```diff
@@ -334,16 +334,27 @@ GridLayout VisualTest::DrawPairsXY(const Sample & sampleX, const Sample & sample
...
+
   const UnsignedInteger dimension = sample.getDimension();
   if (dimension < 2)
     throw InvalidDimensionException(HERE) << "Can only draw clouds from a multivariate sample";
   if (distribution.getDimension() != dimension)
     throw InvalidDimensionException(HERE) << "Distribution dimension does not match the sample dimension";
   GridLayout grid(dimension, dimension);
   const Description description(sample.getDescription());
+  
+  const Scalar xy_Margin = ResourceMap::GetAsScalar("VisualTest-DrawPairsMarginals-AxesMargin");
```
jschueller, line 346: xyMargin ; dont mix snakes & camels :)

mbaudin47 (approved):

lib/src/Uncertainty/StatTests/VisualTest.cxx
```diff
@@ -353,6 +364,13 @@ GridLayout VisualTest::DrawPairsMarginals(const Sample & sample, const Distribut
       const Indices indices = {j, i};
       const Cloud cloud(sample.getMarginal(indices), "blue", "fsquare", "");
```
mbaudin47, line 365:
```suggestion
      const Cloud cloud(sample.getMarginal(indices));
```

mbaudin47 (changes requested):

mbaudin47:
Suggestion: Plot the iso-values of the PDF on the upper right graphs.

m-balesdent:
> Suggestion: Plot the iso-values of the PDF on the upper right graphs. (Do it if you want to only. This can be done later during the Hackaton if needed.)

I tried but the handling of colorbars of 2D PDF is a little bit tricky.

lib/src/Uncertainty/StatTests/VisualTest.cxx
m-balesdent: This is new graph :

image

lib/src/Uncertainty/StatTests/VisualTest.cxx
```diff
@@ -334,29 +334,47 @@ GridLayout VisualTest::DrawPairsXY(const Sample & sampleX, const Sample & sample
...
   if (dimension < 2)
     throw InvalidDimensionException(HERE) << "Can only draw clouds from a multivariate sample";
   if (distribution.getDimension() != dimension)
     throw InvalidDimensionException(HERE) << "Distribution dimension does not match the sample dimension";
   GridLayout grid(dimension, dimension);
   const Description description(sample.getDescription());
+  
+  const Scalar axesMargin = ResourceMap::GetAsScalar("VisualTest-DrawPairsMarginals-AxesMargin");
+  
+  const Point sampleMin = sample.getMin();
```
regislebrun, line 348: Use `const Point sampleMin(sample.getMin());` as `Point` is a complex type. See the section named "Assignment and initialization" here https://openturns.github.io/openturns/latest/developer_guide/coding_rules.html

lib/src/Uncertainty/StatTests/VisualTest.cxx
```diff
@@ -334,29 +334,47 @@ GridLayout VisualTest::DrawPairsXY(const Sample & sampleX, const Sample & sample
...
+  const Point sampleMax = sample.getMax();
```
regislebrun, line 349: same here and for the three following lines

lib/src/Uncertainty/StatTests/VisualTest.cxx
```diff
@@ -334,29 +334,47 @@ GridLayout VisualTest::DrawPairsXY(const Sample & sampleX, const Sample & sample
...
     pdfGraph.setXTitle(i == dimension - 1 ? sample.getDescription()[i] : "");
     grid.setGraph(i, i, pdfGraph);
     for (UnsignedInteger j = 0; j < i; ++ j)
     {
+
       const Indices indices = {j, i};
-      const Cloud cloud(sample.getMarginal(indices), "blue", "fsquare", "");
+      const Cloud cloud(sample.getMarginal(indices));
       Graph graph("", i == dimension - 1 ? description[j] : "", j == 0 ? description[i] : "", true, "topright");
+      const Point minRange = {axesMin[j], axesMin[i]};
```
regislebrun, line 368: same here and for the next two lines

regislebrun (approved):
Only cosmetic changes to do

mbaudin47 (approved):

regislebrun (approved):

Issue #2915, opened 2025-03-31
A getting started example would help new users


mbaudin47:
### Describe the issue

The idea is to consider Scikit's [getting_started.html](https://scikit-learn.org/stable/getting_started.html) as a framework to help new user to start with OpenTURNS.

- The example would be available on the front page, at https://openturns.github.io/www/. 

![Image](https://github.com/user-attachments/assets/b1e4dde3-17a6-47e5-b2c9-6ab80224262d)

- The example would also be available as the first example in the [Examples](https://openturns.github.io/openturns/latest/examples/examples.html) section.

![Image](https://github.com/user-attachments/assets/468e8ac0-2481-4862-a252-3b8b683a62e9)

- The example would require no more than 2 pages i.e. between 1500 and 2000 words. Half of the text would be a script and half would be words.
- We may cover the following methods of the ABC methodology: probabilistic model and inference (step B), propagation (step C), sensitivity analysis (step C'), metamodel.
- The step B would be based on [Exercice-cantilever-beam-fit-distribution.ipynb](https://github.com/mbaudin47/otsupgalilee-eleve/blob/master/2-Quantification/Exercice-cantilever-beam-fit-distribution.ipynb). This section would use the `BetaFactory`. This requires to access a dataset, for example in the cantilever beam use case. It would be accessed through a new `data` attribute. The `drawQQPlot` graph would be shown.
- The step C requires to define a physical model. We could define a `PythonFunction` which explicitly implements the cantilever beam function.
- The step C would use `CompositeRandomVector` and its `getSample()` method. This produces a Monte-Carlo sample. Consider this example [Exercice-Tendance-centrale-cantilever-beam.ipynb](https://github.com/mbaudin47/otsupgalilee-eleve/blob/master/3-Propagation/Exercice-Tendance-centrale-cantilever-beam.ipynb) as a starting point.
- The step C' would use Sobol' indices based on `SobolSimulationAlgorithm`. A starting point for this is [Exercice-Sensibilite-Sobol-cantilever-beam.ipynb](https://github.com/mbaudin47/otsupgalilee-eleve/blob/master/4-Sensibilite/Exercice-Sensibilite-Sobol-cantilever-beam.ipynb)
- The metamodel would be a `LeastSquaresExpansion`. The `MetaModelValidation` graph would be shown.

### Where can the issue be found?

In the doc.

mbaudin47:
@jschueller , @sofianehaddad : Does this roadmap makes sense to you?

[...]

PR #2922, opened 2025-04-04
Misc / doc

jschueller:
- #2915 
- Rename kriging to GPR in theory/usermanual sections


sofianehaddad (approved):

python/doc/examples/getting_started/plot_getting_started.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+# %%
+# Propagate the input distribution through the model
+sample_x = estimated_distribution.getSample(1000)
+sample_y = model(sample_x)
+
+# %%
+# Estimate a non-parametric distribution of the output variable
+histogram = ot.HistogramFactory().build(sample_y)
```
mbaudin47, line 84: This may seem uninteresting to users that may not make understand the difference between drawing the histogram's PDF and computing the histogram. Hence, some readers may not see the difference with [matplotlib.pyplot.hist](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist.html) or [seaborn.kdeplot.html](https://seaborn.pydata.org/generated/seaborn.kdeplot.html). A clarification may be done by inserting an appropriate comment. This can also be done by drawing another plot, e.g. the CDF or printing the list of bins. Perhaps a link to [kernel_smoothing.html](https://openturns.github.io/openturns/latest/theory/data_analysis/kernel_smoothing.html) may help some users that do not know about this method.

python/doc/examples/getting_started/plot_getting_started.py
```diff
@@ -0,0 +1,121 @@
...
+    return [Y]
+
+
+model = ot.PythonFunction(4, 1, fpython)
+
+
+# %%
+# Propagate the input distribution through the model
+sample_x = estimated_distribution.getSample(1000)
+sample_y = model(sample_x)
```
mbaudin47, line 80: The interesting point to notice here is that there is no `for` loop. Moreover, the performance may be improved depending on how the function is implemented. See for example [PythonFunction.html](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.PythonFunction.html)'s `n_cpus` option.

python/doc/examples/getting_started/plot_getting_started.py
```diff
@@ -0,0 +1,121 @@
...
+estimated_copula, _ = ot.FittingTest.BestModelBIC(copula_sample, copula_factories)
+print(estimated_copula)
+
+# %%
+# Assemble the joint distribution from marginals and copula
+estimated_distribution = ot.JointDistribution(estimated_marginals, estimated_copula)
+print(estimated_distribution)
+
+# %%
+# For simplicity in the next steps we assume the distribution is independent
```
mbaudin47, line 56: It may be interesting for some users to know that a [NormalCopula.html](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.NormalCopula.html) can be used if need (without necessarily showing this in the script).

mbaudin47 (changes requested):
The introductory example may help some users: thank you for this interesting example. Inserting some appropriate links within the text may help some readers to understand why the library is adding value compared to other libraries.

python/doc/examples/getting_started/plot_getting_started.py
```diff
@@ -0,0 +1,123 @@
...
+test_y = model(test_x)
+predictions = metamodel(test_x)
+validation = ot.MetaModelValidation(test_y, predictions)
+graph = validation.drawValidation()
+_ = otv.View(graph)
+
+# %%
+# Sensitivity analysis
+# ----------------------
+N = 5000
```
sofianehaddad, line 108: Maybe for the `getting start` data driven methods are prefereable, to reduce the "complexity" for starters.
What do yoy think ?

python/doc/examples/getting_started/plot_getting_started.py
```diff
@@ -0,0 +1,123 @@
...
+# %%
+# Propagate the input distribution through the model
+# Here the function evaluation can benefit parallelization depending on its type, see also n_cpus option from :class:`~openturns.PythonFunction`.
+sample_x = distribution.getSample(1000)
+sample_y = model(sample_x)
+
+# %%
+# Estimate a non-parametric distribution (see :class:`~openturns.KernelSmoothing`) of the output variable
+ks = ot.KernelSmoothing().build(sample_y)
+graph = ks.drawCDF()
```
sofianehaddad, line 87: `pdf` on the left, `cdf` on the right ?
just to highlight we get easily all the needed services

python/doc/examples/getting_started/plot_getting_started.py
```diff
@@ -0,0 +1,123 @@
...
+_ = otv.View(graph)
+
+
+# %%
+# Build a metamodel
+# -----------------
+algo = ot.LeastSquaresExpansion(sample_x, sample_y, distribution)
```
sofianehaddad, line 94: why not, just a small description (1 line) directing to the class and what it does
`LeastSquaresExpansion` is not so explicit...

python/doc/examples/getting_started/plot_getting_started.py
jschueller: yes!

python/doc/examples/getting_started/plot_getting_started.py
jschueller: yes, lets mention pce

python/doc/examples/getting_started/plot_getting_started.py
jschueller: yes, we can use ranksobol

python/doc/examples/getting_started/plot_getting_started.py
```diff
@@ -0,0 +1,126 @@
...
+# -------------------------------------------------
+algo = ot.LeastSquaresExpansion(sample_x, sample_y, distribution)
+algo.run()
+metamodel_result = algo.getResult()
+metamodel = metamodel_result.getMetaModel()
+test_x = distribution.getSample(1000)
+test_y = model(test_x)
+predictions = metamodel(test_x)
+validation = ot.MetaModelValidation(test_y, predictions)
+graph = validation.drawValidation()
```
jschueller, line 105: add R2 score

jschueller:
merged @sofianehaddad

Issue #2908, opened 2025-03-25
DrawParallelCoordinates is broken


mbaudin47:
### What happened?

The axes of a classical parallel coordinate plot are colored. The new version has only black axes which makes the plot useless.

![Image](https://github.com/user-attachments/assets/73dc34e0-12b0-4c22-9302-70764525ebdd)

**Figure 1.** The output from DrawParallelCoordinates  in OT 1.21.

![Image](https://github.com/user-attachments/assets/ead0eee7-e942-424a-a114-67feae41fc73)

**Figure 2.** The output from DrawParallelCoordinates  in OT 1.22.

### How to reproduce the issue?

```shell
See https://openturns.github.io/openturns/1.24/auto_data_analysis/graphics/plot_sensitivity_par_coo_ishigami.html
```

### Version

1.24

### Operating System

unknown

### Installation media

unknown

Issue #2901, opened 2025-03-20
DistributionValidation should check more properties


mbaudin47:
### What is the idea?

The issue #2893 reveal that properties of the quantile are not checked. We should additionnaly check that, for a discrete distribution :

$$
F^{-1}(F(k)) = k
$$

for any $k \in \mathcal{S}(F)$ where $F$ is the cumulated distribution function and $\mathcal{S}(F)$ is the support of the distribution. 

For any distribution (be it continuous or discrete), if $x \in \mathbb{R}$, then:

$$
F^{-1}(F(x)) \leq x.
$$

If $x$ does not belong to the support of $F$, then:

$$
F^{-1}(F(x)) = F^{-1}\left(F \left(\max_{y \in S(F), y \leq x} y\right) \right)
$$

This must be checked for a well chosen collection of integers $k$. For any distribution (be it discrete or continuous), then we must check that:

$$
F(F^{-1}(p)) \geq p
$$

for any $p \in [0,1]$. For any continuous distribution, we must have:

$$
F(F^{-1}(p)) = p
$$

for any $p \in [0,1]$.

PR #2930, opened 2025-04-23
Misc / easter

jschueller:
- #2901
- #2908
- #2934
- #2937
- #2933
- Mark parallel flag of Distribution / Classifier public (also Function, CovModel), was missing in some distributions
- Fix Skellam/Bernoulli distributions quantiles/inverseSurvival
- Add VertexFieldToPointFunction which allows to filter a field function at some node, to be used with PointToPointConnection for example; it is != from the previous VertexFunction attempt I submitted
- Refactor tail dependence coefficients to allow distributions of dim>2 /cc @regislebrun 
- Move 1.20 classes out of experimental + BoundaryMesher from 1.22


sofianehaddad:
why do we want `isParallel` to be public?

jschueller:
In case you wanted to use external parallelization from python like ThreadPoolExecutor or something, is that it @regislebrun ?

jschueller:
@sofianehaddad @regislebrun I added a couple of more fixes plus the classes moved from experimental

jschueller:
hello @sofianehaddad do you agree with these changes ?

PR #2946, opened 2025-05-06
Doc: Add Quantile confidence result tabs

jschueller:
cc @adutfoy @mbaudin47

### DIFF
--- a/python/doc/theory/data_analysis/quantile_confidence_estimation.rst
+++ b/python/doc/theory/data_analysis/quantile_confidence_estimation.rst
@@ -19,8 +19,9 @@ and :math:`X_{(\sampleSize)} = \max (X_1, \dots, X_\sampleSize)` is the maximum.
     X_{(1)} \leq X_{(2)} \leq \dots \leq X_{(\sampleSize)}
 
 
-Smallest rank for an upper bound to the quantile
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


adutfoy:
Thanks for adding the tables!
I think you should remove the labels Case A, Case B and Case C unless you introduce them at the beginning of the doc.

jschueller:
@adutfoy done

adutfoy:
Screenshot_20250530_094936

- ... the unknown quantile $x_{\alpha}$ of level $\alpha$.
- ... using order statistics.
- ... $X_{(k)}$  is th $k$-th minimum of...

adutfoy:
Screenshot_20250530_095818


- Move the general notions before the first case.
- I think it is useless to mention (B) ib the equation (1)
- There are some mistakes in the table Case B: I propose to give the following table: 
Screenshot_20250530_103012

adutfoy:
Screenshot_20250530_110049

- you should add: the ranks $(k_1, k_2)$ such that .... with $(k_2-k_1)$ the smallest
- do not mention (C)
- you should mention the conditions when there is no solution and detail the solution in the trivial cases (if known)

adutfoy:
Screenshot_20250530_112440

- ... such that the equation (4) is satisfied and not (1)
- you did not mention some solutions 
- I propose the following table:
Screenshot_20250530_112800
where you replaces $f_{k,\alpha}(n)$ by its definition $F_{n, \alpha}(k-1)$

jschueller:
ok

adutfoy (approved):

PR #2949, opened 2025-05-07
EGO: Compatibility with new GPR API

jschueller:
now EGO has to go to experimental because of GPRResult class

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
+++ b/lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
@@ -26,6 +26,9 @@
 #include "openturns/MultiStart.hxx"
 #include "openturns/JointDistribution.hxx"
 #include "openturns/Uniform.hxx"
+#include "openturns/GaussianProcessFitter.hxx"
+#include "openturns/GaussianProcessRegression.hxx"
+#include "openturns/GaussianProcessConditionalCovariance.hxx"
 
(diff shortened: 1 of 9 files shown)


sofianehaddad:
Thanks for starting the job ! 

It was in my TODO list

lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
```diff
@@ -129,43 +152,47 @@ class ExpectedImprovementEvaluation : public EvaluationImplementation
...
+    outputSample = gprResult_.getOutputSample();
+  }
+  else
+  {
+    inputSample = krigingResult_.getInputSample();
+    outputSample = krigingResult_.getOutputSample();
+  }
   UnsignedInteger size = inputSample.getSize();
   Point noise(size);
   const Bool hasNoise = metamodelNoise_.getEvaluation().getImplementation()->isActualImplementation();
```
josephmure, line 198: In the GPR case, we could detect a positive nugget factor, which can represent a homoscedastic noise (equivalent to a constant noise function).

Something like this:
```suggestion
  const Bool hasNoise = gprResult_.getCovarianceModel().getNuggetFactor() > ResourceMap::GetAsScalar("CovarianceModel-DefaultNuggetFactor");
```

Note that the [only EGO example in the doc with a noisy function](https://openturns.github.io/openturns/master/auto_numerical_methods/optimization/plot_ego.html?highlight=noise+function) uses a constant noise function, so this example could easily be converted to the new API.

PR #2963, opened 2025-05-30
Translate some GeneralLinearModelAlgorithm & KrigingAlgorithm examples to their GaussianProcess[Fitter/Regression] counterparts

regislebrun:
Also fixed two points in the library:
  * the conditional covariance at a given point could be slightly negative on their diagonal due to roundoff. Thediagonal is now shifted to make all the values positive if needed.
  * the bounds for parameter optimization were updated according to an inspection of the active parameters based on their position. It has been changed in favour of the parameter names, which is more robust. The lower bound for nugget parameter has been set to zero


python/doc/examples/meta_modeling/kriging_metamodel/plot_gpr_advanced.py
```diff
@@ -0,0 +1,332 @@
+"""
+Advanced Gaussian process regression
+====================================
+"""
+
+# %%
+#
+# In this example we will build a metamodel using Gaussian process regression of the :math:`x\sin(x)` function.
+#
+# We will choose the number of learning points, the basis and the covariance model.
```
jschueller: I think we decided to directly move examples instead of adding new ones (ie the old ones should be removed)

regislebrun: @jschueller It looks like we decided to duplicate all the scripts if I understand well the minutes of our last meeting (I double checked with @adutfoy )

jschueller: I dont remember that was the final decision, because @josephmure @m-balesdent and I all just moved examples to the new api without creating duplicates.
We need another viewpoint, @sofianehaddad ?

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessFitter.cxx
```diff
@@ -136,37 +136,38 @@ void GaussianProcessFitter::setCovarianceModel(const CovarianceModel & covarianc
     Point lowerBound(optimizationDimension, ResourceMap::GetAsScalar( "GaussianProcessFitter-DefaultOptimizationLowerBound"));
     Point upperBound(optimizationDimension, ResourceMap::GetAsScalar( "GaussianProcessFitter-DefaultOptimizationUpperBound"));
     // We could set scale parameter if these parameters are enabled.
-    // check if scale is active
-    const Indices activeParameters(reducedCovarianceModel_.getActiveParameter());
-    Bool isScaleActive(true);
-    for (UnsignedInteger k = 0; k < reducedCovarianceModel_.getScale().getSize(); ++k)
+    // check if some scales ar active
```
josephmure:
```suggestion
    // check if some scales are active
```

regislebrun: fixed

adutfoy:
Some few comments: 

- I thought it was not useful anymore to write ot.ResourceMap.Reload()
- in the example 'Advanced GPR': please, move the python comments into the text zone....
- in the example 'GPFitter: configure the optimization solver', you left the comment 'Step 1: Run a first Kriging algorithm.'

python/doc/examples/meta_modeling/kriging_metamodel/plot_gpr_advanced.py
```diff
@@ -1,6 +1,6 @@
 """
-Advanced Kriging
-================
+Advanced Gaussian process regression
+====================================
 """
```
jschueller, line 5: a few examples such as this one are referenced in python/test/CMakeLists.txt, so it has to be modified there too when changing the name eg:
`set_tests_properties (pyinstallcheck_example_plot_kriging_advanced PROPERTIES DISABLED TRUE)`
else cmake will fail (see mindeps job)

Issue #2962, opened 2025-05-26
ConditionedGaussianProcess can not be created from a GaussianProcessRegressionResult


adutfoy:
### What happened?

In order to generate fields from the conditioned Gaussin process estimated by a Gaussian Process Regression algo, we need to add the API: 
```python

ConditionedGaussianProcess(gpr_result, mesh)
```
where gpr_result is a GaussianProcessRegressionResult

The translation of the example https://openturns.github.io/openturns/latest/auto_meta_modeling/kriging_metamodel/plot_kriging_simulate.html#sphx-glr-auto-meta-modeling-kriging-metamodel-plot-kriging-simulate-py awaits for that API.

### How to reproduce the issue?

```shell
read above
```

### Version

1.25dev

### Operating System

Linux

### Installation media

from source

Issue #2951, opened 2025-05-08
The KrigingRandomVector class has no GaussianProcessRegression counterpart


regislebrun:
### What is the idea?

The shift from Kriging classes to GaussianProcessRegression classes should include all the related classes, namely:
  * KrigingEvaluation
  * KrigingGradient
  * KrigingRandomVector

### Why is this needed?

It would be weird to have both terms mixed once the KrigingAlgorithm has disapeared.

jschueller:
#2965

PR #2965, opened 2025-06-02
ConditionedGaussianProcess to support GaussianProcessRegression result & new GP classes

sofianehaddad:
Now we rely also on GaussianProcessRegressionResult to generate Gaussian process sample.

Thus ConditionedGaussianProcess is now in experimental.

- #2962  
- #2951 

This PR is needed to support the migration of the doc/classes/examples to the new API


lib/src/Uncertainty/Process/ConditionedGaussianProcess.cxx
```diff
@@ -211,7 +247,9 @@ void ConditionedGaussianProcess::save(Advocate & adv) const
 void ConditionedGaussianProcess::load(Advocate & adv)
 {
   GaussianProcess::load(adv);
+  adv.loadAttribute("inputDimension_", inputDimension_);
   adv.loadAttribute("krigingResult_", krigingResult_);
+  adv.loadAttribute("gprResult_", gprResult_);
```
jschueller: must be conditional to presence of new attributes

sofianehaddad: indeed, thanks

lib/src/Uncertainty/Process/ConditionedGaussianProcess.cxx
```diff
@@ -211,7 +247,12 @@ void ConditionedGaussianProcess::save(Advocate & adv) const
 void ConditionedGaussianProcess::load(Advocate & adv)
 {
   GaussianProcess::load(adv);
+  adv.loadAttribute("inputDimension_", inputDimension_);
```
jschueller: inputDimension_ too

sofianehaddad: Done.

jschueller:
lets merge so we can continue with gpr stuff

PR #2973, opened 2025-06-16
Misc

jschueller:


lib/src/Base/Algo/LeastSquaresMetaModelSelection.cxx
```diff
@@ -144,15 +145,18 @@ void LeastSquaresMetaModelSelection::run(const DesignProxy & proxy)
...
+      if (basisSequenceFactory_.getImplementation()->currentIndices_.getSize() > smallBasisSize)
       {
-        LOGINFO(OSS() << "Error=" << error << " larger than " << alpha << "*" << minimumError << "=" << alpha * minimumError);
-        break;
-      }
-      if (error > maximumError)
-      {
-        LOGINFO(OSS() << "Error=" << error << " larger than " << maximumError);
-        break;
+        if (!(error <= alpha * minimumError))
```
regislebrun, line 150: Géraud Blatman proposes an additional test to detect overfitting (see the blue box in Figure 5.8 page 123 of https://theses.hal.science/tel-00440197). It is based on the detection of two consecutive inceases of the cross validation error estimate.

lib/src/Uncertainty/Algorithm/Optimization/LOLAVoronoi.cxx
```diff
@@ -84,11 +85,12 @@ String LOLAVoronoi::__repr__() const
 void LOLAVoronoi::computeVoronoiScore() const
 {
   // estimate the Voronoi cell size by Monte Carlo (2.5.1.1, algorithm 2)
-  const Sample voronoiSample(distribution_.getSample(voronoiSamplingSize_));
+  const UnsignedInteger voronoiSamplingSize = std::max(voronoiMinimumSamplingSize_, static_cast<UnsignedInteger>(voronoiSamplingRatio_ * x_.getSize()));
```
regislebrun: voronoiMinimumSamplingSize_ is related to the global budget for the computation of the volumes, and voronoiSamplingRatio_ is in fact the mean number of points used to estimate the volume of a cell for uniformly distributed x_. So its name should reflect this, and also its type (it is an UnsignedInteger). IMO, a minimum value should be 100.

regislebrun (changes requested):
The ratio is indeed a local sampling size

lib/src/Uncertainty/Algorithm/Optimization/LOLAVoronoi.cxx
jschueller: done (using a default value of 10)

sofianehaddad (approved):

PR #2979, opened 2025-06-24
Set optim starting point in plot_gpr_isotropic

josephmure:
Previously the starting point was outside the optimization bounding box.


mbaudin47 (approved):

mbaudin47:
@jschueller : Please merge this PR.

PR #2980, opened 2025-06-27
Fix GPR Fitter optim

sofianehaddad:
If starting point is outside the bounds, we define a new one as the center of the bounds


python/doc/developer_guide/architecture.rst
```diff
@@ -190,7 +190,7 @@ The tools chosen for the development of the platform are:
 +---------------------------------------+---------------------------------------------------------------------------------+-------------------+
 | Special functions (optional)          | `MPC <https://www.multiprecision.org/mpc>`_                                     | 1.1.0             |
 +---------------------------------------+---------------------------------------------------------------------------------+-------------------+
-| Optimization (optional)               | `NLopt <https://nlopt.readthedocs.io/en/latest/>`_                              | 2.6               |
+| Optimization                          | `NLopt <https://nlopt.readthedocs.io/en/latest/>`_                              | 2.6               |
```
jschueller, line 193: I have scenarios where nlopt is too old than 2.6, like the Mageia / centos and other custom embedded build where we want to use the minimum dependencies as possible, so it cant be made mandatory

regislebrun (approved):

m-balesdent:
> hi @sofianehaddad, with Mathieu we are trying parameters normalization instead, cf my adam branch

A complete benchmark has been done this summer on that point. We plan to present these results on next technical committee.

PR #2988, opened 2025-07-07
Add HiGHS solver

jschueller:


lib/src/Base/Optim/HiGHS.cxx
```diff
@@ -0,0 +1,235 @@
...
+{
+}
+
+
+/* Check whether this problem can be solved by this solver */
+// Inherited method: never used by HiGHS because solver_ calls its own checkProblem method.
+void HiGHS::checkProblem(const OptimizationProblem & problem) const
+{
+  // Cannot solve multi-objective problems
+  if (problem.hasMultipleObjective()) throw InvalidArgumentException(HERE) << "Bonmin does not support multi-objective optimization";
```
regislebrun: Highs, not bonmin

lib/src/Base/Optim/HiGHS.cxx
```diff
@@ -0,0 +1,235 @@
...
+  // Nothing to do here
+}
+
+
+/* Constructor that sets starting sample */
+HiGHS::HiGHS(const OptimizationProblem & problem)
+  : OptimizationAlgorithmImplementation(problem)
+
+{
+}
```
regislebrun, line 51: Nothing to do here

lib/src/Base/Optim/HiGHS.cxx
```diff
@@ -0,0 +1,235 @@
...
+{
+  // Cannot solve multi-objective problems
+  if (problem.hasMultipleObjective()) throw InvalidArgumentException(HERE) << "Bonmin does not support multi-objective optimization";
+
+  // No LeastSquaresProblem / NearestPointProblem
+  if (problem.hasResidualFunction() || problem.hasLevelFunction())
+    throw InvalidArgumentException(HERE) << "HiGHS does not support least squares or nearest point problems";
+
+  if (problem.hasEqualityConstraint())
+    throw InvalidArgumentException(HERE) << "HiGHS does not support equality constraints";
```
regislebrun: A way to support it is to mimic Cobyla: f(x)=0 is replaced by f(x)>=-eps and f(x)<=eps for a small eps

lib/src/Base/Optim/HiGHS.cxx
```diff
@@ -0,0 +1,235 @@
...
+  HighsModel model;
+  model.lp_.num_col_ = problemDimension;
+  model.lp_.num_row_ = getProblem().getInequalityConstraint().getOutputDimension();
+
+  // objective function
+  const Point origin(problemDimension, 0.0);
+  model.lp_.sense_ = getProblem().isMinimization() ? ObjSense::kMinimize : ObjSense::kMaximize;
+  const Scalar f0 = getProblem().getObjective()(origin)[0];
+  model.lp_.offset_ = f0;
+  const Matrix grad0(getProblem().getObjective().gradient((origin)));
```
regislebrun: No check on the linearity of the objective function? You could use the isLinear() method here

lib/src/Base/Optim/HiGHS.cxx
```diff
@@ -0,0 +1,235 @@
...
+  if (getProblem().hasBounds())
+    for (int col = 0; col < model.lp_.num_col_; ++ col)
+    {
+      model.lp_.col_lower_[col] = getProblem().getBounds().getLowerBound()[col];
+      model.lp_.col_upper_[col] = getProblem().getBounds().getUpperBound()[col];
+    }
+
+  // Create a Highs instance
+  Highs highs;
+  highs.setOptionValue("output_flag", false);
```
regislebrun: No link with Log levels?

regislebrun (changes requested):
If you are interested in adding other advanced solvers, I have colleagues expert in the field and they know dozens of solvers for various exotic optimization settings (optimal scheduling and so on) ;-)

lib/src/Base/Optim/LinearProblem.cxx
```diff
@@ -0,0 +1,184 @@
...
+
+
+/* From function linearization */
+LinearProblem LinearProblem::Linearize(const OptimizationProblem & problem)
+{
+  if (problem.hasMultipleObjective() || problem.hasLevelFunction() || problem.hasResidualFunction())
+    throw InvalidArgumentException(HERE) << "Cannot linearize a multi-objective / nearest-point / least-squares problem";
+
+  const UnsignedInteger problemDimension = problem.getDimension();
+  const Point origin(problemDimension, 0.0);
```
regislebrun: Is it mandatory to linearize the objective function at the origin or could it be given by the user as an additional argument of the method?

lib/src/Base/Optim/LinearProblem.cxx
```diff
@@ -0,0 +1,184 @@
...
+static const Factory<LinearProblem> Factory_LinearProblem;
+
+/* Default constructor */
+LinearProblem::LinearProblem()
+  : OptimizationProblemImplementation()
+{
+  // Nothing to do
+}
+
+/* Constructor for nearest point problem */
```
regislebrun: No nearest point problem here

lib/src/Base/Optim/LinearProblem.cxx
```diff
@@ -0,0 +1,184 @@
...
+LinearProblem::LinearProblem(const Point & cost, const Interval & bounds,
+                const Matrix & A, const Interval & LU)
```
regislebrun: The name LU is not very informative. BTW, HiGHS outperform the other algorithms for large problems where A is likely to be sparse. At one point SparseMatrix should inherit from Matrix even if many services are not available, at least to be able to use it where only basic algebra is needed.

lib/src/Base/Optim/LinearProblem.cxx
jschueller: good idea, lets add a location parameter

lib/src/Base/Optim/LinearProblem.cxx
jschueller: yes lets change the names

sofianehaddad (approved):

PR #3004, opened 2025-08-28
Histogram: redefined getParametersCollection funtion

arkinn2:
override getParametersCollection function for Histogram in a way more coherent with UserDefined distribution

### DIFF
--- a/lib/src/Uncertainty/Distribution/Histogram.cxx
+++ b/lib/src/Uncertainty/Distribution/Histogram.cxx
@@ -334,6 +334,33 @@ Distribution Histogram::getStandardRepresentative() const
 }
 
 /* Parameters value and description accessor */
+Histogram::PointWithDescriptionCollection Histogram::getParametersCollection() const
+{
+  PointWithDescriptionCollection parameters(3);
+  const UnsignedInteger size = width_.getSize();
+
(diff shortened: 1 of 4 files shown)


regislebrun:
Hi,
The idea behind the getParametersDescription() method is to split the parameters of a distribution between those acting on marginal distributions and those acting on the dependence structure. As a result, it is a collection of points with description with the following convention:
  1. If the distribution is of dimension 1, the collection contains only one point with description
  2. If the distribution is of dimension d>1, the collection contains d+1 points with description (the last point is of size 0 for the case of the independent copula)
As many distributions are of dimension 1, the default behaviour implemented in DistributionImplementation corresponds to 1.
This design has been decided a long time ago and can be changed, but it has to be changed consistently among the many distributions. What is the motivation of the change you propose? BTW, your change may break the following classes:
  * FORMResult
  * AnalyticalResult
  * MarginalTransformationEvaluation
when one of the marginal distributions is an Histogram. Unfortunately it is not checked by the current tests...

jschueller:
indeed getParametersDescription returns the list of parameters for each marginal, so as Histogram is always of dim 1 the current behavior is correct
however you are right UserDefined is not consistent with the others as it does not return a collection of size 1 in dim 1

Issue #2761, opened 2024-09-16
The name of the attributes is not consistent in the use cases


mbaudin47:
### What happened?

There are inconsistencies in the name of the attributes of the use cases: https://openturns.github.io/openturns/latest/usecases/usecases.html

The input distribution has different names:
- In the beam use case (https://openturns.github.io/openturns/latest/usecases/use_case_beam.html) the input distribution is `distribution` ;
- In the wing weight use case (https://openturns.github.io/openturns/latest/usecases/use_case_wingweight.html), its name is `distributionX` ;
- In the Chaboche use case (https://openturns.github.io/openturns/latest/usecases/use_case_chaboche.html), its name is `inputDistribution`.

### How to reproduce the issue?

```shell
This is a code issue.
```

### Version

1.13

### Operating System

unknown

### Installation media

unknown

Issue #2907, opened 2025-03-21
PDF draw of histogram or UserDefined distrubition based on sample from a discrete distribution


pascal2802:
### What happened?

Discrete sample has 4 classes : 
+ 0 : 54
+ 1 : 720
+ 2 : 38 
+ 3 : 6

Produced graph are different depending on the OpenTURNS version used (1.23 or 1.24). 

It seems wrong for UserDefined distribution with ot version 1.23 and 1.24 (only 3 classes instead of 4)
Build histogram seems ok with OpenTURNS 1.23 but wrong with 1.24 (only 1 bin)

OpenTURNS version 1.23

![Image](https://github.com/user-attachments/assets/777fe096-4fcb-4946-b9eb-678b16a647fa)

OpenTURNS version 1.24

![Image](https://github.com/user-attachments/assets/e6395561-7bf3-426a-bb7c-6e7d68c54335)

### How to reproduce the issue?

```python
#!/usr/bin/env python3
import openturns as ot
import openturns.viewer as otv

# load sample from a discrete distribution with only 4 possible values : 0, 1, 2 or 3
sample = ot.Sample.ImportFromCSVFile("sample.csv")

# Count number of element in each class
numberOfeltInClass = ot.Point(4)
for i in range(sample.getSize()):
    for j in range(numberOfeltInClass.getSize()):
        if sample[i, 0] == j:
            numberOfeltInClass[j] += 1
print(numberOfeltInClass)

# build an histogram from the sample
histogram = ot.HistogramFactory().build(sample[:,0])
graph = histogram.drawPDF()

#Build a discrete user defined distribution 
dist = ot.UserDefined(sample)
curve = dist.drawPDF()
graph.add(curve)
print(dist)

# Display the PDF graph
view = otv.View(graph)
view.show()
```

### Version

1.23 and 1.24

### Operating System

Linux

### Installation media

conda

### Additional Context

[sample.csv](https://github.com/user-attachments/files/19397789/sample.csv)

mbaudin47:
Here is the UserDefined distribution:
```
UserDefined({x = [0], p = 0.0660147}, {x = [1], p = 0.880196}, {x = [2], p = 0.0464548}, {x = [3], p = 0.00733496})
```
We see that the class "3" has a small but nonzero probability.

regislebrun:
Hi!
[...]

PR #3016, opened 2025-09-25
Misc

jschueller:


python/test/t_Object_pickle.py
```diff
@@ -23,3 +23,26 @@
     obj2 = pickle.load(src)
     print(("object: " + str(obj)))
     print(("same: " + str(obj2 == obj) + "\n"))
+
+
```
sofianehaddad, line 27: # comment : ticket  2920

python/src/usecases/flood_model.py
```diff
@@ -67,10 +67,9 @@ class FloodModel:
 
     model : :class:`~openturns.ParametricFunction`
         The flood model.
-        The function has input dimension 4 and output dimension 1.
-        More precisely, we have  :math:`\vect{X} = (Q, K_s, Z_v, Z_m)` and
-        :math:`Y = H`.
-        Its parameters are :math:`\theta = (B, L)`.
+        The function has input dimension 4 and output dimension 3.
```
mbaudin47: Not 4: 8 (my original message was wrong, sorry for that).
```suggestion
        The function has input dimension 8 and output dimension 3.
```

mbaudin47 (changes requested):

mbaudin47:
The commit 9dd072df4e475878a262c1021e9742d6f44f8751 fixes #2761, which should be mentioned both in the PR description.
The problem may be more easily fixed if a new **UseCaseModel()** class was developed. Its constructor might be based on a **function** and a **distribution**. This way, class inheritance will enforce the name of the attributes in all examples, whatever the name in the implementation.

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessFitter.cxx
```diff
@@ -395,7 +395,14 @@ Scalar GaussianProcessFitter::maximizeReducedLogLikelihood()
   }
   catch (const NotDefinedException &) // setStartingPoint is not defined for the solver
   {
-    // Nothing to do if setStartingPoint is not defined
+    // Define starting point for the optimization as the center of the bounds if necessary
+    Sample initialPoints(solver_.getStartingSample());
+    for (UnsignedInteger i = 0; i < initialPoints.getSize(); ++ i)
+    {
+      if (!optimizationBounds_.contains(initialPoints[i]))
+        initialPoints[i] = (optimizationBounds_.getUpperBound() + optimizationBounds_.getLowerBound())/2;
```
sofianehaddad: small optimizations :
*  you can compute once `(optimizationBounds_.getUpperBound() + optimizationBounds_.getLowerBound())/2;`
*  you can avoid the creation of points either by extending the `contains` method using `iterator`

jschueller: ok for the first item

jschueller:
@regislebrun would you be able to review this ?

lib/src/Base/Func/DualLinearCombinationEvaluation.cxx
```diff
@@ -297,13 +299,19 @@ void DualLinearCombinationEvaluation::setFunctionsCollectionAndCoefficients(cons
         if (!(std::abs(currentCoefficient[j]) > epsilon))
         {
           currentCoefficient[j] = 0.0;
-          LOGWARN(OSS() << "set the component " << j << " of contributor " << i << "=" << currentCoefficient[j] << " to zero as it is too small");
+          if (removedComponent.getSize() < 20)
```
regislebrun, line 302: What is the rational behind the value of 20?

lib/src/Base/Func/DualLinearCombinationEvaluation.cxx
```diff
@@ -297,13 +299,19 @@ void DualLinearCombinationEvaluation::setFunctionsCollectionAndCoefficients(cons
...
+            removedComponent.add(OSS() << i << ":" << j);
         }
       coefficients_.add(currentCoefficient);
       functionsCollection_.add(functionsCollection[i]);
     }
-    else LOGWARN(OSS() << "removed the contributor " << i << "=" << functionsCollection[i] << " from the linear combination as its coefficient " << absoluteCoefficients[i] << " is too small.");
+    else if (removedContributor.getSize() < 20)
```
regislebrun, line 308: What is the rational behind the value of 20?

lib/src/Base/Func/DualLinearCombinationEvaluation.cxx
```diff
@@ -297,13 +299,19 @@ void DualLinearCombinationEvaluation::setFunctionsCollectionAndCoefficients(cons
...
+      removedContributor.add(i);
   }
+  if (removedContributor.getSize())
+    LOGINFO(OSS() << "Pruned the linear combinations contributors at indices " << removedContributor.__str__().substr(0, 50) << "... wrt the minimum coefficient=" << epsilon);
```
regislebrun, line 312: What is the rational behind the value of 50?

lib/src/Base/Func/DualLinearCombinationEvaluation.cxx
```diff
@@ -297,13 +299,19 @@ void DualLinearCombinationEvaluation::setFunctionsCollectionAndCoefficients(cons
...
+  if (removedComponent.getSize())
+    LOGINFO(OSS() << "Pruned the linear combinations components at indices " << removedComponent.__str__().substr(0, 50) << "... wrt the minimum coefficient=" << epsilon);
```
regislebrun, line 314: same here

lib/src/Base/Func/DualLinearCombinationEvaluation.cxx
jschueller: its just to format the message to make it short to read, we probably dont need to add a key for that

jschueller:
> Only minor remarks. Did you changed the 2D version of drawPDF() for discrete distributions with small support?

yes, I show the full support when small enough, see #2907, do you agree ?
```
 if (isDiscrete())
  {
    // fully draw the support if small enough
    if (getSupport().getSize() <= ResourceMap::GetAsUnsignedInteger("Distribution-SmallSupport"))
    {
      xMin = range_.getLowerBound()[0];
      xMax = range_.getUpperBound()[0];
    }
```

Issue #3000, opened 2025-07-21
Access of input/output samples from monte carlo simulation


m-balesdent:
### What is the idea?

Hi, 

It seems that when performing a reliability problem with `Monte-Carlo` (for example [here](https://openturns.github.io/openturns/latest/auto_reliability_sensitivity/reliability/plot_estimate_probability_monte_carlo.html#sphx-glr-auto-reliability-sensitivity-reliability-plot-estimate-probability-monte-carlo-py)), it is not possible to get the Input / Ouput samples, although they are available with more sophisticated methods such as importance sampling, subset, etc. 

I think it could be really interesting to add these methods to the basic `ProbabilitySimulationAlgorithm` with `MonteCarlo` (and others).

m-balesdent:
> in the example, changing
> 
> model = sm.model
> 
> by
> 
> model = ot.MemoizeFunction(sm.model)
> 
> allows retrieving the used data. In parallel, I can not see why in `Subset` or `Importance sampling`, we can retrieve the data. What are you thinking about [@m-balesdent](https://github.com/m-balesdent) ?

The idea is to get them directly in C++ and not via python. The goal is to build custom classes in C++ (e.g. AK) and add them in the library. We are currently working on a generic AK class that will enable to be used with all the reliability algorithms of the library as argument of this class. But for that purpose, we need the same getters for the reliability algorithms.

Issue #2948, opened 2025-05-07
The FittingAlgorithm.run(leastSquaresMethod, outputSample) method is hidden in Python


mbaudin47:
### What is the idea?

The FittingAlgorithm.run(leastSquaresMethod, outputSample) method is hidden in Python.

This is because of a SWIG directive:

https://github.com/openturns/openturns/blob/dd4f77f327d07934d76bc1bcaf2cedf9f05ba6f0/lib/src/Base/Algo/openturns/FittingAlgorithm.hxx#L69

### Why is this needed?

This prevents from implementing efficiently the orthogonal matching pursuit in Python:

https://gist.github.com/mbaudin47/87a09578aef2e38b498f2f5c5cda193b

PR #3026, opened 2025-10-16
Some bugfixes

sofianehaddad:
- #2948  
 - #3000


lib/src/Uncertainty/Algorithm/Simulation/ProbabilitySimulationAlgorithm.cxx
```diff
@@ -99,6 +113,17 @@ Sample ProbabilitySimulationAlgorithm::computeBlockSample()
   return event_.getSample(blockSize_);
 }
 
+/* Keep event sample */
+void ProbabilitySimulationAlgorithm::setKeepSample(const Bool keepSample)
+{
+  if ((!getEvent().isComposite()) && (keepSample))
```
jschueller: && keepSample)

python/test/t_ProbabilitySimulationAlgorithm_std.py
```diff
@@ -133,6 +133,10 @@ def stop():
     myAlgo.setMaximumOuterSampling(250)
     myAlgo.setBlockSize(4)
     myAlgo.setMaximumCoefficientOfVariation(0.1)
+    try:
+        myAlgo.setKeepSample(True)
+    except:
```
jschueller:
```suggestion
    except Exception:
```

python/test/t_ProbabilitySimulationAlgorithm_inoutsample.py
```diff
@@ -0,0 +1,49 @@
...
+algo.run()
+
+# Stream out the result
+print("MonteCarlo result=", algo.getResult())
+
+# input / output sampls
+input_sample = algo.getInputSample()
+output_sample = algo.getInputSample()
+
+print(f"Sample size = {len(input_sample)}")
```
jschueller: one should prefer assert over expout files

python/test/t_ProbabilitySimulationAlgorithm_inoutsample.py
```diff
@@ -0,0 +1,50 @@
...
+algo.run()
+
+# Stream out the result
+print("MonteCarlo result=", algo.getResult())
+
+# input / output sampls
+input_sample = algo.getInputSample()
+output_sample = algo.getInputSample()
+
+ott.assert_almost_equal(len(input_sample), 2670)
```
jschueller: simple assert is enough to test an integer value

python/test/t_ProbabilitySimulationAlgorithm_inoutsample.py
```diff
@@ -0,0 +1,50 @@
...
+# We create an Event from this RandomVector
+event = ot.ThresholdEvent(output, ot.Less(), -3.0)
+
+# We create a Monte Carlo algorithm
+experiment = ot.MonteCarloExperiment()
+# event.isComposite() --> automatically add a MonteCarloExperiment
+algo = ot.ProbabilitySimulationAlgorithm(event)
+algo.setMaximumOuterSampling(500)
+algo.setBlockSize(10)
+algo.setMaximumCoefficientOfVariation(0.05)
```
jschueller: maybe disable cov criterion so that you can check the sample size is all the simulations

python/test/CMakeLists.txt
```diff
@@ -809,6 +809,7 @@ ot_pyinstallcheck_test (Wilks_std)
 ot_pyinstallcheck_test (SimulationSensitivityAnalysis_std)
 ot_pyinstallcheck_test (ProbabilitySimulationAlgorithm_std)
 ot_pyinstallcheck_test (ProbabilitySimulationAlgorithm_draw)
+ot_pyinstallcheck_test (ProbabilitySimulationAlgorithm_inoutsample)
```
jschueller: IGNOREOUT

python/test/t_ProbabilitySimulationAlgorithm_inoutsample.py
sofianehaddad: no, it is smaller as the default `cov` is `0.10`

python/test/t_ProbabilitySimulationAlgorithm_inoutsample.py
sofianehaddad: ok let's do that

python/test/t_ProbabilitySimulationAlgorithm_inoutsample.py
jschueller: disabling it completely would yield an obvious reference for the expected size (eg 5000)

PR #3029, opened 2025-10-16
Null PDF at finite lower bounds

JPelamatti:
Corrects PDF evaluations for several distributions so as to have a null value at finite lower bounds. Adds a test to ensure this when possible.

### DIFF
--- a/lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/EmpiricalBernsteinCopula.cxx
@@ -241,9 +241,9 @@ Scalar EmpiricalBernsteinCopula::computePDF(const Point & point) const
 {
   const UnsignedInteger dimension = getDimension();
   if (point.getDimension() != dimension) throw InvalidArgumentException(HERE) << "Error: the given point must have dimension=" << dimension << ", here dimension=" << point.getDimension();
-  if (binNumber_ == 1) return 1.0;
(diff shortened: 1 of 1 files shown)


lib/include/openturns/OTtestcode.hxx
```diff
@@ -445,6 +445,12 @@ public:
     checkGeneral();
     checkRealization();
     checkComparison();
+    
+    if (enablePDFAtlowerBound_)
+    {
+        checkFiniteLowerBoundPDF();
```
jschueller:
```suggestion
      checkFiniteLowerBoundPDF();
```

python/test/t_SmoothedUniform_std.py
```diff
@@ -103,4 +103,5 @@
 ot.Log.Show(ot.Log.TRACE)
 ot.RandomGenerator.SetSeed(1)
 validation = ott.DistributionValidation(distribution)
+validation.skipPDFAtLowerBound()  # Temporary and should be removed for proper testing
```
jschueller:
```suggestion
validation.skipPDFAtLowerBound()  # FIXME: https://github.com/openturns/openturns/issues/3028
```

lib/include/openturns/OTtestcode.hxx
```diff
@@ -494,6 +500,10 @@ public:
   {
     enablePDF_ = false;
   }
+  void skipPDFAtLowerBound()
```
jschueller, line 503: must be documented in python/src/OTtestcode_doc.i.in

Issue #2925, opened 2025-04-05
We cannot evaluate a univariate polynomial on a sample


mbaudin47:
### What is the idea?

We can create an univariate polynomial e.g. using `HermiteFactory` of any polynomial degree. Then we can draw it using the `draw()` method. But this polynomial is not a plain `ot.Function`. Therefore, it does not provide other features that we may expect, e.g. evaluate its value on a `ot.Sample`.

### Why is this needed?

We may be interested by computing the Gram matrix from a Hermite basis. This could be done using the following code if the evaluation of a `ot.Sample` was available.

```python
orthogonalFunctionFactory = ot.HermiteFactory()
distribution = orthogonalFunctionFactory.getMeasure()
sampleSize = 20
maximumDimension = 10
gramMatrix = ot.SymmetricMatrix(maximumDimension)
inputSample = distribution.getSample(sampleSize)
for i in range(maximumDimension):
    functionI = orthogonalFunctionFactory.build(i)
    functionIValue = functionI(inputSample).asPoint()
    for j in range(i + 1):
        functionJ = orthogonalFunctionFactory.build(j).asPoint()
        functionJValue = functionJ(inputSample)
        gramMatrix[i, j] = functionIValue.dot(functionJValue) / sampleSize
```

Instead, this fails with:
```
TypeError: in method 'OrthogonalUniVariatePolynomial___call__', argument 2 of type 'OT::Scalar'
```

### Additional Context

Example 1. With a Hermite polynomial.
```
import openturns as ot

orthogonalPolynomialFactory = ot.HermiteFactory()
polynomial = orthogonalPolynomialFactory.build(10)
distribution = orthogonalPolynomialFactory.getMeasure()
inputSample = distribution.getSample(10)
value = polynomial(inputSample)
```
produces:
```
TypeError: in method 'UniVariateFunction___call__', argument 2 of type 'OT::Scalar'
```

Example 2. With a polynomial.
```
import openturns as ot

polynomialFactory = ot.MonomialFunctionFactory()
polynomial = polynomialFactory.build(10)
inputSample = ot.Uniform().getSample(10)
value = polynomial(inputSample)
```
produces:
```
TypeError: in method 'UniVariateFunction___call__', argument 2 of type 'OT::Scalar'
```

regislebrun:
[...]

Issue #2942, opened 2025-05-01
EV3 don't throw on unknown function


regislebrun:
### What happened?

The elementary functions supported by the two backends ExprTk and MuParser include the expm1 and log1p functions. These functions are unknown to Ev3, which should throw an exception on these functions, which should in turn make SymbolicFunction switch to finite differences. But Ev3 silently ignore the unknown functions and return a wrong formula for the symbolic derivative.

### How to reproduce the issue?

```shell
If you run this code

import openturns as ot
ot.Log.Show(ot.Log.ALL)
for expr in ["2*expm1(3*x)", "2*log1p(3*x)"]:
    f = ot.SymbolicFunction("x", expr)
    print("f=", f)
    print("f=", repr(f))
    x0 = [0.4 / 3.0]
    print("f(", x0, ")=", f(x0)[0])
    print("f'(", x0, ")=", f.gradient(x0)[0, 0])
    print("f''(", x0, ")=", f.hessian(x0)[0, 0, 0])

you get

f= [x]->[2*expm1(3*x)]
f= class=SymbolicFunction name=Unnamed implementation=class=FunctionImplementation name=Unnamed description=[x,y0] evaluationImplementation=class=SymbolicEvaluation name=Unnamed inputVariablesNames=[x] outputVariablesNames=[y0] formulas=[2*expm1(3*x)] gradientImplementation=class=SymbolicGradient name=Unnamed evaluation=class=SymbolicEvaluation name=Unnamed inputVariablesNames=[x] outputVariablesNames=[y0] formulas=[2*expm1(3*x)] hessianImplementation=class=SymbolicHessian name=Unnamed evaluation=class=SymbolicEvaluation name=Unnamed inputVariablesNames=[x] outputVariablesNames=[y0] formulas=[2*expm1(3*x)]
f( [0.13333333333333333] )= 0.9836493952825407
DBG - d(3*x)/d(x)=3
f'( [0.13333333333333333] )= 3.0
DBG - First variable=x, derivative=3
DBG - d2(3*x)/d(x)d(x)=0
f''( [0.13333333333333333] )= 0.0
f= [x]->[2*log1p(3*x)]
[...]

PR #3038, opened 2025-10-22
Hackathon 2025 10

regislebrun:
This PR fixes issues #2925 and #2942. It also add support for the floor, ceil, trunc and round functions in Ev3.


lib/src/Base/Func/openturns/UniVariatePolynomialImplementation.hxx
```diff
@@ -67,6 +67,7 @@ public:
   /** UniVariatePolynomialImplementation are evaluated as functors */
   Scalar operator() (const Scalar x) const override;
   Complex operator() (const Complex z) const;
+  Sample operator() (const Sample & x) const;
```
jschueller:
```suggestion
  Sample operator() (const Sample & x) const override;
```

regislebrun: Fixed. It was useless to redefine this method here.

mbaudin47 (approved):

PR #3062, opened 2025-12-08
Meshing algorithms

jschueller:
- Add PolygonMesher to triangulate a 2-d polygon (using embedded earcut.hpp library)
```
mesher = otexp.PolygonMesher()
vertices = [[0.0, 0.0], [0.0, 3.0], [1.0, 2.0], [0.5, 0.5], [1.0, 0.5], [2.0, 2.0], [2.0, 0.0]]
triangulation = mesher.build(vertices)
```

Figure_1

- Add ConvexHullMesher to build the convex hull a of set of points (external qhull library):
```
points = ot.Normal(2).getSample(15)
mesher = otexp.ConvexHullMesher()
convexHull = mesher.build(points) 
```
Figure_1


lib/src/Base/Geom/CloudMesher.cxx
```diff
@@ -0,0 +1,109 @@
...
+{
+  return __repr__();
+}
+
+/* Here is the interface that all derived class must implement */
+
+Mesh PolygonMesher::build(const Sample & polygonVertices) const
+{
+#ifdef OPENTURNS_HAVE_CGAL
+  Triangulation triangulation;
```
regislebrun: A check on the sample dimension & size?

regislebrun (commented):
~polytri.py~, good news!

lib/src/Base/Geom/CloudMesher.cxx
jschueller: yes, I noticed already

lib/src/Base/Geom/CloudMesher.cxx
```diff
@@ -0,0 +1,108 @@
...
+
+BEGIN_NAMESPACE_OPENTURNS
+
+CLASSNAMEINIT(CloudMesher)
+static const Factory<CloudMesher> Factory_CloudMesher;
+
+
+/* Default constructor */
+CloudMesher::CloudMesher()
+  : PersistentObject()
```
sofianehaddad: We have : 
 - `IntervalMesher`
 - `BoundaryMesher`
 - `LevelSetMesher`
 - `CloudMesher`

But we don't have any common parent class. Is there any specific reason for that ?

jschueller: yes, there is no common api as the input objects differ each time

PR #3069, opened 2026-01-10
Improve LOLAVoronoi and NearestNeighbourAlgorithm

regislebrun:
This PR proposes the two following improvements:
  1. The NearestNeighbourAlgorithmImplementation::query(const Sample & sample) method is evaluated in parallel , resulting into a huge speedup e.g for KDTree and the computation of the Voronoi score in LOLAVoronoi.
  2. The LeastSquaresMethod used to compute the LOLA score is moved outside of the loop over the neighbors, the KDTree queries to select new candidates is done in one shot, using potential parallelism in its implementation

### DIFF
--- a/python/doc/theory/numerical_methods/distribution_realization.rst
+++ b/python/doc/theory/numerical_methods/distribution_realization.rst
@@ -59,7 +59,7 @@ Several classical techniques exist:
       cheap function :math:`m` such that :math:`m(u) \geq \rho(u)`: one
       checks if :math:`u\geq m(u)` and directly reject :math:`y` if the
       test is positive (quick rejection test). The combination of quick
-      acceptation and quick rejection is called a *squeeze*.
+      acceptance and quick rejection is called a *squeeze*.
 
    -  The test :math:`u\leq \rho(u)` can be replaced by an equivalent
       one but much more computationally efficient.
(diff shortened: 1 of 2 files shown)


lib/src/Uncertainty/Algorithm/Optimization/LOLAVoronoi.cxx
```diff
@@ -327,15 +328,18 @@ Sample LOLAVoronoi::generate(const UnsignedInteger size) const
     const Point width(d, neighbourhoodMaximumDistance);
     const Interval bounds(xi - width, xi + width);
     Sample voronoiSample;
-    const UnsignedInteger voronoiSamplingSize = std::max(voronoiMinimumSamplingSize_, voronoiMeanSamplingSize_ * x_.getSize());
+    const UnsignedInteger voronoiSamplingSize = voronoiMinimumSamplingSize_;
```
jschueller, line 331: this disables the mean number per cell feature, why is that ?

regislebrun: There are two sampling steps:
  1. The sampling done to estimate the relative size of the Voronoi cells. For this, the number of sampling points has to be greater than the size of the current X sample. t is where the mean number of points per cell is relevant.
  2. The sampling of the box neighbourhood of the candidate. For this step, the sampling size doesn't have to be related to the  current size of the X sample.

The change done at this point is for point 2.

PR #3076, opened 2026-01-16
Gaussian Process Regression analytical LOO cross-validation

josephmure:
This PR implements formula (23) from [Ginsbourger and Schärer (2025)](https://www.tandfonline.com/doi/pdf/10.1080/10618600.2024.2353633) in a new class called `GaussianProcessRegressionCrossValidation`.

This class is adapted from `FunctionalChaosValidation` but does not actually implement K-fold validation (note that Ginsbourger and Schärer 2025 give the formulas for K-fold validation so implementation *is* possible) and sticks to Leave-One-Out validation.

It only works with scalar outputs.

### DIFF
--- a/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
@@ -131,7 +131,7 @@ Sample GaussianProcessRegressionCrossValidation::computeGPRLeaveOneOutPrediction
 
   // Compute unscaled residuals using the already computed rho : Sigmatilde y = L^-T rho
   const GaussianProcessFitterResult gpfResult(gaussianProcessRegressionResult.getGaussianProcessFitterResult());
-  const Point rho(gpfResult.getRho());
+  const Point rho(gpfResult.getStandardizedOutput());
   const TriangularMatrix covarianceCholeskyFactor(gpfResult.getCholeskyFactor()); // L
   const TriangularMatrix covarianceCholeskyFactorTranspose(covarianceCholeskyFactor.transpose()); // L^T
   Point residuals(covarianceCholeskyFactorTranspose.solveLinearSystem(rho)); // L^-T rho = (Sigma^-1  - Sigma^-1 F (F^T Sigma^-1 F)^-1 F^T Sigma^-1 ) y
--- a/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GaussianProcessFitterResult.hxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GaussianProcessFitterResult.hxx
@@ -95,6 +95,12 @@ public:
   /** process accessor */
   Process getNoise() const;
 
+  /** Method that returns the covariance factor - lapack */
+  TriangularMatrix getCholeskyFactor() const;
+
+  /** rho accessor */
+  Point getStandardizedOutput() const;
+
   /** Method save() stores the object through the StorageManager */
   void save(Advocate & adv) const override;
 
@@ -113,22 +119,15 @@ protected:
   // GaussianProcessRegression could use Cholesky setters
   friend class GaussianProcessRegression;
 
-  // GaussianProcessRegressionCrossValidation could use getRho()
-  friend class GaussianProcessRegressionCrossValidation;
-
   /** Accessor to the Cholesky factor*/
   void setCholeskyFactor(const TriangularMatrix & covarianceCholeskyFactor,
                          const HMatrix & covarianceHMatrix);
-
-  /** Method that returns the covariance factor - lapack */
-  TriangularMatrix getCholeskyFactor() const;
 
   /** Method that returns the covariance factor - hmat */
   HMatrix getHMatCholeskyFactor() const;
 
   /** rho accessor */
-  Point getRho() const;
-  void setRho(const Point & rho);
+  void setStandardizedOutput(const Point & rho);
 
 
 private:
--- a/python/src/GaussianProcessFitterResult_doc.i
+++ b/python/src/GaussianProcessFitterResult_doc.i
@@ -198,4 +198,27 @@ Notes
 -----
 As the same basis is used for each marginal output, each :math:`\vect{\beta}^\ell` vector is of dimension
 :math:`b`, the size of the functional basis.)RAW"
+
 // ---------------------------------------------------------------------
+
+%feature("docstring") OT::GaussianProcessFitterResult::getCholeskyFactor
+R"RAW(Accessor to the Cholesky factor of the covariance matrix.
+
+Returns
+-------
+L : :class:`~openturns.TriangularMatrix`
+    Cholesky factor :math:`\mat{L}_{\vect{p}^*}` of the covariance matrix, which can thus be written :math:`\mat{L}_{\vect{p}^*} \mat{L}_{\vect{p}^*}^T`.)RAW"
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::GaussianProcessFitterResult::getStandardizedOutput
+R"RAW(Accessor to the standardized output.
+
+Returns
+-------
+rho : :class:`~openturns.Point`
+    Standardized output vector.
+
+Notes
+-----
+The standardized output vector is defined as :math:`\mat{L}_{\vect{p}^*}^{-1}(\vect{y} - \vect{m}_{\vect{\beta}^*(\vect{p}^*)})`.)RAW"
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,7 @@
 === Python module ===
 
 === Miscellaneous ===
+ * New GaussianProcessFitterResult accessors: getCholeskyFactor and getStandardizedOutput
 
 
 
--- a/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessFitter.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessFitter.cxx
@@ -334,7 +334,7 @@ void GaussianProcessFitter::run()
   reducedCovarianceModelCopy.setActiveParameter(covarianceModel_.getActiveParameter());
 
   result_ = GaussianProcessFitterResult(inputSample_, outputSample_, metaModel, F_, basis_, beta_, reducedCovarianceModelCopy, optimalLogLikelihood, method_);
-  result_.setRho(rho_);
+  result_.setStandardizedOutput(rho_);
 
   // The scaling is done there because it has to be done as soon as some optimization has been done, either numerically or through an analytical formula
   if (keepCholeskyFactor_)
--- a/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessFitterResult.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessFitterResult.cxx
@@ -152,13 +152,13 @@ TriangularMatrix GaussianProcessFitterResult::getCholeskyFactor() const
 }
 
 /** rho accessor */
-Point GaussianProcessFitterResult::getRho() const
+Point GaussianProcessFitterResult::getStandardizedOutput() const
 {
   return rho_;
 }
 
 /** rho accessor */
-void GaussianProcessFitterResult::setRho(const Point & rho)
+void GaussianProcessFitterResult::setStandardizedOutput(const Point & rho)
 {
   rho_ = rho;
 }
--- a/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegression.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegression.cxx
@@ -112,7 +112,7 @@ Point GaussianProcessRegression::computeGamma() const
   // Get cholesky factor & rho from gaussian fitter result
   LOGDEBUG("Solve L^t.gamma = rho");
 
-  const Point rho(gaussianProcessFitterResult_.getRho());
+  const Point rho(gaussianProcessFitterResult_.getStandardizedOutput());
 
   const GaussianProcessFitterResult::LinearAlgebra algebraMethod = gaussianProcessFitterResult_.getLinearAlgebraMethod();
   if (algebraMethod == GaussianProcessFitterResult::HMAT)


jschueller:
I'd be in favor of not exposing the kfold

mbaudin47:
There is a Python implementation available at:

https://github.com/mbaudin47/hpcuqtraining/blob/master/2021/Scripts/Metamodel/TP4b_Kriging.ipynb

Here is the relevant part:
```python
def computeLOO(inputSample, outputSample, krigingResult):
    """
    Compute the Leave One out prediction analytically (from Dubrule 1983 paper).
     """
    inputSample = np.array(inputSample)
    outputSample = np.array(outputSample)

    # get covariance model
    cov = krigingResult.getCovarianceModel()
    # get coefficient and compute trend
    basis = krigingResult.getBasisCollection()[0]
    F1 = krigingResult.getTrendCoefficients()[0]
    size = inputSample.shape[0]
    p = F1.getDimension()
    F = np.ones((size, p))
    for i in range(p):
        F[:, i] = np.hstack(basis.build(i)(normalized_inputSample))
    # Calcul de y_loo
    K = np.array(cov.discretize(normalized_inputSample))
    Z = np.zeros((p, p))
    S = np.vstack([np.hstack([K, F]), np.hstack([F.T, Z])])
    S_inv = np.linalg.inv(S)
    B = S_inv[:size:, :size:]
    B_but_its_diag = B * (np.ones(B.shape) - np.eye(size))
    B_diag = np.atleast_2d(np.diag(B)).T
    y_loo = (- np.dot(B_but_its_diag / B_diag, outputSample)).ravel()
    return ot.Sample(np.vstack(y_loo))

def computeQ2(inputSample, outputSample, krigingResult):
    """
    Compute the Q2 using the analytical loo prediction.
    """
    y_loo = np.hstack(computeLOO(inputSample, outputSample, krigingResult))
    # Calcul du Q2
    delta = (np.hstack(outputSample) - y_loo)
    Q2 = 1 - np.mean(delta**2)/np.var(outputSample)
    return Q2
  ```

File: python/src/GaussianProcessRegressionCrossValidation_doc.i [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+>>> gpf = ot.GaussianProcessFitter(
+...     inputTrain, outputTrain, covariance_kernel, basis
+... )
+>>> gpf.run()
+>>> gpf_result = gpf.getResult()
+>>> gpr = ot.GaussianProcessRegression(gpf_result)
+>>> gpr.run()
+>>> gpr_result = gpr.getResult()
+
+Validate the polynomial chaos expansion using leave-one-out cross-validation.
```
sofianehaddad: Bad copy/paste

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+  Point scales(covarianceCholeskyFactorInverseSquared.getImplementation()->genVectProd(Point(outputSample.getSize(), 1.0), true)); // diagonal elements of L^-T L^-1
+
+  // If the trend is estimated, the scales (which are homogeneous to precisions) must be diminished
+  const UnsignedInteger  basisSize = gaussianProcessRegressionResult.getBasis().getSize();
+  if (basisSize > 0)
+  {
+    const Matrix regressionMatrix(gaussianProcessRegressionResult.getRegressionMatrix()); // F
+    const Matrix Phi(covarianceCholeskyFactor.solveLinearSystem(regressionMatrix)); // Phi = L^-1 F
+    const Matrix Phitranspose(Phi.transpose());
+    CovarianceMatrix PhitransposePhi((Phitranspose * Phi).getImplementation()); // Phi^T Phi (not to be used, later modified in-place)
```
sofianehaddad: use `Phitranspose.computeGram()`

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+    TriangularMatrix phi(PhitransposePhi.computeCholesky()); // Phi^T Phi =: phi phi^T (phi not to be used, later modified in-place)
+    Matrix auxiliary(phi.solveLinearSystemInPlace(Phitranspose * covarianceCholeskyFactorInverse)); // auxiliary := phi^-1 Phi^T L^-1 has basisSize rows
+    // Subtract the diagonal elements of auxiliary^T auxiliary from the scales
+    auxiliary.squareElements();
+    scales -= auxiliary.getImplementation()->genVectProd(Point(basisSize, 1.0), true);
```
sofianehaddad, line 166: These scales are infact the $\sqrt{\sigma_{-i}}$ that could be stored for the diagnosis of normality

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+
+/* Compute cross-validation Leave-One-Out metamodel predictions */
+Sample GaussianProcessRegressionCrossValidation::ComputeMetamodelLeaveOneOutPredictions(
+  const GaussianProcessRegressionResult & gaussianProcessRegressionResult)
+  //const LeaveOneOutSplitter & splitter)
+{
+  const Sample outputSample(gaussianProcessRegressionResult.getOutputSample());
+
+  // Implement second formula from Equation (32) in (Ginsbourger, 2023 preprint) which is equivalent to Dubrule (1983)
+
```
sofianehaddad, line 123: Please add some comments in order to ease the maintenance

First we compute $\tilde{Q}$ using already $\rho$ attribute as $L^{—T} \rho = \tilde{Q} y$

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+  const TriangularMatrix covarianceCholeskyFactorTranspose(covarianceCholeskyFactor.transpose()); // L^T
+  Point residuals(covarianceCholeskyFactorTranspose.solveLinearSystem(rho)); // L^-T rho = (Sigma^-1  - Sigma^-1 F (F^T Sigma^-1 F)^-1 F^T Sigma^-1 ) y
+
+  // Compute scales of residuals
+  const TriangularMatrix covarianceCholeskyFactorInverse(covarianceCholeskyFactor.inverse().getImplementation()); //L^-1
+  TriangularMatrix covarianceCholeskyFactorInverseSquared(covarianceCholeskyFactorInverse);
+  covarianceCholeskyFactorInverseSquared.squareElements();
+  Point scales(covarianceCholeskyFactorInverseSquared.getImplementation()->genVectProd(Point(outputSample.getSize(), 1.0), true)); // diagonal elements of L^-T L^-1
+
+  // If the trend is estimated, the scales (which are homogeneous to precisions) must be diminished
```
sofianehaddad:
```suggestion
  // If the trend is estimated, the scales (which are homogeneous to precisions) must be diminished
  // For the right part we use $\phi$ to compute the Cholesky form of $\phi^t \phi$
  // and get $L^{—T}  \phi^T L^{-t}_{\phi} L_{\phi} \phi^T L{-1}$
  // Setting  $B:= L_{\phi} \phi^T L{-1}$ the previous writes $B^T B$ 
  // and we get the diagonal elements as $\sum_j B^2_{ji}$
```

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+  // Implement second formula from Equation (32) in (Ginsbourger, 2023 preprint) which is equivalent to Dubrule (1983)
+
+  // Compute unscaled residuals
+  const GaussianProcessFitterResult gpfResult(gaussianProcessRegressionResult.getGaussianProcessFitterResult());
+  const Point rho(gpfResult.getRho());
+  const TriangularMatrix covarianceCholeskyFactor(gpfResult.getCholeskyFactor()); // L
+  const TriangularMatrix covarianceCholeskyFactorTranspose(covarianceCholeskyFactor.transpose()); // L^T
+  Point residuals(covarianceCholeskyFactorTranspose.solveLinearSystem(rho)); // L^-T rho = (Sigma^-1  - Sigma^-1 F (F^T Sigma^-1 F)^-1 F^T Sigma^-1 ) y
+
+  // Compute scales of residuals
```
sofianehaddad:
```suggestion
  // Compute scales of residuals
  // we need to compute the diagonal of the matrix $\tilde{Q}$. This diagonal is the sum of the left
  // diagonal block e.g. $Q$ and sum diagonal of the right block, e.g. $Q F \left(F^T Q F\right)^{—1}F^t Q$
  // For the first part we can easily notice that $Q_{i,i} = \sum_{j} (L^{-1}_{j,i})^2$
```

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+/* LOO constructor */
+GaussianProcessRegressionCrossValidation::GaussianProcessRegressionCrossValidation(const GaussianProcessRegressionResult & gaussianProcessRegressionResult,
+    const LeaveOneOutSplitter & splitter)
+  : MetaModelValidation(gaussianProcessRegressionResult.getOutputSample()
+                        , ComputeMetamodelLeaveOneOutPredictions(gaussianProcessRegressionResult))
+  , gaussianProcessRegressionResult_(gaussianProcessRegressionResult)
+  , splitter_ (splitter)
+{
+  const UnsignedInteger sampleDimension = gaussianProcessRegressionResult.getOutputSample().getDimension();
+  if (sampleDimension != 1)
```
sofianehaddad: This test should be done in `ComputeMetamodelLeaveOneOutPredictions` as the method is called when instanciating the object

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+{
+  return splitter_;
+}
+
+/* Compute cross-validation Leave-One-Out metamodel predictions */
+Sample GaussianProcessRegressionCrossValidation::ComputeMetamodelLeaveOneOutPredictions(
+  const GaussianProcessRegressionResult & gaussianProcessRegressionResult)
+  //const LeaveOneOutSplitter & splitter)
+{
+  const Sample outputSample(gaussianProcessRegressionResult.getOutputSample());
```
sofianehaddad, line 117: Here we suppose the used linear algebra to be `LAPACK`.
What about `HMAT`?

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+  const GaussianProcessRegressionResult &,
+  const KFoldSplitter &)
+{
+  throw NotYetImplementedException(HERE) << "Error: cannot perform KFold cross-validation of Gaussian Process Regression.";
+}
+
+/* Method save() stores the object through the StorageManager */
+void GaussianProcessRegressionCrossValidation::save(Advocate & adv) const
+{
+  PersistentObject::save(adv);
```
jschueller: MetaModelValidation

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+  adv.saveAttribute( "gaussianProcessRegressionResult_", gaussianProcessRegressionResult_ );
+}
+
+/* Method load() reloads the object from the StorageManager */
+void GaussianProcessRegressionCrossValidation::load(Advocate & adv)
+{
+  PersistentObject::load(adv);
```
jschueller: MetaModelValidation

mbaudin47:
> I'd be in favor of not exposing the kfold

I do not agree based on the following arguments.
- (In favor of avoiding K-Fold implementation) The theory states that the leave-one-out (LOO) cross-validation (CV) is always better that K-Fold: the bias is lower and the variance is lower (there is no bias-variance trade-off here).
- (In favor of avoiding K-Fold implementation) For polynomial chaos expansion, the benchmark done at [Polynomial chaos cross-validation](https://github.com/openturns/openturns/pull/2330#issuecomment-3035577886) shows that the analytical LOO CV is always faster than the other methods. This is based on a numerical simulation and on the complexity analysis.
- (In favor of implementing K-Fold)  For Kriging, we do not know which of LOO or K-Fold is faster: only a benchmark will reveal this, backed with a theoretical analysis. For some reason (and I do not have any specific point to make here), it may happen that analytical K-Fold CV may be faster than analytical LOO : only a benchmark will show this.
- (In favor of implementing K-Fold)  For research purposes, it might be interesting to experiment with both K-Fold and LOO CV. If the both methods are available, we can compare them. If only one of these methods is available, then we must implement it in Python (and OpenTURNS may be helpful for this), but the provided implementation cannot be used.

In the end, I would be highly interested in having a analytical K-Fold implementation.

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+  : GaussianProcessRegressionCrossValidation(gaussianProcessRegressionResult, LeaveOneOutSplitter(gaussianProcessRegressionResult.getOutputSample().getSize()))
+{
+  // Nothing to do
+}
+
+/* LOO constructor */
+GaussianProcessRegressionCrossValidation::GaussianProcessRegressionCrossValidation(const GaussianProcessRegressionResult & gaussianProcessRegressionResult,
+    const LeaveOneOutSplitter & splitter)
+  : MetaModelValidation(gaussianProcessRegressionResult.getOutputSample()
+                        , ComputeMetamodelLeaveOneOutPredictions(gaussianProcessRegressionResult))
```
mbaudin47: This is as good as can be. This way, the LOO code is shared by PCE and Kriging. The code is then as factored and as efficient as possible. +1

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
```diff
@@ -0,0 +1,172 @@
...
+
+/* K-Fold constructor */
+GaussianProcessRegressionCrossValidation::GaussianProcessRegressionCrossValidation(const GaussianProcessRegressionResult & gaussianProcessRegressionResult,
+    const KFoldSplitter & splitter)
+  : MetaModelValidation(gaussianProcessRegressionResult.getOutputSample()
+                        , ComputeMetamodelKFoldPredictions(gaussianProcessRegressionResult, splitter))
+  , gaussianProcessRegressionResult_(gaussianProcessRegressionResult)
+  , splitter_ (splitter)
+{
+  throw NotYetImplementedException(HERE) << "Error: cannot perform KFold cross-validation of Gaussian Process Regression.";
```
mbaudin47, line 76: I do not understand this: since `MetamodelValidation::ComputeMetamodelKFoldPredictions` is _already implemented_, why preventing the user for accessing it?

python/src/GaussianProcessRegressionCrossValidation_doc.i
josephmure: Fixed

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: Done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
sofianehaddad: `MetamodelValidation::ComputeMetamodelKFoldPredictions(inSample, outSample, residuals, projectionMatrix, splitter)` is designed for pure regression problems with the signature. It is not adapted to GP!

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: OK, done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: Ok, I have written something along these lines.

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: I have proposed something along these lines.

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: Do you want to store this in an attribute?

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: Done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: I don't know how to do that, so I am throwing an exception for the time being.

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: Done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: Done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
sofianehaddad: ok. At least raise an exception when the used linear algebra is HMAT !  I will implement it later

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
sofianehaddad: yes

josephmure: I have done this, and also created an accessor. Side effect: the computation method can no longer be static, so I also changed its name.

josephmure: And... apparently this produces a segfault.

lib/test/t_GaussianProcessRegressionCrossValidation_std.cxx
```diff
@@ -0,0 +1,100 @@
...
+    
+    // Create training sample
+    Sample x_train(7, 1);
+    x_train[0] = Point({1.0});
+    x_train[1] = Point({3.0});
+    x_train[2] = Point({4.0});
+    x_train[3] = Point({6.0});
+    x_train[4] = Point({7.9});
+    x_train[5] = Point({11.0});
+    x_train[6] = Point({11.5});
```
sofianehaddad: The worst way !
Each call to `x_train[i]` creates useless `Point`s
you can build a Sample using `Sample:::BuildFromPoint` and create easily your `Point` like this:
`const Point pt = {1.0, 3.0, 4.0, 6.0, 7.9, 11.0, 11.5};`

lib/test/t_GaussianProcessRegressionCrossValidation_std.cxx
```diff
@@ -0,0 +1,100 @@
...
+    Sample y_train = g(x_train);
+    
+    // Setup basis and covariance model
+    Basis basis = ConstantBasisFactory(1).build();
```
sofianehaddad: const

lib/test/t_GaussianProcessRegressionCrossValidation_std.cxx
```diff
@@ -0,0 +1,100 @@
...
+    fitted_covariance_model.setActiveParameter(Indices()); // no parameter optimization in CV
+    
+    for (UnsignedInteger i = 0; i < x_train.getSize(); ++i)
+    {
+        Sample x_train_cv = Sample(x_train);
+        x_train_cv.erase(i);
+        Sample y_train_cv = Sample(y_train);
+        y_train_cv.erase(i);
+        
+        GaussianProcessFitter fitter_algo_cv(x_train_cv, y_train_cv, fitted_covariance_model, basis);
```
sofianehaddad, line 66: instead of Indices() above, use  `fitter_algo_cv.setOptimizeParameters(false)`;

josephmure: It may be a matter of taste, but it seems to me the current code is better because it allows me to state the intent (no optimization of the covariance model) before getting into the "for" loop, while your way requires to do it inside the "for" loop.

lib/test/t_GaussianProcessRegressionCrossValidation_std.cxx
josephmure: OK, done.

lib/test/t_GaussianProcessRegressionCrossValidation_std.cxx
josephmure: Done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
sofianehaddad: because of the way you call the ctor..., see [my fix](https://github.com/sofianehaddad/openturns/commit/4d28dbb830213c885070bb8a8e77a4b1f53ac1a9)

josephmure: I am sorry @sofianehaddad but I really don't manage to create the attribute or the accessor. Everything I have tried has led to segfaults or errors. I believe the issue is that this `ComputeMetamodelLeaveOneOutPredictions` is really supposed to be static given that it is called at the start of the constructor. Therefore it cannot set attributes.

josephmure: I am reverting this change. For later reference, if we want to retrieve this commit, it is: https://github.com/openturns/openturns/commit/7fc2c0ae2a97eadbeb2b60250a31b0b88c40539d

sofianehaddad (approved):

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GaussianProcessFitterResult.hxx
```diff
@@ -110,9 +110,12 @@ protected:
   // GaussianProcessRegressionResult could use Cholesky setters
   friend class GaussianProcessRegressionResult;
 
-  // GaussianProcessRegressionResult could use Cholesky setters
+  // GaussianProcessRegression could use Cholesky setters
   friend class GaussianProcessRegression;
 
+  // GaussianProcessRegressionCrossValidation could use getRho()
+  friend class GaussianProcessRegressionCrossValidation;
```
jschueller: @sofianehaddad maybe we could make accessors to rho and/or cholesky public ?

python/src/experimental_module.i
```diff
@@ -79,6 +79,7 @@
 /* Uncertainty/Algorithm/Metamodel */
 %include LinearModelValidation.i
 %include EfficientGlobalOptimization.i
+%include GaussianProcessRegressionCrossValidation.i
```
jschueller, line 82: could you mention the new class in changelog ?

jschueller (approved):

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GaussianProcessFitterResult.hxx
sofianehaddad: We can.

lib/test/t_GaussianProcessRegressionCrossValidation_std.cxx
sofianehaddad: the impact is the same, but let's keep it like that

python/src/experimental_module.i
josephmure: Done

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GaussianProcessRegressionCrossValidation.cxx
josephmure: Problem solved by @sofianehaddad's commit

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GaussianProcessFitterResult.hxx
jschueller: @josephmure would you mind making accessors public instead of using friend for rho and cholesky here ?

josephmure: OK, but we also need docstrings then. I will see what I can write about rho.

jschueller: Maybe it is the log-determinant of C ? not sure

josephmure: Actually it is a standardized version of the output. Since `getRho` is now exposed, I have renamed it `getStandardizedOutput` which I think is more explicit (and also the corresponding `setRho` is renamed `setStandardizedOutput` to keep our accessors consistent, even though it stays protected).

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GaussianProcessFitterResult.hxx
```diff
@@ -113,22 +119,15 @@ protected:
   // GaussianProcessRegression could use Cholesky setters
   friend class GaussianProcessRegression;
 
-  // GaussianProcessRegressionCrossValidation could use getRho()
-  friend class GaussianProcessRegressionCrossValidation;
```
jschueller: can you try to also remove the other friend declaration above to use the new cholesky accessor ?

josephmure: This produces an error;

```
GaussianProcessRegression.cxx:120:63: error: ‘OT::HMatrix OT::GaussianProcessFitterResult::getHMatCholeskyFactor() const’ is protected within this context
  120 |     return gaussianProcessFitterResult_.getHMatCholeskyFactor().solveLower(rho, true);
```

Indeed, `getHMatCholeskyFactor` is still protected, and I am not sure we should make it public.

python/src/GaussianProcessFitterResult_doc.i
```diff
@@ -198,4 +198,23 @@ Notes
...
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::GaussianProcessFitterResult::getStandardizedOutput
+R"RAW(Accessor to the standardized output.
+
+Returns
+-------
+rho : :class:`~openturns.Point`
+    Standardized output vector :math:`\mat{L}_{\vect{p}^*}^{-1}(\vect{y} - \vect{m}_{\vect{\beta}^*(\vect{p}^*)})`.)RAW"
```
sofianehaddad: please split this and add the desctiption in a "notes" section

lib/src/Uncertainty/Algorithm/MetaModel/Kriging/openturns/GaussianProcessFitterResult.hxx
```diff
@@ -113,22 +119,15 @@ protected:
...
-
   /** Accessor to the Cholesky factor*/
   void setCholeskyFactor(const TriangularMatrix & covarianceCholeskyFactor,
                          const HMatrix & covarianceHMatrix);
 
-  /** Method that returns the covariance factor - lapack */
-  TriangularMatrix getCholeskyFactor() const;
-
   /** Method that returns the covariance factor - hmat */
   HMatrix getHMatCholeskyFactor() const;
```
sofianehaddad, line 128: we can also make it public

python/src/GaussianProcessFitterResult_doc.i
josephmure: Done.

PR #3084, opened 2026-01-27
Fix EV3 parser

regislebrun:
If a known constant such as pi_, _pi, e_ or e_ is in the expression, it is silently ignored in OT 1.25 (as well as an unknown function names).
In OT 1.26, it is seen as an unknown variable. As a result, an exception is thrown, EV3 is unable to compute the symbolic derivative and the gradient computation falls back to finite differences.
Now, the known constants are explicitely checked in ExpressionParser::IsVariableName() and EV3 is able to compute the symbolic derivative of expressions with known constants.


lib/src/Base/Diff/Ev3/parser.cxx
```diff
@@ -70,6 +70,11 @@ std::string ExpressionParser::GetVariableName(const int vid)
 
 bool ExpressionParser::IsVariableName(const std::string & vname)
 {
+  // First check if vname is a known constant
+  std::map<std::string, double>::const_iterator it;
+  for (it = table_.begin(); it != table_.end(); ++it)
+    if (vname == it->first) return false;
```
jschueller: I believe it can be written in a more compact way:
```
if (table_.count(vname) > 0)
  return false;
```

jschueller: also testing is missing

regislebrun: I was not aware of the count method (too lazy to read https://en.cppreference.com/w/cpp/container/map.html to the end)...

Issue #3095, opened 2026-02-07
Deprecate ProbabilitySimulationResult.getConfidenceLength()


mbaudin47:
### What is the idea?

When we want to estimate a probability, we may want to compute a confidence interval. The  `ProbabilitySimulationResult.getConfidenceLength()` method returns the length of the confidence interval.

<img width="530" height="183" alt="Image" src="https://github.com/user-attachments/assets/b55bf234-bf85-4cf3-9332-30ef1fd37e6e" />

This is why the code necessary to compute the confidence interval must divide the length by 2, then add and subtract the half length to the probability estimate (see [plot_axial_stressed_beam_quickstart.html](https://openturns.github.io/openturns/latest/auto_reliability/reliability_analysis/plot_axial_stressed_beam_quickstart.html?highlight=getconfidencelength)):

```python
pfHat = result.getProbabilityEstimate()
confidenceLevel = 0.95
intervalLength = result.getConfidenceLength(confidenceLevel)
pfHatInterval = [pfHat - intervalLength / 2, pfHat + intervalLength / 2]
print(
    f"{100.0 * confidenceLevel}% confidence interval = {pfHatInterval}"
)
```

This code is error-prone:
- the user migh forgot to divide the length by 2,
- the arithmetic based on `pfHat` may be wrong.

## Proposed Solution

We can get the distribution of the estimator and then compute a confidence interval:
```python
confidenceLevel = 0.95
pfHatDistribution = result.getProbabilityDistribution()
pfHatInterval = pfHatDistribution.computeBilateralConfidenceInterval(confidenceLevel)
print(
    f"{100.0 * confidenceLevel}% confidence interval = {pfHatInterval}"
)
```

## Comparison with ExpectationSimulationResult

Compare this to the `ExpectationSimulationResult` (see [ExpectationSimulationResult](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.ExpectationSimulationResult.html?highlight=expectationsimulationresult#openturns.ExpectationSimulationResult.getExpectationDistribution)): it has no `getConfidenceLength()` method. Instead, it provides the `getExpectationDistribution()` method which, in turn, makes it possible to compute a confidence interval.

### Conclusion

My conclusion is that the `ProbabilitySimulationResult.getConfidenceLength()` method should be removed:
- the `ProbabilitySimulationResult.getProbabilityDistribution()` method already does the job,
- this would be consistent with `ExpectationSimulationResult`, which has no such method.

[...]

Issue #2617, opened 2024-04-19
KDTree queryK returns incorrect results


jschueller:
### What happened?

the legacy KDTree returns repeated indices (without nanoflann)

### How to reproduce the issue?

```shell
import openturns as ot
sample = [[0.0, -1.0/3.0],
          [-0.5, -1.0/3.0],
          [0.5, -0.77777777777],
          [-0.75, -0.111111111],
          [0.25, 0.55555555555],
          [-0.25, -0.55555555555]]
tree = ot.KDTree(sample)
x = sample[1]
sorted = True
k = 6
indices = tree.queryK(x, k, sorted)
print(indices)
```
=> [1,3,0,4,0,0]
expected: [1,5,3,0,2,4]

also incorrect with k=5|sorted=False

### Version

master

### Operating System

all

### Installation media

from source

PR #3097, opened 2026-02-09
Misc

jschueller:
- #3095
- #3108
- #3114
- #3115 
- #3128
- #3135 
- Deprecate SoizeGhanemFactory
- Deprecate Collection.find in favor of Collection.index (Python side only) to mimic built-in list.index operator
- Deprecate assert_almost_equal(Indices) : no need to test integers with tolerances, use assert and ==
- Deprecate Graph showAxes/legendPosition arguments in ctor (showAxes should be True in most cases, prefer setLegendPosition to set the legend)
- meshing:
  * KDTree: Add radius search
  * #2617
  * Mesh: Add convex flag
  * Uniform/TruncatedOverMesh: protected attributes

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
+++ b/lib/src/Uncertainty/Algorithm/Optimization/EfficientGlobalOptimization.cxx
@@ -22,7 +22,6 @@
 #include "openturns/PersistentObjectFactory.hxx"
 #include "openturns/SpecFunc.hxx"
 #include "openturns/DistFunc.hxx"
-#include "openturns/KrigingAlgorithm.hxx"
 #include "openturns/MultiStart.hxx"
 #include "openturns/JointDistribution.hxx"
 #include "openturns/Uniform.hxx"


lib/src/Uncertainty/Bayesian/CalibrationResult.cxx
```diff
@@ -265,7 +266,8 @@ GridLayout CalibrationResult::drawParameterDistributions() const
       const Scalar priorPDFMax = priorJ.drawPDF().getDrawable(0).getData().getMax()[1];
       const Scalar postPDFMax = posteriorJ.drawPDF().getDrawable(0).getData().getMax()[1];
       // use log pdf if scales are very different
-      useLogScale = std::abs(std::log(postPDFMax) - std::log(priorPDFMax)) > std::log(10.0);
+      const Scalar maxPDFRatio = ResourceMap::GetAsScalar("CalibrationResult-LogScalePDFRatioThreshold");
+      useLogScale = std::abs(std::log(postPDFMax) - std::log(priorPDFMax)) > std::log(maxPDFRatio);
```
mbaudin47: This is the wrong way to fix the bug because it only makes the code more complicated without solving the initial problem. See the discussion at #3115 where I suggest to:
- remove the log-scale,
- remove xMin and xMax.

Then:
- the bugs disappear,
- the code is simpler.

python/src/PythonEvaluation.cxx
```diff
@@ -186,7 +186,7 @@ Point PythonEvaluation::operator() (const Point & inP) const
     ScopedPyObjectPointer execName(convert< String, _PyString_ >("_exec"));
     result = PyObject_CallMethodObjArgs(pyObj_, execName.get(), point.get(), NULL);
     if (! result.get())
-      PyErr_SetString(PyExc_RuntimeError, "_exec did not return any value");
+      handleException();
```
mbaudin47, line 189: +1

lib/src/Uncertainty/Bayesian/CalibrationResult.cxx
jschueller: ok

lib/src/Uncertainty/Bayesian/CalibrationResult.cxx
```diff
@@ -295,14 +297,16 @@ GridLayout CalibrationResult::drawParameterDistributions() const
...
+    Drawable postPDF(posteriorJ.drawPDF().getDrawable(0));
     if (upperRightGraph)
       postPDF.setLegend(bayesian_ ? "Posterior" : "Calibrated");
     else
       postPDF.setLegend("");
     postPDF.setColor(posteriorColor_);
     postPDF.setLineStyle(ResourceMap::GetAsString("CalibrationResult-PosteriorLineStyle"));
     graph.add(postPDF);
+    if (useLogScale)
+      graph.setLogScale(GraphImplementation::LOGY);
```
mbaudin47: This is useless in my opinion. This is because there are two calibration cases.

1. In Least Squares calibration, the starting point is represented as a single point. So there is no need to switch to the log-scale here, because the single point can always be clearly seen and compared to the distribution of the parameter after calibration.
 
2. In Gaussian Bayesian calibration, the Gaussian prior is defined by the user. So there is no need to _automatically_ switch to log-scale on the Y axis, because the user is supposed to define sounded values of the standard deviations.

Different methods were wrongly used to solve the issue.

- The first was to use (xMin, xMax) to plot the flat prior shown in #1594 in the context of least squares calibration:

Image

  This bug is now solved, since the initial point is shown as a single point, instead of a flat Gaussian with a MaxScalar standard deviation.

- The second was to use the log-PDF to compare the prior and posterior distribution in the Gaussian Bayesian calibration.

This produces:

Image

- The third (in this commit) is to plot the PDF and use the log-scale in the Y axis (only if needed, based on the difference of the prior and posterior log-PDFs).

This will produce:

Image

Actually, there is no need for these methods. The raw graph is fine:

Image

To make this clearer, I share the script at https://gist.github.com/mbaudin47/470100aa8a78090bfb904708f4f92832 so that you and @regislebrun can experiment with the different settings. From your point of view, all you have to do is find a use-case where using the log-scale on the Y axis brings any value.

I do not claim that the log-scale is useless: my point is that the user can set this log-scale afterwards, if needed. Getting automatically the log-scale plot is, from my point of view, unexpected and unnecessary in most cases.

jschueller: ok

mbaudin47: This is not ok.

mbaudin47 (changes requested):

PR #3103, opened 2026-02-17
Documentation in Symbolic evaluation and new constructors in LeastSquareExpansion

### LINKED ISSUE #2716: SymbolicEvaluation::setIn/OutputDescription does not update in/outputVariablesNames_
### What happened?

We would expect the inputDescription_  (resp. outputDescription_) attribute to be the same as the inputVariablesNames_ (resp. outputVariablesNames_) attribute, but this is not the case.

In fact, I am not sure why we need the inputVariablesNames_ and outputVariablesNames_ attributes in the first place. Maybe that is the real issue.

```shell
import openturns as ot
f = ot.SymbolicFunction(["x1", "x2"], ["x1 + x2", "x1 - x2"])
eval = f.getEvaluation().getImplementation()
eval.setOutputDescription(["f1", "f2"])
print(eval.getOutputDescription()) # output : [f1,f2] which is OK
print(eval.getOutputVariablesNames()) # output : [y0,y1] when [f1,f2] is expected
```

### Version

1.23

### Operating System

Linux

### Installation media

conda

### Additional Context

_No response_

### LINKED ISSUE #2972: The LeastSquaresExpansion class should accept a list of indices
### What is the idea?

The LeastSquaresExpansion class should accept a list of indices instead of the maximum dimension of the basis. 

Currently, the full constructor of LeastSquaresExpansion is (see https://github.com/openturns/openturns/blob/f1d942ece90d478fa2459faaf8de4dc5ee38428a/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx#L90):
```cpp
LeastSquaresExpansion(const Sample & inputSample,
    const Point & weights,
    const Sample & outputSample,
    const Distribution & distribution,
    const OrthogonalBasis & basis,
    const UnsignedInteger basisSize,
    const String & methodName)
```
Notice that the `basisSize` input argument is used to create the full list of indices (see  https://github.com/openturns/openturns/blob/f1d942ece90d478fa2459faaf8de4dc5ee38428a/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx#L105):
```cpp
  // The active functions. By default all the functions are active
  activeFunctions_ = Indices(basisSize_);
  activeFunctions_.fill();
```
This prevents from using a pre-defined list of indices. Hence, the class can only be used for creating a full least-squares PCE. 

### Why is this needed?

If the list of indices is known beforehand (e.g. based on a prior PCE or an oracle), then we may want to give to the class this list: this is not possible with the current API.

@regislebrun : I think that this may help to design new model selection algorithms.

### Additional Context

_No response_

christophedenis1:
- closes #2716 
- closes #2972

### DIFF
--- a/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
@@ -107,7 +107,7 @@ LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
   activeFunctions_.fill();
 }
 
-/* Nouveau constructeur : sélection explicite d'indices (sans poids) */
+/* Constructor */
 LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
     const Sample & outputSample,
     const Distribution & distribution,
@@ -125,7 +125,7 @@ LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
   setActiveFunctions(activeFunctions);
 }
 
-/* Nouveau constructeur : sélection explicite d'indices (avec poids) */
+/* Constructor */
 LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
     const Point & weights,
     const Sample & outputSample,
--- a/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
+++ b/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
@@ -71,7 +71,7 @@ public:
                         const UnsignedInteger basisSize = ResourceMap::GetAsUnsignedInteger("FunctionalChaosAlgorithm-BasisSize"),
                         const String & methodName = ResourceMap::GetAsString("LeastSquaresExpansion-DecompositionMethod"));
 
-  /** New constructor: explicit index selection (without weights) */
+  /** Constructor */
   LeastSquaresExpansion(const Sample & inputSample,
                         const Sample & outputSample,
                         const Distribution & distribution,
@@ -79,7 +79,7 @@ public:
                         const Indices & activeFunctions,
                         const String & methodName = ResourceMap::GetAsString("LeastSquaresExpansion-DecompositionMethod"));
 
-  /** New constructor: explicit index selection (without weights) */
+  /** Constructor */
   LeastSquaresExpansion(const Sample & inputSample,
                         const Point & weights,
                         const Sample & outputSample,


lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
```diff
@@ -1,4 +1,4 @@
-//                                               -*- C++ -*-
+/*                                               -*- C++ -*-
```
josephmure: Please revert this change.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
```diff
@@ -71,6 +71,23 @@ public:
                         const UnsignedInteger basisSize = ResourceMap::GetAsUnsignedInteger("FunctionalChaosAlgorithm-BasisSize"),
                         const String & methodName = ResourceMap::GetAsString("LeastSquaresExpansion-DecompositionMethod"));
 
+  /** Nouveau constructeur : sélection explicite d'indices (sans poids) */
```
josephmure: Please write in English

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
```diff
@@ -71,6 +71,23 @@ public:
...
+  LeastSquaresExpansion(const Sample & inputSample,
+                        const Sample & outputSample,
+                        const Distribution & distribution,
+                        const OrthogonalBasis & basis,
+                        const Indices & activeFunctions,
+                        const String & methodName = ResourceMap::GetAsString("LeastSquaresExpansion-DecompositionMethod"));
+
+  /** Nouveau constructeur : sélection explicite d'indices (avec poids) */
```
josephmure: Here too.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
```diff
@@ -118,4 +135,4 @@ private:
 
 END_NAMESPACE_OPENTURNS
 
-#endif /* OPENTURNS_LEASTSQUARESEXPANSION_HXX */
+#endif /* OPENTURNS_LEASTSQUARESEXPANSION_HXX */
```
josephmure, line 138: Please restore the empty line at the end.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -1,4 +1,4 @@
-//                                               -*- C++ -*-
+//                                              -*- C++ -*-
```
josephmure: Same as before

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -29,6 +29,7 @@
 #include "openturns/DistributionTransformation.hxx"
 #include "openturns/LeastSquaresMethod.hxx"
 #include "openturns/IdentityFunction.hxx"
+#include "openturns/Point.hxx"
```
josephmure: Is this include really necessary?

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -38,6 +39,21 @@ CLASSNAMEINIT(LeastSquaresExpansion)
 
 static const Factory<LeastSquaresExpansion> Factory_LeastSquaresExpansion;
 
+/* Removed the anonymous namespace, keep internal linkage with static */
```
josephmure: This method is not needed and should be removed.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -107,6 +123,43 @@ LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
...
+    const Distribution & distribution,
+    const OrthogonalBasis & basis,
+    const Indices & activeFunctions,
+    const String & methodName)
+  : LeastSquaresExpansion(inputSample,
+                          Point(inputSample.getSize(), 1.0 / inputSample.getSize()),
+                          outputSample,
+                          distribution,
+                          basis,
+                          ComputeBasisSizeFromActive(activeFunctions),
```
josephmure:
```suggestion
                          activeFunctions.normInf() + 1,
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -107,6 +123,43 @@ LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
...
+    const Distribution & distribution,
+    const OrthogonalBasis & basis,
+    const Indices & activeFunctions,
+    const String & methodName)
+  : LeastSquaresExpansion(inputSample,
+                          weights,
+                          outputSample,
+                          distribution,
+                          basis,
+                          ComputeBasisSizeFromActive(activeFunctions),
```
josephmure:
```suggestion
                          activeFunctions.normInf() + 1,
```

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -223,4 +276,4 @@ void LeastSquaresExpansion::load(Advocate & adv)
 
 
 
-END_NAMESPACE_OPENTURNS
+END_NAMESPACE_OPENTURNS
```
josephmure, line 263: The empty line should be restore.

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+>>> ev = f.getEvaluation().getImplementation()
+>>> ev.setOutputDescription(['f1', 'f2'])
+>>> ev.getOutputDescription()
+>>> ev.getOutputVariablesNames()
+>>> ev.getFormulas()
+"
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::clone
```
josephmure: clone is not exposed

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Virtual constructor.
+
+Returns
+-------
+clone : :class:`~openturns.SymbolicEvaluation`
+    A copy of the object."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::operator==
```
josephmure: Operators are not directly documented in Sphinx.

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+Returns
+-------
+outP : :class:`~openturns.Point`
+    Output point.
+outS : :class:`~openturns.Sample`
+    Output sample."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::getInputDimension
```
josephmure: Already documented in the parent class EvaluationImplementation

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Accessor for input dimension.
+
+Returns
+-------
+dim : int
+    Input dimension."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::getOutputDimension
```
josephmure: Same

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Accessor for output dimension.
+
+Returns
+-------
+dim : int
+    Output dimension."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::getMarginal
```
josephmure: Same

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Set the input description.
+
+This method updates the base class input description. It also applies a
+non-regressive synchronization rule between the base description and the internal
+parser variable names: internal names are updated only if they were still equal to
+the previous description."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::setOutputDescription
```
josephmure: Already documented in the parent class, but it may be useful to redocument it here, as you have done, because there is an ambiguity with the output variable names get/set methods.

You need to state the parameters of the method.

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Accessor to the symbolic formulas.
+
+Returns
+-------
+formulas : sequence of `str`
+    The formulas, one per output component, expressed using the parser variable names."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::setInputDescription
```
josephmure: Already documented in the parent class, but it may be useful to redocument it here, as you have done, because there is an ambiguity with the input variable names get/set methods.

You need to state the parameters of the method.

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Set the output description.
+
+This method updates the base class output description. It also applies a
+non-regressive synchronization rule between the base description and the internal
+parser variable names: internal names are updated only if they were still equal to
+the previous description."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::isLinear
```
josephmure: Already documneted

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Linearity accessor.
+
+Returns
+-------
+isLinear : bool
+    Whether the evaluation is linear."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::isLinearlyDependent
```
josephmure: Same

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+    Index of the output component.
+
+Returns
+-------
+isDependent : bool
+    Whether the specified component is linearly dependent."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::isParallel
```
josephmure: Same

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Whether it is safe to call in parallel.
+
+Returns
+-------
+isParallel : bool
+    Whether the evaluation can be called in parallel."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::setCheckOutput
```
josephmure: Same

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Enable/disable invalid values check on the output.
+
+Parameters
+----------
+checkOutput : bool
+    Whether to check invalid values in the output."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::save
```
josephmure: Same

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+"Store the object through the StorageManager."
+
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::load
```
josephmure: Same

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+  They are accessed with :meth:`getInputDescription` / :meth:`getOutputDescription`
+  and can be modified with :meth:`setInputDescription` / :meth:`setOutputDescription`.
+
+* The *input/output variables names* are internal names used by the symbolic parser
+  (ExprTk) to bind variables in the formulas. They are accessed with
+  :meth:`getInputVariablesNames` and :meth:`getOutputVariablesNames`.
+
+These two layers may differ. In particular, changing the output description does not
+necessarily change the internal parser variable names, because formulas must remain
+consistent with the symbols expected by ExprTk.
```
josephmure:
```suggestion
consistent with the symbols expected by the parser.
```

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+
+The overridden :meth:`setInputDescription` and :meth:`setOutputDescription` implement
+a non-regressive synchronization rule: internal parser names are updated from the new
+description only if they were still equal to the previous description. This prevents
+overwriting user-provided parser names.
```
josephmure: Does it really work this way?

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+Examples
+--------
+Define and evaluate a multi-output symbolic evaluation:
+
+>>> import openturns as ot
+>>> inNames = ['x0', 'x1']
+>>> outNames = ['y0', 'y1']
+>>> formulas = ['x0 + x1', 'x0 * x1']
+>>> ev = ot.SymbolicEvaluation(inNames, outNames, formulas)
+>>> ev([1.0, 2.0])
```
josephmure, line 58: Careful, if this is supposed to print something it might cause a test failure.
And if it does not print, then it should!

.gitignore
```diff
@@ -1,3 +1,16 @@
...
+
+# Editor/backup files
+*~
+#*#
+*.swp
+*.swo
+*.old
+*_old
+*_oldold
+*_old*
```
josephmure: Are all these gitignores really useful?

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -38,6 +39,21 @@ CLASSNAMEINIT(LeastSquaresExpansion)
 
 static const Factory<LeastSquaresExpansion> Factory_LeastSquaresExpansion;
 
+/* Removed the anonymous namespace, keep internal linkage with static */
+static UnsignedInteger ComputeBasisSizeFromActive(const Indices & activeFunctions)
```
regislebrun: This static method is useless: it is equivalent to `activeFunctions.normInf() + 1`.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -107,6 +123,43 @@ LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
...
+    const Distribution & distribution,
+    const OrthogonalBasis & basis,
+    const Indices & activeFunctions,
+    const String & methodName)
+  : LeastSquaresExpansion(inputSample,
+                          Point(inputSample.getSize(), 1.0 / inputSample.getSize()),
+                          outputSample,
+                          distribution,
+                          basis,
+                          activeFunctions.normInf() + 1,
```
regislebrun, line 122: as it can be seen here...

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -107,6 +123,43 @@ LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
...
+    const Distribution & distribution,
+    const OrthogonalBasis & basis,
+    const Indices & activeFunctions,
+    const String & methodName)
+  : LeastSquaresExpansion(inputSample,
+                          weights,
+                          outputSample,
+                          distribution,
+                          basis,
+                          activeFunctions.normInf() + 1,
```
regislebrun, line 141: and here

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+-----
+This class maintains two different naming layers:
+
+* The *input/output descriptions* (handled by :class:`~openturns.EvaluationImplementation`)
+  are user-level labels (e.g. for display, reporting, downstream processing).
+  They are accessed with :meth:`getInputDescription` / :meth:`getOutputDescription`
+  and can be modified with :meth:`setInputDescription` / :meth:`setOutputDescription`.
+
+* The *input/output variables names* are internal names used by the symbolic parser
+  (ExprTk) to bind variables in the formulas. They are accessed with
```
regislebrun: MuParser is also a valid parser

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+  :meth:`getInputVariablesNames` and :meth:`getOutputVariablesNames`.
+
+These two layers may differ. In particular, changing the output description does not
```
regislebrun: Changing either the input description or the output description has *no impact* on the variables names

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+necessarily change the internal parser variable names, because formulas must remain
+consistent with the symbols expected by the parser.
+
+The overridden :meth:`setInputDescription` and :meth:`setOutputDescription` implement
```
regislebrun: These methods are *not* overriden in this class

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,271 @@
...
+a non-regressive synchronization rule: internal parser names are updated from the new
```
regislebrun: ??? could you explain this with more details? A link to the specific C++ part of the class would help for the review (not in the docstring but in your reply)

python/src/SymbolicEvaluation_doc.i
regislebrun: This method is *not* overriden! Please remove this part.

python/src/SymbolicEvaluation_doc.i
regislebrun: Here too!

regislebrun (changes requested):
Nice contribution, but with some errors. Could you please double check the actual C++ code and your documentation?

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
christophedenis1: Done

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
christophedenis1: done

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
christophedenis1: done

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
christophedenis1: done

python/src/SymbolicEvaluation_doc.i
christophedenis1: You are right. I will avoid referring to ExprTk specifically here and replace it with a more generic wording such as “symbolic parser”, so that the docstring remains valid independently of the backend parser.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove this misleading wording and keep only the distinction between descriptions and parser variable names, without suggesting any implicit synchronization rule.

python/src/SymbolicEvaluation_doc.i
christophedenis1: You are right. This method is not overridden in SymbolicEvaluation, so I should not document any class-specific behavior here. I will remove this block.

python/src/SymbolicEvaluation_doc.i
christophedenis1: After checking the class again, I agree that this statement should not be in the docstring. I do not have a SymbolicEvaluation-specific C++ override supporting this behavior, so I will remove the claim rather than keep an undocumented assumption.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Good point. After re-reading the implementation, I agree that the current wording is too strong and suggests a behavior that is not explicitly implemented in this class. I will remove it.

python/src/SymbolicEvaluation_doc.i
christophedenis1: I will add the explicit returned value so that the example is a valid doctest.

python/src/SymbolicEvaluation_doc.i
christophedenis1: You are right. Since clone is not exposed in Python, documenting it here does not make sense. I will remove this block.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Understood. I will remove the operator== docstring block, since it is not documented this way in the Python/Sphinx API.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. This method is already documented in the parent class and does not need a specific SymbolicEvaluation docstring. I will remove it.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove this redundant block as well.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove it since it is already documented in the parent API.

christophedenis1: Thanks. Since the method is not overridden in SymbolicEvaluation, I will remove this block rather than redocument it here.

python/src/SymbolicEvaluation_doc.i
christophedenis1: You are right. This method is not overridden in SymbolicEvaluation. I will remove this block.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove it.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove it.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove it.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove it.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove it.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Agreed. I will remove this redundant block.

python/src/SymbolicEvaluation_doc.i
christophedenis1: Thanks. Since the method is not overridden in SymbolicEvaluation, I will remove this block rather than redocument it here.

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
christophedenis1: maybe paramatrization to be changed in VSCODE (to be checked)

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -1,4 +1,4 @@
-//                                               -*- C++ -*-
+//                                               -*- C++ -*- */
```
christophedenis1: to be modified

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -29,6 +29,7 @@
 #include "openturns/DistributionTransformation.hxx"
 #include "openturns/LeastSquaresMethod.hxx"
 #include "openturns/IdentityFunction.hxx"
+//#include "openturns/Point.hxx"
```
christophedenis1: comment to be deleted

python/src/SymbolicEvaluation_doc.i
```diff
@@ -0,0 +1,109 @@
...
+// ---------------------------------------------------------------------
+
+%feature("docstring") OT::SymbolicEvaluation::getFormulas
+R"RAW(Accessor to the symbolic formulas.
+
+Returns
+-------
+formulas : sequence of `str`
+    The formulas, one per output component.
+)RAW"
```
christophedenis1: add newline at the end

josephmure:
The build_linux machine had errors, but I think they are unrelated to this PR. What do you think @jschueller ?

jschueller:
yes, probably just out of sync with master, but will have to rebase to validate doc changes

python/src/LeastSquaresExpansion_doc.i
jschueller, line 43: add activeFunctions

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -107,6 +107,43 @@ LeastSquaresExpansion::LeastSquaresExpansion(const Sample & inputSample,
...
+                          weights,
+                          outputSample,
+                          distribution,
+                          basis,
+                          activeFunctions.normInf() + 1,
+                          methodName)
+{
+  setActiveFunctions(activeFunctions);
+}
+
```
jschueller, line 146: no tests

lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/openturns/LeastSquaresExpansion.hxx
jschueller, line 58: other proposal: replace basisSize by the activeFunctions

jschueller:
thanks, symbolic changes moved to my branch

PR #3129, opened 2026-03-11
Improved SpecFunc

### LINKED ISSUE #2992: SpecFunc::Clip01 does not manage for a tolerance
### What is the idea?

I want to compute a probability which is inside the $[\epsilon, 1 - \epsilon]$ interval, where $\epsilon = 10^{-15}$. This should be a job for `SpecFunc.Clip01`, but the function does not provide a tolerance. I can only project the probability into $[0, 1]$.

In the next example, the value of `p` is 1 and the expected value of `cliped_p` is `1.0 - 1.0e-15`.

```python
import openturns as ot
p = 1.0
cliped_p = ot.clip01(p)  # Ok, but does nothing!
tail = False
epsilon = 1.0e-15
cliped_p = ot.clip01(p, tail, epsilon)  # Does not exist yet.
```

### Why is this needed?

I consider the logistic regression model. This involves the likelihood of a binomial model (see [^Wasserman2006] page 65). The PDF of a binomial is $P(Y = y) = p^y (1 - p)^{1 - y}$ where $Y \sim \mathcal{B}(p)$ and $p \in [0,1]$ is the probability. The likelihood of the logistic regression model is:

$$
L(\boldsymbol{\beta}) = \prod_{i = 1}^n p_i^{y_i} (1 - p_i)^{1 - Y_i}.
$$

The maximum likelihood estimator maximizes the log-likelihood:

$$
\log(L(\boldsymbol{\beta})) = \sum_{i = 1}^n y_i \log(p_i) + (1 - y_i) \log(1 - p_i)
$$

for $\boldsymbol{\beta}$.

In my case some probabilities $p_i$ are exactly equal to 0 or 1. Then I compute the logarithm of $p_i$ or $1 - p_i$, depending on the label $y_i$ (0 or 1). If $p_i = 0$ (and the label $y_i = 0$) or $p_i = 1$ (and the label $y_i= 1$), this leads to log(0) which produces a _math domain error_. 

### Additional Context

The workaround is:
```python
import numpy as np
p = 1.0
epsilon = 1.0e-15
cliped_p = np.clip(p, epsilon, 1.0 - epsilon)
```

[^Wasserman2006]: Wasserman, L. (2006). All of Nonparametric Statistics. Springer, New York, NY.

regislebrun:
The clip(), BesselInu and logBesselInu() functions have been added. The first one allows to project a given scalar onto a given interval, which fixes #2992. The two other functions are based on https://arxiv.org/html/2409.08729v1 to allow for arbitrarily large nu and x. These functions are needed to compute the Von-Mises Fisher multivariate density distribution in large dimension.


python/test/t_SpecFunc_std.py
```diff
@@ -45,3 +45,142 @@
...
+# General case
+#
+x = 15
+nu = 12.5
+y = ot.SpecFunc.LogBesselInu(x, nu)
+yref = 7.6416656645837741856
+assert abs(1 - y / yref) < 1e-15
+y = ot.SpecFunc.BesselInu(x, nu)
+yref = 2083.2108588531338025
+assert abs(1 - y / yref) < 1e-14
```
jschueller: I would try to write these in a loop to deduplicate the code:
```
v = [[15,12.5,7.64166, 2083.21],
        [5, 13.5, -11.067, 1.560997],
        ...]
for x, nu, ref_log, ref in v:
    y_log = ot.SpecFunc.LogBesselInu(x, nu)
    y = ot.SpecFunc.BesselInu(x, nu)
    ...
```
maybe include the tolerances as well if they need to change
also I would use assert_almost_equal with a null absolute tolerance:
`ott.assert_almost_equal(y, yref, 1e-15, 0.0)`

lib/src/Base/Func/SpecFunc/openturns/SpecFunc.hxx
```diff
@@ -76,6 +76,9 @@ OT_API Scalar LogBesselI0(const Scalar x);
 // Modified first kind Bessel function of order 1: BesselI1(x) = \sum_{m=0}\infty\frac{1}{m!(m+1)!}\left(\frac{x}{2}\right)^{2m+1}
 OT_API Scalar BesselI1(const Scalar x);
 OT_API Scalar LogBesselI1(const Scalar x);
+// Modified first kind Bessel function of order nu: BesselInu(x) = \sum_{m=0}\infty(x/2)^{\nu+2m}/(m!Gamma(\nu+m+1))
+OT_API Scalar BesselInu(const Scalar x, const Scalar nu);
+OT_API Scalar LogBesselInu(const Scalar x, const Scalar nu);
```
jschueller, line 81: we need to add these into python/doc/user_manual/functional_modeling.rst

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -236,6 +236,150 @@ Scalar SpecFunc::DeltaLogBesselI10(const Scalar x)
...
+// Andreas Plesner, Hans Henrik Brandenborg Sørensen, Søren Hauberg,
+// "Accurate Computation of the Logarithm of Modified Bessel Functions on GPUs"
+Scalar SpecFunc::LogBesselInu(const Scalar x, const Scalar nu)
+{
+  if (!(nu > 0.0)) throw InvalidArgumentException(HERE) << "Expected a positive value for nu, got nu=" << nu;
+  if (x <= 0.0) return LowestScalar;
+  const Scalar logX = std::log(x);
+  const Scalar logNu = std::log(nu);
+  Scalar value = 0.0;
+#define MU_CODE(kMax) \
```
jschueller: please avoid macros for readability and safety, a lambda would be practical here (rewritten by llm):
```
const Scalar asymptoticMu = [&](UnsignedInteger kMax)
  {
    Scalar value = x - 0.5 * logX - LOGSQRT2PI;

    Scalar sum = 0.0;
    Scalar term = 1.0;
    const Scalar mu = 4.0 * nu * nu;

    for (UnsignedInteger k = 1; k <= kMax; ++k)
    {
      term *= -(mu - (2 * k - 1) * (2 * k - 1)) / (k * 8.0 * x);
      sum += term;
    }

    value += std::log(std::abs(1.0 + sum));
    return value;
  };
...
if (...)
  return asymptoticMu(4);

File: lib/src/Base/Func/SpecFunc/SpecFunc.cxx [hunk 4]
```diff
@@ -236,6 +236,150 @@ Scalar SpecFunc::DeltaLogBesselI10(const Scalar x)
...
+#define U13 (-1.4679261247695617e10 * (t + 0.99685899317493083)) * (t + 0.98410163838326928) * (t + 0.95406147056411588) * (t + 0.90418742233360804) * (t + 0.83366587749170256) * (t + 0.74328958239963429) * (t + 0.63485626061198897) * (t + 0.51083917920458419) * (t + 0.37420880072317122) * (t + 0.22830704344590139) * (t + 0.76737872420980019e-1) * (t - 0.76737872420980019e-1) * (t - 0.22830704344590139) * (t - 0.37420880072317122) * (t - 0.51083917920458419) * (t - 0.63485626061198897) * (t - 0.74328958239963429) * (t - 0.83366587749170256) * (t - 0.90418742233360804) * (t - 0.95406147056411588) * (t - 0.98410163838326928) * (t - 0.99685899317493083) * (t * t + 2.0002534348020824 * t + 1.0002541228185118) * (t * t - 2.0002534348020824 * t + 1.0002541228185118)
+#define U_PREAMBLE \
```
jschueller: same here

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
regislebrun: I had to replace `const Scalar asymptoticMu` by `auto asymptoticMu` to make it work. I didn't knew this kind of constructions, I am still stuck with C++ from the last century!
Do I have to try something like this for the U case? I don't see how I can do this (I would have to return value, t and tOverNu).

jschueller: yes! we should use constexpr:
```
constexpr double U1(double t) {
    return (-0.20833333333333333 * (t + 0.77459666924148338))
         * (t - 0.77459666924148338);
}

constexpr double U2(double t) {
    return (0.33420138888888889 * (t + 0.99337556982854302))
         * (t + 0.46174124492817127)
         * (t - 0.46174124492817127)
         * (t - 0.99337556982854302);
}
```

lib/src/Base/Func/SpecFunc/openturns/SpecFunc.hxx
regislebrun: done

python/test/t_SpecFunc_std.py
regislebrun: done

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
regislebrun: èdone

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
regislebrun: Here I don't know how to do that (for U_PREAMBLE)

jschueller: this looks ok (llm suggestion):
```

auto uPreamble = [&](Scalar x, Scalar nu, Scalar value)
{
  const Scalar xp = x / nu;
  const Scalar invt = std::sqrt(1.0 + xp * xp);
  const Scalar t = 1.0 / invt;
  const Scalar tOverNu = t / nu;
  const Scalar eta = invt + std::log(xp / (1.0 + invt));

  value += eta * nu - 0.5 * logNu - LOGSQRT2PI - 0.25 * log1p(xp * xp);

  return std::tuple{value, tOverNu};
};
if (((x > 274.2377) && (nu > 0.3)) || (nu > 163.6993))
{
  auto [value2, tOverNu] = uPreamble(x, nu, value);
  const Scalar sum = tOverNu * (U1 + tOverNu * (U2 + tOverNu * (U3 + tOverNu * U4)));
  return value2 + log1p(sum);
}
...
```

regislebrun: done

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -236,6 +236,196 @@ Scalar SpecFunc::DeltaLogBesselI10(const Scalar x)
...
+  };
+  auto U13 = [](const Scalar t)
+  {
+    return (-1.4679261247695617e10 * (t + 0.99685899317493083)) * (t + 0.98410163838326928) * (t + 0.95406147056411588) * (t + 0.90418742233360804) * (t + 0.83366587749170256) * (t + 0.74328958239963429) * (t + 0.63485626061198897) * (t + 0.51083917920458419) * (t + 0.37420880072317122) * (t + 0.22830704344590139) * (t + 0.76737872420980019e-1) * (t - 0.76737872420980019e-1) * (t - 0.22830704344590139) * (t - 0.37420880072317122) * (t - 0.51083917920458419) * (t - 0.63485626061198897) * (t - 0.74328958239963429) * (t - 0.83366587749170256) * (t - 0.90418742233360804) * (t - 0.95406147056411588) * (t - 0.98410163838326928) * (t - 0.99685899317493083) * (t * t + 2.0002534348020824 * t + 1.0002541228185118) * (t * t - 2.0002534348020824 * t + 1.0002541228185118);
+  };
+  auto uPreamble = [&](const Scalar x, const Scalar nu)
```
jschueller: x is already captured by "[&]", try:
```suggestion
  auto uPreamble = [&](const Scalar nu)
```

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -236,6 +236,196 @@ Scalar SpecFunc::DeltaLogBesselI10(const Scalar x)
...
+    
+    const Scalar value = eta * nu - 0.5 * logNu - LOGSQRT2PI - 0.25 * log1p(xp * xp);
+    
+    return std::tuple{t, tOverNu, value};
+  };
+  
+  // Third case, use U4
+  if (((x > 274.2377) && (nu > 0.3)) || (nu > 163.6993))
+    {
+      auto [t, tOverNu, value] = uPreamble(x, nu);
```
jschueller:
```suggestion
      auto [t, tOverNu, value] = uPreamble(nu);
```

lib/src/Base/Func/SpecFunc/SpecFunc.cxx
```diff
@@ -236,6 +236,196 @@ Scalar SpecFunc::DeltaLogBesselI10(const Scalar x)
...
+  if (((x > 35.9074) && (nu > 0.6)) || (nu > 20.1534))
+    {
+      auto [t, tOverNu, value] = uPreamble(x, nu);
+      const Scalar sum = tOverNu * (U1(t) + tOverNu * (U2(t) + tOverNu * (U3(t) + tOverNu * (U4(t) + tOverNu * (U5(t) + tOverNu * (U6(t) + tOverNu * (U7(t) + tOverNu * (U8(t) + tOverNu * U9(t)))))))));
+      return value + log1p(sum);
+    }
+  // Sixth case, use U13
+  if (((x > 19.6931) && (nu > 0.7)) || (nu > 12.6964))
+    {
+      auto [t, tOverNu, value] = uPreamble(x, nu);
```
jschueller: the other ones as well

regislebrun: done

PR #3140, opened 2026-03-24
Fix formatting in GaussianProcessFitter_doc.i

mbaudin47:
The goal of this PR is to fix this sentence. It lets the user thinks that the equation (1) is inside the current `GaussianProcessFitter` help page, but it is not: the actual equation is located inside a theory help page.

image

**Figure 1.** Before the commit.

image

**Figure 2.** After the commit.


mbaudin47:
The current commit produces:

image

This is not the expected output. @jschueller : Any other idea?

regislebrun (approved):

PR #3145, opened 2026-03-24
FCA: Add domination method flag

jschueller:
see #3099

PR #3151, opened 2026-04-01
Improved [LeastSquares|Integration]Expansion

regislebrun:
When all the parameters are given to the constructor of these classes, there is no need to call a constructor of the base class which trigger the construction of costly objects. It is fixed now.


lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/LeastSquaresExpansion.cxx
```diff
@@ -137,23 +140,28 @@ void LeastSquaresExpansion::run()
       transformedInputSample = transformation_(inputSample_);
     }
     FunctionCollection functions(basisSize_);
+    std::cerr << "Build basis" << std::endl;
```
jschueller: debug

regislebrun: fixed

jschueller (approved):

jschueller:
moved to #3145 since it conflicts

PR #3155, opened 2026-04-04
Fixed CompositeDistribution

regislebrun:
The computePDF() method generates nan/inf when evaluated at a point where the function derivative is exactly zero.


python/test/t_CompositeDistribution_std.py
```diff
@@ -131,3 +131,6 @@
 validation.skipMinimumVolumeLevelSet()  # slow
 validation.skipPDFAtLowerBound()  # Testing that the PDF is null at its lower bound if finite is a complex task
 validation.run()
+
+# Check that the PDF is zero where it is given by a fraction with null denominator
+ott.assert_almost_equal(ot.Uniform(-5.0, 5.0).sqr().computePDF(0.0), 0.0)
```
jschueller: maybe we can assert ot.Uniform(-5.0, 5.0).sqr().computePDF(0.0) == 0.0

sofianehaddad (approved):

PR #3158, opened 2026-04-06
Fix/clang18 correlationmatrix eq

prudhomm:
use explicit Matrix casts instead of CorrelationMatrix == CorrelationMatrix to avoid compilation error using clang 18


jschueller:
hello Christophe,

Could you tell us more about the error you got ? which version did you use ? master or v1.26 ?
I tried clang 18.1.8 on linux on both v1.26 and master and got no error, maybe its apple clang ?
Also maybe e180e97 already fixed this (in master).

jschueller:
We are releasing a new version soon @prudhomm, so dont hesitate to give us more info to help getting this fixed before the target date

Issue #2709, opened 2024-06-27
ExperimentIntegration should inherit from IntegrationAlgorithmImplementation


mbaudin47:
### What happened?

In OpenTURNS 1.23, ExperimentIntegration  inherits from PersistentObject. This make it impossible to use it as a standard IntegrationAlgorithm. One of the problems is that the IntegrationAlgorithm.integrate(function, interval) method cannot be directly used with a WeightedExperiment: it must define how the interval is used. One method is to integrate:

$$
\int_{I} \frac{g(x)}{w(x)}} dx
$$

### How to reproduce the issue?

```shell
TODO
```

### Version

1.23

### Operating System

other

### Installation media

other

jschueller:
@mbaudin47 in the second formula w0(z) is both at numerator and denominator

mbaudin47:
> @mbaudin47 in the second formula w0(z) is both at numerator and denominator

This is the correct formula in my opinion :

In this method, the integrand is the ratio $\frac{g(T(z))}{w_0(z)}$, which is integrated on the interval $I_0$ with respect to the weight function $w_0$.

jschueller:
I mean it can be simplified right ?

mbaudin47:
> I mean it can be simplified right ?

Assuming we use a quadrature rule associated with the weight function $f_0$ for the integrand $h$, the integral we consider is $\int_{I_0} h(z) f_0(z) dz$. Therefore, the weight $f_0$ does not simplify because the weight and nodes of the quadrature take $f_0$ into account.

jschueller:
#3165

PR #3165, opened 2026-04-15
Improved ExperimentIntegration

regislebrun:
Now, it inherits from the IntegrationAlgorithmImplementation class. 
A new integrate(function, interval) method has been defined as the integral of the function weighted by the distribution linearly scaled to have its numerical range equal to the given interval.


jschueller (approved):
nice, just mention #2709

Issue #3072, opened 2026-01-15
RandomMixture has a wrong name > LinearCombinationDistribution


mbaudin47:
### What is the idea?

The `RandomMixture` implements a linear combination of independent distributions.

The `WeightedSumDistribution` name would be a better name than `RandomMixture`. This is because a **Mixture** is a very different object (see [Mixture_distribution](https://en.wikipedia.org/wiki/Mixture_distribution) and [Mixture.html](https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.Mixture.html)).

### 1. The purpose of the class

The class currently designated as `RandomMixture` is designed to compute the probability distribution resulting from a linear combination of independent random variables. It facilitates the representation of a random variable  defined by the following mathematical expression:

$$Y = \sum_{i=1}^{n} a_i X_i$$

In this equation,  represents a set of independent random distributions and  represents a set of deterministic coefficients or weights. The class provides the necessary logic to perform the convolution of these distributions while accounting for the scaling factors. It allows the user to evaluate the probability density function, the cumulative distribution function, and the statistical moments of the resulting distribution.

### 2. The current class name and source of confusion

The current name, `RandomMixture`, is mathematically inaccurate and creates significant confusion for the following reasons:

* **Mathematical inconsistency**: In probability theory and statistics, a "Mixture Distribution" refers to a convex combination of probability density functions. This is expressed as $f_Y(y) = \sum w_i f_{X_i}(y)$. The class does not implement this operation. It implements a sum of random variables.
* **Terminology conflict**: A mixture distribution and a sum of random variables are distinct concepts. A mixture corresponds to a weighted average of densities, whereas a sum of variables corresponds to the convolution of their distributions.
* **Redundancy**: The prefix "Random" is unnecessary because any mixture or sum of distributions is inherently a stochastic object.

Using the term "Mixture" for a sum of variables leads users to select the wrong class for their specific modeling needs.

### Why is this needed?

The wrong name implies confusions.

mbaudin47:
Let's sum up the discussion so far.

| Proposed Name | Main advantage |
| --- | --- |
| **LinearCombinationDistribution** | This name provides the most precise mathematical description of the operation. |
| **WeightedSumDistribution** | This name uses simple and direct terminology that is accessible to all practitioners. |
| **ConvolutionDistribution** | This name highlights the formal mathematical operation used to derive the distribution of the sum. |

The name **LinearCombinationDistribution** is the most suitable choice in my opinion.

[...]

Issue #2519, opened 2024-01-22
UserDefined => DiscreteNonParametric


jschueller:
### What is the idea?

the UserDefined distribution name is rather vague, I propose to rename to Categorical or generalized Bernoulli,
that's what we can find in the litterature:
https://en.wikipedia.org/wiki/Categorical_distribution

Julia distribution lib uses Categorical:
https://juliastats.org/Distributions.jl/stable/univariate/#Distributions.Categorical

regislebrun:
The current distribution, while being probably misnamed, is *not* a categorical distribution, but a discrete distribution over $\mathbb{R}^n$. As such, it has a well defined expectation, a covariance matrix, a cumulative distribution function. A categorical distribution is defined over an unordered set and the concept of cumulative distribution function is meaningless for these distributions.

regislebrun:
Something like FiniteSupportDiscreteDistribution?

jschueller:
I misread the definition: Categorical was used when the support is the set of natural integers,
but when the support is arbitrary in R they name it "DiscreteNonParametric":
https://juliastats.org/Distributions.jl/stable/univariate/#Distributions.DiscreteNonParametric
what do you think of that ?

sofianehaddad:
`CustomDiscreteDistribution` ?
Does it mean also that all `UserDefinedXXX` classes should be renamed ?

jschueller:
if we can find a meaningful name

jschueller:
also worth noting chatgpt suggestion: WeightedDiscreteDistribution

1. Most accurate / standard in probability theory
DiscreteEmpiricalDistribution
WeightedDiscreteDistribution
FiniteDiscreteDistribution

👉 Best general-purpose replacements.

2. If it’s meant to represent data-driven statistics (very common in OpenTURNS use)
EmpiricalDistribution
WeightedEmpiricalDistribution
SampleBasedDistribution

👉 Emphasizes “built from data”.

3. If you want to stress the mathematical structure
AtomicDistribution
DiscreteAtomicDistribution

👉 “atomic” = sum of Dirac masses (more formal, less user-friendly)

4. If you want a very explicit API-style name
DiscretePointsDistribution
PointMassDistribution
WeightedPointDistribution

👉 Very clear but slightly less “statistical” sounding.

🏆 My recommendations (best balance)

If I had to pick one:

⭐ Best overall
WeightedDiscreteDistribution

Why:

explicit
standard statistical meaning
no ambiguity
[...]

Issue #2480, opened 2023-11-20
sphinx 6+ compatibility


jschueller:
### What happened?

the search is broken with sphinx 6+
a some point the default search will have to be used instead of the button selection as sphinx<6 wont be installable forever
the selection would have to be reimplemented using the new apis

### How to reproduce the issue?

```shell
compatibility
```

### Version

master

### Operating System

unknown

### Installation media

unknown

jschueller:
TODO: drop jquery (sphinxcontrib.jquery)

PR #3167, opened 2026-04-16
Misc

jschueller:
- #2480
- #2519
- #2716
- #3072 
- #3170
- SWIG stuff


lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -306,6 +307,33 @@ Distribution TruncatedDistribution::getSimplifiedVersion() const
   return useSimplifiedVersion_ ? simplifiedVersion_ : *this;
 }
 
+void TruncatedDistribution::computeMean() const
+{
+  if (useSimplifiedVersion_)
+  {
+    if (!isAlreadyComputedMean_)
```
regislebrun: No need to test this, as this function is called only by getMean(), which has this protection

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -306,6 +307,33 @@ Distribution TruncatedDistribution::getSimplifiedVersion() const
...
+  }
+  else
+    DistributionImplementation::computeMean();
+}
+
+void TruncatedDistribution::computeCovariance() const
+{
+  if (useSimplifiedVersion_)
+  {
+    if (!isAlreadyComputedCovariance_)
```
regislebrun: same here

regislebrun (approved):
Nice! Only a minor change for me.

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -223,13 +223,14 @@ Bool TruncatedDistribution::hasSimplifiedVersion(Distribution & simplified) cons
   }
   // If no truncation
   const Interval range(getRange());
-  if (localDistribution.getRange() == range)
+  if (localDistribution.getRange().getLowerBound() == range.getLowerBound()
```
regislebrun: The tests should be reorganized like in PR#3168 in order to address the most radical situations first

regislebrun (changes requested):

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
jschueller: maybe it would be easier to apply my 2 lines fix in your branch, with the small test also

regislebrun: Done.

regislebrun (approved):

PR #3168, opened 2026-04-17
Improve truncated distribution

### LINKED ISSUE #3162: Regression in TruncatedDistribution moments on multivariate KernelSmoothing between 1.23 and 1.24 or later
### What happened?

I observed a significant slowdown in OpenTURNS 1.24 compared to 1.23 when using a
`TruncatedDistribution` built from a multivariate kernel-smoothed distribution.

More precisely, the slowdown does **not** come from:

- `KernelSmoothing().build(sample)` itself
- the raw KDE distribution returned by `build(sample)`

The slowdown appears specifically when wrapping the KDE into:

```python
    ks = ot.KernelSmoothing()
    kde = ks.build(sample)
    trunc_kde = ot.TruncatedDistribution(kde, prior_range)
```

and then calling:
  + getMean()
  + getCovariance()

on the resulting truncated distribution.
The same operations are much faster in 1.23 on the exact same sample.
From the public API documentation, TruncatedDistribution appears to expose the same
constructors and the same general interface in 1.23 and 1.24, so this looks like an
internal behavioral/performance regression rather than an API-level change.

The following script reproduces the issue on my side.
It uses a fixed 3D sample of size 13 (exported from my application), builds a KDE with
KernelSmoothing, then compares the raw KDE against the truncated version.
[conditional_sample_1p23.csv](https://github.com/user-attachments/files/26644672/conditional_sample_1p23.csv)

# Expected behavior
I would expect TruncatedDistribution(KernelSmoothing(sample), interval) to have
roughly similar performance in 1.23 and 1.24 for moment-related methods such as
getMean() and getCovariance(), especially on such a small sample (n=13, dim=3).
The raw KDE itself does not show any slowdown, so the slowdown appears to come
specifically from the truncated wrapper.

```Python
# MWE_KS_TRUNC_1.py
import time
import openturns as ot


def load_sample():
    sample = ot.Sample.ImportFromCSVFile("conditional_sample_1p23.csv")
    return sample


def make_prior_range():
    lower = [-0.01, -0.01, 2.0]
    upper = [ 0.01,  0.01, 4.0]
    interval = ot.Interval(lower, upper)
    return interval


def bench_distribution(label, dist):
    print(f"\n--- {label} ---")
    print("class =", dist.getClassName())

    t0 = time.perf_counter()
    m = dist.getMean()
    t1 = time.perf_counter()
    print(f"getMean       = {t1 - t0:.6f}s  {m}")

    t2 = time.perf_counter()
    c = dist.getCovariance()
    t3 = time.perf_counter()
    print(f"getCovariance = {t3 - t2:.6f}s")

    x = dist.getMean()
    t4 = time.perf_counter()
    p = dist.computePDF(x)
    t5 = time.perf_counter()
    print(f"computePDF    = {t5 - t4:.6f}s  {p}")


def main():
    print("OpenTURNS version:", ot.__version__)
    sample = load_sample()
    print("size =", sample.getSize(), "dim =", sample.getDimension())
    print("mean =", sample.computeMean())
    print("cov  =", sample.computeCovariance())

    ks = ot.KernelSmoothing()

    t0 = time.perf_counter()
    kde = ks.build(sample)
    t1 = time.perf_counter()
    print(f"\nKS build = {t1 - t0:.6f}s")
    print("KS class =", kde.getClassName())

    bench_distribution("KDE raw", kde)

    prior_range = make_prior_range()

    t2 = time.perf_counter()
    trunc_kde = ot.TruncatedDistribution(kde, prior_range)
    t3 = time.perf_counter()
    print(f"\nTruncatedDistribution build = {t3 - t2:.6f}s")
    print("Truncated class =", trunc_kde.getClassName())

    bench_distribution("KDE truncated", trunc_kde)


if __name__ == "__main__":
    main()
```

### Version

OpenTURNS version: 1.24
size = 13 dim = 3
mean = [-0.000335486,-0.000198225,2.94412]
cov  = [[  6.91801e-07 -1.13473e-07 -2.64386e-05 ]
 [ -1.13473e-07  5.01673e-07 -1.92541e-05 ]
 [ -2.64386e-05 -1.92541e-05  0.00817199  ]]

KS build = 0.000063s
KS class = Distribution

--- KDE raw ---
class = Distribution
getMean       = 0.000009s  [-0.000335486,-0.000198225,2.94412]
getCovariance = 0.000007s
computePDF    = 0.000007s  651706.1989233252

TruncatedDistribution build = 0.000041s
Truncated class = TruncatedDistribution

--- KDE truncated ---
class = TruncatedDistribution
getMean       = 1.136807s  [-0.000335486,-0.000198225,2.94412]
getCovariance = 1.529823s
computePDF    = 0.000022s  651706.1989232965

OpenTURNS version: 1.23
size = 13 dim = 3
mean = [-0.000335486,-0.000198225,2.94412]
cov  = [[  6.91801e-07 -1.13473e-07 -2.64386e-05 ]
 [ -1.13473e-07  5.01673e-07 -1.92541e-05 ]
 [ -2.64386e-05 -1.92541e-05  0.00817199  ]]

KS build = 0.000069s
KS class = Distribution

--- KDE raw ---
class = Distribution
getMean       = 0.000008s  [-0.000335486,-0.000198225,2.94412]
getCovariance = 0.000007s
computePDF    = 0.000009s  651706.1989233252

TruncatedDistribution build = 0.000032s
Truncated class = TruncatedDistribution

--- KDE truncated ---
class = TruncatedDistribution
getMean       = 0.001286s  [-0.000335486,-0.000198225,2.94412]
getCovariance = 0.031008s
computePDF    = 0.000820s  651706.1989232958

### Operating System

Linux

### Installation media

other

### Additional Context

A few remarks that may help narrow down the issue:

  1. KernelSmoothing().build(sample) alone is fast in both 1.23 and 1.24.
  2. getMean() / getCovariance() on the raw KDE are also fast in both versions.
  3. The slowdown appears only on the truncated KDE.
  4. computePDF() on the truncated distribution remains fast, so the issue seems specific to moment-related computations rather than to all methods.
  5. The sample is multivariate (dim=3), so this is not related to the 1D log-transform path introduced in KernelSmoothing 1.24.
  6. Public issues around TruncatedDistribution suggest that multivariate truncated distributions can be subtle/costly in some contexts, but I could not find this particular regression documented.

regislebrun:
Now, the simplification algorithm takes into account the KernelMixture case.
It improves things only if the dimension is greater than or equal to 3.
It fixes #3162.

### DIFF
--- a/lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
+++ b/lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
@@ -31,6 +31,8 @@
 #include "openturns/JointDistribution.hxx"
 #include "openturns/BlockIndependentDistribution.hxx"
 #include "openturns/BlockIndependentCopula.hxx"
+#include "openturns/KernelMixture.hxx"
+#include "openturns/Mixture.hxx"
 #include "openturns/Dirichlet.hxx"
 #include "openturns/Beta.hxx"
 
@@ -180,7 +182,27 @@ Distribution TruncatedDistribution::dispatchTruncation(const Collection<Distribu
 /* Get the simplified version (or clone the distribution) */
 Bool TruncatedDistribution::hasSimplifiedVersion(Distribution & simplified) const
 {
+  // Delve into the antecedents until we get something which is not a truncated distribution
+  Distribution localDistribution(distribution_);
+  String kind(localDistribution.getImplementation()->getClassName());
+  UnsignedInteger level = 1;
+  while (kind == "TruncatedDistribution")
+  {
+    const TruncatedDistribution * truncatedDistribution = dynamic_cast<const TruncatedDistribution *>(localDistribution.getImplementation().get());
+    localDistribution = truncatedDistribution->getDistribution();
+    kind = localDistribution.getImplementation()->getClassName();
+    ++ level;
+    // nested truncation intervals have already been intersected during the nested range computations
+  }
   // n-D case
+  // If no truncation
+  const Interval range(getRange());
+  if ((localDistribution.getRange().getLowerBound() == range.getLowerBound()) &&
+      (localDistribution.getRange().getUpperBound() == range.getUpperBound()))
+  {
+    simplified = localDistribution;
+    return true;
+  }
   JointDistribution *p_joint = dynamic_cast<JointDistribution *>(distribution_.getImplementation().get());
   if (p_joint && p_joint->hasIndependentCopula())
   {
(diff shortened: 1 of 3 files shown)


python/test/t_TruncatedDistribution_std.py
jschueller, line 233: add this test:
```
# check simplification logic only checks numerical range
dist = ot.Gumbel(1.0, 0.0)
numericalRange = ot.Interval(dist.getRange().getLowerBound(), dist.getRange().getUpperBound())
truncated = ot.TruncatedDistribution(dist, numericalRange)
assert truncated.getSimplifiedVersion() == dist
```

python/test/t_TruncatedDistribution_std.py
```diff
@@ -214,7 +214,6 @@ def cleanCovariance(inCovariance):
     ot.BlockIndependentDistribution([ot.Normal(2), ot.Normal(2)]),
     ot.BlockIndependentCopula([ot.NormalCopula(2), ot.NormalCopula(2)]),
     ot.Dirichlet([0.7, 0.3]),
-    ot.TruncatedDistribution(ot.Normal(0.0, 1.0), ot.Interval(-1.0, 1.0)),
```
jschueller, line 217: why did you remove this one ? it checks the simplification goes into truncatednormal

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
```diff
@@ -180,7 +182,27 @@ Distribution TruncatedDistribution::dispatchTruncation(const Collection<Distribu
...
+    localDistribution = truncatedDistribution->getDistribution();
+    kind = localDistribution.getImplementation()->getClassName();
+    ++ level;
+    // nested truncation intervals have already been intersected during the nested range computations
+  }
   // n-D case
+  // If no truncation
+  const Interval range(getRange());
+  if ((localDistribution.getRange().getLowerBound() == range.getLowerBound()) &&
+      (localDistribution.getRange().getUpperBound() == range.getUpperBound()))
```
jschueller, line 201: mention #3162 in the commit message

python/test/t_TruncatedDistribution_std.py
regislebrun: I compared t_TruncatedDistribution_std.py from your branch and from master!

lib/src/Uncertainty/Distribution/TruncatedDistribution.cxx
regislebrun: Already mentioned!

python/test/t_TruncatedDistribution_std.py
regislebrun: fixed

python/test/t_TruncatedDistribution_std.py
regislebrun: done

jschueller (approved):

PR #3173, opened 2026-04-24
Add MulticollinearityAnalysis (LMG, PMVD, Johnson, VIF)

NTH-OT:
Add a new MulticollinearityAnalysis class to compute sensitivity indices relevant when predictors are correlated:

- LMG index
- PMVD index
- Johnson index
- VIF metric

## Summary by CodeRabbit

* **New Features**
  * Added multicollinearity analysis with LMG, PMVD, Johnson, and VIF measures.
  * Added exact and Monte Carlo estimation, bootstrap confidence intervals, and parallel computation support.
  * Exposed the analysis through the Python API.
  * Added the Ames Housing dataset and an end-to-end analysis example.

* **Documentation**
  * Added API documentation, usage guidance, and dataset references.

* **Tests**
  * Added C++ and Python coverage for indices, estimates, validation, and error handling.


josephmure (approved):

josephmure (changes requested):
This PR should also receive the actual developed code.

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
```diff
@@ -0,0 +1,85 @@
+//                                               -*- C++ -*-
+/**
+ *  @brief MulticollinearityAnalysis implements computation of various
+ *  statistics used in a multicollinear context
+ *
+ *  Copyright 2005-2026 Airbus-EDF-IMACS-ONERA-Phimeca
```
josephmure:
```suggestion
 *  Copyright 2005-2026 Airbus-EDF-ONERA-Phimeca
```

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
```diff
@@ -0,0 +1,85 @@
...
+                            const Sample & secondSample);
+
+  /** Virtual constructor */
+  MulticollinearityAnalysis * clone() const override;
+
+  /** String converter */
+  String __repr__() const override;
+
+  /** Compute LMG and PMVD indices */
+  void computeLMGPMVD(PointWithDescription & lmgOut, PointWithDescription & pmvdOut) const;
```
josephmure:
```suggestion
  void computeLmgPmvd(PointWithDescription & lmgOut, PointWithDescription & pmvdOut) const;
```

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
```diff
@@ -0,0 +1,85 @@
...
+
+  /** Estimate LMG and PMVD indices via a Monte Carlo method */
+  void estimateLMGPMVDMonteCarlo(PointWithDescription & lmgOut, PointWithDescription & pmvdOut, const UnsignedInteger iterations) const;
```
josephmure:
```suggestion
  void estimateLmgPmvdMonteCarlo(PointWithDescription & lmgOut, PointWithDescription & pmvdOut, const UnsignedInteger iterations) const;
```

python/test/t_MulticollinearityAnalysis_std.py
```diff
@@ -0,0 +1,67 @@
...
+inputDistribution = ot.Normal([0.0, 0.0], [sigma1, sigma2], corrMatrix)
+inputSample = inputDistribution.getSample(sampleSize)
+
+# Build the output sample
+linearFunction = ot.LinearFunction([0.0, 0.0], [0.0], ot.Matrix(1, 2, [beta1, beta2]))
+noiseDistribution = ot.Normal(0.0, sigmaEps)
+noiseSample = noiseDistribution.getSample(sampleSize)
+outputSample = linearFunction(inputSample) + noiseSample
+
+analysis = ot.MulticollinearityAnalysis(inputSample, outputSample)
```
josephmure: Please add a test with only the inputSample and VIF computation.
The exception should also be tested when LMG/PMVD or Johnson computation is attempted.

python/test/t_MulticollinearityAnalysis_std.py
```diff
@@ -0,0 +1,67 @@
...
+print(f"Computed LMG = [{lmg[0]}, {lmg[1]}]")
+ott.assert_almost_equal(lmg, [lmg1, lmg2], 1e-2, 0.0)
+print(f"Theoretical PMVD = [{pmvd1}, {pmvd2}]")
+print(f"Computed PMVD = [{pmvd[0]}, {pmvd[1]}]")
+ott.assert_almost_equal(pmvd, [pmvd1, pmvd2], 1e-2, 0.0)
+
+# LMG and PMVD indices (Monte Carlo estimation)
+lmgMC, pmvdMC = analysis.estimateLMGPMVDMonteCarlo(1000)
+print(f"Estimated LMG = [{lmgMC[0]}, {lmgMC[1]}]")
+ott.assert_almost_equal(lmgMC, [0.84, 0.15], 1e-10, 0.0)
```
josephmure: we should compare to the exact value

python/test/t_MulticollinearityAnalysis_std.py
```diff
@@ -0,0 +1,67 @@
...
+# Build the output sample
+linearFunction = ot.LinearFunction([0.0, 0.0], [0.0], ot.Matrix(1, 2, [beta1, beta2]))
+noiseDistribution = ot.Normal(0.0, sigmaEps)
+noiseSample = noiseDistribution.getSample(sampleSize)
+outputSample = linearFunction(inputSample) + noiseSample
+
+analysis = ot.MulticollinearityAnalysis(inputSample, outputSample)
+
+# LMG and PMVD indices
+lmg, pmvd = analysis.computeLMGPMVD()
```
josephmure: The computed / estimated values should have clearly different names "compute_lmg" or "estimated_lmg"...

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+#include "openturns/TBBImplementation.hxx"
+
+BEGIN_NAMESPACE_OPENTURNS
+
+CLASSNAMEINIT(MulticollinearityAnalysis)
+
+static const Factory<MulticollinearityAnalysis> Factory_MulticollinearityAnalysis;
+
+/* Extract the top-left part of a symmetric matrix */
+static void extract(const SymmetricMatrix & source, SymmetricMatrix & target)
```
josephmure, line 38: Note to self: this looks like it should be implemented by the Matrix class

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+}
+
+class LmgPmvdAlgorithm;
+
+/**
+ * @class LmgPmvdTBBBody
+ *
+ * TBB body to compute the variances associated with combinations of input variables
+ */
+class LmgPmvdTBBBody
```
josephmure: Note to self: elsewhere in OT this is called "policy", why is that ?

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+ * Class for computation of LMG and PMVD indices
+ */
+class LmgPmvdAlgorithm
+{
+public:
+
+  /* Constructor */
+  LmgPmvdAlgorithm(const CovarianceMatrix & covMatrix)
+    : dimension_(covMatrix.getDimension() - 1)
+    , numberOfCombinations_(1 << dimension_)
```
josephmure:
```suggestion
    , numberOfCombinations_(1 << dimension_) // 2^dimension_
```

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+    for (UnsignedInteger i = 0; i < dimension_; i++)
+      covXY_(i, 0) = covMatrix(i, dimension_);
+    varY_ = covMatrix(dimension_, dimension_);
+  }
+
+  void run()
+  {
+    // First step: compute the variance associated with every combination of input variables
+    LmgPmvdTBBBody body(*this);
+    TBBImplementation::ParallelForIf(dimension_ >= 12, 1, numberOfCombinations_, body, 1024);
```
josephmure: Please add a ResourceMap key for the threshold dimension

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+
+    // Compute LMG indices
+    for (UnsignedInteger i = 0; i < dimension_; i++)
+      lmg_[i] = computeLmg(i);
```
josephmure: I would rather have computeLmg be of return type void, to be consistent with computePmvd

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: This should be a struct based on other places in the codebase that use TBB

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+    pmvd_ /= totalWeigtht_ * varY_;
+  }
+
+  void getResult(Point & lmg, Point & pmvd) const
+  {
+    lmg = lmg_;
+    pmvd = pmvd_;
+  }
+
+private:
```
josephmure: This can be public, this class is not publicly available anyway.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+      << " secondSample=" << secondSample_;
+  return oss;
+}
+
+/* Compute LMG and PMVD indices */
+void MulticollinearityAnalysis::computeLmgPmvd(PointWithDescription & lmg, PointWithDescription & pmvd) const
+{
+  checkInputSample();
+  checkOutputSample();
+  if (!(firstSample_.getDimension() <= 28)) throw InvalidDimensionException(HERE) << "Error: input sample dimension must be at most 28";
```
josephmure: This should be a ResourceMap key.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,626 @@
...
+
+  LmgPmvdAlgorithm algo(computeCovariance());
+  algo.run();
+  algo.getResult(lmg, pmvd);
```
josephmure: getResult should be removed in favor of 2 distinct method: getLmg() and getPmvd()

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,154 @@
+"""
+Compute indices in a multicollinear context
```
josephmure:
```suggestion
Compute sensitivity indices in a multicollinear context
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,154 @@
...
+===========================================
```
josephmure:
```suggestion
=======================================================
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,154 @@
...
+# - Johnson index
+# - VIF metric
+#
+# We will use the :class:`~openturns.MulticollinearityAnalysis` class.
+
+import openturns as ot
+import openturns.viewer as otv
+import matplotlib
+import os
+from pathlib import Path
```
josephmure: This should not be necessary

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,705 @@
...
+
+  PointWithDescription result(dimension);
+  for (UnsignedInteger i = 0; i < dimension; i++)
+    result[i] = johnson(i, 0);
+  result.setDescription(firstSample_.getDescription());
+  return result;
+}
+
+/* Compute VIF metric */
+PointWithDescription MulticollinearityAnalysis::computeVif() const
```
josephmure:
```suggestion
PointWithDescription MulticollinearityAnalysis::computeVIF() const
```
No reason not to write in uppercase here

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,705 @@
...
+  // Perform a linear regression
+  LinearModelAlgorithm algo(firstSample_, outputSample);
+  const LinearModelResult linearModelResult(algo.getResult());
+  const SymmetricMatrix gramInverse(linearModelResult.getLeastSquaresMethod().getGramInverse());
+  const Scalar SSE = linearModelResult.getSampleResiduals().asPoint().normSquare(); // Sum of squared errors
+  const Scalar df = linearModelResult.getDegreesOfFreedom();
+  const SymmetricMatrix V(gramInverse * (SSE / df));
+  const CorrelationMatrix R(covarianceToCorrelation(removeRowAndColumn(V, 0)));
+  const Scalar detR = R.computeDeterminant();
+  if (detR == 0.0) throw NotDefinedException(HERE) << "Error: the matrix is singular";
```
josephmure:
```suggestion
  if (!(detR > 0.0)) throw NotDefinedException(HERE) << "Error: the matrix is singular, its determinant is " << detR;
```

josephmure:
@coderabbit review

josephmure:
@coderabbit full review

NTH-OT:
The SWIG compilation fails on CircleCI with strange errors like this:

> typ_modulePYTHON_wrap.cxx:4014:7: error: ‘PyInt_Check’ was not declared in this scope

I have no idea why (it used to work a few weeks ago).

There is also a test (DistFunc_orthant) which is suddenly failing on macOS with this error:

> InvalidArgumentException : Error: expected a probability level in [0, 1] for qNormal, got p=-0.524034

Again, I have no idea why.

josephmure:
> The SWIG compilation fails on CircleCI with strange errors like this:
> 
> > typ_modulePYTHON_wrap.cxx:4014:7: error: ‘PyInt_Check’ was not declared in this scope
> 
> I have no idea why (it used to work a few weeks ago).
> 
> There is also a test (DistFunc_orthant) which is suddenly failing on macOS with this error:
> 
> > InvalidArgumentException : Error: expected a probability level in [0, 1] for qNormal, got p=-0.524034
> 
> Again, I have no idea why.

This seems unrelated to the changes of this PR. Try to rebase your branch on the current master, this may fix the issue.

NTH-OT:
I did the rebasing and everything is now working, thanks.

josephmure (approved):

python/test/CMakeLists.txt
```diff
@@ -873,6 +874,8 @@ if (MATPLOTLIB_FOUND)
       set_tests_properties (pyinstallcheck_example_${TESTNAME} PROPERTIES ENVIRONMENT "${PYINSTALLCHECK_ENVIRONMENT}")
     endif ()
   endforeach ()
+  # Sometimes the following test takes a bit more than 100 seconds
+  set_tests_properties (pyinstallcheck_example_plot_multicollinearity PROPERTIES TIMEOUT 150)
```
jschueller: try reduce the simulation budget instead

python/test/t_MulticollinearityAnalysis_std.py
```diff
@@ -0,0 +1,78 @@
+#! /usr/bin/env python
+
+import openturns as ot
+from openturns.experimental import MulticollinearityAnalysis
```
jschueller: the convention is to import as otexp

python/test/CMakeLists.txt
josephmure:
```suggestion
```

python/test/t_MulticollinearityAnalysis_std.py
josephmure:
```suggestion
import openturns.experimental as otexp
```

python/test/t_MulticollinearityAnalysis_std.py
```diff
@@ -0,0 +1,78 @@
...
+inputDistribution = ot.Normal([0.0, 0.0], [sigma1, sigma2], corMatrix)
+inputSample = inputDistribution.getSample(sampleSize)
+
+# Build the output sample
+linearFunction = ot.LinearFunction([0.0, 0.0], [0.0], ot.Matrix(1, 2, [beta1, beta2]))
+noiseDistribution = ot.Normal(0.0, sigmaEps)
+noiseSample = noiseDistribution.getSample(sampleSize)
+outputSample = linearFunction(inputSample) + noiseSample
+
+analysis = MulticollinearityAnalysis(inputSample, outputSample)
```
josephmure:
```suggestion
analysis = otexp.MulticollinearityAnalysis(inputSample, outputSample)
```

python/test/t_MulticollinearityAnalysis_std.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Johnson index
+johnson_computed = analysis.computeJohnson()
+print(f"Computed Johnson = [{johnson_computed[0]}, {johnson_computed[1]}]")
+ott.assert_almost_equal(
+    johnson_computed, lmg_computed, 1e-10, 0.0
+)  # In 2D, Johnson and LMG indices are identical
+
+# Check that an exception is raised when outputSample is not provided
+analysis = MulticollinearityAnalysis(inputSample)
```
josephmure:
```suggestion
analysis = otexp.MulticollinearityAnalysis(inputSample)
```

python/test/t_MulticollinearityAnalysis_std.py
```diff
@@ -0,0 +1,78 @@
...
+with ott.assert_raises(TypeError):
+    analysis.computeLmgPmvd()
+with ott.assert_raises(TypeError):
+    analysis.estimateLmgPmvdMonteCarlo(1000)
+with ott.assert_raises(TypeError):
+    analysis.computeJohnson()
+
+# VIF metric
+analysis = MulticollinearityAnalysis(inputSample)
```
josephmure:
```suggestion
analysis = otexp.MulticollinearityAnalysis(inputSample)
```

python/src/MulticollinearityAnalysis_doc.i
```diff
@@ -0,0 +1,106 @@
...
+    This sample is used to compute the LMG, PMVD and Johnson indices; the VIF metric ignores it.
+
+See also
+--------
+:class:`~openturns.CorrelationAnalysis`
+
+Examples
+--------
+>>> import openturns as ot
+>>> from openturns.experimental import MulticollinearityAnalysis
```
josephmure:
```suggestion
>>> import openturns.experimental as otexp
```

python/src/MulticollinearityAnalysis_doc.i
```diff
@@ -0,0 +1,106 @@
...
+>>> ot.RandomGenerator.SetSeed(0)
+>>> corMatrix = ot.CorrelationMatrix(3, [1.0, 0.5, 0.5, 0.5, 1.0, 0.5, 0.5, 0.5, 1.0])
+>>> distribution = ot.Normal([0.0] * 3, [1.0] * 3, corMatrix)
+>>> firstSample = distribution.getSample(100)
+>>> model = ot.SymbolicFunction(['x', 'y', 'z'], ['x + 3 * y - 5 * z'])
+>>> secondSample = model(firstSample)
+>>> analysis = MulticollinearityAnalysis(firstSample, secondSample)
```
josephmure:
```suggestion
>>> analysis = otexp.MulticollinearityAnalysis(firstSample, secondSample)
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,180 @@
...
+# - LMG index
+# - PMVD index
+# - Johnson index
+# - VIF metric
+#
+# We will use the :class:`~openturns.experimental.MulticollinearityAnalysis` class.
+
+import openturns as ot
+import openturns.viewer as otv
+from openturns.experimental import LinearModelValidation, MulticollinearityAnalysis
```
josephmure:
```suggestion
import openturns.experimental as otexp
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,180 @@
...
+
+data = ames_housing.AmesHousing().data
+Y = data[:, 0]  # Extract the sale price
+X = data[:, 1 : data.getDimension()]  # Extract the other features
+
+# %%
+# We fit a linear regression model to predict Y, e.g. sale prices from the other features, and we compute the R² score to check that the model is relevant:
+
+regression = ot.LinearModelAlgorithm(X, Y).getResult()
+r2 = LinearModelValidation(regression).computeR2Score()
```
josephmure:
```suggestion
r2 = otexp.LinearModelValidation(regression).computeR2Score()
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,180 @@
...
+def compute_quantity(X, Y, kind):
+    """
+    Compute a quantity for the given dataset
+
+    Parameters
+    - X: the input sample
+    - Y: the output sample
+    - kind: the type of quantity to compute ("LMG", "PMVD", "Johnson" or "VIF")
+    """
+    analysis = MulticollinearityAnalysis(X, Y)
```
josephmure:
```suggestion
    analysis = otexp.MulticollinearityAnalysis(X, Y)
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,180 @@
...
+    lb = boot.computeQuantilePerComponent((1.0 - alpha) / 2.0)
+    ub = boot.computeQuantilePerComponent(1.0 - (1.0 - alpha) / 2.0)
+    interval = ot.Interval(lb, ub)
+    create_graph(kind, X.getDescription(), mean, interval)
+
+
+# %%
+# LMG and PMVD indices:
+
+analysis = MulticollinearityAnalysis(X, Y)
```
josephmure:
```suggestion
analysis = otexp.MulticollinearityAnalysis(X, Y)
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,180 @@
...
+# %%
+print(pmvd)
+
+# %%
+bootstrap(X, Y, "PMVD")
+
+# %%
+# Johnson index:
+
+analysis = MulticollinearityAnalysis(X, Y)
```
josephmure:
```suggestion
analysis = otexp.MulticollinearityAnalysis(X, Y)
```

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,180 @@
...
+
+# %%
+bootstrap(X, Y, "Johnson")
+
+# %%
+# We notice that the Johnson indices are close to the LMG ones, (note that in two-dimensional cases, they are proven to be strictly equal).
+#
+# VIF metric:
+
+analysis = MulticollinearityAnalysis(X)
```
josephmure:
```suggestion
analysis = otexp.MulticollinearityAnalysis(X)
```

python/test/t_MulticollinearityAnalysis_std.py
josephmure: done

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
```diff
@@ -0,0 +1,98 @@
...
+
+  /** Virtual constructor */
+  MulticollinearityAnalysis * clone() const override;
+
+  /** String converter */
+  String __repr__() const override;
+
+  /** Compute LMG and PMVD indices */
+  void computeLmgPmvd(PointWithDescription & lmgOut, PointWithDescription & pmvdOut) const;
+
```
jschueller, line 60: for consistency acronyms usually use uppercase in the api (ARMA, SQP, HSIC, ...)
if two objects are returned we can use "And"

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
```diff
@@ -0,0 +1,98 @@
...
+  /** Estimate LMG and PMVD indices via a Monte Carlo method */
+  void estimateLmgPmvdMonteCarlo(PointWithDescription & lmgOut, PointWithDescription & pmvdOut, const UnsignedInteger iterations) const;
```
jschueller, line 62: usually the method name doesnt differentiate the estimator type, its always computeSomething

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
josephmure: This causes problems in the present case, because we are joining 2 acronyms: LMG and PMVD, so we would get computeLMGPMVD

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
josephmure: @jschueller I do not understand your comment: how should the method be called in your opinion?

jschueller: yes, sorry, my remark was to change "estimate" part into "compute", the rest is independent

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
jschueller: hence my second remark: use "And" in-between

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+}
+
+class LmgPmvdAlgorithm;
+
+/**
+ * @class LmgPmvdTBBBody
+ *
+ * TBB body to compute the variances associated with combinations of input variables
+ */
+struct LmgPmvdTBBBody
```
jschueller, line 56: yes, it usually has the Policy suffix instead of Body

josephmure: This was a question we had. Why does OT call this a Policy? Is there a specific reason?

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,180 @@
...
+
+# %%
+# Then we add a function that builds a graph:
+
+
+def create_graph(title, names, mean, interval):
+    dimension = mean.getDimension()
+    palette = ot.Drawable.BuildDefaultPalette(2)
+    graph = ot.Graph(title, "", "")
+    matplotlib.rc("xtick", labelsize=8.0)  # reduce the font size of the X labels
```
jschueller, line 81: this seems to be a global setting and may affect other examples run in the same process batch, use local plot parameters instead

python/src/MulticollinearityAnalysis_doc.i
```diff
@@ -0,0 +1,106 @@
...
+component of a random vector over a single one-dimensional variable,
+in the case where the components are correlated.
+Refer to :ref:`regression_importance_indices`.
+
+Available constructors:
+    MulticollinearityAnalysis(*firstSample*)
+        This constructor supports only :meth:`computeVIF`.
+
+    MulticollinearityAnalysis(*firstSample, secondSample*)
+        This constructor supports all indices.
```
jschueller, line 19: since the two ctor only differ by the secondSample as pure data argument, we could drop the ctors list and just mark secondSample as optional

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
jschueller: its only because of the name of the argument

josephmure: ???

python/test/CMakeLists.txt
josephmure: Apparently this is not necessary, I was able to remove these lines without causing any issue.

python/src/MulticollinearityAnalysis_doc.i
josephmure: done

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: I wonder if we should not change this in the whole library. However this is not the purpose of this PR so I have changed it to Policy.

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
josephmure: Well, I would debate this since one is an exact computation while the other is an estimate, but changing it to "compute" has the advantage of working better with autocomplete, so I have done as you suggest.

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
josephmure: ok, done

lib/src/Base/Type/MatrixImplementation.cxx
```diff
@@ -1119,7 +1119,7 @@ Scalar MatrixImplementation::computeDeterminantSymInPlace()
   if (nbRows_ == 1) return (*this)(0, 0);
   if (nbRows_ == 2) return (*this)(0, 0) * (*this)(1, 1) - (*this)(1, 0) * (*this)(1, 0);
   Scalar sign = 0.0;
-  const Scalar logAbsoluteDeterminant = computeLogAbsoluteDeterminantInPlace(sign);
+  const Scalar logAbsoluteDeterminant = computeLogAbsoluteDeterminantSymInPlace(sign);
```
jschueller, line 1122: this change needs its own commit

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  for (UnsignedInteger i = 0; i < size; i += 2)
+    outputSample(i, 0) = 1.0;
+
+  // Perform a linear regression
+  LinearModelAlgorithm algo(firstSample_, outputSample);
+  const LinearModelResult linearModelResult(algo.getResult());
+  const SymmetricMatrix gramInverse(linearModelResult.getLeastSquaresMethod().getGramInverse());
+  const CorrelationMatrix R(covarianceToCorrelation(removeRowAndColumn(gramInverse, 0)));
+  const Scalar detR = R.computeDeterminant();
+  if (!(detR > 0.0)) throw NotDefinedException(HERE) << "Error: the matrix is singular, its determinant is " << detR;
```
jschueller: seems we could avoid the dummy regression and determinant here

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,179 @@
...
+import openturns as ot
+import openturns.viewer as otv
+import openturns.experimental as otexp
+from openturns.usecases import ames_housing
+
+# %%
+# We will take the X and Y samples from the :ref:`Ames Housing<use-case-ames-housing>` dataset
+# (which contains data about houses in the city of Ames):
+
+data = ames_housing.AmesHousing().data
```
jschueller, line 28: this example can be costly and can timeout (exact indices dim 10, bootstrap):
	731 - pyinstallcheck_example_plot_multicollinearity (Timeout)
maybe choose another sample ?

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
```diff
@@ -0,0 +1,98 @@
...
+  void checkOutputSample() const;
+
+  /** Compute the covariance matrix of the full sample (input + output) */
+  CovarianceMatrix computeCovariance() const;
+
+  /** Remove a row and a column from a symmetric matrix */
+  SymmetricMatrix removeRowAndColumn(const SymmetricMatrix & matrix, const UnsignedInteger rowCol) const;
+
+  /** Convert a covariance matrix to a correlation matrix */
+  CorrelationMatrix covarianceToCorrelation(const SymmetricMatrix & matrix) const;
```
jschueller: looks these should be static fonctions

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  /* Extract the top-left part of a square matrix */
+  void extract2D(const SquareMatrix & source, SquareMatrix & target) const
+  {
+    const MatrixImplementation & sourceImpl = *source.getImplementation();
+    MatrixImplementation & targetImpl = *target.getImplementation();
+    const UnsignedInteger size = targetImpl.getDimension();
+    for (UnsignedInteger i = 0; i < size; i++)
+    {
+      for (UnsignedInteger j = 0; j <= i; j++)
+        targetImpl(i, j) = sourceImpl(i, j);
```
jschueller: use std::copy

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  }
+
+  /* Extract the top part of a vector matrix */
+  void extract1D(const Matrix & source, Matrix & target) const
+  {
+    const MatrixImplementation & sourceImpl = *source.getImplementation();
+    MatrixImplementation & targetImpl = *target.getImplementation();
+    const UnsignedInteger size = targetImpl.getNbRows();
+    for (UnsignedInteger i = 0; i < size; i++)
+      targetImpl(i, 0) = sourceImpl(i, 0);
```
jschueller: use std::copy

josephmure:
Copying old SHA for reference: 765aa66ada338f3f520b7c5f5df25f8f9a4ff6d9

lib/src/Base/Type/MatrixImplementation.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    {
+      if (i == 0)
+        var1 = 0.0;
+      else
+        var1 = var2;
+      SquareMatrix & mat2D = matrices2D_[i];
+      Matrix & mat1D = matrices1D_[i];
+      extract2D(Linv, mat2D);
+      extract1D(covXYperm_, mat1D);
+      const MatrixImplementation M1(mat1D.getImplementation()->genProd(*mat2D.getImplementation(), true, true));
```
regislebrun, line 402: Why using MatrixImplementation instead of Matrix

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
mbaudin47, line 28: The Ames Housing dataset is:

https://www.kaggle.com/datasets/shashanknecrothapa/ames-housing-dataset

It has 2930 observations. It may be faster if using half of the sample.

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
```diff
@@ -0,0 +1,179 @@
...
+# %%
+# The R² score is relatively close to 1, which means that the linear regression model is relevant.
+# We can now compute the LMG and PMVD indices, the Johnson index and the VIF metric.
+
+# %%
+# We will also perform bootstraps and plot the results, so we start by defining some helper functions.
+# First, we define one that computes the desired quantity:
+
+
+def compute_quantity(X, Y, kind):
```
mbaudin47: The name of the function may be more accurate, say "compute_sensitivity_index()" or "compute_multicollinearity_metric"?

python/doc/theory/reliability_sensitivity/regression_importance_indices.rst
```diff
@@ -0,0 +1,162 @@
...
+can be used instead.
+
+Linear regression setting
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Consider the centered linear model
+
+.. math::
+
+    Y = \vect{X}^{\mathsf{T}}\vect{a} + \varepsilon,
```
mbaudin47: Use "\top" instead of {\mathsf{T}}.

python/doc/theory/reliability_sensitivity/regression_importance_indices.rst
```diff
@@ -0,0 +1,162 @@
...
+    \qquad \vect{X}=(X_1,\ldots,X_d),
```
mbaudin47: Please use the notations introduced in https://github.com/openturns/openturns/blob/master/python/doc/math_notations.sty. For example, use \inputDim and \inputRV.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    varY_ = covMatrix(dimension_, dimension_);
+    if (!(varY_ > 0.0)) throw InvalidArgumentException(HERE) << "Error: the output variance must be positive";
+  }
+
+  void run()
+  {
+    // First step: compute the variance associated with every combination of input variables
+    const LMGAndPMVDTBBPolicy body(*this);
+    const UnsignedInteger threshold = ResourceMap::GetAsUnsignedInteger("MulticollinearityAnalysis-DimensionThresholdForLMGAndPMVDParallelization");
+    TBBImplementation::ParallelForIf(dimension_ >= threshold, 1, numberOfCombinations_, body, 1024);
```
mbaudin47, line 105: What is 1024 for?

jschueller: this is the grain size, this looks a reasonable default unless we want to fine-tune the parallelization

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+class LMGAndPMVDAlgorithm;
+
+/**
+ * @class LMGAndPMVDTBBPolicy
+ *
+ * TBB body to compute the variances associated with combinations of input variables
+ */
+struct LMGAndPMVDTBBPolicy
+{
+  LMGAndPMVDTBBPolicy(LMGAndPMVDAlgorithm & algo)
```
regislebrun, line 59: algo is not a const ref because it is modified?

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+ * Class for computation of LMG and PMVD indices
+ */
+class LMGAndPMVDAlgorithm
+{
+public:
+
+  /* Constructor */
+  LMGAndPMVDAlgorithm(const CovarianceMatrix & covMatrix)
+    : dimension_(covMatrix.getDimension() - 1)
+    , numberOfCombinations_(1ul << dimension_) // 2^dimension_
```
regislebrun, line 85: What if dimension_>63?

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  }
+
+  Point getPmvd() const
+  {
+    return pmvd_;
+  }
+
+  void computeLmg()
+  {
+    for(UnsignedInteger i = 0; i < dimension_; i++)
```
regislebrun: use ++i and not i++

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    {
+      Point var1(dimension_);
+      Point var2(dimension_);
+      for (UnsignedInteger c = 1; c < numberOfCombinations_; c++)
```
regislebrun: use ++c and not c++

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    UnsignedInteger k = 0;
+    UnsignedInteger mask = 0;
+    for (UnsignedInteger i = 0; i < dimension_; i++)
+    {
+      const UnsignedInteger c = 1ul << i;
+      potentials[c] = computeWorth(c);
+      if (potentials[c] == 0.0)
+      {
+        // The variable has no potential, it will be ignored in the rest of the computation
+        k++;
```
regislebrun: use ++k and not k++

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+        mask |= c;
+      }
+    }
+
+    // Compute the potential for combinations of size 2, then 3, and so on
+    for (UnsignedInteger size = 2; size <= dimension_ - k; size++)
```
regislebrun: Use ++size and not size++

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+          UnsignedInteger i = 0;
+          UnsignedInteger j = 0;
+          do
+          {
+            const UnsignedInteger d = c & ~(1ul << i);
+            if (d != c)
+            {
+              // We have a sub-combination
+              sum += 1 / potentials[d];
+              j++;
```
regislebrun: ++j

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+            }
+            i++;
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+          const Scalar worth = computeWorth(c);
+          if (!(worth > 0.0)) throw InvalidArgumentException(HERE) << "Invalid worth value: " << worth;
+          potentials[c] = worth / sum;
+        }
+      }
+    }
+
+    // Now that all potentials have been calculated, we can compute the PMVD indices
+    const UnsignedInteger all = (numberOfCombinations_ - 1) & ~mask;
+    for (UnsignedInteger i = 0; i < dimension_; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+void LMGAndPMVDTBBPolicy::getIndices(const UnsignedInteger combination, const UnsignedInteger size, Indices & indices) const
+{
+  UnsignedInteger i = 0;
+  UnsignedInteger j = 0;
+  while(j < size)
+  {
+    if (combination & (1ul << i))
+    {
+      // Bit found, store the index
+      indices[j++] = i;
```
regislebrun: No. Never. 
```
indices[j] = i;
++j;
```

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    }
+    i++;
```
regislebrun: ++i;

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  }
+}
+
+/* Extract a subset of a square matrix */
+void LMGAndPMVDTBBPolicy::extract2D(const SquareMatrix & source, const Indices & indices, SquareMatrix & target) const
+{
+  const MatrixImplementation & sourceImpl = *source.getImplementation();
```
regislebrun, line 298: What is the purpose of using MatrixImplementation instead of Matrix here and in all the other places?

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  MatrixImplementation & targetImpl = *target.getImplementation();
+  const UnsignedInteger size = targetImpl.getDimension();
+  for (UnsignedInteger i = 0; i < size; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  {
+    for (UnsignedInteger j = 0; j <= i; j++)
```
regislebrun: ++j

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  }
+}
+
+/* Extract a subset of a column matrix */
+void LMGAndPMVDTBBPolicy::extract1D(const Matrix & source, const Indices & indices, Matrix & target) const
+{
+  const MatrixImplementation & sourceImpl = *source.getImplementation();
+  MatrixImplementation & targetImpl = *target.getImplementation();
+  const UnsignedInteger size = targetImpl.getNbRows();
+  for (UnsignedInteger i = 0; i < size; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    , covXYperm_(dimension_, 1)
+    , nbPermutations_(0)
+    , totalWeight_(0.0)
+    , lmg_(dimension_)
+    , pmvd_(dimension_)
+    , delta_(dimension_)
+  {
+    // Extract needed parts from the covariance matrix
+    extract(covMatrix, covXX_);
+    for (UnsignedInteger i = 0; i < dimension_; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    }
+  }
+
+  void run()
+  {
+    fullVar_ = computeFullVariance();
+    // Draw a random permutation for each iteration
+    const KPermutationsDistribution distribution(dimension_, dimension_);
+    Indices permutation(dimension_);
+    for (UnsignedInteger i = 0; i < iterations_; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    {
+      const Point realization(distribution.getRealization());
+      for (UnsignedInteger j = 0; j < dimension_; j++)
```
regislebrun: ++j

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  Point getPmvd() const
+  {
+    return pmvd_;
+  }
+
+  void addPermutation(const Indices & permutation)
+  {
+    permutate2D(permutation, covXX_, covXXperm_);
+    permutate1D(permutation, covXY_, covXYperm_);
+    const TriangularMatrix L = covXXperm_.computeCholesky();
```
regislebrun: use copy constructor

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    const SquareMatrix Linv = L.inverse();
```
regislebrun: use copy constructor

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    Scalar weight = 1.0;
+    Scalar var1 = 0.0;
+    Scalar var2 = 0.0;
+    for (UnsignedInteger i = 0; i < dimension_; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    totalWeight_ += weight;
+  }
+
+  /* Apply a permutation to a square matrix */
+  void permutate2D(const Indices & permutation, const SquareMatrix & source, SquareMatrix & target) const
+  {
+    const MatrixImplementation & sourceImpl = *source.getImplementation();
+    MatrixImplementation & targetImpl = *target.getImplementation();
+    const UnsignedInteger size = target.getDimension();
+    for (UnsignedInteger i = 0; i < size; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    {
+      for (UnsignedInteger j = 0; j < size; j++)
```
regislebrun: ++j

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    }
+  }
+
+  /* Apply a permutation to a column matrix */
+  void permutate1D(const Indices & permutation, const Matrix & source, Matrix & target) const
+  {
+    const MatrixImplementation & sourceImpl = *source.getImplementation();
+    MatrixImplementation & targetImpl = *target.getImplementation();
+    const UnsignedInteger size = target.getNbRows();
+    for (UnsignedInteger i = 0; i < size; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+      targetImpl(i, 0) = sourceImpl(permutation[i], 0);
+  }
+
+  /* Extract the top-left part of a square matrix */
+  void extract2D(const SquareMatrix & source, SquareMatrix & target) const
+  {
+    const MatrixImplementation & sourceImpl = *source.getImplementation();
+    MatrixImplementation & targetImpl = *target.getImplementation();
+    const UnsignedInteger size = targetImpl.getDimension();
+    for (UnsignedInteger i = 0; i < size; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    {
+      for (UnsignedInteger j = 0; j <= i; j++)
```
regislebrun: ++j

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    }
+  }
+
+  /* Extract the top part of a vector matrix */
+  void extract1D(const Matrix & source, Matrix & target) const
+  {
+    const MatrixImplementation & sourceImpl = *source.getImplementation();
+    MatrixImplementation & targetImpl = *target.getImplementation();
+    const UnsignedInteger size = targetImpl.getNbRows();
+    for (UnsignedInteger i = 0; i < size; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  return oss;
+}
+
+/* Compute LMG and PMVD indices */
+void MulticollinearityAnalysis::computeLMGAndPMVD(PointWithDescription & lmg, PointWithDescription & pmvd) const
+{
+  checkInputSample();
+  checkOutputSample();
+  UnsignedInteger maxDimension = ResourceMap::GetAsUnsignedInteger("MulticollinearityAnalysis-MaximumInputDimensionForLMGAndPMVD");
+  maxDimension = std::min(maxDimension, 31ul); // Make sure we don't overflow 32-bit integers
```
regislebrun, line 532: ??? maxDimension is an UnsignedInteger, there is no way it can overflow 32-bits integers. BTW I don' understand why it would matter as it is compared to firstSample_.getDimension(), which is also an UnsignedInteger.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  Sample joinedSample(firstSample_);
+  joinedSample.stack(secondSample_);
+  const CorrelationMatrix corMatrix(joinedSample.computeLinearCorrelation());
+
+  // Extract needed parts of the correlation matrix
+  const UnsignedInteger dimension = firstSample_.getDimension();
+  CorrelationMatrix corXX(dimension);
+  extract(corMatrix, corXX);
+  Matrix corXY(dimension, 1);
+  for (UnsignedInteger i = 0; i < dimension; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+    corXY(i, 0) = corMatrix(i, dimension);
+
+  // Perform the eigenvalues decomposition of the corXX matrix
+  SquareMatrix Phi;
+  const Point eigenvalues(corXX.computeEV(Phi));
+
+  SquareMatrix Delta(dimension);
+  for (UnsignedInteger i = 0; i < dimension; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  LinearModelAlgorithm algo(firstSample_, outputSample);
+  const LinearModelResult linearModelResult(algo.getResult());
+  const SymmetricMatrix gramInverse(linearModelResult.getLeastSquaresMethod().getGramInverse());
+  const CorrelationMatrix R(covarianceToCorrelation(removeRowAndColumn(gramInverse, 0)));
+  const Scalar detR = R.computeDeterminant();
+  if (!(detR > 0.0)) throw NotDefinedException(HERE) << "Error: the matrix is singular, its determinant is " << detR;
+
+  const UnsignedInteger dimension = firstSample_.getDimension();
+  PointWithDescription result(dimension);
+  for(UnsignedInteger i = 0; i < dimension; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  {
```
regislebrun: Bad. We create d matrices of dimension (d-1) to compute their determinant, plus the determinant of the full matrix. It is much simpler (and more efficient) to compute its inverse and take the diagonal elements.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+/* Compute the covariance matrix of the full sample (input + output) */
+CovarianceMatrix MulticollinearityAnalysis::computeCovariance() const
+{
+  Sample joinedSample(firstSample_);
+  joinedSample.stack(secondSample_);
+  return joinedSample.computeCovariance();
+}
+
+/* Remove a row and a column from a symmetric matrix */
+SymmetricMatrix MulticollinearityAnalysis::removeRowAndColumn(const SymmetricMatrix & matrix, const UnsignedInteger rowCol) const
```
regislebrun: useless

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  }
+  return result;
+}
+
+/* Convert a covariance matrix to a correlation matrix */
+CorrelationMatrix MulticollinearityAnalysis::covarianceToCorrelation(const SymmetricMatrix & matrix) const
+{
+  const UnsignedInteger size = matrix.getDimension();
+  CorrelationMatrix result(size);
+  for (UnsignedInteger i = 0; i < size; i++)
```
regislebrun: ++i

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,709 @@
...
+  {
+    for (UnsignedInteger j = 0; j < i; j++)
```
regislebrun: ++j

regislebrun (changes requested):
I checked only the C++ library, neither the tests nor the documentation. Several coding rules are infringed, and some algorithmic points have to be fixed. Otherwise, good job.

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
NTH-OT: Question: is it really necessary to execute the examples as unit tests? Sphinx also runs them, so they are actually executed twice.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
NTH-OT: Yes, algo is modified.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
NTH-OT: The dimension is enforced to be less than 32, so numberOfCombinations is always correctly defined.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
NTH-OT: It's faster. Matrix is a wrapper over MatrixImplementation, so every time you use it, you get an overhead. If you need to access 100 coefficients, you will pay the overhead 100 times.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: However, this maximum dimension is controlled by the "MulticollinearityAnalysis-MaximumInputDimensionForLMGAndPMVD" ResourceMap key.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
NTH-OT: Combinations of input variables are stored in integers (as bit masks), and integers are 32 bits on some platforms, which is why the dimension is enforced to be less than 32. Besides, using the algorithm with a dimension greater than 32 would require a very large amount of memory (dozens or hundreds of gigabytes) and would be extremely slow, so it wouldn't be reasonable anyway (and the Monte Carlo estimation can be used in case of high dimension).

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
NTH-OT: The default key value is 28, but there is a check on line 534 to make sure it's less than 32.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
NTH-OT: I wasn't aware of this. I took the algorithm to compute the VIF metric from an R package (as agreed with EDF).

This looks like a great improvement indeed (the timeout issue has now disappeared).

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
```diff
@@ -0,0 +1,98 @@
...
+  : public PersistentObject
+{
+  CLASSNAME
+public:
+
+  /** Default constructor */
+  MulticollinearityAnalysis();
+
+  /** Constructor with input sample (will only allow one to compute the VIF metric) */
+  MulticollinearityAnalysis(const Sample & firstSample);
```
jschueller: explicit

python/src/MulticollinearityAnalysis_doc.i
```diff
@@ -0,0 +1,100 @@
...
+
+.. warning::
+    This class is experimental and likely to be modified in future releases.
+    To use it, import the ``openturns.experimental`` submodule.
+
+
+Provides sensitivity analysis methods to determine the influence of each
+component of a random vector over a single one-dimensional variable,
+in the case where the components are correlated.
+Refer to :ref:`regression_importance_indices`.
```
jschueller: use :any: directive instead of ref in docstrings (see developer guide)

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: Indeed, but I have now changed the algorithm anyway. And I think it really is faster: the example no longer timeouts on the Windows CI machine and takes about 7 seconds to complete (the additional change regarding LMG/PMVD joint bootstrap probably helped too)

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: indeed, I have removed it

python/src/MulticollinearityAnalysis_doc.i
josephmure: done

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: Quoting @NTH-OT : 

>It's faster. Matrix is a wrapper over MatrixImplementation, so every time you use it, you get an overhead. If you need to access 100 coefficients, you will pay the overhead 100 times.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: Yes! This has been changed to be more efficient.

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
josephmure: I think the problem is solved now: the more efficient VIF computation, coupled with the simultaneous bootstrap of LMG and PMVD indices has shortened the execution time considerably.

lib/src/Base/Stat/openturns/MulticollinearityAnalysis.hxx
josephmure: actually I am dropping this function

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
josephmure: this is no longer necessary I think (cf. VIF computation improvements)

python/doc/examples/data_analysis/sample_analysis/plot_multicollinearity.py
josephmure: Thanks for the suggestions, I have chosen "compute_multicollinearity_metric"

python/doc/theory/reliability_sensitivity/regression_importance_indices.rst
josephmure: actually, there is an even better macro for the matrix transpose: \Tr

python/doc/theory/reliability_sensitivity/regression_importance_indices.rst
josephmure: done

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
regislebrun: Ok. A benchmark confirm this? A 2% penalty is not the same as a 300% penalty. For readability I accept 2%

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,661 @@
...
+PointWithDescription MulticollinearityAnalysis::computeVIF() const
+{
+  checkInputSample();
+
+  const UnsignedInteger dimension = firstSample_.getDimension();
+
+  // Correlation matrix of the predictors — no dummy regression needed
+  const CorrelationMatrix R = firstSample_.computeLinearCorrelation();
+
+  // Guard against (near-)singularity
```
regislebrun: It is fragile to check the determinant for this. You'd better rely on LAPACK detection of singularity, which is translated into a `NotDefinedException` at the C++ level. BTW it avoid to compute the LU factorization twice.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
```diff
@@ -0,0 +1,661 @@
...
+  const Scalar detR = R.computeDeterminant();
+  if (!(detR > 0.0))
+    throw NotDefinedException(HERE) << "Error: the matrix is singular, its determinant is " << detR;
+
+  // VIFs are the diagonal of the inverse correlation matrix
+  const SymmetricMatrix Rinv = R.inverse();
```
regislebrun: So here, a try/catch

regislebrun (changes requested):
Still a little step to climb

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
NTH-OT: Here MatrixImplementation is used because I needed to call genProd() with the extra parameters (transposeLeft and transposeRight). Matrix doesn't give access to them.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: Ok, I am removing this guard and trusting LAPACK to do it for us.

lib/src/Base/Stat/MulticollinearityAnalysis.cxx
josephmure: done

josephmure:
@jschueller Since my rebase pyinstallcheck_Object_std fails on all machines.

NTH-OT:
> Since my rebase pyinstallcheck_Object_std fails on all machines.

Looks like this line is missing in MulticollinearityAnalysis.i:

%copyctor OT::MulticollinearityAnalysis;

The error message could probably be improved by changing "serialization failures" to "missing copy constrictors".

PR #3176, opened 2026-04-28
Update confidence interval of a quantile: doc and code

mbaudin47:
The goal of this PR is to:
- clarify the documentation of the `QuantileConfidence`;
- fix the bug #3178;
- fix the bug #3182;
- provide methods to compute the probability of coverage of a confidence interval of a quantile, given the sample size and the rank.

## Update of the doc

The help for `computeUnilateralConfidenceInterval` states that "_True indicates the interval is bounded by a lower value_". This is confusing because we do not want to bound an interval: we want to bound the quantile. The fact that the method computes a bound of the interval is not relevant here. The correct sentence is: "True indicates the **quantile** is bounded by a lower value". 

In the API help page of `computeUnilateralMinimumSampleSize`, updated the name of the rank parameter, so that it matches the name in the C++ source code. Also, updated the `tail` name in the source code and changed it `lower_bounded` in order to match the name in the doc.

image

**Figure 1.** The name of the variables in the current API documentation is not consistent with the name of the variables in the C++ source code, which creates confusions.

## Compute the probability of coverage

This PR provides two new methods:
- `computeBilateralCoverage(size, rank1, rank2)`: computes the coverage probability for a two-sided confidence interval;
- `computeUnilateralCoverage(size, rank, lower_bounded)`: computes the coverage probability for a one-sided confidence interval.

The motivation for this development is to provide a direct and efficient way to compute the coverage probability of quantile confidence intervals without requiring an actual dataset. Mathematically, this probability depends solely on the sample size, the selected ranks, and the quantile level; it is entirely independent of the empirical values within a sample. If this direct computation feature is missing, a user is forced to rely on an inefficient workaround to obtain the coverage metric: they must first determine the sample size, generate a dummy sample of that exact size, and then pass this artificial data into the existing interval computation methods merely to extract the probability. By evaluating the underlying Binomial distribution directly, the new implementation eliminates the need to allocate and process dummy samples, making the interface more logical and computationally efficient.

For example, consider the quantile level 5% and the probability of coverage 95%. Assume that we want to compute the probability of coverage of the smallest observation in a sample of size 59, i.e. the probability that the smallest observation in the sample is a lower bound of the quantile. Then the current API forces us to generate a dummy sample of size 59 and use `computeUnilateralConfidenceIntervalWithCoverage()`. With the new `computeUnilateralCoverage()`, we can directly compute the probability of coverage.

The next script is a workaround using the current API based on a dummy sample.
```python
import openturns as ot
import openturns.experimental as otexp

# %%
alpha = 0.05 # 5% quantile
beta = 0.95 # 95% confidence
algo = otexp.QuantileConfidence(alpha, beta)
lower_bounded = True
size = 59

# %%
dummy_sample = ot.Sample(size, 1)  # <- Fake sample
lower_bounded = True
ci, coverage = algo.computeUnilateralConfidenceIntervalWithCoverage(dummy_sample)
print(coverage)  # And ignore ci, which is not used in this case
```
This prints 0.9726281425903597.

The new interface prevents from generating a dummy sample.
```python
rank = 0
lower_bounded = True
coverage = algo.computeUnilateralCoverage(size, rank, lower_bounded)
print(coverage)
```

## Summary by CodeRabbit

* **New Features**
  * Added public methods to compute unilateral and bilateral coverage probabilities.
  * Added configuration entry QuantileConfidence-ProbabilityEpsilon (default 1.0e-15).

* **Improvements**
  * Stricter input validation with clearer, parameter-specific errors.
  * Renamed boolean convention from "tail" to "lowerBounded" for clarity.
  * Improved bilateral rank search, coverage computation, and asymptotic interval handling.

* **Documentation**
  * Expanded and clarified API docs with updated examples.

* **Tests**
  * Reorganized and extended tests, including empirical coverage checks and minimum-sample-size validations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openturns/openturns/pull/3176?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -33,6 +33,8 @@
  * LOLAVoronoi moved out of experimental
  * Deprecated GaussianProcessConditionalCovariance::operator() for getMarginalDistribution
  * Deprecated ProcessSample.setField(Field, int) in favor ProcessSample.setField(int, Field)
+ * Deprecated QuantileConfidence.computeUnilateralConfidenceIntervalWithCoverage(Sample) in favor of QuantileConfidence.computeUnilateralCoverage(UnsignedInteger, UnsignedInteger, Bool)
+ * Deprecated QuantileConfidence.computeBilateralConfidenceIntervalWithCoverage(Sample) in favor of computeBilateralCoverage(UnsignedInteger, UnsignedInteger, UnsignedInteger)
 
 === Documentation ===
 
--- a/lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
@@ -53,10 +53,12 @@ QuantileConfidence::QuantileConfidence(const Scalar alpha, const Scalar beta)
   , alpha_(alpha)
   , beta_(beta)
 {
-  if (!(alpha >= 0.0) || !(alpha <= 1.0))
-    throw InvalidArgumentException(HERE) << "Quantile level must be in [0, 1]";
-  if (!(beta >= 0.0) || !(beta <= 1.0))
-    throw InvalidArgumentException(HERE) << "Confidence level must be in [0, 1]";
+  if (!(alpha >= 0.0 && alpha <= 1.0))
+    throw InvalidArgumentException(HERE)
+      << "Quantile level must be in [0, 1], but alpha = " << alpha;
+  if (!(beta >= 0.0 && beta <= 1.0))
+    throw InvalidArgumentException(HERE)
+      << "Confidence level must be in [0, 1], but beta = " << beta;
 }
 
 /* Virtual constructor */
@@ -79,28 +81,34 @@ String QuantileConfidence::__str__(const String & /*offset*/) const
 }
 
 // compute rank k given by the beta-level quantile of B(n, alpha)
-UnsignedInteger QuantileConfidence::computeUnilateralRank(const UnsignedInteger size, const Bool tail) const
+UnsignedInteger QuantileConfidence::computeUnilateralRank(const UnsignedInteger size, const Bool lowerBounded) const
 {
-  const UnsignedInteger minimumSize = computeUnilateralMinimumSampleSize(0, tail);
+  const UnsignedInteger minimumSize = computeUnilateralMinimumSampleSize(0, lowerBounded);
   if (size < minimumSize)
-    throw InvalidArgumentException(HERE) << "Cannot compute unilateral rank as size (" << size << ") is lower than minimum size (" << minimumSize << ")";
+    throw InvalidArgumentException(HERE) 
+        << "Cannot compute unilateral rank as size (" << size 
+        << ") is lower than minimum size (" << minimumSize << ")";
 
   const Binomial binomial(size, alpha_);
-  const Scalar p = tail ? 1.0 - std::pow(alpha_, 1.0 * size) : 1.0 - std::pow(1.0 - alpha_, 1.0 * size);
+  const Scalar p = lowerBounded ? 1.0 - std::pow(1.0 - alpha_, 1.0 * size) : 1.0 - std::pow(alpha_, 1.0 * size);
 
   if (p < beta_)
   {
-    if (tail)
-      throw InvalidArgumentException(HERE) << "Cannot compute rank as parameters do not satisfy 1 - alpha^n >= beta";
+    if (lowerBounded)
+      throw InvalidArgumentException(HERE)
+          << "Cannot compute rank as parameters do not satisfy 1 - (1 - alpha)^n >= beta. "
+          << "Increase the size, or decrease the confidence level.";
     else
-      throw InvalidArgumentException(HERE) << "Cannot compute rank as parameters do not satisfy 1 - (1 - alpha)^n >= beta";
+      throw InvalidArgumentException(HERE)
+          << "Cannot compute rank as parameters do not satisfy 1 - alpha^n >= beta. "
+          << "Increase the size, or decrease the confidence level.";
   }
-
-  const UnsignedInteger rank = binomial.computeQuantile(beta_, tail)[0];
+  const UnsignedInteger rank = binomial.computeQuantile(beta_, lowerBounded)[0];
   return rank;
 }
 
 // compute argmin (k1, k2) P_Xd(]k1, k2]) under constraint P_Xd(]k1, k2])>=beta, with Xd~B(n, alpha)
+// and 1 <= k1 <= k2 <= n
 Indices QuantileConfidence::computeBilateralRank(const UnsignedInteger size) const
 {
   const UnsignedInteger minimumSize = computeBilateralMinimumSampleSize();
@@ -141,36 +149,40 @@ Indices QuantileConfidence::computeBilateralRank(const UnsignedInteger size) con
 }
 
 // compute interval of the form [X_k; +inf[ or ]-inf; X_k] from unilateral rank k
-Interval QuantileConfidence::computeUnilateralConfidenceInterval(const Sample & sample, const Bool tail) const
+Interval QuantileConfidence::computeUnilateralConfidenceInterval(const Sample & sample, const Bool lowerBounded) const
 {
   Scalar coverageOut = -1.0;
-  return computeUnilateralConfidenceIntervalWithCoverage(sample, coverageOut, tail);
+  return computeUnilateralConfidenceIntervalWithCoverage(sample, coverageOut, lowerBounded);
 }
 
 // compute interval of the form [X_k; +inf[ or ]-inf; X_k] from unilateral rank k and the actual coverage
-Interval QuantileConfidence::computeUnilateralConfidenceIntervalWithCoverage(const Sample & sample, Scalar & coverageOut, const Bool tail) const
+Interval QuantileConfidence::computeUnilateralConfidenceIntervalWithCoverage(const Sample & sample, Scalar & coverageOut, const Bool lowerBounded) const
 {
+  LOGWARN("QuantileConfidence.computeUnilateralConfidenceIntervalWithCoverage(Sample) is deprecated, use QuantileConfidence.computeUnilateralCoverage(UnsignedInteger, UnsignedInteger, Bool)");
+  if (sample.getSize() == 0)
+    throw InvalidArgumentException(HERE)
+      << "Expected a non-empty sample, but size is zero";
   if (sample.getDimension() != 1)
-    throw InvalidArgumentException(HERE) << "Expected a sample of dimension 1";
+    throw InvalidArgumentException(HERE)
+      << "Expected a sample of dimension 1, but dimension = "
+      << sample.getDimension();
   Point lowerBound(1, -SpecFunc::MaxScalar);
   Point upperBound(1, SpecFunc::MaxScalar);
   Interval::BoolCollection finiteLowerBound(1, false);
   Interval::BoolCollection finiteUpperBound(1, false);
   const Sample sortedSample(sample.sort());
-  const UnsignedInteger k = computeUnilateralRank(sample.getSize(), tail);
-  const Binomial binomial(sample.getSize(), getAlpha());
-  if (tail)
+  const UnsignedInteger k = computeUnilateralRank(sample.getSize(), lowerBounded);
+  if (lowerBounded)
   {
     lowerBound[0] = sortedSample(k, 0);
     finiteLowerBound[0] = true;
-    coverageOut = binomial.computeComplementaryCDF(k);
   }
   else
   {
     upperBound[0] = sortedSample(k, 0);
     finiteUpperBound[0] = true;
-    coverageOut = binomial.computeCDF(k);
   }
+  coverageOut = computeUnilateralCoverage(sample.getSize(), k, lowerBounded);
   return Interval(lowerBound, upperBound, finiteLowerBound, finiteUpperBound);
 }
 
@@ -184,15 +196,61 @@ Interval QuantileConfidence::computeBilateralConfidenceInterval(const Sample & s
 // compute interval of the form [X_k1; X_k2] from bilateral ranks k1, k2 with actual coverage
 Interval QuantileConfidence::computeBilateralConfidenceIntervalWithCoverage(const Sample & sample, Scalar & coverageOut) const
 {
+  LOGWARN("QuantileConfidence.computeBilateralConfidenceIntervalWithCoverage(Sample) is deprecated, use QuantileConfidence.computeBilateralCoverage(UnsignedInteger, UnsignedInteger, UnsignedInteger)");
+  if (sample.getSize() == 0)
+    throw InvalidArgumentException(HERE)
+      << "Expected a non-empty sample, but size is zero";
   if (sample.getDimension() != 1)
-    throw InvalidArgumentException(HERE) << "Expected a sample of dimension 1";
+    throw InvalidArgumentException(HERE)
+      << "Expected a sample of dimension 1 but dimension = "
+      << sample.getDimension();
   const Indices rank(computeBilateralRank(sample.getSize()));
-  const Binomial binomial(sample.getSize(), getAlpha());
-  coverageOut = binomial.computeCDF(rank[1]) - binomial.computeCDF(rank[0]);
+  coverageOut = computeBilateralCoverage(sample.getSize(), rank[0], rank[1]);
   const Sample sortedSample(sample.sort());
   return Interval(Point({sortedSample(rank[0], 0)}), Point({sortedSample(rank[1], 0)}));
 }
 
+Scalar QuantileConfidence::computeUnilateralCoverage(const UnsignedInteger size, const UnsignedInteger rank, const Bool lowerBounded) const
+{
+  if (rank >= size)
+    throw InvalidArgumentException(HERE)
+        << "The rank must be strictly less than the sample size, but rank = "
+        << rank << ", size = " << size;
+
+  // Y ~ Binomial(size, alpha)
+  Binomial binomial(size, alpha_);  
+  Scalar cdf;
+
+  if (lowerBounded)
+    // Lower bound coverage: 1 - P(Y <= rank)
+    cdf = binomial.computeComplementaryCDF(rank);
+  else
+    // Upper bound coverage: P(Y <= rank)
+    cdf = binomial.computeCDF(rank);
+  
+  return cdf;
+}
+
+Scalar QuantileConfidence::computeBilateralCoverage(const UnsignedInteger size, const UnsignedInteger rank1, const UnsignedInteger rank2) const
+{
+  if (rank1 > rank2)
+    throw InvalidArgumentException(HERE)
+      << "The lower rank (" << rank1 
+      << ") must be less than or equal to the upper rank (" << rank2 << ").";
+  if (rank2 >= size)
+    throw InvalidArgumentException(HERE)
+      << "The upper rank (" << rank2 
+      << ") must be strictly less than the sample size (" << size << ").";
+
+  // Y ~ Binomial(size, alpha)
+  Binomial binomial(size, alpha_);
+  
+  // P(k1 + 1 <= Y <= k2) = P(Y <= k2) - P(Y <= k1)
+  Scalar cdf_upper = binomial.computeCDF(rank2);
+  Scalar cdf_lower = binomial.computeCDF(rank1);
+  
+  return cdf_upper - cdf_lower;
+}
 
 namespace
 {
@@ -201,11 +259,11 @@ class QuantileConfidenceEvaluation: public EvaluationImplementation
 public:
   QuantileConfidenceEvaluation(const Scalar alpha,
                                const UnsignedInteger rank,
-                               const Bool tail)
+                               const Bool lowerBounded)
     : EvaluationImplementation()
     , alpha_(alpha)
     , rank_(rank)
-    , tail_(tail)
+    , lowerBounded_(lowerBounded)
   {
     // Nothing to do
   }
@@ -217,7 +275,7 @@ public:
 
   Point operator() (const Point & point) const override
   {
-    return {DistFunc::pBeta(rank_ + 1, point[0] - rank_, tail_ ? 1.0 - alpha_ : alpha_)};
+    return {DistFunc::pBeta(rank_ + 1, point[0] - rank_, lowerBounded_ ? alpha_ : 1.0 - alpha_)};
   }
 
   UnsignedInteger getInputDimension() const override
@@ -233,12 +291,12 @@ public:
 private:
   Scalar alpha_ = 0.0;
   UnsignedInteger rank_ = 0;
-  Bool tail_ = true;
+  Bool lowerBounded_ = true;
 }; // class QuantileConfidenceEvaluation
 } // Anonymous namespace
 
 
-UnsignedInteger QuantileConfidence::computeUnilateralMinimumSampleSize(const UnsignedInteger rank, const Bool tail) const
+UnsignedInteger QuantileConfidence::computeUnilateralMinimumSampleSize(const UnsignedInteger tail_rank, const Bool lowerBounded) const
 {
   // Here we have to find the minimal value of N such that
   // 1-\sum_{i=N-r}^N Binomial(i, N)\alpha^i(1-\alpha)^{N-i}>=\beta
@@ -250,13 +308,13 @@ UnsignedInteger QuantileConfidence::computeUnilateralMinimumSampleSize(const Uns
   // Binomial(N, alpha) distribution.
   // Easy case: rank=0, the quantile bound is given by the largest upper statistics. The equation to solve is N=\min{n|1-\alpha^n>=\beta}
   Scalar nApprox = 0.0;
-  const Function wilksConstraint(QuantileConfidenceEvaluation(alpha_, rank, tail).clone());
-  if (rank == 0)
+  const Function wilksConstraint(QuantileConfidenceEvaluation(alpha_, tail_rank, lowerBounded).clone());
+  if (tail_rank == 0)
   {
-    if (tail)
-      nApprox = std::log1p(-beta_) / std::log(alpha_);
-    else
+    if (lowerBounded)
       nApprox = std::log1p(-beta_) / std::log1p(-alpha_);
+    else
+      nApprox = std::log1p(-beta_) / std::log(alpha_);
   }
   else
   {
@@ -267,18 +325,19 @@ UnsignedInteger QuantileConfidence::computeUnilateralMinimumSampleSize(const Uns
     // We compute a reasonable guess for n using a Normal approximation:
     // n*alpha+q_beta*sqrt(n*alpha*(1-alpha))=n-r:
     const Scalar aBeta = DistFunc::qNormal(beta_);
-    UnsignedInteger nMax = static_cast<UnsignedInteger>((rank + 0.5 * (alpha_ * aBeta * aBeta + std::abs(aBeta) * std::sqrt(alpha_ * (4.0 * rank + alpha_ * aBeta * aBeta)))) / (1.0 - alpha_));
+    UnsignedInteger nMax = static_cast<UnsignedInteger>((tail_rank + 0.5 * (alpha_ * aBeta * aBeta + std::abs(aBeta) * std::sqrt(alpha_ * (4.0 * tail_rank + alpha_ * aBeta * aBeta)))) / (1.0 - alpha_));
     // This loop must end as wilksConstraint->1 when n->inf
     while (wilksConstraint(Point({0.0 + nMax}))[0] < beta_)
       // At the beginning of the loop nMax is >= 1 so it increases strictly
       nMax += nMax;
-    nApprox = Brent().solve(wilksConstraint, beta_, rank, nMax);
+    nApprox = Brent().solve(wilksConstraint, beta_, tail_rank, nMax);
   } // rank > 0
   // Here, nApprox can be very close to an integer (in the sense of: the value of the constraint evaluated at the nearest integer is very close to beta_), in which case the correct answer is round(nApprox), or the answer is the next integer value
-  const UnsignedInteger nInf = std::round(std::max(1.0 * rank, nApprox));
+  const UnsignedInteger nInf = std::round(std::max(1.0 * (tail_rank + 1), nApprox));
   const Scalar constraintInf = wilksConstraint(Point({0.0 + nInf}))[0];
   if (std::abs(constraintInf - beta_) < std::sqrt(SpecFunc::Precision)) return nInf;
-  return std::ceil(nApprox);
+  const UnsignedInteger minimumSampleSize = std::max(tail_rank + 1, static_cast<UnsignedInteger>(std::ceil(nApprox)));
+  return minimumSampleSize;
 }
 
 // Find the minimal value of N such that 1 - alpha^n - (1 - alpha)^n >= beta
@@ -308,11 +367,17 @@ Indices QuantileConfidence::computeAsymptoticBilateralRank(const UnsignedInteger
 // compute interval of the form [X_k1; X_k2] from asymptotic bilateral ranks k1, k2
 Interval QuantileConfidence::computeAsymptoticBilateralConfidenceInterval(const Sample & sample) const
 {
+  if (sample.getSize() == 0)
+    throw InvalidArgumentException(HERE)
+      << "Expected a non-empty sample, but size is zero";
+  if (sample.getDimension() != 1)
+    throw InvalidArgumentException(HERE)
+      << "Expected a sample of dimension 1, but dimension = "
+      << sample.getDimension();
   const Indices rank(computeAsymptoticBilateralRank(sample.getSize()));
   const Sample sortedSample(sample.sort());
   return Interval(Point({sortedSample(rank[0], 0)}), Point({sortedSample(rank[1], 0)}));
 }
-
 
 /* Quantile level accessor */
 void QuantileConfidence::setAlpha(const Scalar alpha)
--- a/lib/src/Uncertainty/Algorithm/Simulation/openturns/QuantileConfidence.hxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/openturns/QuantileConfidence.hxx
@@ -47,17 +47,22 @@ public:
   QuantileConfidence * clone() const override;
 
   /** Compute ranks */
-  UnsignedInteger computeUnilateralRank(const UnsignedInteger size, const Bool tail = false) const;
+  UnsignedInteger computeUnilateralRank(const UnsignedInteger size, const Bool lowerBounded = false) const;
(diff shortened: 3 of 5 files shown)


mbaudin47:
@adutfoy , @jschueller : I suggest to review this PR.

lib/src/Uncertainty/Algorithm/Simulation/openturns/QuantileConfidence.hxx
```diff
@@ -47,17 +47,22 @@ public:
   QuantileConfidence * clone() const override;
 
   /** Compute ranks */
-  UnsignedInteger computeUnilateralRank(const UnsignedInteger size, const Bool tail = false) const;
+  UnsignedInteger computeUnilateralRank(const UnsignedInteger size, const Bool lower_bounded = false) const;
```
mbaudin47: Rename lower_bounded into lowerBounded

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -79,121 +81,178 @@ String QuantileConfidence::__str__(const String & /*offset*/) const
...
+
+  // Phase 1: Exponential search to find an exact valid bracket (low, high]
+  UnsignedInteger low = k;
+  UnsignedInteger high = k + 1;
+
+  while (true)
+  {
+    const Scalar probabilityTest = binomial.computeCDF(high);
+
+    if (probabilityTest > probabilityReference)
```
jschueller: we definitely should not use an epsilon, I have a proposal of fix in misc which I believe is better

mbaudin47: Why shouldn't we use an epsilon? What is wrong with this algorithm?

Here is the proof that the $\epsilon$ parameter is useful. I counted the number of evaluations of the Binomial  CDF (FEval) and quantile (QEval) with $\epsilon=0$ and $\epsilon = 10^{-15}$. This is the number of function calls which are directly evaluated in the computeBilateralRank and searchProbabilityJump. This is not the total number of CDF calls (evaluating the quantile function from Binomial.computeQuantile() can lead to several evaluations of the CDF).

|       | With $\epsilon = 0$ | With $\epsilon = 10^{-15}$ |
| ----- | ------------------- | -------------------------- |
| FEval | 1236                | 182                        |
| QEval | 406                 | 54                         |
| $k_1$ | 910                 | 910                        |
| $k_2$ | 935                 | 935                        |

We see that the number of CDF and quantile evaluations is divided by 7.

The reason for this improvement is as follows. Suppose $\epsilon=0$.
- We accept k1 = 0, k2 = 935, with coverage p = 0.916306
- The first jump in the CDF is found at k1 = 512, where p1 = 3.95253e-323. This does not improve the coverage, so it is rejected.
- Then the next jump is found at k1 = 513, where p1 = 7.16395e-322. This does not improve the coverage, so it is rejected.
- Then lots of iterations like these are performed without any acceptance of the pair $(k_1, k_2)$, because the coverage does not improve. 
- This happens until k1 = 861, where p1 = 1.18087e-16. Here we accept k1 = 861, k2 = 935, where the coverage p = 0.916306 (this is a small decrease of the coverage).
- Some iterations are performed until we finally accept k1 = 910, k2 = 935, with coverage p = 0.903463.
- A few iterations are performed, but they do not improve the coverage.

Suppose $\epsilon=10^{-15}$.
- We accept k1 = 0, k2 = 935, with coverage p = 0.916306
- Then the first **significant** jump is found at k1 = 864, where p1 = 1.85384e-15. Here we accept k1 = 864, k2 = 935, with coverage p = 0.916306
- The remaining iterations are the same as with $\epsilon=0$.

Notice that lots of iterations are saved in the first part of the algorithm, where the binomial is almost flat and does not bring any significant coverage.

mbaudin47: I improved the algorithm in order to make so that the CDF of the Binomial is never evaluated twice with the same value of k1 or k2. Here are the updated results.

|       | With $\epsilon = 0$ | With $\epsilon = 10^{-15}$ |
| ----- | ------------------- | -------------------------- |
| FEval | 428                 | 78                         |
| QEval | 406                 | 54                         |
| $k_1$ | 910                 | 910                        |
| $k_2$ | 935                 | 935                        |

jschueller: indeed this epsilon really makes a difference
it should not be arbitrary though, I propose to take the machine epsilon times beta, since that we pass a delta of cdf + beta in the quantile function

mbaudin47: Scaling epsilon to beta is a good idea indeed... in theory. In practice, beta is in the [0.5, 1] interval. So $\epsilon \beta$ is actually close to $\epsilon$ in this case.

This is why I resolve this discussion.

mbaudin47:
At this point, for the size n=975 example, the number of CDF evaluations is equal to 164 and the number of quantile evaluations is equal to 1. The values of $k$ in the range [865, 917] are evaluated twice: first as the leading edge of the window (k2) and later as the trailing edge (k1). We can cache the values of the CDF in this range, which can save approximately 30% of the CDF evaluations. This can be done by caching all values from Quantile(epsilon) to Quantile(1 - epsilon). I am not sure that the savings are worth, compared to the memory requirements.

Here is a benchmark between PR #3176 and PR #3187.

|       | PR 3176 (MBN)                                                                   | PR 3187 (JSR)                                                                   |
| ----- | ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
|       | [Gist Link](https://gist.github.com/mbaudin47/35f8413ce68311af577976b2b8b61b99) | [Gist Link](https://gist.github.com/mbaudin47/0f55b57d1fa6589481e6de54e5bdff10) |
| $k_1$ | 910                                                                             | 910                                                                             |
| $k_2$ | 935                                                                             | 935                                                                             |
| FEval | 164                                                                             | 186                                                                             |
| QEval | 1                                                                               | 55                                                                              |
| Lines of C++ code | 74 | 94 |

**Table 1.** Results and performance of two algorithms to solve the problem presented at #3182.

Both algorithms provide the same results in this case. This shows that the PR #3176 provides a more efficient algorithm than PR #3187, with an improvement of 55x for the number of quantile evaluations. If also shows that the code is significantly shorter.

We can improve over the PR #3176 by using a cache to save 30% of CDF evaluations. But this will cost some memory. The method may be based on pre-evaluating the Binomial CDF for all k between $F_{n,\alpha}^{-1}(\epsilon)$ and $F_{n,\alpha}^{-1}(1 - \epsilon)$, where $F_{n,\alpha}$ is the CDF of the $\text{Binomial}(n,\alpha)$. If $\epsilon = 10^{-15}$, $n=1000$ and $\alpha = 0.95$ for example, this amounts to allocate an array with length 107, which is manageable (less than 1KB). With $n = 10^6$, the length of the array grows to 3363 (24kB). Also, this requires a little more complicated algorithm and will be more complicated to maintain. My conclusion is that the result is not worth the effort.

mbaudin47:
@jschueller , @adutfoy : I suggest to review this PR.

jschueller:
@coderabbitai reivew

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -54,9 +54,11 @@ QuantileConfidence::QuantileConfidence(const Scalar alpha, const Scalar beta)
   , beta_(beta)
 {
   if (!(alpha >= 0.0) || !(alpha <= 1.0))
```
josephmure:
```suggestion
  if (!(alpha >= 0.0 && alpha <= 1.0))
```
Equivalent but easier to read IMO.

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -54,9 +54,11 @@ QuantileConfidence::QuantileConfidence(const Scalar alpha, const Scalar beta)
...
-    throw InvalidArgumentException(HERE) << "Quantile level must be in [0, 1]";
+    throw InvalidArgumentException(HERE)
+      << "Quantile level must be in [0, 1], but alpha = " << alpha;
   if (!(beta >= 0.0) || !(beta <= 1.0))
```
josephmure:
```suggestion
  if (!(beta >= 0.0 && beta <= 1.0))
```

python/src/QuantileConfidence_doc.i
```diff
@@ -6,79 +6,94 @@ Refer to :any:`quantile_confidence_estimation`.
...
+This class estimates bounds of the quantile of level
+:math:`\alpha \in [0,1]` of the random variable :math:`X`
 with a confidence greater than :math:`\beta` using order statistics.
 
-Let :math:`x_{\alpha}` be the unknown quantile of level :math:`\alpha` of the random variable :math:`X` of dimension 1.
-Let :math:`(X_0, \dots, X_{\sampleSize-1})` be a sample of independent and identically distributed variables according to :math:`X`.
+Let :math:`x_{\alpha}` be the unknown quantile of level :math:`\alpha`
+of the random variable :math:`X` of dimension 1.
+Let :math:`(X_0, \dots, X_{\sampleSize - 1})` be a sample of independent
+and identically distributed variables according to :math:`X`.
```
josephmure:
```suggestion
Let :math:`(X_0, \dots, X_{\sampleSize - 1})` be independent
copies of :math:`X`.
```

I think this is clearer and better terminology: https://math.stackexchange.com/a/2179361

python/src/QuantileConfidence_doc.i
```diff
@@ -6,79 +6,94 @@ Refer to :any:`quantile_confidence_estimation`.
...
 
+Note: Python uses 0-based indexing.
+In textbook notation, order statistics often use 1-based indexing where
+:math:`X_{(k)}` with :math:`1 \leq k \leq \sampleSize` denotes the k-th smallest value.
+In this class, we use 0-based indexing where the textbook's :math:`X_{(k)}`
+corresponds to :math:`X_{(k - 1)}` in our notation.
 
 Examples
 --------
+In the next example, we compute a confidence interval
```
josephmure:
```suggestion
In the following example, we compute a confidence interval
```

python/src/QuantileConfidence_doc.i
```diff
@@ -6,79 +6,94 @@ Refer to :any:`quantile_confidence_estimation`.
...
 
 // ---------------------------------------------------------------------
 
 %feature("docstring") OT::QuantileConfidence::computeUnilateralRank
-R"RAW(Evaluate the rank of a quantile lower bound.
+R"RAW(Evaluate the rank of a quantile lower or upper bound.
 
-The lower rank :math:`k_{low}` is the largest rank :math:`k`  with :math:`0 \leq k \leq \sampleSize -1` such that:
+If `lowerBounded` is `True`, the lower rank :math:`k_{low}` is the largest rank :math:`k` with
+:math:`0 \leq k \leq \sampleSize - 1` such that:
```
josephmure:
```suggestion
:math:`0 \leq k < \sampleSize` such that:
```

I think this notation is more intuitive for Python users because they are used to it : array[0:n] is commonly used to get the elements of an array with index between 0 and n-1.

python/src/QuantileConfidence_doc.i
```diff
@@ -6,79 +6,94 @@ Refer to :any:`quantile_confidence_estimation`.
...
 
-In other words, the interval :math:`\left[ X_{(k_{low})}, +\infty\right[` is a unilateral confidence
-interval for the quantile :math:`x_\alpha` with confidence :math:`\beta`.
+In other words, the interval :math:`\left[ X_{(k_{low})}, +\infty\right)` is a
+unilateral confidence interval for the quantile :math:`x_\alpha` with
+confidence :math:`\beta`.
 
-The upper rank :math:`k_{up}` is the smallest rank :math:`k` with :math:`0 \leq k \leq \sampleSize -1` such that:
+If `lowerBounded` is `False`, the upper rank :math:`k_{up}` is the smallest rank :math:`k` with
+:math:`0 \leq k \leq \sampleSize - 1` such that:
```
josephmure:
```suggestion
:math:`0 \leq k \< \sampleSize ` such that:
```

python/src/QuantileConfidence_doc.i
```diff
@@ -88,35 +103,44 @@ rank : int
 
 Examples
 --------
+In the next example, we consider a sample which size is 100.
```
josephmure:
```suggestion
In the following example, we consider a 100 point sample.
```

python/src/QuantileConfidence_doc.i
```diff
@@ -88,35 +103,44 @@ rank : int
...
+8
 )RAW"
 
 // ---------------------------------------------------------------------
 
 %feature("docstring") OT::QuantileConfidence::computeBilateralRank
 R"RAW(Evaluate the ranks of a quantile bilateral bound.
 
-The ranks :math:`k_1, k_2` with :math:`0 \leq k_1, k_2 \leq \sampleSize -1` are defined by:
+The ranks :math:`k_1, k_2` with :math:`0 \leq k_1 \leq k_2 \leq \sampleSize - 1`
```
josephmure:
```suggestion
The ranks :math:`k_1, k_2` with :math:`0 \leq k_1 \leq k_2 < \sampleSize`
```

python/src/QuantileConfidence_doc.i
```diff
@@ -142,45 +166,47 @@ Examples
...
+by the order statistic:
 
 .. math::
 
-    [X_{(k_{low})}, +\infty[
+    \left[X_{(k)}, +\infty\right)
 
-where :math:`k_{low}` is the largest rank with :math:`0 \leq k_{low} \leq \sampleSize - 1` so that:
+where :math:`k` is the largest rank such that
+:math:`0 \leq k \leq \sampleSize - 1` and:
```
josephmure:
```suggestion
:math:`0 \leq k < \sampleSize` and:
```

python/src/QuantileConfidence_doc.i
```diff
@@ -142,45 +166,47 @@ Examples
...
+by the order statistic:
 
 .. math::
 
-    ]-\infty, X_{(k_{up})}]
+    \left(-\infty, X_{(k)}\right]
 
-where :math:`k_{up}` is the smallest rank with :math:`0 \leq k_{up} \leq \sampleSize - 1` so that:
+where :math:`k` is the smallest rank such that
+:math:`0 \leq k \leq \sampleSize - 1` and:
```
josephmure:
```suggestion
:math:`0 \leq k < \sampleSize` and:
```

python/src/QuantileConfidence_doc.i
```diff
@@ -142,45 +166,47 @@ Examples
...
-    Quantile level
-lower_bounded : bool, optional
-    True indicates the interval is bounded by a lower value.
-
-    False indicates the interval is bounded by an upper value.
-
+    Sample of the variable :math:`X`.
+lowerBounded : bool, optional
+    False to compute the upper bound of the quantile,
+    True for the lower bound.
```
josephmure:
```suggestion
    True to compute its lower bound.
```

python/src/QuantileConfidence_doc.i
```diff
@@ -190,28 +216,33 @@ ci : :class:`~openturns.Interval`
 
 Examples
 --------
+In the next example, we compute a confidence interval
```
josephmure:
```suggestion
In the following example, we compute a confidence interval
```

python/src/QuantileConfidence_doc.i
```diff
@@ -190,28 +216,33 @@ ci : :class:`~openturns.Interval`
...
 ----------
 sample : 2-d sequence of float
-    Quantile level
-lower_bounded : bool, optional
-    True indicates the interval is bounded by a lower value.
-    False indicates the interval is bounded by an upper value.
+    Sample of the variable :math:`X`.
+lowerBounded : bool, optional
+    False to compute the upper bound of the quantile,
+    True for the lower bound.
```
josephmure:
```suggestion
    True to compute its lower bound.
```

python/src/QuantileConfidence_doc.i
```diff
@@ -220,43 +251,50 @@ ci : :class:`~openturns.Interval`
...
+The confidence interval for the quantile of level :math:`\alpha` is given by
+the order statistics:
 
 .. math::
 
     [X_{(k_1)}, X_{(k_2)}]
 
-where :math:`(k_1, k_2)`  are the ranks with :math:`0 \leq k_1 \leq k_2 \leq n - 1` defined by:
+where :math:`(k_1, k_2)`  are the ranks with
+:math:`0 \leq k_1 \leq k_2 \leq \sampleSize - 1` defined by:
```
josephmure:
```suggestion
:math:`0 \leq k_1 \leq k_2 < \sampleSize` defined by:
```
There are a few other examples below I believe.

python/src/QuantileConfidence_doc.i
```diff
@@ -301,103 +342,166 @@ ci : :class:`~openturns.Interval`
...
 
 Returns
 -------
 size : int
-    Minimum sample size :math:`n`.
+    Minimum sample size :math:`\sampleSize`.
 
 Examples
 --------
+In the next example, we compute the smallest sample size such that the
```
josephmure:
```suggestion
In the following example, we compute the smallest sample size such that the
```

python/src/QuantileConfidence_doc.i
```diff
@@ -301,103 +342,166 @@ ci : :class:`~openturns.Interval`
...
 )RAW"
 
 // ---------------------------------------------------------------------
 
 %feature("docstring") OT::QuantileConfidence::computeBilateralMinimumSampleSize
-R"RAW(Evaluate the minimum size of the sample for the bilateral min - max confidence interval.
+R"RAW(Evaluate the minimum size of the sample for the bilateral confidence interval.
 
-The method determines the smallest sample size :math:`n` so that:
+The method determines the smallest sample size :math:`\sampleSize` so that:
```
josephmure:
```suggestion
The method determines the smallest sample size :math:`\sampleSize` such that:
```

python/src/QuantileConfidence_doc.i
```diff
@@ -534,3 +656,94 @@ Returns
...
+
+Parameters
+----------
+size : int
+    Sample size :math:`\sampleSize`.
+rank : int
+    Rank :math:`k` of the order statistic.
+lowerBounded : bool, optional
+    False to consider the upper bound of the quantile,
+    True for the lower bound.
```
josephmure:
```suggestion
    True to consider its lower bound.
```

josephmure (changes requested):

python/src/QuantileConfidence_doc.i
mbaudin47: The "size" is the second attribute of a `Sample`. So I would rather use:
```suggestion
In the following example, we consider a sample of size 100.
```

mbaudin47:
@jschueller , @adutfoy : I suggest to review this PR.

jschueller:
> At this point, for the size n=975 example, the number of CDF evaluations is equal to 164 and the number of quantile evaluations is equal to 1. The values of k in the range [865, 917] are evaluated twice: first as the leading edge of the window (k2) and later as the trailing edge (k1). We can cache the values of the CDF in this range, which can save approximately 30% of the CDF evaluations. This can be done by caching all values from Quantile(epsilon) to Quantile(1 - epsilon). I am not sure that the savings are worth, compared to the memory requirements.
> 
> Here is a benchmark between PR #3176 and PR #3187.
> 	PR 3176 (MBN) 	PR 3187 (JSR)
> 	[Gist Link](https://gist.github.com/mbaudin47/35f8413ce68311af577976b2b8b61b99) 	[Gist Link](https://gist.github.com/mbaudin47/0f55b57d1fa6589481e6de54e5bdff10)
> k 1 	910 	910
> k 2 	935 	935
> FEval 	164 	186
> QEval 	1 	55
> Lines of C++ code 	74 	94
> 
> **Table 1.** Results and performance of two algorithms to solve the problem presented at #3182.
> 
> Both algorithms provide the same results in this case. This shows that the PR #3176 provides a more efficient algorithm than PR #3187, with an improvement of 55x for the number of quantile evaluations. If also shows that the code is significantly shorter.
> 
> We can improve over the PR #3176 by using a cache to save 30% of CDF evaluations. But this will cost some memory. The method may be based on pre-evaluating the Binomial CDF for all k between F n , α − 1 ( ϵ ) and F n , α − 1 ( 1 − ϵ ) , where F n , α is the CDF of the Binomial ( n , α ) . If ϵ = 10 − 15 , n = 1000 and α = 0.95 for example, this amounts to allocate an array with length 107, which is manageable (less than 1KB). With n = 10 6 , the length of the array grows to 3363 (24kB). Also, this requires a little more complicated algorithm and will be more complicated to maintain. My conclusion is that the result is not worth the effort.

I think its about the quantile call, but your algo is significantly slower than in my branch:
```
import openturns as ot
import openturns.testing as ott
import time

beta = 0.95
t0 = time.time()
for rep in range(int(1e3)):
    print(rep)
    alpha = ot.RandomGenerator.Generate()
    algo = ot.QuantileConfidence(alpha, beta)
    n = max(100, ot.RandomGenerator.IntegerGenerate(int(1e9)))
    k1, k2 = algo.computeBilateralRank(n)
    # binomial = ot.Binomial(n, alpha)
    # p = binomial.computeProbability(ot.Interval(k1, k2))
    # print(f"n={n} k1={k1} k2={k2} p={p:.6f}")
    # assert (k1, k2) == k_ref[n]
    # assert p >= beta
t1 = time.time()
print(t1-t0)
```

mbaudin47:
> I think its about the quantile call, but your algo is significantly slower than in my branch: [...]

I confirm this. I slightly modified the script to print intermediate messages (see [benchmark_computeBilateralRank.py](https://gist.github.com/mbaudin47/97f60c020b567a121098676d8894e450#file-benchmark_computebilateralrank-py)).

|             | PR 3176 (MBN) | PR 3187 (Schueller) |
| ----------- | ------------- | ------------------- |
| Performance | 0.08 (it/s)   | 8.01 (it/s)         |

The conclusion is that your algorithm is roughly 100x faster with very large values of $n$. 

There is one problem with your algorithm for some values for which the probability of coverage is not correct:

```
n=674726183, alpha=0.51762756742630334
k1=349193433, k2=349278222
p=0.94999999992866635
```

Edit. I updated the algorithm. The new algorithm has 2.05 (it/s).

mbaudin47:
- Suggestions: Deprecate computeUnilateralConfidenceIntervalWithCoverage() and computeBilateralConfidenceIntervalWithCoverage. It suffices to evaluate computeBilateralCoverage and computeUnilateralCoverage as a post-processing.
- What happens in terms of accuracy if $\alpha \rightarrow 0$ in the expression:

$$
1 - (1 - \alpha)^n
$$

and in:

$$
1 - \alpha^n
$$

josephmure (approved):
Good to go, provided the sample-based method is deprecated.

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -79,121 +81,332 @@ String QuantileConfidence::__str__(const String & /*offset*/) const
...
+  Bool found = false;
+  Scalar pBest = 2.0;
+  UnsignedInteger k1Best = 0;
+  UnsignedInteger k2Best = size;
+
+  const Scalar probabilityEpsilon = ResourceMap::GetAsScalar("QuantileConfidence-ProbabilityEpsilon");
+  UnsignedInteger k1 = binomial.computeScalarQuantile(probabilityEpsilon);
+  Scalar p1 = binomial.computeCDF(k1);
+
+  UnsignedInteger lastK2 = -1;
```
regislebrun: UnsignedInteger equals to -1???

lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -1819,6 +1819,10 @@ void ResourceMap::loadDefaultConfiguration()
 
   // viewer.View parameters //
   addAsString("View-ImageFormat", "png");
+
+  // QuantileConfidence parameters //
+  addAsScalar("QuantileConfidence-ProbabilityEpsilon", 1.0e-15);
+  addAsString("QuantileConfidence-EpsilonMethod", "Jump");
```
mbaudin47: Use https://github.com/openturns/openturns/blob/6002aa6e22310a37ffb14326fe5776c5a78a30b6/lib/src/Base/Common/ResourceMap.cxx#L932
to define all the values possible.

mbaudin47: Fixed at cddd0e0e1a604ee8b28d5ce15be69ab6d8191e07

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
mbaudin47: Fixed at cddd0e0e1a604ee8b28d5ce15be69ab6d8191e07

mbaudin47:
@jschueller : I suggest to make a review of computeBilateralRankHybrid at : https://github.com/mbaudin47/openturns/blob/cddd0e0e1a604ee8b28d5ce15be69ab6d8191e07/lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx#L123

mbaudin47:
In order to improve the speed of `computeBilateralRank`, I updated the algorithm and create a Binomial once an for all instead of creating it each time `searchProbabilityJump` is evaluated. This improves the speed by a 4x factor. Another improvement is to compute $k_1$ depending on $k_2$ instead of searching for a significant increase of $F(k_1)$, regardless of $k_2$.

benchmark_alpha_variation_100000000

benchmark_alpha_0 5


Here are updated benchmark results.
- Hybrid algorithm : 8.39 (it/s)
- Bisection (by JSN from #3211) : 7.82 (it/s)
- Jump : 8.42 (it/s)
- Epsilon 0.13 (it/s)
- DoubleJump : 27.79 (it/s)

From the benchmark we conclude that:
- the "epsilon" algorithm is significantly slower than the others; it is especially slow when alpha is close to zero;
- the "jump", "bisection" and "hybrid" have almost the same speed compared to "epsilon";
- the "bisection" algorithm is slighly slower than "jump" and "hybrid";
- "jump" and "hybrid" are almost indistinguishable in terms of elapsed time;
- the faster is "DoubleJump";
- for a given alpha and beta, the log of the elapsed time grows almost linearily depending on the log of the sample size;
- for a given value of n and beta, the elapsed time is larger when alpha is close to 0.5 and smaller when alpha is close to 0 or 1.

mbaudin47:
@adutfoy : I suggest to review this PR, with the exception of computeBilateralRank() (which will come when @jschueller will merge the dedicated PR #3211).

mbaudin47:
@adutfoy : I suggest to review this PR, with the exception of computeBilateralRank() (which will come when @jschueller will merge the dedicated PR https://github.com/openturns/openturns/pull/3211).

adutfoy (approved):

adutfoy (approved):

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -79,30 +81,439 @@ String QuantileConfidence::__str__(const String & /*offset*/) const
...
   }
-
-  const UnsignedInteger rank = binomial.computeQuantile(beta_, tail)[0];
+  const UnsignedInteger rank = binomial.computeQuantile(beta_, lowerBounded)[0];
   return rank;
 }
 
-// compute argmin (k1, k2) P_X([k1, k2]) under constraint P_X([k1, k2])>=beta, with X~B(n, alpha)
 Indices QuantileConfidence::computeBilateralRank(const UnsignedInteger size) const
 {
```
jschueller: revert the computeBilateralRank changes

mbaudin47:
@jschueller : As a tip to review python/test/t_QuantileConfidence_std.py, I suggest to use Meld and ignore the white spaces.

TODO-List:
- [x] Deprecate `QuantileConfidence.computeUnilateralConfidenceIntervalWithCoverage()`  Use `computeUnilateralCoverage()` instead.
- [x] Deprecate `QuantileConfidence.computeBilateralConfidenceIntervalWithCoverage()`. Use `computeBilateralCoverage()` instead.
- [x] Remove unnecessary ResourceMap keys.

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
mbaudin47: Done.

mbaudin47:
@jschueller : This is ready for merging.

mbaudin47:
@jschueller : This is ready for merging in my opinion. I am waiting for this code to be merged so that I can update some Python scripts that are making use of these features.

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -233,12 +291,12 @@ class QuantileConfidenceEvaluation: public EvaluationImplementation
...
   Scalar alpha_ = 0.0;
   UnsignedInteger rank_ = 0;
-  Bool tail_ = true;
+  Bool lowerBounded_ = true;
 }; // class QuantileConfidenceEvaluation
 } // Anonymous namespace
 
 
-UnsignedInteger QuantileConfidence::computeUnilateralMinimumSampleSize(const UnsignedInteger rank, const Bool tail) const
+UnsignedInteger QuantileConfidence::computeUnilateralMinimumSampleSize(const UnsignedInteger tail_rank, const Bool lowerBounded) const
```
jschueller, line 299: tail_rank should be renamed to rank (it can be one side or the other), or at least it must be consistent with the header/docstring

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -267,18 +325,19 @@ UnsignedInteger QuantileConfidence::computeUnilateralMinimumSampleSize(const Uns
     // We compute a reasonable guess for n using a Normal approximation:
     // n*alpha+q_beta*sqrt(n*alpha*(1-alpha))=n-r:
     const Scalar aBeta = DistFunc::qNormal(beta_);
-    UnsignedInteger nMax = static_cast<UnsignedInteger>((rank + 0.5 * (alpha_ * aBeta * aBeta + std::abs(aBeta) * std::sqrt(alpha_ * (4.0 * rank + alpha_ * aBeta * aBeta)))) / (1.0 - alpha_));
+    UnsignedInteger nMax = static_cast<UnsignedInteger>((tail_rank + 0.5 * (alpha_ * aBeta * aBeta + std::abs(aBeta) * std::sqrt(alpha_ * (4.0 * tail_rank + alpha_ * aBeta * aBeta)))) / (1.0 - alpha_));
```
jschueller, line 328: the denominator 1-alpha ignores the value of lowerBounded, is this ok for the upper bound case ?

lib/src/Uncertainty/Algorithm/Simulation/openturns/QuantileConfidence.hxx
```diff
@@ -47,17 +47,22 @@ public:
...
   /** Compute ranks */
-  UnsignedInteger computeUnilateralRank(const UnsignedInteger size, const Bool tail = false) const;
+  UnsignedInteger computeUnilateralRank(const UnsignedInteger size, const Bool lowerBounded = false) const;
   Indices computeBilateralRank(const UnsignedInteger size) const;
 
   /** Compute confidence intervals */
-  Interval computeUnilateralConfidenceInterval(const Sample & sample, const Bool tail = false) const;
-  Interval computeUnilateralConfidenceIntervalWithCoverage(const Sample & sample, Scalar & coverageOut, const Bool tail = false) const;
+  Interval computeUnilateralConfidenceInterval(const Sample & sample, const Bool lowerBounded = false) const;
+  Interval computeUnilateralConfidenceIntervalWithCoverage(const Sample & sample, Scalar & coverageOut, const Bool lowerBounded = false) const;
```
jschueller, line 55: add @deprecated marker for these

ChangeLog
```diff
@@ -33,6 +33,8 @@
  * LOLAVoronoi moved out of experimental
  * Deprecated GaussianProcessConditionalCovariance::operator() for getMarginalDistribution
  * Deprecated ProcessSample.setField(Field, int) in favor ProcessSample.setField(int, Field)
+ * Deprecated QuantileConfidence.computeUnilateralConfidenceIntervalWithCoverage(Sample) in favor of QuantileConfidence.computeUnilateralCoverage(UnsignedInteger, UnsignedInteger, Bool)
+ * Deprecated QuantileConfidence.computeBilateralConfidenceIntervalWithCoverage(Sample) in favor of computeBilateralCoverage(UnsignedInteger, UnsignedInteger, UnsignedInteger)
```
jschueller, line 37: update TODO too

python/src/QuantileConfidence_doc.i
```diff
@@ -190,28 +216,33 @@ ci : :class:`~openturns.Interval`
...
 >>> sample = ot.Gumbel().getSample(100)
 >>> ci = algo.computeUnilateralConfidenceInterval(sample)
+>>> print(ci.getUpperBound()[0])
+3.999...
 )RAW"
 
 // ---------------------------------------------------------------------
 
 %feature("docstring") OT::QuantileConfidence::computeUnilateralConfidenceIntervalWithCoverage
-R"RAW(Evaluate an unilateral confidence interval of a quantile.
```
jschueller, line 204: I would remove the doc of deprecated methods

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -141,36 +149,40 @@ Indices QuantileConfidence::computeBilateralRank(const UnsignedInteger size) con
 }
 
 // compute interval of the form [X_k; +inf[ or ]-inf; X_k] from unilateral rank k
-Interval QuantileConfidence::computeUnilateralConfidenceInterval(const Sample & sample, const Bool tail) const
+Interval QuantileConfidence::computeUnilateralConfidenceInterval(const Sample & sample, const Bool lowerBounded) const
 {
   Scalar coverageOut = -1.0;
-  return computeUnilateralConfidenceIntervalWithCoverage(sample, coverageOut, tail);
+  return computeUnilateralConfidenceIntervalWithCoverage(sample, coverageOut, lowerBounded);
```
jschueller, line 155: should not call deprecated methods here, else you'd get the deprecation warnings even in non-deprecated methods

python/src/QuantileConfidence_doc.i
jschueller, line 329: deprecated, can be removed

jschueller:
-1
we should not change the argument name for consistency with other methods
instead we should just swap the cases in computeUnilateralRank & computeUnilateralMinimumSampleSize
other methods are fine

PR #3179, opened 2026-05-02
Fix formatting in sensitivity_sobol_from_pce.rst

mbaudin47:
Fixes an error alread fixed in #2881, but introduced again in #2883.

The error is this: the input of the metamodel is the physical input random vector **X**, not the standard input random vector **Z**. This is consistent with the model, which input is the physical input random vector **X**. The variance is not changed by an isoprobabilistic transform on the input. This is why the remaining equations, including the Sobol' indices, are ok.


mbaudin47:
@adutfoy , @josephmure : I suggest to review this PR.

mbaudin47:
@adutfoy , @josephmure : I suggest to review this PR.

sofianehaddad (approved):

PR #3190, opened 2026-05-08
Added SklarCopula::computeLogPDF

### LINKED ISSUE #3185: StudentCopulaFactory does not estimate the shape parameter nu
### What happened?

The estimation of a Student copula is done the following way:
  * First, the correlation matrix is estimated using the empirical Kendall tau and the universal relationship 
     with the correlation of elliptical copulas.
  * Then, the $\nu$ parameter is optimized in order to maximize the likelihood of the sample.
     This optimization is delegated to the MaximumLikelihoodFactory class. The starting point is given by the *StudentCopulaFactory-NuStart* key in ResourceMap and the bounds by the *StudentCopulaFactory-NuMin* and *StudentCopulaFactory-NuMax* keys.

For an unknown reason, the last step fails: the optimizer is stuck at the initial value.

```Python
import openturns as ot

copula = ot.StudentCopula(3.5, ot.CorrelationMatrix(3, [1.0, 0.2, 0.3, 0.2, 1.0, 0.4, 0.3, 0.4, 1.0]))
sample = copula.getSample(100000)
estimated = ot.StudentCopulaFactory().buildAsStudentCopula(sample)
print(f"Estimated nu={estimated.getNu()}")

gives:

Estimated nu=5.0
```

### Version

1.28dev

### Operating System

Linux

### Installation media

from source

### Additional Context

_No response_

regislebrun:
This method was missing, resulting into numerical issues in StudentCopulaFactory. It fixes #3185.

### DIFF
--- a/python/test/t_StudentCopulaFactory_std.py
+++ b/python/test/t_StudentCopulaFactory_std.py
@@ -31,3 +31,12 @@ estimatedStudentCopula = factory.buildAsStudentCopula()
 print("Default copula=", estimatedStudentCopula)
 estimatedStudentCopula = factory.buildAsStudentCopula(distribution.getParameter())
 print("StudentCopula from parameters=", estimatedStudentCopula)
+
+copula = ot.StudentCopula(3.5, ot.CorrelationMatrix(3, [1.0, 0.2, 0.3, 0.2, 1.0, 0.4, 0.3, 0.4, 1.0]))
+sample = copula.getSample(1000)
+estimated = ot.StudentCopulaFactory().buildAsStudentCopula(sample)
+
+nuRef = 3.5
+nuEst = estimated.getNu()
+print(f"Estimated nu={nuEst}")
+ott.assert_almost_equal(nuEst, nuRef, 1e-1, 1e-1)
--- a/lib/src/Uncertainty/Model/SklarCopula.cxx
+++ b/lib/src/Uncertainty/Model/SklarCopula.cxx
@@ -148,6 +148,35 @@ Point SklarCopula::computeDDF(const Point & point) const
   return result;
 }
 
+/* Get the log PDF of the distribution
+   F(x_1,\dots,x_n) = C(F_1(x_1),\dots,F_n(x_n))
+   so p(x_1,\dots,x_n) = c(F_1(x_1),\dots,F_n(x_n))\prod_{i=1}^n p_i(x_i)
+   and logp(x_1,\dots,x_n) = logc(F_1(x_1),\dots,F_n(x_n)) + \sum_{i=1}^n logp_i(x_i) */
+Scalar SklarCopula::computeLogPDF(const Point & point) const
+{
+  const UnsignedInteger dimension = getDimension();
+  if (point.getDimension() != dimension)
+    throw InvalidArgumentException(HERE) << "Error: the given point must have dimension " << dimension << ", here dimension=" << point.getDimension();
+
+  // Early exit for the independent case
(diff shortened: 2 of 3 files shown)


jschueller:
could you add the testcase from #3185 ? dont forget to rebase also

python/test/t_StudentCopulaFactory_std.py
```diff
@@ -31,3 +31,12 @@
 print("Default copula=", estimatedStudentCopula)
 estimatedStudentCopula = factory.buildAsStudentCopula(distribution.getParameter())
 print("StudentCopula from parameters=", estimatedStudentCopula)
+
+copula = ot.StudentCopula(3.5, ot.CorrelationMatrix(3, [1.0, 0.2, 0.3, 0.2, 1.0, 0.4, 0.3, 0.4, 1.0]))
+sample = copula.getSample(100000)
```
jschueller: looks like aarch64 timeouts on this test, maybe try reduce the size to 1e4 ?

python/test/t_StudentCopulaFactory_std.py
```diff
@@ -31,3 +31,12 @@
 print("Default copula=", estimatedStudentCopula)
 estimatedStudentCopula = factory.buildAsStudentCopula(distribution.getParameter())
 print("StudentCopula from parameters=", estimatedStudentCopula)
+
```
mbaudin47, line 34: Since a new computeLogPDF() method is introduced, I suggest to unit test it. This can be done by pre-computing a reference value using the naive formula and compute the log. Then the test compares the reference value against the computed one. If the test case is well chosen, then the naive formula will not lose too much accuracy and the test will be relevant.

mbaudin47 (changes requested):
The PR can be improved with a unit test corresponding to the added feature.

python/test/t_StudentCopulaFactory_std.py
regislebrun: There is already a checkLogPDF() method in DistributionValidation, which does exactly this test ;-)! This test is done in both t_SklarCopula_std.py and t_StudentCopula_std.py

mbaudin47: You made a good point. But notice that the current `python/test/t_StudentCopulaFactory_std.py` does not use `DistributionValidation` (you may think that this is automatically done elsewhere, but this is not the case, as far as I understand). I suggest to update the unit test and use it. You may be inspired by the current test in:

https://github.com/openturns/openturns/blob/6e22849d788e09c47cd092c6ce0e871923659130/python/test/t_Uniform_std.py#L108

regislebrun: The bug (and the correction) was *not* in StudentCopulaFactory (nothing has changed in this class). It was a limitation in SklarCopula:computeLogPDF(). I saw it using StudentCopulaFactory, but the actual class to test is SklarCopula or eventually StudentCopula, even if its computeLogPDF() is directly inherited from the SklarCopula class.
IMO it would looks awkward to have a call to StudentCopula::computeLogPDF() or to DistributionValidation inside of t_StudentCopulaFactory_std.py, the correct place is in the corresponding distribution test (t_StudentCopula_std.py) and it is already done.

mbaudin47: Ok, and SklarCopula already tests for the validity of the class.

https://github.com/openturns/openturns/blob/6e22849d788e09c47cd092c6ce0e871923659130/python/test/t_SklarCopula_std.py#L137

So everything seems to be ok here.

mbaudin47 (approved):

PR #3192, opened 2026-05-10
Documentation: improves plot_enumeratefunction example

mbaudin47:
This PR fixes typos introduced in #3145. It also improves and simplifies the example. It uses the viridis palette whenever necessary.

image

**Figure 1.** The multi-indices from linear enumeration in the current doc (master).

image

**Figure 2.** The multi-indices from linear enumeration from the current PR (updated).

image

**Figure 3.** The contours of the q-Norm in the current doc (master).

image

**Figure 4.** The contours of the q-Norm from the current PR (updated).

image

**Figure 5.** The multiindices of the hyperbolic rule in the current doc (master).

image

**Figure 6.** The multiindices of the hyperbolic rule from the current PR (updated).

## Summary by CodeRabbit

* **Documentation**
  * Clarified how enumeration-function bijections map within the polynomial-chaos workflow example.

* **Visualization**
  * Refreshed plot styling with per-stratum Viridis colors, consistent titles, and fixed legend placement.
  * Streamlined hyperbolic and anisotropic hyperbolic strata plots with clearer wrappers and updated sizing.
  * Improved weighted \(L_q\) (quasi-norm) rendering using a dedicated builder with input validation, updated contour drawing, and revised \(q\) panel values.

* **Refactor**
  * Refined coefficient-count calculations and standardized figure sizing in the example views.

### DIFF
--- a/python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
+++ b/python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
@@ -4,15 +4,15 @@ Plot enumeration function
 """
 
 # %%
-# This example illustrates the enumeration functions which are ijections between :math:`\Nset` into :math:`\Nset^\inputDim`.
+# This example illustrates the enumeration functions which are bijections between :math:`\Nset` into :math:`\Nset^\inputDim`.
 # Refer to :ref:`enumeration_strategy` to get the precise description of the enumerate functions. We detail here the bijections:
 #
-# - Linear enumeration function
-# - Hyperbolic enumeration function
-# - Anisotropic hyperbolic enumeration function
-# - Infinity norm enumeration function
+# - linear enumeration function;
+# - hyperbolic enumeration function;
+# - anisotropic hyperbolic enumeration function;
+# - infinity norm enumeration function.
 #
-# These bijections are used in the in the :ref:`functional chaos expansion setting <functional_chaos>`.
+# These bijections are used in the :ref:`functional chaos expansion setting <functional_chaos>`.
 # In order to build up a multivariate basis :math:`\{\psi_{\vect{\alpha}},\vect{\alpha} \in \Nset^\inputDim\}`
 # by tensorization of univariate basis terms, we need to enumerate the multi-indices :math:`\vect{\alpha} \in \Nset^\inputDim`.
 # In this example, we interprete the impact of the different enumeration functions within the functional chaos expansion setting.
@@ -20,12 +20,14 @@ Plot enumeration function
 import openturns as ot
 import openturns.viewer as otv
 import math as m
+import matplotlib.colors as mcolors
+import matplotlib.pyplot as plt
 
 # %%
-# The simplest way to generate the multi-indices is to enumerate the terms of increasing length.
+# The simplest way to generate multi-indices is to list them by increasing length.
 # If the basis is polynomial, then the length corresponds to the total degree of the polynomial.
 # Within a strata, the multi-indices are ordered according to the "graded reverse lexicographic
-# ordering" in [sullivan2015]_.
+# ordering" (see [sullivan2015]_).
 # This is named the linear enumeration rule in the library.
 #
 # We print the ordered elements in dimension 4 to illustrate the ordering of the multi-indices.
@@ -63,12 +65,24 @@ def draw_stratas(enum_func, maximum_strata_index):
     graph : openturns.Graph
         Plot of the multi-indices colored by stratas
     """
+    if maximum_strata_index < 1:
+        raise ValueError(
+            f"maximum_strata_index must be >= 1, got {maximum_strata_index}"
+        )
+    cmap = plt.get_cmap("viridis")
+    if maximum_strata_index == 1:
+        viridis_palette = [mcolors.to_hex(cmap(0.0))]
+    else:
+        viridis_palette = [
+            mcolors.to_hex(cmap(i / (maximum_strata_index - 1)))
+            for i in range(maximum_strata_index)
+        ]
     graph = ot.Graph("", "$\\alpha_1$", "$\\alpha_2$", True)
-    if enum_func.__class__.__name__ == "LinearEnumerateFunction":
+    if isinstance(enum_func, ot.LinearEnumerateFunction):
         graph.setTitle("Linear enumeration rule")
-    elif enum_func.__class__.__name__ == "HyperbolicAnisotropicEnumerateFunction":
+    elif isinstance(enum_func, ot.HyperbolicAnisotropicEnumerateFunction):
         graph.setTitle(f"q={enum_func.getQ()}")
-    elif enum_func.__class__.__name__ == "NormInfEnumerateFunction":
+    elif isinstance(enum_func, ot.NormInfEnumerateFunction):
         graph.setTitle("Infinity Norm enumeration rule")
     offset = 0
     for strata_index in range(maximum_strata_index):
@@ -78,10 +92,12 @@ def draw_stratas(enum_func, maximum_strata_index):
         cloud = ot.Cloud(sample_in_layer)
         cloud.setLegend(str(strata_index))
         cloud.setPointStyle("circle")
+        cloud.setColor(viridis_palette[strata_index])
         graph.add(cloud)
     graph.setIntegerXTick(True)
     graph.setIntegerYTick(True)
-    graph.setLegendPosition("upper right")
+    graph.setLegendPosition("upper left")
+    graph.setLegendCorner((1.0, 1.0))
     return graph
 
 
@@ -90,7 +106,7 @@ def draw_stratas(enum_func, maximum_strata_index):
 dim = 2
 enum_func = ot.LinearEnumerateFunction(dim)
 graph = draw_stratas(enum_func, 7)
-view = otv.View(graph, axes_kw={"aspect": "equal"})
+_ = otv.View(graph, axes_kw={"aspect": "equal"}, figure_kw={"figsize": (5, 4)})
 
 # %%
 # When the number of input dimensions of a functional chaos expansion (FCE) increases,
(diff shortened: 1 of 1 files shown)


python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -119,6 +128,27 @@ def draw_stratas(enum_func, maximum_strata_index):
...
+
+
+def build_q_norm_function(weights, q):
+    """Returns a parametric function representing the $q$-norm of a
+    two-dimensional vector weighted by the provided coefficients."""
+    if any(w <= 0 for w in weights):
+        raise ValueError("All weights must be strictly positive.")
+    if len(weights) != 2:
+        raise ValueError(f"The number of weights should be 2, but is {len(weights)}")
+    if q < 0.0 or q > 1.0:
```
regislebrun: q can be greater than 1, and BTW it is the only way to get an actual norm.

regislebrun (commented):
A single change and ok for me

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -119,6 +128,27 @@ def draw_stratas(enum_func, maximum_strata_index):
...
+        raise ValueError(f"The parameter q should be between 0 and 1, but is {q}")
```
mbaudin47:
```suggestion
    if q < 0.0:
        raise ValueError(f"The parameter q should be greater or equal than 0, but is {q}")
```

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -119,6 +128,27 @@ def draw_stratas(enum_func, maximum_strata_index):
...
+def build_q_norm_function(weights, q):
+    """Returns a parametric function representing the $q$-norm of a
+    two-dimensional vector weighted by the provided coefficients."""
+    if any(w <= 0 for w in weights):
+        raise ValueError("All weights must be strictly positive.")
+    if len(weights) != 2:
+        raise ValueError(f"The number of weights should be 2, but is {len(weights)}")
+    if q < 0.0:
+        raise ValueError(f"The parameter q should be greater or equal than 0, but is {q}")
+    q_norm_function = ot.SymbolicFunction(
```
regislebrun, line 150: Does it works when q==0?

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -119,6 +128,27 @@ def draw_stratas(enum_func, maximum_strata_index):
...
+
+def build_q_norm_function(weights, q):
+    """Returns a parametric function representing the $q$-norm of a
+    two-dimensional vector weighted by the provided coefficients."""
+    if any(w <= 0 for w in weights):
+        raise ValueError("All weights must be strictly positive.")
+    if len(weights) != 2:
+        raise ValueError(f"The number of weights should be 2, but is {len(weights)}")
+    if q < 0.0:
+        raise ValueError(f"The parameter q should be greater or equal than 0, but is {q}")
```
mbaudin47:
```suggestion
    if q <= 0.0:
        raise ValueError(f"The parameter q should be greater than 0, but is {q}")
```

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
mbaudin47: You are right, it does not work:
```
>>> q_norm_function = build_q_norm_function([1.0, 1.0], 0.0)
>>> q_norm_function([1, 1])
  File "<stdin>", line 1, in <module>
  File "/home/C61372/GITHUB/openturns-quantile-doc/build/install/lib/python3.12/site-packages/openturns/func.py", line 8265, in __call__
    return _func.Function___call__(self, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: NotDefinedException : Cannot evaluate ((w1 * x1)^q + (w2 * x2)^q)^(1 / q) at [x1,x2,w1,w2,q]=[1,1,1,1,0]
```

mbaudin47: Notice that ensuring that q > 0 is necessary, but not sufficient. This is because the q-norm can fail to evaluate when q is close to 0. For example:

```
>>> q_norm_function = build_q_norm_function([1.0, 1.0], 0.0001)
>>> q_norm_function([1, 1])
```

generates an exception. More precisely, the q-norm of the multi-index $(1,..., 1)$ in dimension $d$ with equal and unit weights will generate an overflow if:

$$
q \leq \frac{\log(d)}{\log(f_\max)}
$$

where $f_\max$ is the largest floating point number (i.e. $f_\max \approx 10^{308}$ with 64 bits fp numbers).

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
mbaudin47: This is fixed.

mbaudin47:
@adutfoy : I suggest to review this PR.

mbaudin47:
@jschueller : I suggest to review this PR.

jschueller:
@coderabbitai review

mbaudin47:
@regislebrun : Since your comments have been taken into account, I suggest to update your review.

mbaudin47:
@jschueller, @regislebrun  : I suggest to review this PR.

mbaudin47:
@sofianehaddad : I suggest to review this PR

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -104,72 +120,93 @@ def draw_stratas(enum_func, maximum_strata_index):
...
+        ["x1", "x2", "w1", "w2", "q"], ["((w1 * x1)^q + (w2 * x2)^q)^(1 / q)"]
+    )
+    q_norm_parametric = ot.ParametricFunction(
+        q_norm_function, [2, 3, 4], [weights[0], weights[1], q]
+    )
+    return q_norm_parametric
+
+
+# %%
+# The hyperbolic enumeration function is based on the q-quasi-norm.
```
regislebrun, line 160: You can be more precise: it is a quasi-norm for $q\in(0,1)$ and a norm for $q\geq 1$

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -104,72 +120,93 @@ def draw_stratas(enum_func, maximum_strata_index):
...
-grid.setGraph(
-    0, 1, draw_stratas(ot.HyperbolicAnisotropicEnumerateFunction(dim, 0.75), 7)
-)
-grid.setGraph(
-    1, 0, draw_stratas(ot.HyperbolicAnisotropicEnumerateFunction(dim, 0.5), 7)
-)
-grid.setGraph(
-    1, 1, draw_stratas(ot.HyperbolicAnisotropicEnumerateFunction(dim, 0.25), 7)
-)
+grid.setGraph(0, 0, drawHyperbolicStratas(dim, 1.0))
```
regislebrun, line 205: Instead of 4 value all $\leq 1$, you can replace e.g $q=0.25$ by $q=2$

regislebrun (changes requested):
Good job. I am not very fond of dense figures, but it is a matter of taste. The only point is to take into account the fact that one may use $q>1$ by adding this case to the illustrations. After all, the inf norm is the limit of the q norm when $q\rightarrow+\infty$.

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -182,79 +219,63 @@ def qnorm(x):
...
     enum_func = ot.HyperbolicAnisotropicEnumerateFunction(dim, q)
-    number_of_coeff_array = ot.Sample(degree_maximum, 2)
-    for p in range(1, 1 + degree_maximum):
-        number_of_coeff_array[p - 1, 0] = p
-        number_of_coeff_array[p - 1, 1] = enum_func.getMaximumDegreeCardinal(p)
+    data = [
+        [p, enum_func.getMaximumDegreeCardinal(p)] for p in range(1, 1 + degree_maximum)
+    ]
+    number_of_coeff_array = ot.Sample(data)
     cloud = ot.Cloud(number_of_coeff_array)
```
jschueller: I think you could pass data directly, no need to cast

jschueller (approved):

adutfoy:
The improvements are ok for me! Thanks for them!

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
mbaudin47: Fixed at 45984fb44cbd84e7fa1d52bfc203190d3e638b87

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
mbaudin47: Fixed at 45984fb44cbd84e7fa1d52bfc203190d3e638b87

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
mbaudin47: This is interesting indeed.

mbaudin47: Fixed at c68958d6606bab0280bbb5fb3ee2333e5c2032cb

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_enumeratefunction.py
```diff
@@ -104,72 +120,95 @@ def draw_stratas(enum_func, maximum_strata_index):
...
-grid.setGraph(
-    1, 0, draw_stratas(ot.HyperbolicAnisotropicEnumerateFunction(dim, 0.5), 7)
-)
-grid.setGraph(
-    1, 1, draw_stratas(ot.HyperbolicAnisotropicEnumerateFunction(dim, 0.25), 7)
-)
+grid.setGraph(0, 0, drawHyperbolicStratas(dim, 1.0))
+grid.setGraph(0, 1, drawHyperbolicStratas(dim, 0.75))
+grid.setGraph(1, 0, drawHyperbolicStratas(dim, 0.5))
+grid.setGraph(1, 1, drawHyperbolicStratas(dim, 2.0))
```
jschueller, line 208: cant we use the same set of q values than for the q-norm contours ? they have no q>1 case

mbaudin47: You are absolutely right: this was inconsistent. This is fixed at 1c02adbeb43e8d08fa2a6cb163b2a11456d7fc45

mbaudin47:
Here are the updated plots.

image

image

@jschueller : I suggest to merge this PR.

PR #3200, opened 2026-05-17
Add more documentation for domination method

dannys4:
I found that the `domination` method for the PCE was missing at least one reference, and some important details in the case of unbounded domains. I thought I'd just go ahead and make a PR since it would take less time than to go back-and-forth on an issue. Feel free to warp these contributions in any which way.

### DIFF
--- a/python/doc/examples/surrogate_modeling/polynomial_chaos/plot_chaos_ishigami_dependent_input.py
+++ b/python/doc/examples/surrogate_modeling/polynomial_chaos/plot_chaos_ishigami_dependent_input.py
@@ -150,7 +150,10 @@ view = otv.View(graph)
 # -------------------------------
 # Now, we want to use the domination method, which means that the basis created by the
 # adaptive strategy is used to project the model. This basis is not orthonormal to
-# :math:`\mu_{\inputRV}`.
+# :math:`\mu_{\inputRV}`. For more information on domination, see, e.g., [jakeman2019]_.
+# Note that domination methods require the basis' measure to be a constant multiple of a density
+# that is greater than the input distributions density---for unbounded domains, this entails
+# choosing a basis orthogonal w.r.t. a density that has appropriate tail decay.
 #
 # We use the :meth:`~openturns.FunctionalChaosAlgorithm.setUseDomination` method.
 # We implement the same steps as before, until the validation graph.
--- a/python/doc/bibliography.rst
+++ b/python/doc/bibliography.rst
@@ -230,6 +230,9 @@ Bibliography
 .. [jackson1991] Jackson, J. E. (1991).
     *A user’s guide to principal components.*
     John Wiley & Sons.
+.. [jakeman2019] Jakeman J., Franzelin F., Narayan A., Eldred M., Pflüger D. (2019)
+    *Polynomial chaos expansions for dependent random variables*,
+    Computer Methods in Applied Mechanics and Engineering, 351, pp.643-666.
 .. [janon2014] Janon A., Klein T., Lagnoux-Renaudie A., Prieur C., *Asymptotic
     normality and efficiency of two Sobol index estimators*,
     ESAIM: Probability and Statistics, EDP Sciences, 2014, 18, pp.342-364.
--- a/python/doc/theory/meta_modeling/functional_chaos.rst
+++ b/python/doc/theory/meta_modeling/functional_chaos.rst
@@ -240,14 +240,15 @@ We can also choose **a basis which is not orthonormal with respect to**
 which is orthonormal with respect to an *instrumental measure* :math:`\tilde{p} \neq \mu_{\inputRV}`.
 This instrumental measure is such that:
 
-- the support of :math:`\tilde{p}` is larger than the support of :math:`\mu_{\inputRV}`,
+- if :math:`\mu_{\inputRV}` has bounded support, the support of :math:`\tilde{p}` is larger than the support of :math:`\mu_{\inputRV}`,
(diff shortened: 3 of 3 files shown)


dannys4:
ref: https://github.com/openjournals/joss-reviews/issues/10484.

regislebrun (commented):
Hi Dannys4!

Many thanks for this contribution.

What is called "Domination" in Jaleman's paper is indeed not what is implemented in OpenTURNS, despite the name!

In Jakeman's paper, the dependent joint distribution of the input is simply replaced by another distribution with independent marginal distributions (and with additional technical details such as the tail behaviour and the boundedness of the likelihood ratio)). Then, the input sample is generated according to this new distribution, and the function to approximate is evaluated on this sample. It is simply a new approximation problem, posed on the same function but using another inner product. This way:

  * the initial distribution totally disappear from the problem
  * the Gram matrix has asymptotically (ie when the input sample size goes to infinity) orthonormal columns so the least squares problem is well-conditioned
  * if the ratio of the two density functions is bounded then we get an approximation which is close to the optimal (in the initial distribution norm) up to this bound

In OpenTURNS, we adopt another point of view. We keep the input sample generated according to the dependent input distribution (in many applications, the input sample is imposed as well as the corresponding output sample), and we project the function we want to approximate on a space given by a functional basis which is no more orthonormal wrt this dependent distribution (but which is orthonormal wrt an instrumental distribution). We do this by using the sample generated by the dependent input distribution. It is what Nora Lüthen called the domination method in her paper https://arxiv.org/abs/2002.01290, hence the name of the method in OpenTURNS. This way:

  * the projection is done using the inner product associated to the dependent distribution
  * the Gram matrix may be ill conditioned as the columns are no more asymptotically orthonormal, but this is mitigated by the use of a basis orthonormal wrt another well-chosen distribution (we avoid eg the troubles linked to the use of the canonical basis, which is not orthonormal regardless of the input distribution)
  * the drawback is that we must solve a least-squares problem: the solution is no more given by suitably chosen inner products as the basis is no more orthonormal wrt the input distribution. It is not an actual issue in a Monte Carlo sampling setting, as the least-squares estimator is the most efficient estimator (in a statistical sense) of the coefficients.

We found this during a comparative analysis of both papers. We wanted to answer Nora's questions regarding the better performance she get using what she called domination (which is what we implemented) instead of an orthonormal basis built wrt to inner product defined by the dependent input distribution, in subspace selection setting. It is not obvious to see that her domination method differ from Jakeman's one, but once you read the details the difference becomes clear. A detailed analysis of these two papers is the object of an upcoming paper from our side!

dannys4:
Sorry, could you be more specific with the Lüthen article? I looked in that one and found nothing called the "domination method"---perhaps you meant a different PCE-based article from her? I found [this part of the UQLab documenation](https://ethz.ch/content/dam/ethz/special-interest/baug/ibk/risk-safety-and-uncertainty-dam/publications/reports/RSUQ-2023-001.pdf), which references [Torre 2019](https://www.sciencedirect.com/science/article/pii/S0021999119302220), presumably in reference to what they call "`aPCEonX`" described in section 2.4. Would this be a more appropriate reference, then?

Additionally, I think that the consideration of the tails is necessary regardless of the approach---I believe that using a Hermite basis on a measure that decays, e.g., as $\exp(-t)$, will still produce problematic behavior in your scenario.

adutfoy:
Hi Dannys4!

You're right—we jumped a bit too quickly into the terminology of the method Nora used (projecting orthogonally onto the input distribution using a basis orthogonal to another distribution), and calling it “Domination” is confusing. We'll change the name of our method in the library.

Regarding the second point you mentioned, we are well aware of the compatibility issues between the initial measure and the instrumental measure—that is, the measure for which the basis of functions is orthonormal. In the code, we reformulate the problem so that the support of the two distributions coincides, that is, so that the quantiles of level epsilon and (1-epsilon) of the marginal distributions of the two measures are equal, with epsilon = 1e-12:

https://github.com/openturns/openturns/blob/6e22849d788e09c47cd092c6ce0e871923659130/lib/src/Uncertainty/Algorithm/MetaModel/FunctionalChaos/FunctionalChaosAlgorithm.cxx#L251

This condition is not sufficient from a theoretical standpoint (the decay rates at the tails of the two distributions can be very different), but in practice, all distributions are truncated to their numerical support.
It is therefore appropriate to specify these conditions in the documentation, explain what the library does, and highlight its limitations.

regislebrun (approved):
Thanks for the contribution, the tail behaviour is of uttermost importance.

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_chaos_ishigami_dependent_input.py
```diff
@@ -150,7 +150,10 @@
 # -------------------------------
 # Now, we want to use the domination method, which means that the basis created by the
 # adaptive strategy is used to project the model. This basis is not orthonormal to
-# :math:`\mu_{\inputRV}`.
+# :math:`\mu_{\inputRV}`. For more information on domination, see, e.g., [jakeman2019]_.
+# Note that domination methods require the basis' measure to be a constant multiple of a density
+# that is greater than the input distributions density---for unbounded domains, this entails
```
jschueller, line 155: --- is ok in latex as it expands to a long dash, but for rst a colon or comma would be better

jschueller (approved):

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_chaos_ishigami_dependent_input.py
adutfoy: No problem for this long dahs: we will correct it afetr the merge.

Issue #3182, opened 2026-05-02
QuantileConfidence.computeBilateralRank() can fail


mbaudin47:
### What happened?

Given the sample size $n$, the quantile level $\alpha \in (0,1)$ and the confidence level $\beta \in [1/2,1)$, the `QuantileConfidence.computeBilateralRank()` method returns two integers $k_1$ and $k_2$ such that:

$$
\mathbb{P} \left( X_{(k_1)} \leq x_\alpha \leq X_{(k_2)} \right) \geq \beta.
$$

Let $\alpha = 0.95$ and $\beta = 0.90$. If $n=974$, the algorithm works fine and the returned indices are $k_1=915$ and $k_2=938$. The probability of coverage is then $\mathbb{P} \left( X_{(k_1)} \leq x_\alpha \leq X_{(k_2)} \right)=0.9000410346812894$ which is ok.

But if $n=975$, then the returned indices are $k_1=975$ and $k_2=975$. The correct indices are $k_1=910$, $k_2=935$ which lead to a coverage of $p=0.9034630959860642$.

### How to reproduce the issue?

```Python
# %%
import openturns as ot
import openturns.experimental as otexp
from packaging import version

# %%
print(ot.__version__)

# %%
print("computeBilateralRank")
alpha = 0.95
beta = 0.90
algo = otexp.QuantileConfidence(alpha, beta)
size = 975
k1, k2 = algo.computeBilateralRank(size)
print(f"alpha={alpha}, beta={beta}, size={size}")
print(f"k1={k1}, k2={k2}")
```

### Version

1.27

### Operating System

unknown

### Installation media

unknown

mbaudin47:
To explore the bug, I used a modified C++ version of the code. This amounts to inserting `LOGDEBUG` statements in the code.
```cpp
Indices QuantileConfidence::computeBilateralRank(const UnsignedInteger size) const
{
  const UnsignedInteger minimumSize = computeBilateralMinimumSampleSize();
  if (size < minimumSize)
    throw InvalidArgumentException(HERE) << "Cannot compute bilateral rank as size (" << size << ") is lower than minimum size (" << minimumSize << ")";

  // find the indices of interval ]k1, k2] with smallest probability >=beta according to a Binomial(n=size, alpha)
  const Binomial binomial(size, alpha_);
  Scalar pBest = SpecFunc::MaxScalar;
  UnsignedInteger k1Best = size;
  UnsignedInteger k2Best = size;
[...]

PR #3211, opened 2026-06-01
Fix QuantileConfidence::computeBilateralRank

jschueller:
#3182

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bilateral confidence interval search and probability evaluation for greater numerical robustness and correctness in edge cases, reducing spurious failures.

* **Tests**
  * Expanded regression and randomized tests for bilateral ranks, adding fixed edge cases and stress checks to validate interval ordering and probability constraints.

### DIFF
--- a/lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
+++ b/lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
@@ -100,67 +100,44 @@ UnsignedInteger QuantileConfidence::computeUnilateralRank(const UnsignedInteger 
   return rank;
 }
 
-// compute argmin (k1, k2) P_X([k1, k2]) under constraint P_X([k1, k2])>=beta, with X~B(n, alpha)
+// compute argmin (k1, k2) P_Xd(]k1, k2]) under constraint P_Xd(]k1, k2])>=beta, with Xd~B(n, alpha)
 Indices QuantileConfidence::computeBilateralRank(const UnsignedInteger size) const
 {
   const UnsignedInteger minimumSize = computeBilateralMinimumSampleSize();
   if (size < minimumSize)
     throw InvalidArgumentException(HERE) << "Cannot compute bilateral rank as size (" << size << ") is lower than minimum size (" << minimumSize << ")";
 
-  // find the indices of interval ]k1, k2] with smallest probability >=beta according to a Binomial(n=size, alpha)
   const Binomial binomial(size, alpha_);
-  Scalar pBest = SpecFunc::MaxScalar;
-  UnsignedInteger k1Best = size;
-  UnsignedInteger k2Best = size;
-  Scalar p1Prev = 0.0;
+  Scalar pBest = 2.0;
+  Indices kBest;
   UnsignedInteger k1 = 0;
   while (k1 < size)
   {
-    // find upper bound of cdf increment location
-    Scalar p1 = binomial.computeCDF(k1);
-    UnsignedInteger inc = 1;
-    UnsignedInteger k1Prev = k1;
-    while (p1 == p1Prev)
-    {
-      k1Prev = k1;
-      k1 += inc;
-      p1 = binomial.computeCDF(k1);
-      inc *= 2;
-    }
-
-    // find exact cdf increment location
-    UnsignedInteger k1Max = k1;
-    UnsignedInteger k1Min = k1Prev;
-    while (k1Max - k1Min > 1)
-    {
-      const UnsignedInteger k1Mid = (k1Min + k1Max) / 2;
-      const Scalar p1Mid = binomial.computeCDF(k1Mid);
-      if (p1Mid > p1Prev)
-        k1Max = k1Mid;
-      else
-        k1Min = k1Mid;
-    }
-    k1 = k1Max;
-
-    // we have to stop when the probability will cross the bound
-    if (p1 + beta_ > 1.0)
+    // update CDF at k1, stop when it crosses the bound
+    const Scalar p1 = binomial.computeCDF(k1);
+    if (p1 + beta_ >= 1.0)
       break;
 
-    // we know P([k1, k2])>=beta which gives k2 directly from k1(p1)
+    // we know P(]k1, k2])>=beta which gives k2 directly from k1(p1)
     const UnsignedInteger k2 = binomial.computeScalarQuantile(p1 + beta_);
-    p1Prev = p1;
+    const Scalar p2 = binomial.computeCDF(k2);
 
-    // we compute P(k1<X<=k2)=P(k1+1<=X<=k2), CDF(k2)-CDF(k1) also works
-    const Scalar p = binomial.computeProbability(Interval(k1 + 1, k2));
+    // we compute P(k1<Xd<=k2)=CDF(k2)-CDF(k1) and check for improvement
+    const Scalar p = p2 - p1;
     LOGDEBUG(OSS() << "k1=" << k1 << " k2=" << k2 << " p=" << p);
     if ((p >= beta_) && (p < pBest))
     {
       pBest = p;
-      k1Best = k1;
-      k2Best = k2;
+      kBest = {k1, k2};
     }
-  } // while k1
-  return {k1Best, k2Best};
+
+    // k1 advances right before the next CDF bump at qB(p2-beta), ensure it stays stricly increasing
+    k1 = std::max(binomial.computeScalarQuantile(p2 - beta_) - 1.0, k1 + 1.0);
+  }
+
+  if (!kBest.getSize())
+    throw InternalException(HERE) << "Cannot compute bilateral rank: no valid interval found for size=" << size;
+  return kBest;
 }
 
 // compute interval of the form [X_k; +inf[ or ]-inf; X_k] from unilateral rank k
--- a/python/test/t_QuantileConfidence_std.py
+++ b/python/test/t_QuantileConfidence_std.py
@@ -94,8 +94,8 @@ k_ref = {
 }
 for n in k_ref.keys():
     k1, k2 = algo.computeBilateralRank(n)
-    binomial = ot.Binomial(n, alpha)
-    p = binomial.computeProbability(ot.Interval(k1, k2))
+    binom = ot.Binomial(n, alpha)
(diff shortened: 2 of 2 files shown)


mbaudin47:
Here is a review of the code at 3890b36dd2d572afb83831698f3d4972b09223bb. I focus on the defects from my point of view.

### Inability to Include Zero in the Lower Bound

#### 1. Code Segment

```cpp
UnsignedInteger k1 = 0;
Scalar p1 = binomial.computeCDF(k1);
// ...
while (k1 < size)
{
  // find upper bound of cdf increment location
// ...
```

#### 2. The Problem

The value of $k_1$ is initialized in the statement `UnsignedInteger k1 = 0;`. When the `while` loop begins, it would be logical to test whether this value of $k_1$ is suitable, meaning whether it can lead to a $k_2$ value associated with the required coverage level. However, the algorithm does not test this value of $k_1$: instead, a probability increment is searched for starting from this value of $k_1$. This is an algorithm design issue, because if $k_1 = 0$ is suitable, then the search is unnecessary or harmful. Worse, in the event of a difficulty with the algorithm, the value $k_1 = 0$ can be skipped to move to the next value of $k$ leading to a CDF jump. This may be detrimental if $\alpha$ is close to zero, because most of the mass of the binomial is located near small values of $k$ in this case.

#### 3. The Solution

The solution consists of swapping the sections of the algorithm. First, the `while` loop must start by testing the current value of $k_1$: this corresponds to the code `if (p1 + beta_ > 1.0)`. Then, after testing the value of $k_1$, the next CDF jump can be searched for using the algorithm `// find upper bound of cdf increment location`.

### Inadequate Failure Handling

#### 1. Code Segment

```cpp
UnsignedInteger k1Best = size;
UnsignedInteger k2Best = size;
```

#### 2. The Problem

At the beginning of the algorithm, the values of $k_1$ and $k_2$ are initialized. However, no verification of the search's success is performed. Consequently, if the algorithm fails to find a value for $(k_1, k_2)$, the initial values are returned. The variables `k1Best` and `k2Best` are initialized to `size`. If no pair numerically satisfies the condition $p \ge \beta$ (which can occur in conjunction with premature stopping), the loop ends without ever updating these variables. The function silently returns the array `{size, size}` without throwing an exception. This creates a degenerate and invalid interval for downstream computations.

#### 3. The Solution

The solution consists of using a boolean to detect the success of the algorithm. If the algorithm fails to find a pair $(k_1, k_2)$ satisfying the constraint, an exception must be thrown.

### Numerical Stability of the Exponential Search

#### 1. Code Segment

```cpp
// find upper bound of cdf increment location
UnsignedInteger inc = 1;
UnsignedInteger k1Prev = k1;
while (p1 < p1Prev + epsilon)
{
  k1Prev = k1;
  k1 = std::min(k1 + inc, size);
  p1 = binomial.computeCDF(k1);
  inc *= 2;
}
```

#### 2. The Problem

The exponential leap `inc *= 2` creates an integer overflow vulnerability prior to the execution of the boundary clamp. If the multiplier `inc` grows sufficiently large during a prolonged search, the evaluation of `k1 + inc` within the `std::min(k1 + inc, size)` statement will exceed the maximum capacity of `UnsignedInteger`. This arithmetic overflow forces the value to wrap around before the `std::min` function can clamp it to `size`, destroying the index progression and causing unpredictable algorithmic behavior.

#### 3. The Solution

Implement an arithmetic overflow guard before performing the addition. The assignment must structurally verify that the available capacity of the integer type can safely absorb the exponential increment before attempting the bounds check.

```cpp
k1 = (inc < std::numeric_limits<UnsignedInteger>::max() - k1) ? std::min(k1 + inc, size) : size;
```

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -112,20 +112,27 @@ Indices QuantileConfidence::computeBilateralRank(const UnsignedInteger size) con
...
   UnsignedInteger k1Best = size;
   UnsignedInteger k2Best = size;
-  Scalar p1Prev = 0.0;
+  UnsignedInteger k2Prev = 0;
+  Scalar p1Prev = -1.0;
   UnsignedInteger k1 = 0;
+  Scalar p1 = binomial.computeCDF(k1);
+  Scalar p2Prev = 0.0;
+  // we pass increments of cdf in the scale of beta to quantile function
+  const Scalar epsilon = SpecFunc::ScalarEpsilon * beta_;
```
mbaudin47: Computing epsilon based on the value of beta is not relevant here because beta is close to 1 in most cases.

```suggestion
  const Scalar epsilon = SpecFunc::ScalarEpsilon;
```

mbaudin47:
This version seems excellent from my point of view. I will make several tests today et let you know about the results.

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -100,67 +100,44 @@ UnsignedInteger QuantileConfidence::computeUnilateralRank(const UnsignedInteger
...
-  UnsignedInteger k2Best = size;
-  Scalar p1Prev = 0.0;
+  Scalar pBest = 2.0;
+  Indices kBest;
   UnsignedInteger k1 = 0;
   while (k1 < size)
   {
-    // find upper bound of cdf increment location
+    // update CDF at k1, stop when it crosses the bound
     Scalar p1 = binomial.computeCDF(k1);
```
mbaudin47:
```suggestion
    const Scalar p1 = binomial.computeCDF(k1);
```

mbaudin47 (changes requested):

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -100,67 +100,44 @@ UnsignedInteger QuantileConfidence::computeUnilateralRank(const UnsignedInteger
...
       pBest = p;
-      k1Best = k1;
-      k2Best = k2;
+      kBest = {k1, k2};
     }
-  } // while k1
-  return {k1Best, k2Best};
+
+    // k1 advances right before the next CDF bump at qB(p2-beta), ensure it stays stricly increasing
+    k1 = std::max(binomial.computeScalarQuantile(p2 - beta_) - 1.0, k1 + 1.0);
```
mbaudin47, line 135: This line is ok because the output of computeScalarQuantile is subtracted the floating point number 1.0. Hence, even if the output of computeScalarQuantile is the UnsignedInteger 0, it is implicitly converted into a floating point number. Therefore, there is no risk of wrap-around here. It seems to me that a small comment may clarify this point in order to make future maintenance easier. Otherwise, this subtle point may be overlooked.

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -100,67 +100,44 @@ UnsignedInteger QuantileConfidence::computeUnilateralRank(const UnsignedInteger
...
+  }
+
+  if (!kBest.getSize())
```
mbaudin47, line 138: This implicitly converts an integer into a boolean. I suggest this:
```suggestion
  if (kBest.getSize()==0)
```

jschueller: this is very common, imho its ok

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
```diff
@@ -100,67 +100,44 @@ UnsignedInteger QuantileConfidence::computeUnilateralRank(const UnsignedInteger
...
   if (size < minimumSize)
     throw InvalidArgumentException(HERE) << "Cannot compute bilateral rank as size (" << size << ") is lower than minimum size (" << minimumSize << ")";
 
-  // find the indices of interval ]k1, k2] with smallest probability >=beta according to a Binomial(n=size, alpha)
   const Binomial binomial(size, alpha_);
-  Scalar pBest = SpecFunc::MaxScalar;
-  UnsignedInteger k1Best = size;
-  UnsignedInteger k2Best = size;
-  Scalar p1Prev = 0.0;
+  Scalar pBest = 2.0;
```
mbaudin47, line 111: This initialization is confusing.

```suggestion
  Scalar pBest = SpecFunc::MaxScalar;
```

mbaudin47:
The case where the confidence level is equal to the difference of CDF of the optimal ranks is very special: it might be useful to include it into the unit tests. In one of my algorithms, this generates an infinite loop. I suggest to add it into the unit tests.
```python
"""
Compute the confidence level such that it is exactly
equal to the difference of known ranks.
"""
import openturns as ot

n = int(1.0e6)
k1_ref = 49625
k2_ref = 50533
alpha = 0.05

binomial = ot.Binomial(n, alpha)
beta = binomial.computeCDF(k2_ref) - binomial.computeCDF(k1_ref)

algo = ot.QuantileConfidence(alpha, beta)
k1, k2 = algo.computeBilateralRank(n)
assert k1 == k1_ref
assert k2 == k2_ref
p = algo.computeBilateralCoverage(n, k1, k2)
assert p >= beta
````

lib/src/Uncertainty/Algorithm/Simulation/QuantileConfidence.cxx
jschueller: any value above 2 is ok

mbaudin47: What is magic about 2?

jschueller:
ok, included

mbaudin47:
@adutfoy I suggest to review this PR

mbaudin47 (approved):

jschueller:
@regislebrun @sofianehaddad  we would need one more reviewer; this blocks #3176

mbaudin47:
@regislebrun @sofianehaddad @adutfoy Can you please review this PR?

mbaudin47:
@regislebrun @sofianehaddad @adutfoy Can you please review this PR?

adutfoy:
This PR is ok for me!

Issue #3085, opened 2026-02-01
The taylor_moments theory help page has issues.


mbaudin47:
### Describe the issue

The [taylor_moments](https://openturns.github.io/openturns/latest/theory/reliability_sensitivity/taylor_moments.html) theory help page has issues.

## Issues

<img width="802" height="65" alt="Image" src="https://github.com/user-attachments/assets/4d5fa8dd-7c3d-46a3-a46f-3389cd822b8f" />

- There is no function $h$ involved here.

<img width="1172" height="191" alt="Image" src="https://github.com/user-attachments/assets/cf0ac41f-cb16-40dc-a531-9406631bfe3d" />

- There are unnecessary round brackets before and after the partial derivatives of the function g. The correct expression for the value of the partial derivative of $g$ at point $\mu$ is:

$$
\frac{\partial g}{\partial x_i}(\boldsymbol{\mu}).
$$

  There is no need for extra round brackets. There are 20 extra round brackets in the page.

- There is an extra point at the end of the equation. This is an error since the sentence is continuing.

- Although this is not a strict error, the double sum may not be the best possible expression:

$$
\sum_{i,j = 1}^d  \textrm{ (Questionable.)}
$$

 The correct way to express the sum is:

$$
\sum_{i = 1}^d \sum_{j = 1}^d  \textrm{ (Perfect.)}
$$

  There are 6 such double sums in the page.

<img width="904" height="236" alt="Image" src="https://github.com/user-attachments/assets/44056c89-2a89-4cb5-987f-ecb940b70081" />

- In the case 2, dimension $d_Y$, the expression is wrong because the indices $k$ and $\ell$ should be expressed in the left hand side, and not on the right hand side. The correct expression is:

$$
\text{Cov}(Y_k, Y_\ell) = \sum_{i = 1}^d \sum_{j = 1}^d c_{ij} \frac{\partial g_k}{\partial x_i}(\boldsymbol{\mu}) \frac{\partial g_\ell}{\partial x_j}(\boldsymbol{\mu})
$$

  for $1 \leq k, \ell \leq d_Y$.

## Use matrix expressions

In order to obtain a more compact, yet accurate, representation of the Taylor expansion, I would suggest to use vectors and matrices:

$$
[...]

Issue #2408, opened 2023-09-30
The LeastSquaresExpansion and IntegrationExpansion doc help pages are incomplete


mbaudin47:
### What happened?

The doc of [LeastSquaresExpansion](http://openturns.github.io/openturns/latest/user_manual/response_surface/_generated/openturns.experimental.LeastSquaresExpansion.html) is unspecified. Similarly, the [IntegrationExpansion](http://openturns.github.io/openturns/latest/user_manual/response_surface/_generated/openturns.experimental.IntegrationExpansion.html) help page is unspecified. More precisely, the link between the input arguments and the output is unspecified in the help page. Their doc should be either point to a full help page (such as `FunctionalChaosAlgorithm` for example) or be completed. Their link with respect to one another is undocumented. 

### How to reproduce the issue?

```shell
Not relevant
```

### Version

1.21

### Operating System

all

### Installation media

unknown

mbaudin47:
@regislebrun : You may want to share your thoughts on this topic.

Issue #1511, opened 2020-06-02
The LeastSquaresMethod classes have doc issues


mbaudin47:
The LeastSquaresMethod class has issues.

- [ ] The purpose of the `weights` in the constructor in unspecified.
- [ ] The purpose of the `indices` in the constructor in unspecified.

- [ ] The meaning of $\Psi$, $U$, $b$ and $y$ is unspecified in the mathematical equation of the `LeastSquaresMethod` API help page. The Euclidian norm index 2 should be written in the norm. The vector $y$ should be in bold face. The `weights` do not appear and neither do the `indices`.

![image](https://github.com/openturns/openturns/assets/31351465/16de647e-3e89-448c-9259-ec5fb0c9875f)

- [ ] The mathematical equations in many methods are confusing. For example, the `LeastSquaresMethod.getGramInverse()` method has stars instead of the invisible matrix-matrix product. Moreover, the function states that the output is $c$, while the mathematical equation sets $G^{-1}$.

![image](https://github.com/openturns/openturns/assets/31351465/2143402a-ebf5-4016-9be8-9ef77d4aec15)

- [ ] The `getHDiag()` defines $H$, but the function returns $d$.

![image](https://github.com/openturns/openturns/assets/31351465/2e9abdc7-bfb6-45b1-93fb-666560cec81c)

- [ ] The `getGramInverseDiag()` method should use the `\operatorname{diag}` command, instead of `diag` in the LaTeX source.

![image](https://github.com/openturns/openturns/assets/31351465/fe4b16c5-b3d2-41fe-82f3-cd0957178e37)

- [ ] The `solve()` method does not define what is $M$. Moreover, the vector $b$ is unrelated to the vector $y$ in the main doc of the class.

![image](https://github.com/openturns/openturns/assets/31351465/0200cca5-8440-4056-840d-ad89377de8c6)

- [ ] The `update()` method has no example. Its purpose is unspecified.

![image](https://github.com/openturns/openturns/assets/31351465/fa65304c-b777-4fbe-9a0f-a956cf5690c4)

- [ ] The `CholeskyMethod`, `SVDMethod` and `QRMethod` do not have any mathematical equations specifying the purpose of these classes.

[...]

PR #3212, opened 2026-06-02
Hackathon 2026: doc

### LINKED ISSUE #2385: The doc of LeastSquaresMethod is wrong
- [ ] The doc of `LeastSquaresMethod` is related to the PCE. In the current doc, the `LeastSquaresMethod` (and its children classes SVD, Cholesky and QR)  is related to the PCE. This can be seen for example on the general API help page of the PCE.

image

This is a bug, because the linear least squares method does not depend on the PCE. The PCE, however, can use the least squares method to estimate its coefficients. Hence, the `LeastSquaresMethod` and its children should be located in the `LinearModel` API section:

image

- [ ] The mathematical equation is:

image

The `weights` and `indices` are not specified in this equation. Moreover, the Euclidian norm is unspecified as the index of the norm. More generally, this help page has all the flaws already presented in #2408.

- [ ] The dimension of the `weights` is unspecified:

image

Its mathematical meaning is undefined.

- [ ] The dimension of the `indices` is unspecified:

image

Its mathematical meaning is undefined.

- [ ] The wording "A priori" is poorly chosen:

image

This is an expression relevant to Bayesian computing, which is not related to the least squares method. Moreover, the dimension of the design matrix is unspecified.

### LINKED ISSUE #3071: The theory help page of random_mixture has a bug
### Describe the issue

There is an inconsistent equation:

Image

### Where can the issue be found?

https://openturns.github.io/openturns/latest/theory/probabilistic_modeling/random_mixture.html#random-mixture

### Additional Context

_No response_

### LINKED ISSUE #3086: The TaylorExpansionMoments class has issues
### Describe the issue

The `TaylorExpansionMoments` class has issues:
- in the doc,
- in the code.

More precisely, I suggest to limit the features of the class so that the method `getImportanceFactors()` and `drawImportanceFactors()` raise an exception if the input random vector has a non-trivial dependency.

### In the theory documentation

The [taylor_importance_factors.html](https://openturns.github.io/openturns/latest/theory/reliability_sensitivity/taylor_importance_factors.html) help page states that, assuming that the marginals of $\boldsymbol{X}$ are independent, the importance factors are:

$$
\eta_i = \frac{\left(\frac{\partial g}{\partial x_i} (\boldsymbol{\mu})\right)^2 \sigma_i^2}{\text{Var}(Y)}  \qquad \text{(1)}
$$

for $1 \leq i \leq d$ where $\sigma_i^2 = \text{Var}(X_i)$ is the variance of the $i$-th input.
In the previous definition, we assume that $\text{Var}(Y) > 0$ otherwise we cannot partition the variance and the sensitivity analysis is impossible.

This is **correct**.

We always have:

$$
\eta_i \geq 0, \qquad 1 \leq i \leq d
$$

because the numerator is a square and the denominator is nonnegative, by hypothesis.

The previous inequality holds regardless of the accuracy of the Taylor expansion in terms of function approximation of the model $g$.

Moreover, if the order 1 Taylor expansion is exact (and if the marginals of $\boldsymbol{X}$ are independent), then:

$$
\text{Var}(Y) = \sum_{i=1}^d \left(\frac{\partial g}{\partial x_i} (\boldsymbol{\mu})\right)^2 \sigma_i^2.
$$

By definition of the importance factor $\eta_i$, this implies:

$$
\text{Var}(Y) = \sum_{i=1}^d \text{Var}(Y) \eta_i.
$$

Since $\text{Var}(Y) > 0$ by hypothesis, we can simplify the previous equation. Hence the sum of the importance factors is equal to 1:

$$
\sum_{i = 1}^d \eta_i = 1.
$$

This implies:

$$
0 \leq \eta_i \leq 1, \quad 1 \leq i \leq d.
$$

### In the source code

The source code at [TaylorExpansionMoments.cxx#L224](https://github.com/openturns/openturns/blob/e2ef35815466a59118ceb09fe0dacc33ccd83250/lib/src/Uncertainty/Algorithm/TaylorExpansionMoments.cxx#L224) implements the following formula:

$$
\eta_i = \frac{\frac{\partial g}{\partial x_i}(\boldsymbol{\mu})}{\text{Var}(Y)} \cdot \sum_{j=1}^{d} \frac{\partial g}{\partial x_j}(\boldsymbol{\mu}) \text{Cov}(X_i, X_j)  \qquad \text{(2)}
$$

for $1 \leq i \leq d$. 

By definition, if the order 1 Taylor expansion is exact, then the sum of these importance factors is equal to 1:

$$
\sum_{i = 1}^d \eta_i = 1.
$$

Compare the equations (1) and (2). If the input marginals are independent, then the equation (2) simplifies and we get the equation (1).

### The problem

The problem is: the equation (2) should not be used. Indeed, in the case where the inputs are dependent, then the sensitivity analysis of the Taylor expansion approximation is complicated. For example, the property that each importance factor is in the $[0,1]$ interval is lost: if $\text{Cov}(X_i, X_j) < 0$ for some $i$ and $j$, then the importance factor $\eta_i$ can be lower than zero.

Here is an example from Bertrand Iooss:

Let $Z = X_1 + 2 X_2$  where $X_1 \sim \mathcal{N}(0;1)$, $X_2 \sim \mathcal{N}(0;1)$ and $r(X1,X2) = -0.9$. Then $V=\text{Var}(Z) = 1.4$, $\eta_1 = \frac{-0.8}{V}$ and $\eta_2 = \frac{2.2}{V}$.

The purpose of the LMG indices is to solve this issue (see [^Clouvel2024]).

### Conclusion

My conclusion is that the code should raise an exception in the case where the inputs have a non-trivial dependency. Otherwise, it should implement the equation (1).

[^Clouvel2024]: Clouvel, L., Iooss, B., Chabridon, V., Idrissi, M. I., & Robin, F. (2024). An overview of variance-based importance measures in the linear regression context: comparative analyses and numerical tests.

### Where can the issue be found?

https://openturns.github.io/openturns/latest/user_manual/_generated/openturns.TaylorExpansionMoments.html

### Additional Context

The [taylor_importance_factors.html](https://openturns.github.io/openturns/latest/theory/reliability_sensitivity/taylor_importance_factors.html) help page has issues.

Image

The input of $g$ is $\boldsymbol{X}$, not $\boldsymbol{x}$. This is because the input random vector is random, not deterministic.

### LINKED ISSUE #3166: The doc of KernelMixture can be improved
### Describe the issue

The KernelMixture class has specific implementations of getMean and computeCovariance(). But the doc does not reflect the implementation.

The formula of the mean is presented at:

https://github.com/openturns/openturns/blob/fd6499b3ea97d4367882695592e60637643d274c/lib/src/Uncertainty/Distribution/KernelMixture.cxx#L640

$$
\text{PDF}(x) = C\sum_{i=1}^N \prod_{j=1}^n K\left( \frac{X^i_j-x_j}{h_j}\right),
$$

where:

$$
 C = \frac{1}{N \prod_{k=1}^n h_k} \mu_j = \int_{-\infty}^{\infty} x_j\frac{1}{Nh_j}\sum_{i=1}^N K((x_j - X^i_j) / h_j) dx_j
$$

$$
C = \int_{-\infty}^{\infty} \frac{1}{Nh_j} \sum_{i=1}^N (h_j t + X^i_j)K(t) h_j dt 
$$

$$
C = \frac{1}{Nh_j} \sum_{i=1}^N X^i_j + \int_{-\infty}^{\infty} \frac{1}{N} \sum_{i=1}^N h_jt K(t) dt
$$

$$
C = \mu_{\text{sample}_j} + h_j \mu_K
$$

The formula of the covariance is presented at:

https://github.com/openturns/openturns/blob/fd6499b3ea97d4367882695592e60637643d274c/lib/src/Uncertainty/Distribution/KernelMixture.cxx#L662

$$
\text{Cov}(k_M) = (1-1/N) \text{Cov}(\text{sample}) + \text{Cov}(k) \text{diag}(h_i^2)
$$

### Where can the issue be found?

dodo

### Additional Context

_No response_

### LINKED ISSUE #3206: The theory help page for linear regression has errors
### Describe the issue

Image

First, the notations are inconsistent, because we cannot take the variable $X$ to the power $i$. Let us assume that it means that $X^i$ is the $i$-th marginal of the random vector $X$.

This means that we predict an input $X_j$ from another input $X_i$. This is not what we mean by regression in general. The correct equation is:

$$
Y = a_0 + \sum_{j \in \{ j_1,\ldots,j_K \} } a_j X_j + \varepsilon
$$

where $Y$ is the variable to predict.

Image

This definition of the $R^2$ is unknown.

https://github.com/openturns/openturns/blob/6002aa6e22310a37ffb14326fe5776c5a78a30b6/python/doc/theory/data_analysis/regression_analysis.rst?plain=1#L118

Fix mean notation in $SS_{TOT}$. Line 119 uses $\bar{y}_j$, but $\bar{y}$ is a scalar sample mean in this section. The expression should be $\left(y_j-\bar{y}\right)^2$.

### Where can the issue be found?

https://openturns.github.io/openturns/latest/theory/data_analysis/linear_regression.html

### Additional Context

_No response_

adutfoy:
Hackathonn Printemps 2026:

- Taylor moments: Fixed #3086 and #3085
- Linear model doc improved: I gathered two theory files in one: "Linear regression" and "Regression analysis". I fixed #3206
- Improved doc of the KernelMixture by adding the formula of the mean and the covariance: Fixed #3166
- Fixed #3071
- Improved the API of least squares classes: Fixed #2385 and other modifications
- Improved doc of leastSquaresExpansion and IntegrationEXpansion : #2408

## Summary by CodeRabbit

* **Documentation**
  * Clarified Taylor importance factors and moments with consistent input/output notation; noted possible negative factors and linear-model simplification.
  * Reorganized regression material into a dedicated linear-regression chapter (estimation, tests, distributions, examples); removed the older regression page.
  * Expanded KernelMixture mean/covariance explanation and fixed several math/typo issues across theory pages.
  * Improved numerous method docstrings and examples for least-squares/linear-methods.


python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,133 @@
...
+
+    \bar{y} = \frac{1}{\sampleSize} \sum_{j = 1}^\sampleSize y_j.
+
+The total sum of squares (see [baron2014]_ page 398) is:
+
+.. math::
+
+    SS_{TOT}
+    = \sum_{j = 1}^\sampleSize \left(y_j - \bar{y}_j\right)^2
+    = \left(\vect{y} - \bar{\vect{y}}\right)^T \left(\vect{y} - \bar{\vect{y}}\right).
```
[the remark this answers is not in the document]
adutfoy: no ...

python/src/CholeskyMethod_doc.i
```diff
@@ -11,27 +11,34 @@ Available constructors:
...
+    The design proxy that builds the design matrix
 weight : sequence of float
-    Output weights
+    The output weights
 indices : sequence of int
-    Indices allowed in the basis
+    The indices allowed in the basis
 design : 2-d sequence of float
-    A priori known design matrix
+    The apriori known design matrix
```
mbaudin47: The english word is "prior".

adutfoy: I removed the term prior as resuqted in #2385

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
-where :math:`\varepsilon` describes a random variable with zero mean
-and standard deviation :math:`\sigma` independent of the input variables
-:math:`X^i`. For given values of :math:`X^{j_1}`,…,\ :math:`X^{j_K}`,
-the average forecast of :math:`X^i` is denoted by :math:`\widehat{X}^i`
-and is defined as:
+
+where :math:`a_1, ..., a_{p} \in \Rset` are scalar coefficients,
+:math:`\epsilon` a random
+noise with zero mean and constant variance :math:`\sigma^2` independent from
+the random variables :math:`\varphi_j(\inputRV)` and :math:`(\varphi_j)_{1 \leq j \leq p}`  a functional basis
```
mbaudin47: Only one blank space is necessary here. Moreover, a verb helps to articulate the sentence.
```suggestion
the random variables :math:`\varphi_j(\inputRV)` and :math:`(\varphi_j)_{1 \leq j \leq p}` is a functional basis
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+such that :math:`\varphi_j : \Rset^{\inputDim} \rightarrow \Rset`.
+
+The forecast output random variable is defined by:
+
+.. math::
+   :label: linearOutputRV
+
+   \widehat{Y} = \sum_{j=1}^{p} a_j \varphi_j(\inputRV).
```
mbaudin47, line 29: The forecast is based on the estimated coefficients, not the exact ones.


```suggestion
   \widehat{Y} = \sum_{j=1}^{p} \hat{a}_j \varphi_j(\inputRV).
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+
+the linear forecast values of the :math:`\vect{x}_i`.
+
+Then, the residuals defined by:
+
+.. math::
+
+    \epsilon_i  = y_i -\widehat{y}_i
+
+are independent realizations of  :math:`\epsilon`. To ensure the independence of the residuals, the random variables
```
mbaudin47: In classical regression, the inputs :math:`(\varphi_j(\inputRV))_{1 \leq j \leq p}` are not random variables. The requirement is that the design matrix has full rank, i.e., its columns must be linearly independent. In classical ordinary regression (based on ordinary least squares), the independence assumption is that the observations of the residuals are independent.

mbaudin47 (changes requested):

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+:math:`(\varphi_j(\inputRV))_{1 \leq j \leq p}` need to be independent.
+
+The linear coefficients :math:`(a_1, \ldots, a_{p})`  and the variance :math:`\sigma^2` are estimated by maximizing the
```
mbaudin47:
```suggestion
The linear coefficients :math:`(a_1, \ldots, a_{p})` and the variance :math:`\sigma^2` are estimated by maximizing the
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+likelihood of the residuals according to the distribution of :math:`\epsilon`.
+
+If the noise  :math:`\varepsilon` follows a normal distribution with zero mean, the maximization of the residuals
```
mbaudin47, line 52: I suggest to check the doubled spaces in the text, since it appears that there are several of them here.
```suggestion
If the noise :math:`\varepsilon` follows a normal distribution with zero mean, the maximization of the sample
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+likelihood is equivalent to solving the least-squares problem (see [bingham2010]_ theorem 1.8 page 22):
+
+.. math::
+    :label: linRegCoef
+
+    \left\{ \widehat{a}_1,\ldots,\widehat{a}_{p} \right\} = \textrm{argmin} \sum_{i=1}^{\sampleSize} \left( y_i - \widehat{y}_i\right)^2
+
```
mbaudin47: This blank line is unnecessary.
```suggestion
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+and :math:`(y_1, \dots, y_\sampleSize)` the associated output values.
+We denote by:
+
+.. math::
+
+   \widehat{y}_i =  \sum_{j=1}^{p} a_j \varphi_j(\vect{x}_i)
+
+the linear forecast values of the :math:`\vect{x}_i`.
+
+Then, the residuals defined by:
```
mbaudin47: The residuals are `y_i -\widehat{y}_i` are dependent because the estimated coefficients, `(\hat{a}_i)_{1 \leq i \leq n}`, are dependent because they are constrained by the least squares fit. The true errors, `y_i - \sum_j a_j \varphi_j (\vect{x}_i)`, are independent realizations of the noise.

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+
+   \widehat{Y} = \sum_{j=1}^{p} a_j \varphi_j(\inputRV).
+
+Let :math:`(\vect{x}_1, \dots, \vect{x}_\sampleSize)` be a sample of the input random vector :math:`\inputRV`
+and :math:`(y_1, \dots, y_\sampleSize)` the associated output values.
+We denote by:
+
+.. math::
+
+   \widehat{y}_i =  \sum_{j=1}^{p} a_j \varphi_j(\vect{x}_i)
```
mbaudin47:
```suggestion
   \widehat{y}_i = \sum_{j=1}^{p} a_j \varphi_j(\vect{x}_i)
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+    = \sum_{j = 1}^\sampleSize \left(y_j - \bar{y}_j\right)^2
+    = \left(\vect{y} - \bar{\vect{y}}\right)^T \left(\vect{y} - \bar{\vect{y}}\right).
 
-The estimators for the regression coefficients
-:math:`\widehat{a}_0,\widehat{a}_1,\ldots,\widehat{a}_{K}`, and the
-standard deviation :math:`\sigma` are obtained from a sample of
-:math:`(X^i,X^{j_1},\ldots,X^{j_K})`, that is a set of :math:`\sampleSize` values
-:math:`(x_1^i,x_1^{j_1},\ldots,x_1^{j_K})`,…,\ :math:`(x_\sampleSize^i,x_\sampleSize^{j_1},\ldots,x_\sampleSize^{j_K})`.
-They are determined via the least-squares method:
+where :math:`\bar{\vect{y}} = (\bar{y}, ..., \bar{y})^T \in \Rset^{\sampleSize}`.
```
mbaudin47: I suggest to remove this equation. 
If, however, you decide to maintain it, it should be properly typeset.
```suggestion
where :math:`\bar{\vect{y}} = (\bar{y}, \ldots, \bar{y})^T \in \Rset^{\sampleSize}`.
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
-coefficient is said to be “significant” if and only if
-:math:`\alpha_{\ell \textrm{lim}}` is lower than a value
-:math:`\alpha` chosen by the user (typically 5% or 10%). The lower the
-*p*-value, the more significant the coefficient.
+.. math::
+
+    SS_{\text{ERR}}
+    = \sum_{j = 1}^\sampleSize \left(y_j -\widehat{y}_j\right)^2
+    = \left(\vect{y} -\widehat{\vect{y}}\right)^T \left(\vect{y} -\widehat{\vect{y}}\right).
+
```
mbaudin47:
```suggestion
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +139,155 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+Variance
+~~~~~~~~
+
+The unbiased estimator of the variance is (see [baron2014]_ page 400,
+[bingham2010]_ page 67):
+
+.. math::
+
+   \widehat{\sigma}^2 = \frac{SS_{\text{ERR}}}{\sampleSize - p}.
+
```
mbaudin47, line 146:
```suggestion
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +139,155 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+~~~~~~~~~~~~
+We assume that the residual :math:`\varepsilon` follows a normal distribution with zero mean.
+
+The F-statistic is based on the hypothesis that all coefficients
+are simultaneously zero (see [baron2014]_ page 400).
+More precisely, the ANOVA F-test is based on the hypothesis:
+
+.. math::
+
+    H_0 : a_1 = . . . = a_p = 0
```
mbaudin47:
```suggestion
    H_0 : a_1 = \cdots = a_p = 0
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +139,155 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+   \widehat{y} = \Tr{\vect{x}} \vect{\hat{a}}
+
+where :math:`\vect{x} = \Tr{(\Psi_1(\vect{x}_0),\dots,\Psi_p(\vect{x}_0))}`.
+
+In case of normally-distributed errors, it follows the distribution:
+
+.. math::
+    :label: linRegPredictDist
+
+    \widehat{y} \sim \cN\left(\Tr{\vect{x}} \vect{a}, \Tr{\vect{x}} (\Tr{\mat{\Psi}} \mat{\Psi})^{-1} \vect{x} \sigma^2\right)
```
mbaudin47:
```suggestion
    \widehat{y} \sim \cN\left(\Tr{\vect{x}} \vect{a}, \Tr{\vect{x}} (\Tr{\mat{\Psi}} \mat{\Psi})^{-1} \vect{x} \sigma^2\right).
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +139,155 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+
+Moreover:
+
+.. math::
+    :label: linRegOutputDist
+
+    Y -\widehat{y} \sim \cN\left(0, \left(1 + \Tr{\vect{x}} \left(\Tr{\mat{\Psi}} \mat{\Psi}\right)^{-1} \vect{x}\right) \sigma^2\right)
+
+
```
mbaudin47:
```suggestion
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +139,155 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+    \widehat{\sigma}^2 \sim \cN\left(\sigma^2, \frac{2  \sigma^4}{\sampleSize - p}\right)
+
+The prediction on a given input vector :math:`\vect{x}_0` is defined by:
+
+.. math::
+
+   \widehat{y} = \Tr{\vect{x}} \vect{\hat{a}}
+
+where :math:`\vect{x} = \Tr{(\Psi_1(\vect{x}_0),\dots,\Psi_p(\vect{x}_0))}`.
```
mbaudin47: In the beginning of the text, the columns are denoted `\varphi_j`. Therefore, the document should stick to this unique notation.
```suggestion
where :math:`\vect{x} = \Tr{(\varphi_1(\vect{x}_0), \dots, \varphi_p(\vect{x}_0))}`.
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +139,155 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+
+The T-test is based on the hypothesis that one single coefficient is zero.
+More precisely, let :math:`\hat{\sigma}^2` be the estimator of the variance
+(see [baron2014]_ page 400):
+
+.. math::
+
+   \widehat{\sigma}^2 = \frac{SS_{\text{ERR}}}{\sampleSize - p}.
+
+The variance of the estimator of the parameters is:
```
mbaudin47, line 202: This is the first time the matrix $\Psi$ is introduced: its definition must be clarified.


```suggestion
Let us define the design matrix :math:`\mat{\Psi}` of dimensions :math:`\sampleSize \times p`. 
This matrix evaluates each of the :math:`p` basis functions on the :math:`\sampleSize` observations of the input sample. 
Its elements are defined by:

.. math::

    \mat{\Psi}_{i,j} = \varphi_j(\vect{x}_i)

for :math:`i = 1, \dots, \sampleSize` and :math:`j = 1, \dots, p`. 
Consequently, the ordinary least squares problem can be expressed in matrix form, and the normal equations guarantee that the matrix :math:`\Tr{\mat{\Psi}}\mat{\Psi}` is invertible provided the basis functions are linearly independent.
The variance of the estimator of the parameters is:
```

python/doc/theory/data_analysis/linear_regression.rst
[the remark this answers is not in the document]
mbaudin47: Actually, @coderabbitai is right on this point. The vector $\boldsymbol{y}$ represents the vector of observations. Still, the sample mean, $\bar{y}$, is a scalar.

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+
+The total sum of squares (see [baron2014]_ page 398) is:
 
 .. math::
 
-    \begin{aligned}
-       \widehat{X}^i = a_0 + \sum_{j \in \{ j_1,\ldots,j_K \} } a_j X^j
-     \end{aligned}
+    SS_{\text{TOT}}
+    = \sum_{j = 1}^\sampleSize \left(y_j - \bar{y}_j\right)^2
```
mbaudin47:
```suggestion
    = \sum_{j = 1}^\sampleSize \left(y_j - \bar{y}\right)^2
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+    = \left(\vect{y} - \bar{\vect{y}}\right)^T \left(\vect{y} - \bar{\vect{y}}\right).
```
mbaudin47: @coderabbitai is correct on the issue below: I suggest to unresolve the comment to reveal the discussion.

mbaudin47: This vector equation, which expresses the total sum of squares as a dot product, is ok. Isn't it @coderabbitai? Since it generates confusions, I suggest to remove it. The definition based on the sum of squares does not introduce any confusion.

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+where :math:`\bar{\vect{y}} = (\bar{y}, ..., \bar{y})^T \in \Rset^{\sampleSize}`.
+The regression sum of squares is:
 
 .. math::
 
-    \begin{aligned}
-       \left\{ \widehat{a}_0,\widehat{a}_1,\ldots,\widehat{a}_{K} \right\} = \textrm{argmin} \sum_{k=1}^\sampleSize \left[ x^i_k - a_0 - \sum_{j \in \{ j_1,\ldots,j_K \} } a_j x^j_k \right]^2
-     \end{aligned}
+    SS_{\text{REG}}
+    = \sum_{j = 1}^\sampleSize \left(\hat{y}_j - \bar{y}_j\right)^2
```
mbaudin47:
```suggestion
    = \sum_{j = 1}^\sampleSize \left(\hat{y}_j - \bar{y}\right)^2
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
+
+    \left\{ \widehat{a}_1,\ldots,\widehat{a}_{p} \right\} = \textrm{argmin} \sum_{i=1}^{\sampleSize} \left( y_i - \widehat{y}_i\right)^2
+
+
+Refer to  :any:`least_squares` to get details on the resolution of the problem.
+
+In addition, considering a normal noise enables to make tests of
+significance (see [rawlings2001]_ page 3).
+In particular, this enables to use the F-test and T-test detailed in the next sections.
+and the maximum likelihood method are equivalent.
```
mbaudin47: This is the result of unfinalized merging the two documents: the segment "_and the maximum likelihood method are equivalent_" has no meaning when it follows "_in the next sections_".

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
-     \end{aligned}
+    R^2 = \frac{SS_{\text{REG}}}{SS_{\text{TOT}}}.
 
-where :math:`\overline{x}^i` denotes the empirical mean of the sample
-:math:`\left\{ x^i_1,\ldots,x_\sampleSize^i  \right\}`.
+
+The coefficient of determination measures the part
+of the variance explained by the linear regression model.
 
 Thus, :math:`0 \leq R^2 \leq 1`. A value close to 1 indicates a good fit
```
mbaudin47: This is wrong. We have $R^2 \leq 1$. But $R^2$ can be negative. It happens if the predictions of the model are worse than a constant, equal to the sample mean.

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,75 +3,134 @@
...
 of the linear model, whereas a value close to 0 indicates that the
```
mbaudin47:
```suggestion
of the linear model, whereas a value close to 0 (or negative) indicates that the
```

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: I do not agree: the meta model is defined with theoretical coefficients and then the coefficients are estimated. This is the way all the meta models are presented, no?

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: The likelihood which is built is the produt of the normal pdf applied on the residuals.... so the algorithm maximizes the likelihood of the residuals with respect to the normal distribution, no?

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: I denoted by 

$$
\widehat{y}_i =  \sum_{j=1}^{p} a_j \varphi_j(\mathbb{x}_i)
$$

the forecast values. So the residuals  $\varepsilon_i$ are independent realizations of the noise. I think we say the same thing.

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok!

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok

adutfoy: The vectorial notation is confusing because the vector $\bar{\mathbb{y}}$ is not natural. I removed the vectporial notations that add nothing more than the summation form.

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes
.....

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: I don't understand! By definition, the numerator and the denominator of $R^2$ are sum of squares... it can not be negative!

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: I have not added the negative value!

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes!

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: I commented this point

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok! I modified the text.

python/doc/theory/data_analysis/linear_regression.rst
mbaudin47: You are right: the original text was wrong (it is likely that I am guilty for this). Still, it must be corrected.

python/doc/theory/data_analysis/linear_regression.rst
mbaudin47: Actually, it is more subtle: sorry for the confusion. Here is a better paragraph.

The coefficient of determination is generally defined as:

.. math::

    R^2 = 1 - \frac{SS_{\text{ERR}}}{SS_{\text{TOT}}}.

For a linear regression model fitted by ordinary least squares that includes an intercept, the total sum of squares is exactly equal to the sum of the regression and error sums of squares. In this specific case, the formula is equivalently written as (see [baron2014]_ page 399):

.. math::

    R^2 = \frac{SS_{\text{REG}}}{SS_{\text{TOT}}}.

The coefficient of determination measures the proportion of the variance explained by the model. When the model includes an intercept and is evaluated on the training data, :math:`0 \leq R^2 \leq 1`. A value close to 1 indicates a good fit, whereas a value close to 0 indicates that the linear model does not provide a relevant forecast. 

However, :math:`R^2` can be negative if the model's predictions are worse than simply predicting the sample mean. This can occur in models without an intercept, models with poorly optimized parameters, or when evaluating predictions on out-of-sample data.

mbaudin47:
Actually, it is more subtle: sorry for the confusion. Here is a better paragraph.

The coefficient of determination is generally defined as:

.. math::

R^2 = 1 - \frac{SS_{\text{ERR}}}{SS_{\text{TOT}}}.
For a linear regression model fitted by ordinary least squares that includes an intercept, the total sum of squares is exactly equal to the sum of the regression and error sums of squares. In this specific case, the formula is equivalently written as (see [baron2014]_ page 399):

.. math::

R^2 = \frac{SS_{\text{REG}}}{SS_{\text{TOT}}}.
The coefficient of determination measures the proportion of the variance explained by the model. When the model includes an intercept and is evaluated on the training data, :math:0 \leq R^2 \leq 1. A value close to 1 indicates a good fit, whereas a value close to 0 indicates that the linear model does not provide a relevant forecast.

However, :math:R^2 can be negative if the model's predictions are worse than simply predicting the sample mean. This can occur in models without an intercept, models with poorly optimized parameters, or when evaluating predictions on out-of-sample data.

python/doc/theory/data_analysis/linear_regression.rst
mbaudin47: The model is Y, uses the exact coefficients and no noise. The outputs y are noisy observations of the model. The metamodel is hat{Y}, uses estimated coefficients and no noise. The equation which uses the estimated coefficients and includes the noise has not particular name, as far as I know.

There is no need to make a difference between the model and the expression of the metamodel using the exact coefficients because there is no truncation in the regression setting, as opposed to PCE. Also, PCE does not involve noise in its most simple setting.

python/doc/theory/meta_modeling/functional_chaos.rst
```diff
@@ -14,8 +14,8 @@ computational cost, in order to obtain accurate results.
 A possible solution to overcome this problem is to project the model
 :math:`\model` in a suitable functional space, such as
 the Hilbert space :math:`L^2\left(\mu_{\inputRV}\right)` of square-integrable functions with
-respect to :math:`\mu_{\inputRV}`. This projection is called a *functional chaos expansion* of
-the model :math:`\model`.
+respect to :math:`\mu_{\inputRV}`. This projection is called a *functional chaos expansion* of
```
jschueller:
```suggestion
respect to :math:`\mu_{\inputRV}`. This projection is called a *functional chaos expansion* of
```

python/doc/theory/numerical_methods/least_squares_resolution.rst
```diff
@@ -3,123 +3,122 @@
...
       \matr{R} \; \vect{a} \, \, = \, \, \matr{Q}^{\mbox{\scriptsize \textsf{T}}} \; \vect{y}
 
-| This upper triangular system can be solved using backwards
-  substitution.
+This upper triangular system can be solved using backwards
+substitution.
 
-| The solving scheme based on Householder QR factorization leads to a
-  relative error that is proportional to:
+The solving scheme based on Householder QR factorization leads to arelative error that is proportional to:
```
jschueller:
```suggestion
The solving scheme based on Householder QR factorization leads to a relative error that is proportional to:
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +131,165 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+
+.. math::
+
+    \mat{\Psi}_{i,j} = \varphi_j(\vect{x}_i)
+
+for :math:`i = 1, \dots, \sampleSize` and :math:`j = 1, \dots, p`.
+
+Consequently, the ordinary least squares problem can be expressed in matrix form, and the normal
+equations guarantee that the matrix :math:`\Tr{\mat{\Psi}}\mat{\Psi}` is invertible provided the basis
+unctions are linearly independent.
```
jschueller:
```suggestion
functions are linearly independent.
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +131,165 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+   \widehat{\sigma}^2 = \frac{SS_{\text{ERR}}}{\sampleSize - p}.
+
+Fisher - test
+~~~~~~~~~~~~~
+We assume that the residual :math:`\varepsilon` follows a normal distribution with zero mean.
+
+The  Fisher’s statistic tests if a coefficient is significantly different from the null value. If not,
+it indicates that the variable :math:`\varphi_i(\inputRV)` does not
+bring valuable information to the forecast :math:`\hat{Y}`. The lower the *p*-value, the more
+significant the coefficient.
```
jschueller: are you describing the T-test instead of the F-test here ?

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +131,165 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+Let :math:`\vect{y} \in \Rset^\sampleSize` be the vector of
+observations and :math:`\vect{\hat{y}} \in \Rset^\sampleSize`
+be the vector of predictions.
+
+The F-statistic is (see [bingham2010]_ Kolodziejcyzk’s theorem 6.5 page 154,
+[baron2014]_ page 400):
+
+.. math::
+
+    f = \frac{SS_{\text{REG}} / (p - 1)}{SS_{\text{ERR}} / (\sampleSize - p)}.
```
jschueller: this should be p instead of p-1 as the model has no explicit intercept

the caution section aknowledges this but the formula itself is inconsistent with the chosen convention

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +131,165 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+    :label: linRegOutputDist
+
+    Y -\widehat{y} \sim \cN\left(0, \left(1 + \Tr{\vect{x}} \left(\Tr{\mat{\Psi}} \mat{\Psi}\right)^{-1} \vect{x}\right) \sigma^2\right).
+
+Some illustrations
+~~~~~~~~~~~~~~~~~~
+By definition, the linear regression model is only relevant for linear
 relationships, as in the following simple example where
-:math:`X^2 = a_0 + a_1 X^1`.
+:math:`Y = a_0 + a_1 X_1`.
```
jschueller, line 289: now we have an intercept which is confusing as the theory model has no explicit intercept

jschueller (approved):

python/src/LeastSquaresMethod_doc.i
```diff
@@ -40,31 +40,38 @@ name : str
...
-    Output weights
+    The design proxy that builds the design matrix
+weight : sequence of float
+    The output weights
 indices : sequence of int
-    Indices allowed in the basis
+    The indices allowed in the basis
 design : 2-d sequence of float
-    A priori known design matrix
+    The known design matrix
```
mbaudin47:
```suggestion
    The design matrix
```

mbaudin47 (changes requested):

mbaudin47:
The first word of the PR is wrong:

image

python/src/LeastSquaresMethod_doc.i
```diff
@@ -40,31 +40,38 @@ name : str
...
 proxy : :class:`~openturns.DesignProxy`
-    Input sample
-weight : sequence of float, optional
-    Output weights
+    The design proxy that builds the design matrix
+weight : sequence of float
+    The output weights
 indices : sequence of int
-    Indices allowed in the basis
+    The indices allowed in the basis
```
mbaudin47:
```suggestion
    The indices in the basis
```

python/src/LeastSquaresMethod_doc.i
```diff
@@ -40,31 +40,38 @@ name : str
     The name of the least-squares method
     Values are 'QR', 'SVD', 'Cholesky'
 proxy : :class:`~openturns.DesignProxy`
-    Input sample
-weight : sequence of float, optional
-    Output weights
+    The design proxy that builds the design matrix
```
mbaudin47:
```suggestion
    The proxy that builds the design matrix
```

mbaudin47:
The link to the produced html doc is broken:

image

image

python/src/LeastSquaresMethodImplementation_doc.i
```diff
@@ -11,34 +11,50 @@ Available constructors:
...
 Notes
 -----
-Solve the least-squares problem:
+This class finds :math:`\widehat{\mat{A}}` that minimizes the objective function  :math:`\cJ`:
 
 .. math::
+    :label: def_A_lsmethod
 
-    \vect{a} = \argmin_{\vect{b} \in \Rset^P} ||y - \vect{b}^{\intercal} \vect{\Psi}(\vect{U})||^2
+    \widehat{\mat{A}} = \argmin \cJ(\mat{A})
```
mbaudin47: There is no matrix involved here.

python/src/LeastSquaresMethodImplementation_doc.i
```diff
@@ -11,34 +11,50 @@ Available constructors:
...
 >>> y = [6, 5, 7, 10]
->>> method = ot.LeastSquaresMethod(A)
->>> x = method.solve(y)
->>> print(x)
+>>> method = ot.LeastSquaresMethod(design_matrix)
+>>> A = method.solve(y)
+>>> print(A)
 [3.5,1.4]
 )RAW"
 %enddef
```
mbaudin47, line 44: There are many other issues lower in the help page: Please consider the issues identified at #1511 and #2385.

mbaudin47:
As part of a task on calibration, we are going to improve the documentation of the `LeastSquaresMethod` class. However, I noticed that you are working on this topic in the current pull request. To avoid conflicts in Git, could you please revert your changes in the files `python/src/LeastSquaresMethod_doc.i` and `python/src/LeastSquaresMethodImplementation_doc.i`?

This will ensure we do not run into any issues during the merge.

Thank you!

adutfoy:
I restored both files as requested by @mbaudin47 .
I propose that @mbaudin47 checks the question raised by Julien about the presence of an intercept or not in the formula within the ongoing work on the regression documentation.
@mbaudin47 and @jschueller : may you have a look on this new version of the PR?

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: Ok. We have to check that point, maybe within the ongoing work on the regression done by @mbaudin47 ?

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: Yes, I corrected the formula.

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: Yes you are right. The test was written two times!!

python/src/ApproximationAlgorithmImplementationFactory_doc.i
```diff
@@ -36,21 +36,15 @@ R"RAW(Build the approximation.
 Parameters
 ----------
 x : 2-d sequence of float
-    The input random observations :math:`\left\{\vect{X}^{(1)}, ..., \vect{X}^{(n)}\right\}` 
-    where :math:`\vect{X}=(X_1, \dots, X_{n_X})^T` is the input of the physical
-    model, :math:`n_X` is the input dimension and :math:`n` is the sample size.
+    The input sample :math:`\left\{\vect{x}^{(1)}, ..., \vect{x}^{(\sampleSize)}\right\}`
+    where :math:`\vect{y}\in \Rset^\inputDim`
```
jschueller:
```suggestion
    where :math:`\vect{x}\in \Rset^\inputDim`
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,86 +3,289 @@
...
+
+.. math::
+
+    R^2 = \frac{SS_{\text{REG}}}{SS_{\text{TOT}}}.
+
+
+The coefficient of determination measures the proportion of the variance explained by the model. When the model includes an intercept and is evaluated on the training data,
+:math:0 \leq R^2 \leq 1. A value close to 1 indicates a good fit, whereas a value close to 0 indicates that the linear model does not provide a relevant forecast.
+
+However, :math:R^2 can be negative if the model's predictions are worse than simply predicting the sample mean. This can occur in models without an intercept,
```
jschueller:
```suggestion
However, :math:`R^2` can be negative if the model's predictions are worse than simply predicting the sample mean. This can occur in models without an intercept,
```

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -3,86 +3,289 @@
...
+For a linear regression model fitted by ordinary least squares that includes an intercept, the total sum of squares is exactly equal to the sum of the regression and error
+sums of squares. In this specific case, the formula is equivalently written as (see [baron2014]_ page 399):
+
+.. math::
+
+    R^2 = \frac{SS_{\text{REG}}}{SS_{\text{TOT}}}.
+
+
+The coefficient of determination measures the proportion of the variance explained by the model. When the model includes an intercept and is evaluated on the training data,
+:math:0 \leq R^2 \leq 1. A value close to 1 indicates a good fit, whereas a value close to 0 indicates that the linear model does not provide a relevant forecast.
```
jschueller:
```suggestion
:math:`0 \leq R^2 \leq 1`. A value close to 1 indicates a good fit, whereas a value close to 0 indicates that the linear model does not provide a relevant forecast.
```

python/doc/theory/reliability_sensitivity/taylor_moments.rst
```diff
@@ -33,25 +27,24 @@ equal to the variance of an input variable (:math:`X_i`).
 Case 1: :math:`\outputDim=1`, :math:`Y = \model(\inputRV) \in \Rset`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In this section, we analyse the special case where :math:`q = 1` and :math:`Y = h(\vect{X}) \in \Rset`.
+In this section, we analyse the special case where :math:`q = 1` and :math:`Y = \model(\inputRV) \in \Rset`.
```
jschueller:
```suggestion
In this section, we analyse the special case where :math:`\outputDim = 1` and :math:`Y = \model(\inputRV) \in \Rset`.
```

python/src/KernelMixture_doc.i
```diff
@@ -19,6 +19,27 @@ For continuous atoms, the probability density function :math:`\inputRV` is defin
...
+    \mu^j = \bar{s}^j + h^j \mu_K
+
+where :math:`\bar{s}^j` is the :math:`j`-th component of the mean vector of the sample points
+:math:`\vect{s}` and :math:`\mu_K` the mean of the kernel.
+
+The covariance matrix :math:`\Cov{\inputRV}` is defined by:
+
+.. math::
+
+    \Cov{\inputRV} = \left( 1 - \dfrac{1}{\sampleSize} \right) \Cov{\vect{s}} + \sigma_K
```
jschueller: sigmaK^2 then if its the stddev

jschueller (approved):
minor comments

python/src/ApproximationAlgorithmImplementationFactory_doc.i
adutfoy: yes

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes

python/doc/theory/data_analysis/linear_regression.rst
mbaudin47: The upcoming work is on `LeastSquaresMethod`, not regression. So updating the regression help pages is ok.

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -80,9 +131,165 @@ allows one to detect significant values of :math:`R^2`. Again, a
...
+    :label: linRegOutputDist
+
+    Y -\widehat{y} \sim \cN\left(0, \left(1 + \Tr{\vect{x}} \left(\Tr{\mat{\Psi}} \mat{\Psi}\right)^{-1} \vect{x}\right) \sigma^2\right).
+
+Some illustrations
+~~~~~~~~~~~~~~~~~~
+By definition, the linear regression model is only relevant for linear
 relationships, as in the following simple example where
-:math:`X^2 = a_0 + a_1 X^1`.
+:math:`Y = a_0 + a_1 X_1`.
```
mbaudin47, line 292: Actually, the correct equation is:
```
:math:`Y = a_1 + a_2 X_1 + \epsilon`.
```
because the indices in the equation `RandomLinearModel` start from 1 and the observations are noisy.

python/doc/theory/data_analysis/linear_regression.rst
mbaudin47: > now we have an intercept which is confusing as the theory model has no explicit intercept

This depends on the basis used. Here, the equation $Y = a_1 + a_2 X_1 + \epsilon$ fits to the regression model with $\varphi_1(X_1) = 1$ and $\varphi_2(X_1) = X_1$. So there is no confusion.

The example is, however, rather elliptic. It might help to clarify the basis used in this example and the value of the exact parameters. It may also help to know the value of the standard deviation of the Gaussian error.

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -107,16 +314,15 @@ relationships, as in the following simple example where
...
 
 
 In this second example (still in dimension 1), the linear model is not
 relevant because of the exponential shape of the relation. But a linear
 approach would be useful on the transformed problem
-:math:`X^2 = a_0 + a_1 \exp X^1`. In other words, what is important is
-that the relationship between :math:`X^i` and the variables
-:math:`X^{j_1}`,…,\ :math:`X^{j_K}` is linear with respect to the
-regression coefficients :math:`a_j`.
+:math:`Y = a_0 + a_1 \exp X_1`.
```
mbaudin47, line 325: The noise is missing here and the first index is 1, according to equation `RandomLinearModel`.

jschueller:
@adutfoy can you rebase and address the comments by michael ?

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: Yes you are right. The meta model is Y and we speak about its estimation here...

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: ok. I modified the examples to fit to the new documentation wich introduces the use of a functional basis. I also mentioned the model we try to estimate.

python/src/LeastSquaresMethod_doc.i
adutfoy: ok

python/src/LeastSquaresMethod_doc.i
adutfoy: ok

python/src/LeastSquaresMethod_doc.i
adutfoy: ok

python/src/LeastSquaresMethodImplementation_doc.i
adutfoy: If y is a vector, then A is a matrix, no?

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes.I improved the presentation of the example.

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes. I improved the presentation of the example.

python/src/LeastSquaresMethodImplementation_doc.i
adutfoy: #1511 and #2385 focuse on LeastSquaresMethod ==> I let it aside, as requested.

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -315,10 +317,12 @@ relationships, as in the following simple example where
     otv.View(graph)
 
 
-In this second example (still in dimension 1), the linear model is not
-relevant because of the exponential shape of the relation. But a linear
-approach would be useful on the transformed problem
-:math:`Y = a_0 + a_1 \exp X_1`.
+Example 2: We consider now the output random variable defined by :math:`Y = \exp (0.5X_1)+ \epsilon` where the random noise :math:`\varepsilon` is distributed according to the standard normal
```
jschueller: line too long

python/doc/theory/data_analysis/linear_regression.rst
```diff
@@ -315,10 +317,12 @@ relationships, as in the following simple example where
...
+distribution and :math:`X_1` is distributed according to the Triangular(1.0, 5.0, 10.0) distribution.
+
+In that case, the default functional basis is not relevant because of the exponential shape of the relation. The following graph shows it clearly. We should have used the functional basis  composed of
```
jschueller: line too long, split at .

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes...

python/doc/theory/data_analysis/linear_regression.rst
adutfoy: yes...

jschueller:
great, now it also needs a rebase

mbaudin47 (approved):

jschueller:
#3314

PR #3213, opened 2026-06-02
Tolerate zero division

### LINKED ISSUE #1328: GeneralLinearModelAlgorithm::computeReducedLogLikelihood code question
I am not sure what [this line](https://github.com/josephmure/openturns/blob/0ae3e83643240ddac3d7e3fe9dbe0ff9485c8103/lib/src/Uncertainty/Algorithm/MetaModel/Kriging/GeneralLinearModelAlgorithm.cxx#L748) is supposed to do.
The block of code immediately above is:

```
  const Scalar epsilon = rho_.normSquare();
  LOGDEBUG(OSS(false) << "epsilon=||rho||^2=" << epsilon);

  if (epsilon <= 0) lastReducedLogLikelihood_ = SpecFunc::LogMinScalar;
```
It seems to me that epsilon is necessarily >=0 due to its construction as the square of a norm. 
Could epsilon be 0? Because the covariance matrix is positive definite, that would imply that $y - F \beta$ is null. Is that not checked somewhere else? That feels like something the user should be warned about.

Here is a simple example where there exists $\beta$ such that $y = F \beta$:

```
import openturns as ot
sampleX = [[1.0], [2.0], [3.0], [4.0], [5.0], [6.0], [7.0], [8.0]]
sampleY = [[1.0], [1.0], [1.0], [1.0], [1.0], [1.0], [1.0], [1.0]]

basis = ot.ConstantBasisFactory(1).build()
covarianceModel = ot.SquaredExponential([1.0])
algo = ot.KrigingAlgorithm(sampleX, sampleY, covarianceModel, basis)
algo.run()
```
The last line, ``algo.run()`` produces the following warnings:
```
WRN - Switch to finite difference to compute the gradient at point=class=Point name=Unnamed dimension=1 values=[2.00066]
WRN - Warning! TNC algorithm failed to converge. The error message is User requested end of minimization
```
This is to be expected since this example was designed to result in a null $y - F \hat{\beta}$.

Now let us compute the reduced log-likelihood at several points:
`rll = algo.getReducedLogLikelihoodFunction()`

`rll([0.1])` produces an error:
```

  File "<ipython-input-35-555d5b56cca2>", line 1, in <module>
    rll([0.1])

  File "/home/C39575/anaconda3/lib/python3.7/site-packages/openturns/func.py", line 10003, in __call__
    return _func.Function___call__(self, *args)

TypeError: InvalidArgumentException : In CovarianceModelImplementation::setAmplitude, the component 0 of amplitude is non positive
```

On the other hand, the reduced log likelihood can be evaluated at other points:

```
print(rll([0.5]))
[285.259]

print(rll([1.0]))
[298.123]

print(rll([2.0]))
[315.479]
```

I do not understand how these values are computed. Mathematically speaking, the reduced likelihood should not even be properly defined in such a case.

josephmure:
## Summary by CodeRabbit

* **Bug Fixes**
  * Amplitude validation now allows zero values.
  * Covariance normalization check improved to treat non-positive or non-finite results as degenerate.
  * Point division/scaling no longer errors on zero divisors or zero-size cases.

* **Tests**
  * Added a regression test to validate log-likelihood behavior on a constant dataset.

* **Chores**
  * .gitignore updated to exclude build artifacts, editor backups, and Pixi-related files.

Closes #1328

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
 build*/
 *~
 *kate-swp
+
+# pixi environments
+.pixi/*
+!.pixi/config.toml
+pixi.lock
+pixi.toml
+
--- a/lib/src/Base/Type/Point.cxx
+++ b/lib/src/Base/Type/Point.cxx
@@ -281,7 +281,6 @@ Point & Point::operator *=(const Scalar scalar)
 Point operator /(const Point & point,
                  const Scalar scalar)
 {
-  if (!(scalar > 0.0 || scalar < 0.0)) throw InvalidArgumentException(HERE) << "Error: cannot divide by 0.";
   int n = point.getDimension();
   double alpha = 1.0 / scalar;
   Point result(point);
@@ -294,7 +293,6 @@ Point operator /(const Point & point,
 /*  In-place division operator */
 Point & Point::operator /=(const Scalar scalar)
 {
-  if (!(scalar > 0.0 || scalar < 0.0)) throw InvalidArgumentException(HERE) << "Error: cannot divide by 0.";
   int n = getDimension();
   double alpha = 1.0 / scalar;
   int one = 1;
--- a/python/test/t_GaussianProcessFitter_std.py
+++ b/python/test/t_GaussianProcessFitter_std.py
@@ -264,8 +264,28 @@ def bugfix_optim_no_feasible():
     ott.assert_almost_equal(squared_epsilon, [7.248e-06, 298.4, 0.9051], 5e-1, 1e-3)
 
 
-if __name__ == "__main__":
+def bugfix_loglikelihood_exception_issue_1328():
+    """
+    This function tests the reduced log-likelihood function of the GaussianProcessFitter
+    to ensure it returns -inf in degenerate cases.
+    """
+    sampleX = [[1.0], [2.0], [3.0], [4.0], [5.0], [6.0], [7.0], [8.0]]
+    sampleY = [[1.0], [1.0], [1.0], [1.0], [1.0], [1.0], [1.0], [1.0]]
 
+    basis = ot.ConstantBasisFactory(1).build()
+    covarianceModel = ot.SquaredExponential([1.0])
+    algo = ot.GaussianProcessFitter(sampleX, sampleY, covarianceModel, basis)
+    algo.run()
+    rll = algo.getReducedLogLikelihoodFunction()
+    rll_value = rll([0.1])[0]
+    assert (
+        rll_value < 300.0
+    ), f"Log-likelihood should theoretically be [-inf], and is tolerated to be < [300.0], but is {rll_value}"
+    surrogate_model_value = algo.getResult().getMetaModel()([4.5])[0]
+    ott.assert_almost_equal(surrogate_model_value, 1.0, 1e-15, 0.0)
+
+
+if __name__ == "__main__":
     ot.RandomGenerator.SetSeed(0)
     sampleSize = 40
     inputDimension = 1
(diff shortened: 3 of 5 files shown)


josephmure:
@coderabbitai review

josephmure:
@coderabbit review

python/test/t_GaussianProcessFitter_std.py
```diff
@@ -264,8 +264,23 @@ def bugfix_optim_no_feasible():
...
+    """
+    sampleX = [[1.0], [2.0], [3.0], [4.0], [5.0], [6.0], [7.0], [8.0]]
+    sampleY = [[1.0], [1.0], [1.0], [1.0], [1.0], [1.0], [1.0], [1.0]]
 
+    basis = ot.ConstantBasisFactory(1).build()
+    covarianceModel = ot.SquaredExponential([1.0])
+    algo = ot.GaussianProcessFitter(sampleX, sampleY, covarianceModel, basis)
+    algo.run()
+    rll = algo.getReducedLogLikelihoodFunction()
+    assert rll([0.1]) == [-float("inf")], f"Log-likelihood should be [-inf], but is {rll([0.1])}"
```
josephmure:
```suggestion
    rll_value = rll([0.01])
    assert rll_value == [
        -float("inf")
    ], f"Log-likelihood should be [-inf], but is {rll_value}"
```

lib/src/Base/Type/Point.cxx
```diff
@@ -281,7 +281,6 @@ Point & Point::operator *=(const Scalar scalar)
 Point operator /(const Point & point,
                  const Scalar scalar)
 {
-  if (!(scalar > 0.0 || scalar < 0.0)) throw InvalidArgumentException(HERE) << "Error: cannot divide by 0.";
```
regislebrun, line 284: Disagree. In specific contexts, dividing by zero makes sense, but not in the general case. Deal with zero division on a case by case basis.

lib/src/Base/Type/Point.cxx
```diff
@@ -294,7 +293,6 @@ Point operator /(const Point & point,
 /*  In-place division operator */
 Point & Point::operator /=(const Scalar scalar)
 {
-  if (!(scalar > 0.0 || scalar < 0.0)) throw InvalidArgumentException(HERE) << "Error: cannot divide by 0.";
```
regislebrun, line 297: Disagree. In specific contexts, dividing by zero makes sense, but not in the general case. Deal with zero division on a case by case basis.

.gitignore
```diff
@@ -1,3 +1,10 @@
 build*/
 *~
 *kate-swp
+
```
regislebrun, line 4: WTF?

regislebrun (changes requested):
I disagree with the way you allow for zero amplitude in covariance models. It is not as simple as allowing for division by zero for Point. Deal with the zero amplitude case locally, ie at the CovarianceModel&Co level.

.gitignore
josephmure: These are useful if you use a pixi environment as a substitute for conda. As the environment is defined in the working directory rather than on a system-wide basis, it stores files in the folder which you do not want to commit.

lib/src/Base/Type/Point.cxx
josephmure: But to deal with zero division on a case by case basis you cannot forbid it at the Point level.

regislebrun: Of course you can. Either you trap the exception, or you test locally for zero division before to divide your Point by your Scalar

.gitignore
regislebrun: IMO it is local behaviour, not to be enforced at the project level (or we will have exclusions for such or such other environment, a never ending list)

PR #3218, opened 2026-06-04
Accessor event in simulation algorithms

m-balesdent:
This PR proposes to add a setter accessor of event for the reliability algorithms.

This accessor is needed for the active learning Kriging class currently under development.

## Summary by CodeRabbit

* **New Features**
  * Simulation algorithms now support updating the event after creation via a public setter, allowing an algorithm instance to accept a new event and be rerun.

* **Documentation**
  * Added Python docstring for the new event setter to document its usage.

* **Tests**
  * Added deterministic unit tests that verify event updates for multiple simulation methods and confirm expected probability estimates after updates.


lib/src/Uncertainty/Algorithm/Simulation/DirectionalSampling.cxx
```diff
@@ -65,6 +65,17 @@ DirectionalSampling * DirectionalSampling::clone() const
   return new DirectionalSampling(*this);
 }
 
+/*  Event accessor */
+void DirectionalSampling::setEvent(const RandomVector & event)
+{
+  if (!event.isEvent() || !event.isComposite()) throw InvalidArgumentException(HERE) << "DirectionalSampling requires a composite event";
```
jschueller: isEvent is already check in the parent setEvent

lib/src/Uncertainty/Algorithm/Simulation/DirectionalSampling.cxx
jschueller, line 59: looks like we could use setEvent to factorize

lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
```diff
@@ -61,6 +61,24 @@ NAIS * NAIS::clone() const
   return new NAIS(*this);
 }
 
+/*  Event accessor */
+void NAIS::setEvent(const RandomVector & event)
+{
+
+  const Bool previousSense = getEvent().getOperator()(0, 1);
```
jschueller: wrong translation sens->direction

lib/src/Uncertainty/Algorithm/Simulation/DirectionalSampling.cxx
m-balesdent: we use the syntax as it is in the constructor of the class. which lines do you want to change?

lib/src/Uncertainty/Algorithm/Simulation/NAIS.cxx
m-balesdent: fixed

lib/src/Uncertainty/Algorithm/Simulation/DirectionalSampling.cxx
m-balesdent: fixed

lib/src/Uncertainty/Algorithm/Simulation/DirectionalSampling.cxx
m-balesdent: By the way, I'm not sure that this could be factorized, because `standardEvent_` is not implemented by all the classes that use `setEvent`, see for example `NAIS` or `PhysicalSpaceCrossEntropyImportanceSampling`

m-balesdent:
Can be closed because included in PR #3262

PR #3219, opened 2026-06-04
add warning for NAIS and adaptive IS 

m-balesdent:
Hello,

`SubsetSampling`, `NAIS` , `StandardSpaceCrossEntropyImportanceSampling`, and `PhysicalSpaceCrossEntropyImportanceSampling` do not use `MaximumCoefficientOfVariation` parameter. 

For `SubsetSampling`, a warning already exists to alert the user when he wants to set this parameter to a value. 

This PR proposes to add this feature to `NAIS` , `StandardSpaceCrossEntropyImportanceSampling`, and `PhysicalSpaceCrossEntropyImportanceSampling` 

## Summary by CodeRabbit

* **Bug Fixes**
  * Algorithms now warn when a configured maximum coefficient of variation will not be used as a termination criterion.
  * Attempts to set the maximum coefficient of variation are explicitly rejected and now raise an error to prevent misuse.
  * Minor wording fix in the warning message (singular vs. plural) for clearer user-facing text.


sofianehaddad:
hmmm but what if user change the resource map key considering it will influence the convergence ?
I suggest :
  - set the coeff to `0` and not the default value
  - throw an exception with the setter

m-balesdent:
> what if user change the resource map key considering it will influence the convergence ?

I think this parameter is never used by the algorithm.

m-balesdent:
> hmmm but what if user change the resource map key considering it will influence the convergence ? I suggest :
> 
>     * set the coeff to `0` and not the default value
> 
>     * throw an exception with the setter

exception with the setter has been added to `SubsetSampling`, `NAIS` and `CrossEntropyImportanceSampling`.

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
jschueller: drop this then ?

m-balesdent: Yes !

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
```diff
@@ -70,6 +70,13 @@ SubsetSampling * SubsetSampling::clone() const
 }
 
 
+/* Setter for MaximumCoefficientOfVariation. */
+void SubsetSampling::setMaximumCoefficientOfVariation(const Scalar maximumCoefficientOfVariation)
```
jschueller:
```suggestion
void SubsetSampling::setMaximumCoefficientOfVariation(const Scalar)
```

jschueller (approved):

lib/src/Uncertainty/Algorithm/Simulation/SubsetSampling.cxx
m-balesdent: done

regislebrun (approved):

jschueller:
applied to master manually because of the conflicts

PR #3225, opened 2026-06-09
LLM1

jschueller:
move trivial fixes from #3187

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numerical robustness: avoided divide-by-zero, handled singular/zero cases, fixed incorrect cached flags and persistence/restoration behavior.

* **Improved Input Validation**
  * Tightened checks for dimensions, sample sizes, bounds, orientations and parameters; added explicit errors for invalid or degenerate inputs.

* **Behavioral Changes**
  * More deterministic error handling (exceptions or error codes) in previously undefined or silent-failure scenarios; clearer error messages.

* **Tests**
  * Added/updated tests to verify new guards on random generator and input validations.


lib/src/Uncertainty/Distribution/Normal.cxx
```diff
@@ -118,7 +118,7 @@ Normal::Normal(const Point & mean,
     {
       sigma[i] = std::sqrt(cii);
       for (UnsignedInteger j = 0; j < i; ++ j)
-        if (sigma[j] > 0.0)
+        if ((sigma[i] > 0.0) && (sigma[j] > 0.0))
```
regislebrun: The test on sigma[i] should be done outside of the j loop, and this loop should be skipped if sigma[i]<=0

regislebrun (approved):
Only a minor improvement. A good bunch of bugs fixed!

PR #3226, opened 2026-06-09
LLM2

jschueller:
moved from #3187

## Release Notes

**Bug Fixes**
- Fixed computation of field input means and distribution parameter edge cases (degenerate probabilities, boundary values)
- Corrected probability and CDF calculations across multiple distributions including Binomial, Geometric, Clayton, and Pareto
- Resolved mean, kurtosis, and quantile calculations in truncated normal and student distributions
- Fixed hypergeometric probability table generation

**Tests**
- Enhanced test coverage for hypergeometric distribution and complex matrix operations
- Added compound distribution test scenarios with mixed component types


lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
```diff
@@ -216,7 +216,7 @@ Point PiecewiseHermiteEvaluation::derivate(const Point & inP) const
   {
     if (enableExtrapolation_)
     {
-      return values_[iRight];
+      return Point(values_.getDimension());
```
regislebrun: same here

lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
```diff
@@ -202,7 +202,7 @@ Point PiecewiseHermiteEvaluation::derivate(const Point & inP) const
   {
     if (enableExtrapolation_)
     {
-      return values_[iLeft];
+      return Point(values_.getDimension());
```
regislebrun: It was not supposed to be triggered by a key in ResourceMap, allowing to choose between an extrapolation with constant values or throwing an exception?  What is special about the null point in this context?

lib/src/Uncertainty/Distribution/DistFunc.cxx
```diff
@@ -797,7 +797,7 @@ UnsignedInteger DistFunc::rHypergeometric(const UnsignedInteger n,
   for (UnsignedInteger x = xMode + 1; x <= xMax; ++x)
   {
     p *= ((1.0 + k - x) / x) * ((1.0 + m - x) / (n + x - m - k));
-    probabilities[k] = p;
+    probabilities[x] = p;
```
regislebrun, line 800: ouch...

lib/src/Uncertainty/Distribution/DistFunc.cxx
```diff
@@ -823,12 +823,12 @@ Indices DistFunc::rHypergeometric(const UnsignedInteger n,
...
   }
   p = probabilities[xMode];
   for (UnsignedInteger x = xMode; x > xMin; --x)
   {
-    p *= (x / (k - x - 1.0)) * ((n + x - m - k) / (m - x - 1.0));
+    p *= (x / (k - x + 1.0)) * ((n + x - m - k) / (m - x + 1.0));
```
regislebrun, line 831: re-ouch

regislebrun (changes requested):
Only a minor point to discuss, otherwise full ok

lib/src/Base/Func/PiecewiseHermiteEvaluation.cxx
jschueller: lets drop that commit for now

PR #3234, opened 2026-06-15
Building reproducibly in Debian

pgrt:
Hello,

Debian recently made it mandatory for its packages to build reproducibly, according to the changes listed on
    https://tests.reproducible-builds.org/debian/index_variations.html

While I trust it is an important achievement for software to build reproducibly, I acknowledge the perspective of packaging is not necessarily the same as the one of openturns as a library to be built from source by its users.

Up to now openturns did not build reproducibly on Debian, I could achieve this by making these changes:
- fixing the locales for the build (the build date is embedded into the .so through the header files, in a way that is sensitive to the locale).
- using "ORIGIN" instead of the build path in the RPATH at build time, to be robust to the directory in which the build is made;
- changing the information in OTconfigure_args.hxx, namely the "system version" which includes the version of the kernel, hence is not robust to a change in the version of the kernel.

I am proposing this pull request which only touches distro/debian, so that you are informed about the changes I did. Maybe you will want to contemplate incorporating them / tweaking them for the main openturns library -- once again, I understand my packaging concerns maybe do not fit yours :)

Best wishes,

Pierre

## Summary by CodeRabbit

* **Chores**
  * Improved Debian build reproducibility by adjusting embedded system version metadata to use the intended build-time value.
  * Enhanced reproducible build locale handling by exporting a consistent UTF-8 locale.
  * Updated Debian CMake configure options related to system version reporting and runtime path behavior.

### DIFF
--- a/distro/debian/patches/build_reproducibility.patch
+++ b/distro/debian/patches/build_reproducibility.patch
@@ -1,0 +1,17 @@
+Description: enhancing build reproducibility: putting the architecture in the
+ system information instead of the kernel version
+Author: Pierre Gruet <pgt@debian.org>
+Forwarded: no
+Last-Update: 2026-06-09
+
+--- a/lib/include/OTconfigure_args.hxx.in
++++ b/lib/include/OTconfigure_args.hxx.in
+@@ -30,7 +30,7 @@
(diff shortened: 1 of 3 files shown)


jschueller:
moved to #3187, thanks

PR #3244, opened 2026-06-26
NonLinearLeastSquaresCalibration: Improve documentation

mbaudin47:
Updated documentation for NonLinearLeastSquaresCalibration, enhancing clarity on parameter descriptions and algorithm behavior.

Fixed the sentence:

> By default, the posterior distribution is evaluated based on a linear approximation of the model at the optimum. 

This is wrong, since the default value of `NonLinearLeastSquaresCalibration-BootstrapSize` is equal to 100.

## Summary by CodeRabbit

* **Documentation**
  * Clarified nonlinear least-squares calibration, including Gaussian-error estimation and posterior evaluation through linearization or kernel-smoothed bootstrap methods.
  * Defined `startingPoint` as the initial optimization value.
  * Updated accessor and mutator descriptions with clearer “Return” and “Set” wording.
  * Clarified that zero bootstrap size corresponds to linear least-squares calibration.
  * Improved residual-function documentation with revised observation types, notation, formatting, and stacked residual-vector details.

### DIFF
--- a/python/src/NonLinearLeastSquaresCalibration_doc.i
+++ b/python/src/NonLinearLeastSquaresCalibration_doc.i
@@ -11,31 +11,43 @@ inputObservations : 2-d sequence of float
 outputObservations : 2-d sequence of float
     The sample of output observations.
 startingPoint : sequence of float
-    The reference value of the parameter, used as the starting point
-    of the optimization.
+    Initial parameter value used as the starting point of the optimization.
 
 Notes
 -----
-NonLinearLeastSquaresCalibration is the minimum variance estimator of the parameter 
-of a given model with no assumption on the dependence of the model with respect 
-to the parameter.
+
+NonLinearLeastSquaresCalibration estimates the parameter of a model by
+minimizing the sum of squared residuals between the model predictions and
+the observed outputs.
+Under standard assumptions on the observation errors (see 
+:any:`code_calibration` for more details), the resulting
+estimator coincides with the maximum likelihood estimator for Gaussian
+errors.
+Unlike :class:`~openturns.LinearLeastSquaresCalibration`, this algorithm
+does not require the model to depend linearly on the parameter.
 
 The prior distribution of the parameter is a noninformative prior
-emulated using a flat :class:`~openturns.Normal` centered on the `startingPoint` and with a variance equal to SpecFunc.Infinity.
+emulated using a flat :class:`~openturns.Normal` centered on the `startingPoint`
+and with a variance equal to SpecFunc.Infinity.
 
-The posterior distribution of the parameter is :class:`~openturns.Normal` and reflects the 
-variability of the optimum parameter depending on the observation sample. 
-By default, the posterior distribution is evaluated based on a linear approximation 
-of the model at the optimum. 
-This corresponds to using the :class:`~openturns.LinearLeastSquaresCalibration` at the optimum, 
-and is named *Laplace approximation* in the Bayesian context. 
-However, if the key `NonLinearLeastSquaresCalibration-BootstrapSize` in the :class:`~openturns.ResourceMap` 
-is set to a nonzero positive integer, then a bootstrap resampling of the observations 
-is performed and the posterior distribution is based on a :class:`~openturns.KernelSmoothing` 
-of the sample of bootstrap optimum parameters. 
+The posterior distribution of the parameter reflects the uncertainty in the
+parameter estimate arising from observation errors, under the assumed model
+(and the prior information).
+If the key `NonLinearLeastSquaresCalibration-BootstrapSize` in the :class:`~openturns.ResourceMap`
+is set to zero, the posterior distribution is evaluated based on a linear approximation
+of the model at the optimum.
+This corresponds to using the :class:`~openturns.LinearLeastSquaresCalibration` algorithm at the optimum.
+However, if the key `NonLinearLeastSquaresCalibration-BootstrapSize` in the :class:`~openturns.ResourceMap`
+is set to a nonzero positive integer, bootstrap resampling of the observations is used to estimate
+the distribution of the calibrated parameters.
+The calibrated parameter distribution is approximated based on a kernel density estimate
+obtained from the bootstrap sample.
+By default, a bootstrap algorithm is used.
 
-The resulting distribution of the output error is a :class:`~openturns.Normal` and is 
-computed from the residuals. 
+The output (observation) error is modeled as a Normal random variable whose variance
+is estimated from the calibration residuals.
+This is performed whatever the value of the `NonLinearLeastSquaresCalibration-BootstrapSize`
+parameter i.e., it is independent from the distribution of the parameter.
 
 If least squares optimization algorithms are enabled, then the algorithm used is the
 first found by `Build` of :class:`~openturns.OptimizationAlgorithm`.
@@ -72,17 +84,17 @@ Calibrate a nonlinear model using non-linear least-squares:
 // ---------------------------------------------------------------------
 
 %feature("docstring") OT::NonLinearLeastSquaresCalibration::getStartingPoint
-"Accessor to the parameter `startingPoint`.
+"Return the starting point used for the optimization.
 
 Returns
 -------
 startingPoint : :class:`~openturns.Point`
-    Parameter startingPoint."
+    Starting point used for the optimization."
 
 // ---------------------------------------------------------------------
 
 %feature("docstring") OT::NonLinearLeastSquaresCalibration::getOptimizationAlgorithm
-"Accessor to the optimization algorithm used for the computation.
+"Return the optimization algorithm used for the computation.
 
 Returns
 -------
@@ -92,7 +104,7 @@ algo : :class:`~openturns.OptimizationAlgorithm`
 // ---------------------------------------------------------------------
 
 %feature("docstring") OT::NonLinearLeastSquaresCalibration::setOptimizationAlgorithm
-"Accessor to the optimization algorithm used for the computation.
+"Set the optimization algorithm used for the computation.
 
 Parameters
 ----------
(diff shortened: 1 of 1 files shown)


jschueller (approved):

python/src/NonLinearLeastSquaresCalibration_doc.i
```diff
@@ -11,28 +11,37 @@ inputObservations : 2-d sequence of float
...
-The prior distribution of the parameter is a noninformative prior
-emulated using a flat :class:`~openturns.Normal` centered on the `startingPoint` and with a variance equal to SpecFunc.Infinity.
-
-The posterior distribution of the parameter is :class:`~openturns.Normal` and reflects the 
-variability of the optimum parameter depending on the observation sample. 
-By default, the posterior distribution is evaluated based on a linear approximation 
+NonLinearLeastSquaresCalibration estimates the parameter of a model by
+minimizing the sum of squared residuals between the model predictions and
+the observed outputs.
+Under standard assumptions on the observation errors, the resulting
```
l-brevault: Maybe we can recall them here: observations errors are independent accross observation samples, homoscedasticity of the observations errors, Gaussian errors, residuals arise only from observation errors so model discrepancy is negligeable, parameters remain constant over all observations, noise affects only outputs, not the model inputs, etc.

It would help the user to know if there problems are satisfying the underlying assumptions.

python/src/NonLinearLeastSquaresCalibration_doc.i
```diff
@@ -11,28 +11,37 @@ inputObservations : 2-d sequence of float
...
+estimator coincides with the maximum likelihood estimator for Gaussian
+errors.
+Unlike :class:`~openturns.LinearLeastSquaresCalibration`, this algorithm
+does not require the model to depend linearly on the parameter.
+
+The prior distribution of the parameter is approximated by a Normal
```
l-brevault: is set to a Normal ? 
The prior is set, it is not really approximated.

python/src/NonLinearLeastSquaresCalibration_doc.i
```diff
@@ -11,28 +11,37 @@ inputObservations : 2-d sequence of float
...
+distribution centered on the `startingPoint` with a very large variance,
```
l-brevault: In the previous documentation, it was stated: "a variance equal to SpecFunc.Infinity." 
which is more clear (if it is the case). 
"very large variance" is vague

python/src/NonLinearLeastSquaresCalibration_doc.i
```diff
@@ -11,28 +11,37 @@ inputObservations : 2-d sequence of float
...
+so that the influence of the prior is negligible compared with that of
+the observations.
+
+The posterior distribution of the parameter reflects the sampling
+variability of the estimated parameter.
```
l-brevault: Adopting a Bayesian interpretation and with the standard assumptions detailed above, the posterior distribution of the parameter reflects the uncertainty in the parameter estimate arising from observation errors, under the assumed model (and the prior information).

python/src/NonLinearLeastSquaresCalibration_doc.i
```diff
@@ -11,28 +11,37 @@ inputObservations : 2-d sequence of float
...
+If the key `NonLinearLeastSquaresCalibration-BootstrapSize` in the :class:`~openturns.ResourceMap` 
+is set to zero, the posterior distribution is evaluated based on a linear approximation
```
l-brevault: (a second-order Taylor expansion)

it could be interesting to mention that it is not just a first-order by a second-order approximation meaning that Hessian matrix is computated based on the model   (which could lead to numerical issues depending on the type of model)

python/src/NonLinearLeastSquaresCalibration_doc.i
```diff
@@ -11,28 +11,37 @@ inputObservations : 2-d sequence of float
...
 This corresponds to using the :class:`~openturns.LinearLeastSquaresCalibration` at the optimum, 
 and is named *Laplace approximation* in the Bayesian context. 
 However, if the key `NonLinearLeastSquaresCalibration-BootstrapSize` in the :class:`~openturns.ResourceMap` 
 is set to a nonzero positive integer, then a bootstrap resampling of the observations 
 is performed and the posterior distribution is based on a :class:`~openturns.KernelSmoothing` 
-of the sample of bootstrap optimum parameters. 
+of the sample of parameter estimates obtained from the bootstrap resamples.
+By default, a bootstrap algorithm is used.
 
 The resulting distribution of the output error is a :class:`~openturns.Normal` and is
```
l-brevault: It is not necessarely be the case if Bootstrap is used for the posterior. It is true for Laplace approximation but not in the more generic case of Bootstrap even if the observation errors are assumed Gaussian.

l-brevault (approved):
Thank you for the clarification of this documentation. 
Some suggested additional improvements, if you agree with it.

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: Introducing the details of the hypotheses in this particular page is not necessarily what we want to do. This is because it implies to introduce the same hypotheses in each of the `LinearLeastSquaresCalibration`, `GaussianLinearCalibration` and `GaussianNonLinearCalibration`. Since the theory is available in the theory help page, I introduce a pointer to the theory help page to solve this.

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: The doc reflects what is actually done in the code:

https://github.com/openturns/openturns/blob/d209b55dee0d0d593283f97566e7999ed85e5a21/lib/src/Uncertainty/Bayesian/NonLinearLeastSquaresCalibration.cxx#L247

Would you suggest to change the code?

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: Ok: I write this exact sentence in the doc.

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: You are right: I restore the earlier sentence, which is clearer.

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: You are right: I restore the earlier sentence, which is clearer.

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: I am confused here: do you mean that the first-order Taylor expansion of the parametric model leads to a second-order Taylor expansion of the cost function?

l-brevault: In fact, I think that there is a confusion in the documentation, it is stated that: 

> This corresponds to using the class:`~openturns.LinearLeastSquaresCalibration` at the optimum, and is named *Laplace approximation* in the Bayesian context.

However, as stated in Wikipedia: "Laplace's approximation or the **quadratic approximation** (QUAP) provides an analytical expression for a posterior probability distribution by fitting a Gaussian distribution with a mean equal to the MAP solution and precision equal to the observed Fisher information."

By definition as you are at the "optimum" (the MAP) with respect to the calibraiton parameters, the gradient of the likelihood (multiplied by the prior) is equal to zero and a second order Taylor expression is used in Laplace approximation to provide an approximate the posterior distribution. Therefore the Hessian matrix of the likelihood (multiplied by the prior) at the optimal \thetastar is computed (which  (which could lead to numerical issues depending on the type of the simulation model).

However, looking at the code, this is not what is done (it is not a Laplace approximation). A linear least square posterior estimation is applied based on the optimal \thetastar: 

`LinearLeastSquaresCalibration::ComputePosteriorAndErrorDistribution(thetaStar, residual.getImplementation()->getData(), gramInverse, 0, tmp);`

Therefore, I would suggest to modifiy the documentation to remove the mention of Laplace approximation and only say that the posterior is approximated based on a linear least square calibration method by fixing the parameters to the optimum \thetastar.

In the future, it could be interesting to implement the Laplace approximation that is more suited for non-linear least squares calibration.

python/src/NonLinearLeastSquaresCalibration_doc.i
l-brevault: The association of the two sentences (about boostrap and output error) are a bit confusing. 

The sentence:

> "By default, a bootstrap algorithm is used.

The resulting distribution of the output error is a Normal and is computed from the residuals."

may suggest that the bootstrap somehow leads to a Normal output-error distribution. The "resulting distribution" may refer to the one obtained from bootstrap.

The bootstrap only estimates uncertainty on the calibrated parameters.
The Normal output-error model is estimated separately from the residuals of the final calibration (for \thetastar).

For more clarity, I would suggest to improve the documentation as:

> if the key `NonLinearLeastSquaresCalibration-BootstrapSize` in the :class:`~openturns.ResourceMap` 
is set to a nonzero positive integer, bootstrap resampling of the observations is used to estimate the distribution of the calibrated parameters. The calibrated parameter distribution is approximated based on a kernel density estimate obtained from the bootstrap samples.
> 
> Independently, the output (observation) error is modeled as a Normal random variable whose variance is estimated from the calibration residuals.

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: Thank you for this insightful analysis. To implement the feature, I created the new issue #3273. Meanwhile, I updated the doc at d723b794e0de75388b2ec2d1576d54c015f91315. This should be ok now.

python/src/NonLinearLeastSquaresCalibration_doc.i
mbaudin47: You are right, this was confusing. This should be good to go after 8fdb6f1c79de6aac0454d541ea93a7a8338a1540.

mbaudin47:
@l-brevault : I think that I took all of your comments in my review. Can you confirm this?

l-brevault:
Yes, Thank you for taking into consideration the different comments.
It is ok for me.

mbaudin47:
@jschueller : This PR is ok for merging.

mbaudin47:
@jschueller : I suggest to merge this PR. Otherwise, this may generate conflicts with future PRs that are going to be created because of new calibrations features that are planned in september.

PR #3247, opened 2026-06-29
LLM7

jschueller:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness for empty/degenerate inputs (better handling of marginal computations and matrix/tau/quantile paths).
  * Tightened XML and HDF5 save/load behavior, including safer sizing, conditional compression, and clearer validation failures.
  * Corrected numerical formulas and edge-case logic in several uncertainty/coplanar computations (e.g., Generalized Pareto characteristic functions; expected-improvement near zero variance).
* **New Features**
  * Save/load now preserves additional regression/calibration-related state when available.
* **Tests**
  * Updated existing expectations and added regressions for Generalized Pareto, Frank copula derivative, Sklar copula upper-tail quantiles, and SimulatedAnnealingLHS.

### DIFF
--- a/lib/src/Base/Stat/SampleImplementation.cxx
+++ b/lib/src/Base/Stat/SampleImplementation.cxx
@@ -1254,6 +1254,7 @@ struct ComputeKendallPolicy
 CorrelationMatrix SampleImplementation::computeKendallTau() const
 {
   if (!(size_ > 0)) throw InternalException(HERE) << "Error: cannot compute the Kendall tau of an empty sample.";
+  if (dimension_ < 2) return CorrelationMatrix(dimension_);
 
   // Use external efficient C implementation of the O(Nlog(N)) or O(N^2) Kendall tau computation depending on the sample size
   const Bool smallCase = size_ < ResourceMap::GetAsUnsignedInteger("Sample-SmallKendallTau");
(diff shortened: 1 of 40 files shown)


regislebrun (approved):

jschueller:
@regislebrun I added a few more commits (it was a draft when you reviewed)

lib/src/Base/Stat/SampleImplementation.cxx
```diff
@@ -1634,7 +1634,7 @@ Scalar SampleImplementation::computeEmpiricalCDF(const Point & point,
 
   const CDFPolicy policy( *this, point, tail );
   ReductionFunctor<CDFPolicy> functor( *this, policy );
-  TBBImplementation::ParallelReduce( 0, size_, functor );
+  TBBImplementation::ParallelReduce(0, size_, functor, 1024);
```
regislebrun, line 1637: Is it reasonable to hard code this value? And to make it independent from the dimension?

regislebrun (commented):
does not depend on the functions being combined in DualLinearCombinationEvaluation (the underlying functions are not always cheap polynomials) and so on.

lib/src/Base/Stat/SampleImplementation.cxx
jschueller: yes, as the underlying operation is very lightweight

for DualLinearCombinationEvaluation, can you think of an example ?

regislebrun: I suppose that you chose 1024 after a benchmark. Can you share it? Is it the same if you have a collection of size ~5000 of ProductPolynomialEvaluation with input dimension~20 or if you have 100 functions with input dimension 5?

jschueller: its not really a fine-grained benchmark, I told the LLM to increase test time and analyze the speedup of each parallel region for 1,2,4,8 cpus with 2 replications against the default grainsize.
My goal was to avoid often spawning all the threads at very small workloads, so this should only discriminate only very lightweight tasks where parallelization is often harmful.

regislebrun: Ok. Go for it.

regislebrun (approved):

Issue #1468, opened 2020-04-27
Karhunen–Loève expansion of a cyclic field


jigarparekh279:
Hi, I want to perform a Karhunen–Loève expansion of a cyclic field. Is it possible to do so with KarhunenLoeveP1Algorithm or KarhunenLoeveSVDAlgorithm?

regislebrun:
Hi,
Could you give me more details on your application? Is it a 1D mesh? Is it uniform or not? In the easy setting of a uniform 1D mesh (ie a circle) you can do it using KarhunenLoeveSVDAlgorithm. All your realizations are periodic, so will be the modes if yoy take care of the weights of each node. The difference between [0,1] and R/Z is that in the first case the end points 0 and 1 have a weight half of the other nodes, while for R/Z they must have the same weight.
Cheers
Régis

jigarparekh279:
Hi Régis,
Thanks for the reply. I'll try to be more elaborate. 
I am solving a stochastic CFD problem. First, I obtain a turbulent viscosity field from a deterministic CFD solution and then I use a log-normal random field to insert randomness into this viscosity. The CFD problem is 2D (x,y) and cyclic in the x-direction, which gives me a cyclic turbulent viscosity field. The following is the _snippet_ of the code where I perform KL expansion of a Gaussian process with a SquaredExponential covariance model -

```py
...
H  = 0.1

lx = 1.0 * H 
ly = 0.5 * H 
l  = [lx, ly]

covModel  = ot.SquaredExponential(l, [1])
eigValK_0 = 1e-2

kleAlgo = ot.KarhunenLoeveP1Algorithm(kleMesh, covModel, eigValK_0)
# sample  = ot.GaussianProcess(covModel, kleMesh).getSample(1000)
# kleAlgo = ot.KarhunenLoeveSVDAlgorithm(sample, eigValK_0)
  
kleAlgo.run()
kleResult = kleAlgo.getResult()  
...
``` 

When I perform KL expansion of this random field, I don't know how can I set the x-direction to give me cyclic modes. Should it be done in the covariance model or in the KL algorithm? 
Let me know if something is not clear.

jigarparekh279:
Hi Régis,
Any comments on my previous post?

regislebrun:
[...]

PR #3252, opened 2026-07-01
Create CurlFreeModel and DivFreeModel

regislebrun:
These models allow to extract the curl-free and div-free parts of a given covariance model. They are well-defined when the given covariance model has an input dimension greater or equal to 2 and an output dimension equal to 1. For an input dimension greater or equal to 4, the well-definitness of these models is a conjecture.

This PR addresses part of the points raised in #3305.

## Summary by CodeRabbit

- **New Features**
  - Added experimental curl-free and divergence-free covariance models for vector-valued random fields.
  - Models are available through the Python `openturns.experimental` module and support covariance evaluation, scalar extraction, scaling, parameter handling, and persistence.

- **Documentation**
  - Added user-manual guidance, API documentation, examples, and a bibliography reference for both models.

- **Tests**
  - Added comprehensive validation and post-installation checks for both experimental models.


lib/src/Base/Stat/CMakeLists.txt
```diff
@@ -40,6 +40,8 @@ ot_add_source_file (FilteringWindowsImplementation.cxx)
 ot_add_source_file (FilteringWindows.cxx)
 ot_add_source_file (Hamming.cxx)
 ot_add_source_file (Hann.cxx)
+ot_add_source_file (CurlFreeModel.cxx)
```
jschueller, line 43: you forgot to add new files

jschueller: maybe we should rename them to contain CovarianceModel

lib/src/Base/Stat/openturns/DivFreeModel.hxx
```diff
@@ -0,0 +1,88 @@
...
+
+  /** String converter */
+  String __repr__() const override;
+
+  /** String converter */
+  String __str__(const String & offset = "") const override;
+
+  /** Underlying model accessor */
+  CovarianceModel getCovarianceModel() const;
+
```
jschueller, line 72: do we need to override parameter accessors ?

lib/src/Base/Stat/CMakeLists.txt
regislebrun: Fixed

regislebrun:
@coderabbitai review

jschueller:
a dedicated example in the gallery would be welcome to understand the big picture ; is this related to #1468 ?

lib/src/Base/Stat/CurlFreeModel.cxx
```diff
@@ -0,0 +1,215 @@
...
+    Point tPlus(t);
+    Point tMinus(t);
+    tPlus[j] += epsilon;
+    tMinus[j] -= epsilon;
+    const Scalar delta = tPlus[j] - tMinus[j];
+    const Matrix gradPlus(model_.partialGradient(s, tPlus));
+    const Matrix gradMinus(model_.partialGradient(s, tMinus));
+    for (UnsignedInteger i = 0; i < inputDimension_; ++i)
+    {
+      result(i, j) = (gradPlus(i, 0) - gradMinus(i, 0)) / delta;
```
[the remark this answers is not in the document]
regislebrun: Fixed: the finite-difference step is now derived per-coordinate from the model scale: epsilon_j = scale[j] * cbrt(eps). This commit also lowers the risk of small/large scales by making the step scale-aware, and the extreme-scales regression blocks (small_scale/large_scale) were added to python/test/t_DivFreeModel_std.py. Additionally, the step now also works when the underlying model is an IsotropicCovarianceModel, which reports a single scale: the code falls back to scale[0] in that case, with a dedicated IsotropicCovarianceModel regression block added to t_DivFreeModel_std.py (this was crashing under -D_GLIBCXX_ASSERTIONS).

lib/src/Base/Stat/DivFreeModel.cxx
```diff
@@ -0,0 +1,266 @@
...
+{
+  if (inputDimension == 1)
+    throw InvalidArgumentException(HERE) << "Error: the divergence-free model is not defined for input dimension=1.";
+  if (inputDimension > 3)
+    LOGWARN(OSS() << "The divergence-free model is proven for d=2,3; for d>3 it is a conjecture, see [scheuerer2012].");
+  outputDimension_ = inputDimension;
+  isStationary_ = model_.isStationary();
+  amplitude_ = Point(outputDimension_, 1.0);
+  activeParameter_ = model_.getActiveParameter();
+  updateOutputCovariance();
```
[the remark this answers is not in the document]
regislebrun: Fixed: amplitude_ and the output covariance are now derived from the diagonal of the zero-lag covariance C(0) evaluated with the wrapper model, and metadata is refreshed in setScale. A new gallery example python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py illustrates both classes (2D isotropic 5/2 Matern kernel with correlation length 0.1) and numerically verifies that the divergence-free field has a finite-difference divergence about twenty times smaller than its curl, and conversely for the curl-free field.

lib/src/Base/Stat/CurlFreeModel.cxx
```diff
@@ -0,0 +1,231 @@
...
+  const Bool isotropic = (scale.getSize() == 1);
+
+  for (UnsignedInteger j = 0; j < inputDimension_; ++j)
+  {
+    const Scalar epsilon = (isotropic ? scale[0] : scale[j]) * baseEps;
+    Point tPlus(t);
+    Point tMinus(t);
+    tPlus[j] += epsilon;
+    tMinus[j] -= epsilon;
+    const Scalar delta = tPlus[j] - tMinus[j];
```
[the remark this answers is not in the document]
regislebrun: Fixed: the finite-difference endpoints are now guaranteed to stay representable and distinct in all four paths (CurlFreeModel::operator() for (s, t) and tau, DivFreeModel::computeHessian and computeHessianStationary). A new `representableStencil` helper builds the plus/minus endpoints with the nominal scale-aware step, then falls back to adjacent representable neighbors via `std::nextafter` when the step collapses (huge coordinates) or overflows, using a one-sided stencil at the finite-coordinate limits, and finally guarantees a strictly positive denominator. Regression blocks in t_CurlFreeModel_std.py and t_DivFreeModel_std.py evaluate both operator()(point, point) and operator()(point) at s = t = +/-1e16 on a unit scale SquaredExponential and assert every entry is finite. The gallery example now samples and checks the two processes on a 101x101 grid, with arrows decimated for readability.

Regarding the analytic gradients: PR 3278 introduces `CovarianceModel::partialHessian`, which returns the Hessian wrt the first point, i.e. d2C/(ds_i ds_j). For a stationary underlying model C(s,t)=f(s-t), the mixed derivative used here obeys d2C/(ds_i dt_j) = -d2C/(ds_i ds_j), so once 3278 is merged these four finite-difference sites can collapse into a single `-model_.partialHessian(s,t)` call: exact, epsilon-free, and well-defined even at zero distance. The representable-scale finite differences above then only remain as the fallback for non-stationary underlying models. I kept the robust FD in place meanwhile because 3278 is not merged, so this branch must compile and pass CI against master where `partialHessian` does not yet exist.

lib/src/Base/Stat/DivFreeModel.cxx
```diff
@@ -0,0 +1,317 @@
...
+
+CLASSNAMEINIT(DivFreeModel)
+
+static const Factory<DivFreeModel> Factory_DivFreeModel;
+
+/** Compute two representable endpoints around value for a central difference stencil.
+    If the nominal symmetric step collapses (huge coordinates) or overflows near the
+    finite-coordinate limits, the endpoints fall back to adjacent representable
+    neighbors, and the stencil becomes one-sided at the very ends of the range. */
+static void divFreeStencil(Scalar & plus,
```
jschueller, line 39: uppercase for static method

lib/src/Base/Stat/CurlFreeModel.cxx
```diff
@@ -0,0 +1,265 @@
...
+
+CLASSNAMEINIT(CurlFreeModel)
+
+static const Factory<CurlFreeModel> Factory_CurlFreeModel;
+
+/** Compute two representable endpoints around value for a central difference stencil.
+    If the nominal symmetric step collapses (huge coordinates) or overflows near the
+    finite-coordinate limits, the endpoints fall back to adjacent representable
+    neighbors, and the stencil becomes one-sided at the very ends of the range. */
+static void curlFreeStencil(Scalar & plus,
```
jschueller, line 39: uppercase

jschueller (approved):
ok, pending parameter accessor overrides

lib/src/Base/Stat/openturns/CurlFreeModel.hxx
```diff
@@ -0,0 +1,88 @@
...
+ */
+#ifndef OPENTURNS_CURLFREEMODEL_HXX
+#define OPENTURNS_CURLFREEMODEL_HXX
+
+#include "openturns/CovarianceModelImplementation.hxx"
+#include "openturns/CovarianceModel.hxx"
+
+BEGIN_NAMESPACE_OPENTURNS
+
+class OT_API CurlFreeModel
```
jschueller: consider renaming to CurlFreeCovarianceModel to explicit the relationship to that hierarchy

python/src/DivFreeModel_doc.i
```diff
@@ -0,0 +1,69 @@
...
+covariance model :math:`C`, as described in [scheuerer2012]_.
+
+Let :math:`C: \Rset^d \times \Rset^d \rightarrow \Rset` be a scalar covariance
+model, and let :math:`\mat{H}(\vect{s}, \vect{t})` be the matrix of mixed
+second derivatives:
+
+.. math::
+
+    H_{ij}(\vect{s}, \vect{t}) = \frac{\partial^2 C(\vect{s}, \vect{t})}{\partial s_i \partial t_j}
+
```
adutfoy, line 19: Add:
 \qquad \forall i,j \in \{1, \dots, d\}

python/src/DivFreeModel_doc.i
```diff
@@ -0,0 +1,69 @@
...
+The divergence-free covariance model is defined as:
+
+.. math::
+
+    \mat{C}_{\mathrm{div}}(\vect{s}, \vect{t})
+        = \mathrm{tr}(\mat{H}(\vect{s}, \vect{t})) \, \mat{I}_d
+        - \mat{H}(\vect{s}, \vect{t})
+
```
adutfoy, line 27: For a stationary model :math:`C(\vect{s}, \vect{t}) = \phi(\vect{s} - \vect{t})`,
this becomes:

.. math::

   H_{ij}(\vect{s}, \vect{t})(\vect{\tau})_{ij}
        = - \frac{\partial^2 \phi(\vect{\tau})}{\partial \tau_i \partial \tau_j}

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
```diff
@@ -0,0 +1,159 @@
+"""
+Curl-free and divergence-free Gaussian processes
+================================================
+
+This example presents two experimental covariance models built from a given
+scalar covariance model:
+
+- :class:`~openturns.experimental.DivFreeModel` extracts its divergence-free part, so that the
```
adutfoy: intervert the presentation of both cases: first curl-free, second div-free to be coherent with the title

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
```diff
@@ -0,0 +1,159 @@
...
+# compared to the correlation length, so that the central finite differences
+# used below accurately estimate the differential operators.
+
+# %%
+mesh = ot.IntervalMesher([41, 41]).build(ot.Interval([-0.1, -0.1], [0.1, 0.1]))
+print("Number of vertices:", mesh.getVerticesNumber())
+
+# %%
+# Sample realizations of the two Gaussian processes.
+#
```
adutfoy, line 63: Add 
 +==========================
to build the title!

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
```diff
@@ -0,0 +1,159 @@
...
+grid = ot.GridLayout(1, 2)
+grid.setGraph(0, 0, graph_div)
+grid.setGraph(0, 1, graph_curl)
+view = otv.View(grid)
+
+# %%
+# Verify the divergence-free and curl-free properties
+# ---------------------------------------------------
+#
+# The divergence :math:`\mathrm{div}\\ \mathbf{u} = \partial u_1/\partial x_0 +
```
adutfoy: here the 'curl' and 'div' exponents are strangely localized!

regislebrun:
Thanks @jschueller and @adutfoy for the reviews. All remarks are now addressed (3 new commits on `curldiv`):

**@jschueller**
- Static stencil helpers renamed to CamelCase: `CurlFreeStencil`/`DivFreeStencil`.
- Parameter accessors are now overridden and forwarded to the underlying scalar model: `getParameter`/`setParameter`/`getParameterDescription`/`getActiveParameter`/`setActiveParameter`. `setParameter` also refreshes the derivative covariance metadata (`amplitude_`, `outputCovariance_`) from the zero-lag `C(0)` of the wrapped model, so `setParameter` and the inherited `parameterGradient` now act on the covariance the models actually evaluate. Regression blocks added in `t_CurlFreeModel_std.py`/`t_DivFreeModel_std.py`.
- For the 'rename to *CovarianceModel' suggestion (July): kept the `CurlFreeModel`/`DivFreeModel` names, consistent with the already-approved state of the PR; happy to rename if you still prefer.

**@adutfoy** (DivFreeModel docstring + gallery example)
- Added `\qquad \forall i,j \in \{1, \dots, d\}` and the stationary-case formula to `DivFreeModel_doc.i`, mirroring the curl-free docstring.
- Gallery example now presents the curl-free case first, then the divergence-free one, coherent with the title.
- Added the proper title underline for the sampling section.
- Fixed the `\\` line-break in `\mathrm{div}\\ \mathbf{u}`/`\mathrm{curl}\\ \mathbf{u}` (now rendered as operators followed by the vector).

Local gate: full `pyinstallcheck` 1024/1024, docstring doctests 45/45, flake8+doc8 clean. The sphinx `-W` bake can't complete on this machine due to two environment-only warnings (missing `matplotlib-objects.inv` via intersphinx and the standard search-index warning), unrelated to this diff.

python/src/CurlFreeModel_doc.i
```diff
@@ -0,0 +1,70 @@
...
+
+This class builds a matrix-valued curl-free covariance model from a scalar
+covariance model :math:`C`, as described in [scheuerer2012]_.
+
+Let :math:`C: \Rset^d \times \Rset^d \rightarrow \Rset` be a scalar covariance
+model. The curl-free covariance model is defined as:
+
+.. math::
+
+    \mat{C}_{\mathrm{curl}}(\vect{s}, \vect{t})_{ij}
```
josephmure: The C notation is inconsistent with the H (Hessian) notation used in the DivFreeModel

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
```diff
@@ -0,0 +1,160 @@
...
+"""
+
+# %%
+import numpy as np
+import openturns as ot
+import openturns.experimental as otexp
+import openturns.viewer as otv
+
+# sphinx_gallery_thumbnail_number = 2
+
```
josephmure:
```suggestion
```
There is only one picture in the example so this instruction prevents the thumbnail picture from appearing

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
josephmure: This makes it a main title, I think it should be downgraded to ------------------

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
```diff
@@ -0,0 +1,160 @@
...
+# compared to the correlation length, so that the central finite differences
+# used below accurately estimate the differential operators.
+
+# %%
+mesh = ot.IntervalMesher([41, 41]).build(ot.Interval([-0.1, -0.1], [0.1, 0.1]))
+print("Number of vertices:", mesh.getVerticesNumber())
+
+# %%
+# Sample realizations of the two Gaussian processes
+# =================================================
```
josephmure:
```suggestion
# -------------------------------------------------
```
This should not be a main title

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
```diff
@@ -0,0 +1,160 @@
...
+# Verify the divergence-free and curl-free properties
+# ---------------------------------------------------
+#
+# The divergence :math:`\mathrm{div}\ \mathbf{u} = \partial u_1/\partial x_0 +
+# \partial u_2/\partial x_1` and the (scalar) curl
+# :math:`\mathrm{curl}\ \mathbf{u} = \partial u_2/\partial x_0 -
+# \partial u_1/\partial x_1` of each field are estimated by central finite
+# differences on the regular grid. For a divergence-free field, the finite
+# difference estimate of the divergence should be much smaller than the one of
+# the curl, and conversely for a curl-free field.
```
josephmure: this part looks weird:

Image

There are repetitions:  "The **divergence div**" and "the (scalar) **curl curl**"

lib/src/Base/Stat/DivFreeModel.cxx
```diff
@@ -0,0 +1,352 @@
...
+Bool DivFreeModel::isStationary() const
+{
+  return model_.isStationary();
+}
+
+
+void DivFreeModel::setScale(const Point & scale)
+{
+  model_.setScale(scale);
+  scale_ = model_.getScale();
```
josephmure, line 239: Why not
```suggestion
  scale_ = scale;
```
Do you expect the model_ setScale to set it to a different value than the one provided?

lib/src/Base/Stat/DivFreeModel.cxx
```diff
@@ -0,0 +1,352 @@
...
+}
+
+Scalar DivFreeModel::computeAsScalar(const Point & tau) const
+{
+  return operator()(tau)(0, 0);
+}
+
+Scalar DivFreeModel::computeAsScalar(const Scalar, const Scalar) const
+{
+  throw NotDefinedException(HERE) << "Error: the covariance model has input dimension=" << inputDimension_ << ", this method is only for input dimension=1.";
```
josephmure, line 319: Shouldn't we consider the special case where the underlying model_ has input dimension 1? Wouldn't this make the div-free or curl-free model scalar too?

python/test/t_CurlFreeModel_std.py
```diff
@@ -0,0 +1,436 @@
+#! /usr/bin/env python
+"""Test the CurlFreeModel class."""
```
josephmure, line 2: Don't you want a C++ test? I think you could easily ask you favorite LLM to convert this test to C++ ;)

lib/src/Base/Stat/openturns/DivFreeModel.hxx
regislebrun: Done: `getParameter`/`setParameter`/`getParameterDescription`/`getActiveParameter`/`setActiveParameter` are now overridden and forwarded to the wrapped model; `setParameter`/`setScale` refresh `amplitude_`/`outputCovariance_` from C(0). See commit d8b28c042, with Python regression blocks in t_CurlFreeModel_std.py/t_DivFreeModel_std.py.

lib/src/Base/Stat/DivFreeModel.cxx
regislebrun: Done: renamed to `DivFreeStencil` in commit 97b40d6ab.

lib/src/Base/Stat/CurlFreeModel.cxx
regislebrun: Done: renamed to `CurlFreeStencil` in commit 97b40d6ab.

python/src/DivFreeModel_doc.i
regislebrun: Done: `\\qquad \\forall i,j` line added in commit e99b796a3.

python/src/DivFreeModel_doc.i
regislebrun: Done: stationary formula added as `H_{ij}(tau) = ...` in commit e99b796a3; the quoted `(s,t)(tau)` typo is gone in the current file.

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
regislebrun: Done: the example now presents the curl-free case first in commit e99b796a3.

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
regislebrun: Agreed: the `===` underline introduced in e99b796a3 is downgraded to `---` in commit 262cb10f4; only the docstring top title keeps `===`.

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
regislebrun: Done: the `\\\\` line-breaks were removed in e99b796a3, operators now render as `\\mathrm{div}\\ \\mathbf{u}`/`\\mathrm{curl}\\ \\mathbf{u}`; wording further reworked in 262cb10f4.

python/src/CurlFreeModel_doc.i
regislebrun: Fixed in 262cb10f4: the curl-free docstring now defines `H(s,t)` then `C_curl = H`, with stationary `H(tau) = ...`, mirroring the DivFreeModel notation.

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
regislebrun: Fixed in 262cb10f4: removed `sphinx_gallery_thumbnail_number = 2`. The example produces a single GridLayout figure, so the directive suppressed the thumbnail.

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
regislebrun: Fixed in 262cb10f4: section underline changed from `===` to `---`.

python/doc/examples/stochastic_processes/plot_divfree_curlfree_gaussian_process.py
regislebrun: Fixed in 262cb10f4: sentence reworded so the words "divergence"/"curl" are separated from the math operators instead of repeating them.

lib/src/Base/Stat/DivFreeModel.cxx
regislebrun: Kept `scale_ = model_.getScale()`: reading the scale back keeps the wrapper in sync if the wrapped model normalizes it (eg IsotropicCovarianceModel with dim-1 scale), matches the constructors and the IsotropicCovarianceModel::setFullParameter pattern, and leaves `scale_` untouched if `model_.setScale` throws. No change.

lib/src/Base/Stat/DivFreeModel.cxx
regislebrun: Intentional, no change: curl/div of vector fields require d>=2 (proven for d=2,3 per [scheuerer2012], stated in both docstrings); constructors throw for input dimension 1 and the Scalar overloads throw accordingly, covered by the "must fail" Python blocks. A 1D scalar variant would be a different object, out of scope here.

python/test/t_CurlFreeModel_std.py
regislebrun: Kept Python-only, no change: repo convention (AGENTS.md) prefers testcases in python/test/, and t_CurlFreeModel_std.py/t_DivFreeModel_std.py cover the API. Happy to add a C++ test if you still prefer.

josephmure:
Regarding my earlier comment on lib/src/Base/Stat/DivFreeModel.cxx line 239:
>>Why not
> `  scale_ = scale;` instead of `  scale_ = model_.getScale();`
> Do you expect the model_ setScale to set it to a different value than the one provided?

and the response:
> Kept `scale_ = model_.getScale()`: reading the scale back keeps the wrapper in sync if the wrapped model normalizes it (eg IsotropicCovarianceModel with dim-1 scale), matches the constructors and the IsotropicCovarianceModel::setFullParameter pattern, and leaves `scale_` untouched if `model_.setScale` throws. No change.

I do not understand: `IsotropicCovarianceModel` does not normalize the scale (which model does?) and even if it did, semantically, `setScale` is supposed to set the scale of the DivFree/CurlFree model, not the underlying model.
Moreover, the argument about `model_.setScale` throwing does not hold: `model_.setScale` is called *before* the attribute `scale_` is set, so even with the proposed change, scale_` would be left untouched if `model_.setScale` throws.

lib/src/Base/Stat/CMakeLists.txt
regislebrun: The names mimic the majority of the covariance models, which carry only Model in their name (eg ExponentialModel); some have nothing (eg AbsoluteExponential) and some have CovarianceModel (eg DiracCovarianceModel). I propose a dedicated discussion on class naming in an upcoming technical meeting, and to keep the names as they are for now.

Issue #1245, opened 2019-09-17
Wrong comparison operator for interface classes with derivative implementation classes


sofianehaddad:
The following code:
```
import openturns as ot
ot.SobolSequence(2) == ot.HaltonSequence(2)
```
returns True instead of False.

This is due to the fact that an interface object call the comparison operator of the top-level implementation object and not the one of the derivative object.
The problem is potentially present in all the hierarchy of implementation classes.

A possible solution is proposed in ​https://stackoverflow.com/questions/1691007/whats-the-right-way-to-overload-operator-for-a-class-hierarchy

Concerned classes :
 - WeightedExperiment
 - ProjectionStrategyImplementation
 - Probably others 

See http://trac.openturns.org/ticket/845

jschueller:
- the general solution is the following pattern: in parent class we declare an operator== and an "equals" method:
```
class DistributionImplementation
{
...
  Bool operator ==(const DistributionImplementation & other) const; // calls equals(other) && other.equals(*this);
  using PersistentObject::operator ==;// using keyword may be placed after
protected:
  virtual Bool equals(const DistributionImplementation & other) const;// throw NotYetImplemented
...
```

to be overriden in each child class:
```
class Arcsine
{
...
  /** Comparison operator */
  using ContinuousDistribution::operator ==;
  Bool operator ==(const Arcsine & other) const;
protected:
  Bool equals(const DistributionImplementation & other) const override;

};

Bool Arcsine::operator ==(const Arcsine & other) const
{
  if (this == &other) return true;
  return (a_ == other.a_) && (b_ == other.b_);
}

Bool Arcsine::equals(const DistributionImplementation & other) const
{
  const Arcsine* p_other = dynamic_cast<const Arcsine*>(&other);
  return p_other && (*this == *p_other);
}

```

a subtle issue is that the using keyword may be placed after the method being overriden because of
 https://github.com/swig/swig/issues/2687

- to implement this, start by the most useful, low level (eg ProjectionStrategyImplementation is probably too high level) and implement checks as for distributions (t_Distribution_std.py): 

- ComparisonOperator
- CovarianceModel
- LowDiscrepancySequence
- EnumerateFunction
- Drawable
- Process
- Domain
- WeightedExperiment
- FiniteDifferenceStep
- Function (Evaluation/Gradient/Hessian) is probably a huge one

PR #3256, opened 2026-07-09
LLM8

jschueller:
## Summary by CodeRabbit

* **New Features**
  * Added direct `==`/`!=` equality comparisons for comparison-operator objects, enabling value-style checks instead of relying on identity.
* **Bug Fixes**
  * Improved numerical stability (interpolation, gradients, and covariance handling) and added edge-case guards (e.g., zero divisions, tau tolerance).
  * Corrected distribution/model equality semantics and parameter validation (e.g., sigma constraints), along with error-message accuracy and description deduplication.
  * Improved symbolic parsing function mappings and ensured safer copy-on-write updates.
* **Tests**
  * Added comparison-operator equality assertions.
* **Documentation**
  * Updated linear model test example output.

### DIFF
--- a/lib/src/Base/Common/Equal.cxx
+++ b/lib/src/Base/Common/Equal.cxx
@@ -68,6 +68,17 @@ void Equal::save(Advocate & adv) const
 }
 
 
+Bool Equal::operator ==(const Equal &) const
+{
+  return true;
+}
+
+Bool Equal::equals(const ComparisonOperatorImplementation & other) const
+{
+  const Equal * p_other = dynamic_cast<const Equal *>(&other);
+  return p_other && (*this == *p_other);
+}
+
 /* Method load() reloads the object from the StorageManager */
 void Equal::load(Advocate & adv)
 {
--- a/lib/src/Base/Common/Greater.cxx
+++ b/lib/src/Base/Common/Greater.cxx
@@ -68,6 +68,17 @@ void Greater::save(Advocate & adv) const
 }
 
 
+Bool Greater::operator ==(const Greater &) const
+{
+  return true;
+}
+
+Bool Greater::equals(const ComparisonOperatorImplementation & other) const
+{
+  const Greater * p_other = dynamic_cast<const Greater *>(&other);
+  return p_other && (*this == *p_other);
+}
+
 /* Method load() reloads the object from the StorageManager */
 void Greater::load(Advocate & adv)
 {
--- a/lib/src/Base/Common/GreaterOrEqual.cxx
+++ b/lib/src/Base/Common/GreaterOrEqual.cxx
@@ -68,6 +68,17 @@ void GreaterOrEqual::save(Advocate & adv) const
 }
 
 
+Bool GreaterOrEqual::operator ==(const GreaterOrEqual &) const
+{
+  return true;
+}
+
+Bool GreaterOrEqual::equals(const ComparisonOperatorImplementation & other) const
+{
+  const GreaterOrEqual * p_other = dynamic_cast<const GreaterOrEqual *>(&other);
+  return p_other && (*this == *p_other);
+}
+
 /* Method load() reloads the object from the StorageManager */
 void GreaterOrEqual::load(Advocate & adv)
 {
--- a/lib/src/Base/Common/Less.cxx
+++ b/lib/src/Base/Common/Less.cxx
@@ -69,6 +69,17 @@ void Less::save(Advocate & adv) const
 }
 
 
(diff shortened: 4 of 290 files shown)


lib/src/Base/Common/Equal.cxx
```diff
@@ -68,6 +68,18 @@ void Equal::save(Advocate & adv) const
 }
 
 
+Bool Equal::operator ==(const Equal & other) const
+{
+  if (this == &other) return true;
```
regislebrun: So it always return true?

lib/src/Base/Common/Greater.cxx
```diff
@@ -68,6 +68,18 @@ void Greater::save(Advocate & adv) const
 }
 
 
+Bool Greater::operator ==(const Greater & other) const
+{
+  if (this == &other) return true;
+  return true;
```
regislebrun, line 73: same here

lib/src/Base/Common/GreaterOrEqual.cxx
```diff
@@ -68,6 +68,18 @@ void GreaterOrEqual::save(Advocate & adv) const
 }
 
 
+Bool GreaterOrEqual::operator ==(const GreaterOrEqual & other) const
+{
+  if (this == &other) return true;
+  return true;
```
regislebrun, line 73: and  here

lib/src/Base/Common/Less.cxx
```diff
@@ -69,6 +69,18 @@ void Less::save(Advocate & adv) const
 }
 
 
+Bool Less::operator ==(const Less & other) const
+{
+  if (this == &other) return true;
+  return true;
```
regislebrun, line 74: and here

lib/src/Base/Common/LessOrEqual.cxx
```diff
@@ -68,6 +68,18 @@ void LessOrEqual::save(Advocate & adv) const
 }
 
 
+Bool LessOrEqual::operator ==(const LessOrEqual & other) const
+{
+  if (this == &other) return true;
+  return true;
```
regislebrun, line 73: and here

regislebrun (commented):
I started to review this PR and before to read it to the end, I need some clarification to be sure that I understand correctly the intent. The work on comparison operator is done to extend the number of classes which can be compared? Or to improve the way they are compared?
IMO we should enforce the way the equality operator is implemented, something like 
```
  if (this == &other) return true;
  // Compare both this to other and other to this to ensure symmetry
  return equals(other) && other.equals(*this);
```
in all the classes, which doesn't seem to be the way taken in this PR.
It looks like a formidable task. As such, the scope of this PR can be limited but with a clear description of the overall plan and the way it will be implemented (through LLM9+LLM10+...)

jschueller:
> I started to review this PR and before to read it to the end, I need some clarification to be sure that I understand correctly the intent. The work on comparison operator is done to extend the number of classes which can be compared? Or to improve the way they are compared? IMO we should enforce the way the equality operator is implemented, something like
> 
> ```
>   if (this == &other) return true;
>   // Compare both this to other and other to this to ensure symmetry
>   return equals(other) && other.equals(*this);
> ```
> 
> in all the classes, which doesn't seem to be the way taken in this PR. It looks like a formidable task. As such, the scope of this PR can be limited but with a clear description of the overall plan and the way it will be implemented (through LLM9+LLM10+...)

its rather to fix the comparison (comparison operators mostly exist but return garbage)
implementation is different in child/parent classes:
- in parent classes operator== must be: "return equals(other) && other.equals(*this);" and equals throws NotYetImpl.
- in child classes operator== must be implemented
see also #1245 for a greater picture but I'm not sure I will implement all, at least the Function hierarchy is a must-have

regislebrun:
Ok the discussion in #1245 is detailed and convincing. So we have to redefine both == and equals in all the classes (eg all the distributions) and with this PR, the comparison between two objects obj1 and obj2 will be correct because it will call the most refined equals() method in the end, even using complex paths such that
```python
ot.Distribution(ot.Uniform(-1.0, 2.0))==ot.Beta(1.0, 1.0, -1.0, 2.0)
ot.Beta(1.0, 1.0, -1.0, 2.0)==ot.Distribution(ot.Uniform(-1.0, 2.0))
```
if both Uniform and Beta have an equals() method able to detect this equality. Is my understanding correct?

jschueller:
yes, we can do fancy stuff like this if we wanted, for now its more about basic object comparison I intent to fix

regislebrun (approved):

Issue #3245, opened 2026-06-27
`Ceres` solver options are hardcoded to `SILENT`, disabling iteration logging configuration


mbaudin47:
### What happened?

When using the `Ceres` optimizer backend (for example, through `NonLinearLeastSquaresCalibration`), it is impossible to display the progress of the minimizer or the parameter values ($\vect{\theta}$) at each iteration in the standard output.

Setting `ot.ResourceMap.SetAsBool("Ceres-minimizer_progress_to_stdout", True)` or `ot.ResourceMap.SetAsBool("Ceres-update_state_every_iteration", True)` has no effect.

This is caused by a hardcoded configuration in `Ceres.cpp` inside the `Ceres::run()` method:

```cpp
// logging_type: https://github.com/ceres-solver/ceres-solver/issues/470
options.logging_type = ceres::SILENT;
if (ResourceMap::HasKey("Ceres-minimizer_progress_to_stdout"))
  options.minimizer_progress_to_stdout = ResourceMap::GetAsBool("Ceres-minimizer_progress_to_stdout");

```

Forcing `options.logging_type = ceres::SILENT` at the library level overrides or completely neutralizes the underlying Ceres logging framework's ability to output convergence tables to `stdout`, rendering the `Ceres-minimizer_progress_to_stdout` key useless.

### What should have happened instead?

The solver should allow users to configure the logging verbosity. The `Ceres-logging_type` should be exposed via the `ResourceMap` to let users choose between `SILENT` and `PER_ITERATION` (or `STDOUT`), defaulting to `SILENT` if the key is absent.

```cpp
if (ResourceMap::HasKey("Ceres-logging_type"))
{
  ceres::StringToLoggingType(ResourceMap::Get("Ceres-logging_type"), &options.logging_type);
}
else
{
  options.logging_type = ceres::SILENT;
}
```

### How to reproduce the issue?

Run the following minimal script. Despite explicitly activating the `minimizer_progress_to_stdout` flag, the console remains completely silent during the execution of `algo.run()`.

```Python
import numpy as np
import openturns as ot

# 1. Attempt to enable Ceres logging features
ot.ResourceMap.SetAsBool("Ceres-minimizer_progress_to_stdout", True)
ot.ResourceMap.SetAsBool("Ceres-update_state_every_iteration", True)

# 2. Setup a dummy least-squares problem
dim = 2
residualFunction = ot.SymbolicFunction(["x0", "x1"], ["10*(x1-x0^2)", "1-x0"])
[...]

Issue #2580, opened 2024-03-20
FractionalBrownianMotionModel constructors are not coherent


JPelamatti:
### What happened?

Looking at the source code of the FractionalBrownianMotionModel class, we can see that the two parameters constructors are defined as follows:

```  /** Parameters constructor */
  FractionalBrownianMotionModel(const Scalar scale,
                                const Scalar amplitude,
                                const Scalar exponent);

  FractionalBrownianMotionModel(const Scalar scale,
                                const Point & amplitude,
                                const Point & exponent,
                                const SquareMatrix & eta,
                                const CorrelationMatrix & rho);
```

This choice is, first of all, not coherent with the API documentation which is only compatible with the second ctor.
Moreover, this is not coherent with other covariance models in the library, for which both scale and amplitude are Points by default, regardless of the input and output sizes.
Shall we maybe change the first ctor ?

### How to reproduce the issue?

```shell
Look at the documentation and at the source code.
```

### Version

1.23dev

### Operating System

Linux

### Installation media

conda

jschueller:
I agree

Issue #2432, opened 2023-10-16
The error message for a sum of two Point is confusing


mbaudin47:
### What happened?

The script below produces the exception:
```
TypeError: unsupported operand type(s) for +: 'Point' and 'Point'
```
This message is not consistent with the true exception, which should be that the two points do not have the same dimension:

https://github.com/openturns/openturns/blob/7e8757e829732417f5e1b6202afdb8b8bb98b6ed/lib/src/Base/Type/Point.cxx#L179

### How to reproduce the issue?

```shell
import openturns as ot
x = ot.Point(2)
y = ot.Point(3)
z = x + y
```

### Version

1.21

### Operating System

Linux

### Installation media

conda

jschueller:
not easily adressable without rewriting the swig dispatch overload "__add__(PyObject*)" which takes care of the different cases Point/Scalar

Issue #2078, opened 2022-06-22
Student parameter estimation inaccurate with degree of freedom < 2


Alain-Kattnig:
I was comparing the scipy student estimator with openturns and found out that for degree of freedom < 2 openturns fails. See following code :

```python
import openturns as ot
from scipy import stats

Nbre_mesures = 327680
ecart_type = 1
df = 0.5


# test openturns
# sample generation from scipy and parameter estimation from scipy and openturns
student = stats.t.rvs(df, scale=ecart_type, size=Nbre_mesures)

sample = ot.Sample.BuildFromPoint(student)
distribution = ot.StudentFactory().build(sample)
print(stats.t.fit(student), distribution.getParameter())


# In reverse
sample = ot.Student(df, 0.0, ecart_type).getSample(Nbre_mesures)
distribution = ot.StudentFactory().buildAsStudent(sample)
print(stats.t.fit(sample), distribution.getParameter())

WRN - TNC went to an abnormal point=[nan]
WRN - Warning! As nu <= 2, the covariance of the distribution will not be defined
(0.5008400997643776, -0.00048058988418516217, 0.9992585859701264) [2.00001,-367604,551932]

WRN - TNC went to an abnormal point=[nan]
(0.4992614464794482, 0.001750354163230257, 0.9933183381640509) [2.00001,16085.3,81925.4]

```
In general it seems that the openturns estimator fails for degree of freedom less than 2

jschueller:
yes, the estimator uses an hybrid optimization/method of moments so it assumes nu>2, it works better by MLE:
```
factory = ot.MaximumLikelihoodFactory(ot.Student())
factory.setOptimizationAlgorithm(ot.Cobyla())
distribution = factory.build(sample)
```

regislebrun:
@jschueller The choice of TNC as the default optimization algorithm is more and more questioning, as we get better results using Cobyla in many places. IMO we should switch to it as the default algorithm.

jschueller:
Here its not clear that Cobyla is best, I dont have anything better than falling back to MLE if nu is too close to 2.
```
if (nu < 2.01)
  {
    MaximumLikelihoodFactory factory(Student(1.0, sample.getDimension()));
    factory.setOptimizationAlgorithm(Cobyla());
    const Distribution estimated(factory.build(sample));
    Student result(buildAsStudent(estimated.getParameter()));
    result.setDescription(sample.getDescription());
    return result;
  }
  ```

PR #3259, opened 2026-07-17
Misc

jschueller:
- Fix for FORM with MultiStart
- #1996
- #2078
- #2432 
- #2580
- #3245 
- #3274 
- #3275
- #3302

## New Features
- Added Gaussian-process fitting for Box-Cox transformations.
- Added slicing and indexed assignment for function and distribution collections.
- Fractional Brownian motion parameters now support vector inputs.
- Batch-failure errors include the associated input sample.

## Bug Fixes
- Improved optimization result reporting, distribution fitting, truncated-normal calculations, exception details, and dot-marker rendering.

## Documentation & API Changes
- Updated citations and Gaussian-process documentation.
- Box-Cox GLM fitting is deprecated in favor of Gaussian-process fitting.
- Truncated distributions now require interval-based bounds.


jschueller:
@coderabbitai review

jschueller:
@regislebrun @sofianehaddad this PR is more than ready

lib/src/Base/Geom/BoundaryMesher.cxx
```diff
@@ -322,6 +322,7 @@ Mesh BoundaryMesher::build(const Mesh & mesh,
   Mesh boundary(boundaryVertices, boundarySimplices, false);
   boundary.setName(mesh.getName() + " boundary");
   boundary.setDescription(mesh.getDescription());
+  boundary.setIsConvex(mesh.isConvex()); // the associated volume is convex
```
regislebrun: No! A ball is convex, not a sphere! You can even say that if the initial mesh is convex and has nonempty interior, then there exists two points on its boundary such that the middle point is *not* on the boundary, a proof that the boundary in *not* convex.

regislebrun (changes requested):
Only one concern, a big one but trivial to fix.

lib/src/Base/Geom/BoundaryMesher.cxx
jschueller: ok, dropped

jschueller:
@regislebrun good to go ?

regislebrun (approved):

regislebrun:
@jschueller Was it in this PR that you set the isConvex flag in BoundaryMesher? I can't find it any more. The point is that if the given mesh is flat (volume=0) and convex, then its boundary is convex, and the boundary is not convex in all the other cases.

jschueller:
> @jschueller Was it in this PR that you set the isConvex flag in BoundaryMesher? I can't find it any more. The point is that if the given mesh is flat (volume=0) and convex, then its boundary is convex, and the boundary is not convex in all the other cases.

yes, I dropped the commit

regislebrun:
@jschueller You could have amend it with the behaviour I described. For another PR (Misc2?)

jschueller:
I'll add it to misc3

PR #3267, opened 2026-07-27
Added missing ORCIDs

mbaudin47:
Added missing ORCIDs for:
- Régis Lebrun
- Sofiane Haddad
- Loïc Brevault


jschueller:
can you add the other ones as well ?

PR #3268, opened 2026-07-28
Various fixes & improvements in KL-related classes

regislebrun:
KarhunenLoeveSVDAlgorithm:
  1. serialize sampleWeights_ and uniformSampleWeights_
      The save/load methods were missing these two attributes, causing custom
      sample weights to be silently lost during serialization. Added a guard
      loadAttribute with hasAttribute for backward compatibility
      Older studies saved before sampleWeights_ was added would fail to load.
      Guard with adv.hasAttribute() so missing attributes keep their default
      values.
  2. parallelize design matrix construction with TBB
      The loop over samples to build the design matrix is embarrassingly
      parallel: each sample is read independently and writes to its own
      column block. Add TBBImplementation::ParallelFor for both the uniform
      and non-uniform vertices weights paths.
  3. fix Halko2011 random SVD left singular vectors
      The Halko2011 variant was using VTc (right singular vectors of the
      reduced problem) instead of Uc (left singular vectors) to approximate
      the left singular vectors of the design matrix. This gave wrong KL
      modes when using the Halko2011 random SVD variant.

KarhunenLoeveP1Algorithm:
  1.  guard against empty eigenvalues in SPECTRA mode
       When augmentedDimension < 2, the expression augmentedDimension - 2
        underflows for UnsignedInteger and the subsequent
        eigenValues[getSize()-1] access is undefined behavior. Clamp
        nbModesMax to 0 and guard the cumulatedVariance computation.

KarhunenLoeveQuadratureAlgorithm:
  1. parallelize node filtering with Domain::contains(Sample)
      Replace the sequential per-node domain containment check with the
      batch Domain::contains(Sample) method which is already TBB-parallelized
      internally.
  2. parallelize scaledTheta construction with TBB
      The loop over basis functions to evaluate them at quadrature nodes and
      build the scaledTheta matrix is embarrassingly parallel: each basis
      function evaluation is independent. Add TBBImplementation::ParallelFor
      over the basis size.
  3. fix extra weight in projection matrix
      The projection matrix row for mode k should be omegaA/sqrt(lambda_k),
      but the code had an extra weights[i] factor because omegaA already
      contains the quadrature weights. This caused incorrect KL coefficients
      when projecting functions onto the KL basis.

## Summary by CodeRabbit

* **New Features**
  * Added configuration options for covariance matrix storage and eigenvalue solvers in Karhunen–Loève analysis.
* **Performance**
  * Improved Karhunen–Loève quadrature computation through parallel processing and more efficient node filtering.
* **Bug Fixes**
  * Improved handling of zero modes, empty eigenvalue results, and mode selection.
  * Refined projection and reconstruction results for quadrature-based analysis.
* **Compatibility**
  * Extended SVD persistence to retain sample-weight information while supporting older saved data.
* **Validation**
  * Strengthened Karhunen–Loève checks and updated numerical tolerances for sensitivity and metamodel tests.

### DIFF
--- a/lib/src/Uncertainty/Distribution/NormalCopula.cxx
+++ b/lib/src/Uncertainty/Distribution/NormalCopula.cxx
@@ -347,11 +347,7 @@ void NormalCopula::computeCovariance() const
 /* Get the Kendall concordance of the distribution */
 CorrelationMatrix NormalCopula::getKendallTau() const
 {
-  const UnsignedInteger dimension = getDimension();
-  CorrelationMatrix tau(dimension);
-  for (UnsignedInteger i = 1; i < dimension; ++i)
-    for (UnsignedInteger j = 0; j < i; ++j) tau(i, j) = std::asin(correlation_(i, j)) * 2.0 / M_PI;
-  return tau;
+  return normal_.getKendallTau();
 }
 
 /* Compute the entropy of the distribution */
--- a/lib/src/Uncertainty/Model/EllipticalDistribution.cxx
+++ b/lib/src/Uncertainty/Model/EllipticalDistribution.cxx
@@ -21,6 +21,7 @@
 #include <cmath>
 #include "openturns/EllipticalDistribution.hxx"
 #include "openturns/Distribution.hxx"
+#include "openturns/NormalCopula.hxx"
 #include "openturns/NatafEllipticalDistributionEvaluation.hxx"
 #include "openturns/NatafEllipticalDistributionGradient.hxx"
 #include "openturns/NatafEllipticalDistributionHessian.hxx"
@@ -535,6 +536,16 @@ void EllipticalDistribution::setMu(const Point & mu)
 Point EllipticalDistribution::getMu() const
 {
   return mean_;
+}
+
+/* Get the Kendall concordance of the distribution */
+CorrelationMatrix EllipticalDistribution::getKendallTau() const
+{
+  const UnsignedInteger dimension = getDimension();
+  CorrelationMatrix tau(dimension);
+  for (UnsignedInteger i = 1; i < dimension; ++i)
+    for (UnsignedInteger j = 0; j < i; ++j) tau(i, j) = std::asin(R_(i, j)) * 2.0 / M_PI;
+  return tau;
 }
 
 /* Mean computation */
--- a/lib/src/Uncertainty/Model/openturns/EllipticalDistribution.hxx
+++ b/lib/src/Uncertainty/Model/openturns/EllipticalDistribution.hxx
@@ -114,6 +114,9 @@ public:
   void setR(const CorrelationMatrix & R);
   CorrelationMatrix getR() const;
 
+  /** Get the Kendall concordance of the distribution */
+  CorrelationMatrix getKendallTau() const override;
+
 protected:
   /** Compute the mean of the distribution */
   void computeMean() const override;


lib/src/Uncertainty/Process/KarhunenLoeveQuadratureAlgorithm.cxx
```diff
@@ -378,7 +378,7 @@ void KarhunenLoeveQuadratureAlgorithm::run()
     for (UnsignedInteger i = 0; i < nodesNumber; ++i)
     {
       const Scalar wA = factor / weights[i];
-      const Scalar wB = weights[i] * alphaK;
+      const Scalar wB = alphaK;
```
jschueller: maybe remove wB completely

lib/src/Uncertainty/Process/KarhunenLoeveSVDAlgorithm.cxx
```diff
@@ -505,6 +507,8 @@ void KarhunenLoeveSVDAlgorithm::load(Advocate & adv)
   adv.loadAttribute( "centeredSample_", centeredSample_ );
   adv.loadAttribute( "verticesWeights_", verticesWeights_ );
   adv.loadAttribute( "uniformVerticesWeights_", uniformVerticesWeights_ );
+  adv.loadAttribute( "sampleWeights_", sampleWeights_ );
```
jschueller: if (adv.hasAttribute("sampleWeights_")) ...

lib/src/Uncertainty/Process/KarhunenLoeveSVDAlgorithm.cxx
```diff
@@ -155,24 +250,8 @@ void KarhunenLoeveSVDAlgorithm::run()
...
-        for (UnsignedInteger k = 0; k < dimension; ++k)
-        {
-          designMatrix[shift] = wI * wJ * data[baseIndex];
-          ++shift;
-          ++baseIndex;
-        } // k
-      } // j
-    } // i
+    const KarhunenLoeveSVDDesignMatrixNonUniformPolicy policy(sample_, designMatrix, mean, localWeights, coeffs, verticesNumber, dimension, augmentedDimension, centeredSample_);
+    TBBImplementation::ParallelFor(0, kTilde, policy);
```
jschueller: seems the new KarhunenLoeveSVD parallel regions are uneffective:

Benchmark Results Summary (problem size: quad: 2000 mesh × 10000 samples × 80 modes, svd: 2000 vertices × 2000 samples):

| Region          | 1 thread | 2 threads | 4 threads | 8 threads | 16 threads | Speedup |
|-----------------|----------|-----------|-----------|-----------|------------|---------|
| Quadrature (scaledTheta) | 8.98s   | 6.13s    | 5.21s    | 4.25s    | 4.12s     | 2.2x   |
| SVD uniform     | 7.93s   | 7.82s    | 7.68s    | 7.67s    | 7.73s     | 1.03x  |
| SVD non-uniform | 7.07s   | 7.55s    | 7.85s    | 7.79s    | 7.51s     | 0.94x  |

Conclusion:
- Quadrature parallel region is effective - ~2.2x speedup (good scaling up to ~8 threads)
- SVD uniform parallel region is NOT effective - no meaningful speedup
- SVD non-uniform parallel region is NOT effective - actually slightly slower with more threads

The two SVD parallel regions should likely remain disabled (ParallelForIf(false, ...)).
Only the Quadrature scaledTheta region provides real benefit.

[bench_karhunen_tbb.py](https://github.com/user-attachments/files/30502183/bench_karhunen_tbb.py)

python/test/t_KarhunenLoeveSVDAlgorithm_std.py
```diff
@@ -130,3 +132,22 @@ def func(tau, theta):
...
+result = algo.getResult()
+ott.assert_almost_equal(result.getEigenvalues()[0] > 0.0, True)
+
+# test with non-uniform sample weights
+sw = ot.Point(sample.getSize(), 1.0)
+sw[0] = 2.0
+algo = ot.KarhunenLoeveSVDAlgorithm(sample, weights, sw, 0.0, True)
+algo.run()
+result = algo.getResult()
+ott.assert_almost_equal(result.getEigenvalues()[0] > 0.0, True)
```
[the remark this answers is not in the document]
regislebrun: fixed

python/test/t_KarhunenLoeveP1Algorithm_std.py
```diff
@@ -70,3 +70,15 @@
...
+if ot.PlatformInfo.HasFeature("spectra"):
+    tiny_mesh = ot.RegularGrid(0.0, 1.0, 2)
+    cov1d = ot.AbsoluteExponential([1.0])
+    ot.ResourceMap.SetAsString("KarhunenLoeveP1Algorithm-EigenvaluesSolver", "SPECTRA")
+    algo0 = ot.KarhunenLoeveP1Algorithm(tiny_mesh, cov1d, 0.0)
+    algo0.run()
+    result0 = algo0.getResult()
+    ott.assert_almost_equal(result0.getEigenvalues().getDimension(), 0)
+    ott.assert_almost_equal(result0.getModes().getSize(), 0)
+    ot.ResourceMap.SetAsString("KarhunenLoeveP1Algorithm-EigenvaluesSolver", "LAPACK")
```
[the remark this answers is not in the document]
regislebrun: Fixed

lib/src/Uncertainty/Process/KarhunenLoeveQuadratureAlgorithm.cxx
regislebrun: fixed

lib/src/Uncertainty/Process/KarhunenLoeveSVDAlgorithm.cxx
regislebrun: The TBB parts of the SVD version have been removed.

python/test/t_KarhunenLoeveP1Algorithm_std.py
```diff
@@ -10,25 +10,30 @@
...
+assert coefficients.getSize() == 10
+assert coefficients.getDimension() == 8
 KLFunctions = result.getModes()
-print("KL functions=", KLFunctions)
-print("KL lift=", result.lift(coefficients[0]))
-print("KL lift as field=", result.liftAsField(coefficients[0]))
+assert len(KLFunctions) == 8
+lift = result.lift(coefficients[0])
+assert lift.getOutputDimension() == 1
+assert len(result.liftAsField(coefficients[0])) == 10
```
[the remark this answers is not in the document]
regislebrun: This is not an actual numeric assertion (ie the comparison between a floating-point value computed by an algorithm and a reference floating point value) so raw assert is ok

python/test/t_KarhunenLoeveQuadratureAlgorithm_std.py
```diff
@@ -17,17 +19,10 @@
...
-print("KL eigenvalues=", lambd)
-process = ot.GaussianProcess(model, KLModes.getMesh())
-sample = process.getSample(10)
-coefficients = result.project(sample)
-print("KL coefficients=", coefficients)
-KLFunctions = result.getModes()
-print("KL functions=", KLFunctions)
-print("KL lift=", result.lift(coefficients[0]))
-print("KL lift as field=", result.liftAsField(coefficients[0]))
+assert len(KLModes) == 1
```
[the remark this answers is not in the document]
regislebrun: This is not an actual numeric assertion (ie the comparison between a floating-point value computed by an algorithm and a reference floating point value) so raw assert is ok

PR #3269, opened 2026-07-29
Fix compile error when build check with msvc

Lunar-YZ:
cmake --build BUILD_PATH --target check
error: M_PI not defined. 

// it is add to OPENTURNS_DEFINITIONS 
if (MSVC OR WINTEL)
  list (APPEND OPENTURNS_DEFINITIONS -D_USE_MATH_DEFINES)
endif ()

Note: ctest will fail because  use sh in windows, but let it compile pass now

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency for internal test builds by ensuring generated tests use the project’s configured compile-time definitions.


jschueller:
thanks, fixed in https://github.com/openturns/openturns/pull/3259/changes/ee9e963c5bbbfa5408ab71fa20b6a7c0a9c9ce9c

PR #3249, opened 2026-06-29
Improved OrthogonalUniVariatePolynomial and OrthogonalBasis

regislebrun:
# Improved OrthogonalUniVariatePolynomial and OrthogonalBasis

## Summary

Introduce `UniVariateDistributionPolynomialFactory` as a replacement for the
deprecated `StandardDistributionPolynomialFactory`. The new class detects
specific orthogonal polynomial families (Hermite, Legendre, Jacobi, Laguerre,
Chebychev, etc.) for any parameterization of the corresponding distribution,
not just the standard one (e.g., Normal(0,1), Uniform(-1,1), Beta on [-1,1]).

This PR also adds an affine transformation framework to `OrthogonalUniVariatePolynomial`
and `OrthogonalUniVariatePolynomialFactory`, enabling correct polynomial
evaluation, root computation, and Gauss quadrature node/weight generation over
non-standard distribution intervals.

### New class: UniVariateDistributionPolynomialFactory

- Inherits from `OrthogonalUniVariatePolynomialFactory`
- Two modes of operation:
  - **Specific family detection**: Recognizes Hermite, Legendre, Jacobi,
    Laguerre, Chebychev, Charlier, Krawtchouk, Meixner families from any
    parameterization of the corresponding distribution
  - **Fallback**: Generic orthonormalization via `AdaptiveStieltjesAlgorithm`
  for unrecognized distributions

### Affine transformation support

- Added `a_`, `b_` members to `OrthogonalUniVariatePolynomial` and
  `OrthogonalUniVariatePolynomialFactory` to represent the affine map
  `z = a_*x + b_`
- `buildCoefficients()` computes monomial coefficients under the affine
  variable transformation, enabling correct coefficient output for non-standard
  factories
- Clenshaw evaluation and root computation properly handle the affine map

### Non-standard parameter constructors

| Factory | New Constructor | Distribution |
|---------|----------------|--------------|
| `HermiteFactory` | `(mu, sigma)` | Normal(mu, sigma) |
| `LegendreFactory` | `(a, b)` | Uniform(a, b) |
| `JacobiFactory` | `(alpha, beta, a, b)` | Beta(alpha, beta) on [a, b] |
| `LaguerreFactory` | `(k, lambda, gamma)` | Gamma(k, lambda, gamma) |
| `ChebychevFactory` | `(a, b)` | Arcsine(a, b) |

### Bug fixes

- `LegendreFactory::getNodesAndWeights()` properly scales from [-1,1] to
[...]

PR #3277, opened 2026-08-14
Improve quadrature

regislebrun:
This PR proposes a rework of two multivariate adaptive integration algorithms: IteratedQuadrature and SimplicialCubature. 

IteratedQuadrature
------------------
- Flatten the recursion to avoid parametric wrappers
- Reuse the integrand over the points of a batch
- Fill the full points of a batch with std::copy
- Evaluate the local integrals over the batch points with a TBB parallel_for_if when the integrand is parallel-safe
  - Let each thread own its integrand with a dedicated prefix, keeping the prefix-based reuse while remaining safe to call concurrently
  - Make the exptk and muparser parsers thread-safe (per-instance mutex and lazy per-thread expressions)
  - Declare the SymbolicEvaluation parallel
  - Guard the Python evaluation with the Python GIL so that it can be called from several threads, while remaining serial to avoid TBB/GIL deadlocks
- Reuse a GaussKronrod workspace for the local integrals: each thread keeps the scratch containers of the underlying GaussKronrod algorithm (sub-interval bounds, values, errors and the local rule samples) and reuses them for all its local integrals, so that the inner loop of the recursion does not allocate anymore
  - Grow the container capacity on the first call only
  - Keep the fallback path for the non-GaussKronrod inner algorithms

SimplicialCubature
------------------
- Precompute the rule points and weights once
  - initializeBasicRule builds the matrix B of barycentric coordinates of all the integration points grouped by generator, and the matrix Wbar of their weight for every (sub)rule
  - computeRulePoints maps these barycentric coordinates to the vertices of a simplex, so each simplex requires only a single call to the integrand followed by a matrix contraction, instead of one call per generator orbit
  - Compute the node number as the actual sum of the orbit sizes
- Batch the rule evaluation over the initial mesh: build the physical rule points of all the simplices into a single flat sample, evaluate the integrand with one call and contract the result per simplex (one function call instead of one per simplex)
- Parallelize with TBB and batch the adaptive refinements
  - Use TBB for the two embarrassingly parallel parts of the initial pass (building of the physical rule points and contraction of the rule), with a per-thread scratch
  - Refine the R worst-error simplices per round in the adaptive loop, where R is limited by the remaining evaluation budget, and evaluate the rule over all the resulting simplices in a single batched call
  - Keep the subdivision itself serial so that the refinement sequence is deterministic and independent of the number of threads, with results identical whatever the number of threads
- Remove the per-call allocations in the inner loop
  - Pass the contraction rule matrix as a per-thread scratch of the contract policy instead of allocating it for every simplex
  - Pass the subdivision scratch (difference point, centroid, function value, difference table and difference-test sample) of the refinement loop by reference instead of allocating them per subregion
- Check the integrand dimension against the mesh dimension, throwing an InvalidArgumentException instead of failing later inside the evaluation
- Guard the empty mesh, returning a zero vector
- Replace the non portable M_PI constant by acos(-1.0)
- Extend the Python testcase to cover the multi-output path, the adaptive refinement on a peaked integrand over an interval, and the accuracy of rule 4 over the canonical simplex

## Summary by CodeRabbit

* **New Features**
  * Improved numerical integration and cubature performance through batching, parallel processing, and reusable workspaces.
  * Added configurable evaluation-block sizing for simplex cubature.
  * Added reusable evaluation buffers for Gauss–Kronrod integration.

* **Bug Fixes**
  * Strengthened thread safety for symbolic and Python-backed evaluations.
  * Improved validation for invalid integration and cubature inputs.
  * Corrected bootstrap Sobol’ confidence-interval handling.

* **Documentation**
  * Documented cubature configuration options and evaluation thread-safety behavior.

* **Tests**
  * Expanded integration, cubature, and parallel-evaluation coverage.


ChangeLog
```diff
@@ -85,7 +85,8 @@
  * KDTree: Add radius search
  * Mesh: Add convex flag
 
-=== Bug fixes (total 27) ===
+=== Bug fixes (total 28) ===
+ * (IteratedQuadrature throws a clear error when the number of bound functions is inconsistent with the integrand dimension)
```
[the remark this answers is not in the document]
regislebrun: fixed

lib/src/Base/Algo/SimplicialCubature.cxx
```diff
@@ -195,15 +245,27 @@ Point SimplicialCubature::integrate(const Function & F, const Mesh & mesh) const
...
+  {
+    for (UnsignedInteger K = range.begin(); K < range.end(); ++ K)
+      computeRulePoints(flatVertices[K], B, allPoints, K * nodeNumber);
+  });
+  const Sample outputs(F(allPoints));
+  Indices initialTargets(flatSize);
+  for (UnsignedInteger K = 0; K < flatSize; ++ K)
+    initialTargets[K] = K;
+  const ContractPolicy contractPolicy(*this, nodeNumber, outputDimension, initialTargets, volume, outputs, Wbar, VLS, AES);
+  TBBImplementation::ParallelForIf(parallel, 0, flatSize, contractPolicy);
```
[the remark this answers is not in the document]
regislebrun: fixed

lib/src/Base/Common/ResourceMap.cxx
```diff
@@ -1835,6 +1835,7 @@ void ResourceMap::loadDefaultConfiguration()
   addAsScalar("SimplicialCubature-DefaultMaximumRelativeError", 1.0e-5);
   addAsUnsignedInteger("SimplicialCubature-DefaultMaximumCallsNumber", 100000);
   addAsUnsignedInteger("SimplicialCubature-DefaultRule", 3);
+  addAsUnsignedInteger("SimplicialCubature-EvaluationBlockSize", 2048);
```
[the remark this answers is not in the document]
regislebrun: fixed

lib/src/Base/Func/SymbolicEvaluation.cxx
```diff
@@ -272,8 +272,9 @@ Bool SymbolicEvaluation::isLinearlyDependent(const UnsignedInteger index) const
 /* Is it safe to call in parallel? */
 Bool SymbolicEvaluation::isParallel() const
 {
-  // neither Exprtk nor muParser are thread-safe
-  return false;
+  // The point evaluations are serialized with a mutex, and the batch
+  // evaluations use one expression per thread
+  return true;
```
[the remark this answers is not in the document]
regislebrun: Fixed

regislebrun:
@coderabbitai do you think all the issues have been fixed now?

lib/test/t_SimplicialCubature_std.cxx
```diff
@@ -64,6 +64,89 @@ int main(int, char *[])
...
+      const Point reference(algo.integrate(f, mesh));
+      for (UnsignedInteger i = 0; i < 3; ++ i)
+      {
+        const Point value(algo.integrate(f, mesh));
+        assert_almost_equal(value, reference);
+        assert_almost_equal(value[0], ref, 1.0e-6, 1.0e-8);
+      }
+      // The refinement must take place for the tie-breaking to be exercised
+      if (!(f.getCallsNumber() > mesh.getSimplicesNumber() * 16))
+        throw TestFailed(OSS() << "No refinement occurred, calls=" << f.getCallsNumber());
```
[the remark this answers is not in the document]
regislebrun: fixed

python/test/t_GaussKronrod_std.py
```diff
@@ -61,18 +51,66 @@
...
+f = ot.SymbolicFunction("x", "sin(x)")
+ref = math.cos(-2.5) - math.cos(4.5)
+error = ot.Point()
+ai = ot.Point()
+bi = ot.Point()
+fi = ot.Sample()
+ei = ot.Point()
+value = algo.integrate(f, -2.5, 4.5, error, ai, bi, fi, ei)
+ott.assert_almost_equal(value[0], ref, 1e-8, 1e-12)
+assert error[0] < algo.getMaximumError()
```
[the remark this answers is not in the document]
regislebrun: I disagree with your analysis. The current test checks that the error is below the prescribed threshold, an assert would only check that the error is close to the threshold and would failed eg if the error is much lower than the threshold.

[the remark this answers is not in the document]
regislebrun: It is not very readable, but ok. Fixed.

regislebrun:
@coderabbitai resume

regislebrun:
@coderabbitai review

lib/test/t_SimplicialCubature_std.cxx
```diff
@@ -64,6 +64,91 @@ int main(int, char *[])
...
+      Point reference;
+      for (const UnsignedInteger blockSize : {2048u, 7u, 1u})
+      {
+        ResourceMap::SetAsUnsignedInteger("SimplicialCubature-EvaluationBlockSize", blockSize);
+        const Point value(algo.integrate(f, mesh));
+        if (reference.getDimension() == 0) reference = value;
+        assert_almost_equal(value[0], refValue, 1.0e-6, 1.0e-8);
+        assert_almost_equal(value, reference, 1.0e-12, 0.0);
+      }
+      ResourceMap::SetAsUnsignedInteger("SimplicialCubature-EvaluationBlockSize", 2048);
```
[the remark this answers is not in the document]
regislebrun: fixed

python/doc/examples/surrogate_modeling/polynomial_chaos/plot_chaos_sobol_confidence.py
```diff
@@ -270,6 +268,11 @@ def computeSobolIndicesConfidenceInterval(fo_sample, to_sample, alpha=0.95):
     to_interval : ot.Interval
         The confidence interval of total order Sobol' indices
     """
+    if fo_sample.getSize() == 0 or to_sample.getSize() == 0:
+        raise ValueError(
+            "Empty bootstrap sample: all Sobol' index estimates were "
+            "rejected. Try increasing the training sample size."
+        )
```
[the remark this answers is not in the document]
regislebrun: fixed

lib/test/t_SimplicialCubature_std.cxx
```diff
@@ -64,6 +64,108 @@ int main(int, char *[])
...
+      bool caught = false;
+      try
+      {
+        algo.integrate(SymbolicFunction(Description(1, "x"), Description(1, "x")), mesh);
+      }
+      catch (const InvalidArgumentException &)
+      {
+        caught = true;
+      }
+      assert_almost_equal(caught ? 1.0 : 0.0, 1.0);
```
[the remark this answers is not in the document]
regislebrun: fixed

lib/test/t_SimplicialCubature_std.cxx
```diff
@@ -64,6 +64,108 @@ int main(int, char *[])
...
+          assert_almost_equal(value, reference, 1.0e-12, 0.0);
+        }
+      }
+      catch (...)
+      {
+        ResourceMap::SetAsUnsignedInteger("SimplicialCubature-EvaluationBlockSize", savedBlockSize);
+        throw;
+      }
+      ResourceMap::SetAsUnsignedInteger("SimplicialCubature-EvaluationBlockSize", savedBlockSize);
+    }
```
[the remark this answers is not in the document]
regislebrun: fixed

lib/src/Base/Algo/SimplicialCubature.cxx
```diff
@@ -172,21 +149,94 @@ Point SimplicialCubature::integrate(const Function & F, const Mesh & mesh) const
...
+      {
+        const UnsignedInteger K = targets_[t];
+        cubature_.computeRuleValueAndError(t * nodeNumber_, volumes_[t], outputDimension_, outputs_, Wbar_, RULE_, value_, error_);
+        VLS_[K] = value_;
+        AES_[K] = error_;
+      }
+    }
+  }; // struct ContractPolicy
+
+  const Bool parallel = (flatSize > 1) && (TBBImplementation::GetThreadsNumber() > 1);
```
jschueller: parallel = (flatSize > 1) && function.isParallel();

lib/src/Base/Func/openturns/SymbolicParserExprTk.hxx
```diff
@@ -72,6 +75,11 @@ private:
   mutable Collection<ExpressionCollection> threadExpressions_;
   mutable Collection<Point> threadStack_;
 
+  // guards the lazy initialization of threadExpressions_ and the point
+  // evaluation which uses the shared expressions_, so that the parser can be
+  // evaluated concurrently from several threads
+  mutable std::shared_ptr<std::mutex> mutex_ = std::make_shared<std::mutex>();
```
jschueller, line 110: this shares mutex_ over copies, copy ctor/clone must create fresh mutexes

python/src/PythonEvaluation.cxx
```diff
@@ -168,6 +192,8 @@ String PythonEvaluation::__str__(const String & ) const
 /* Operator () */
 Point PythonEvaluation::operator() (const Point & inP) const
 {
+  // The evaluation may be called from any thread: hold the GIL
+  const ScopedPyGIL gil;
```
jschueller: correct, but has good chances to have no speedup at all, because of the GIL contention

instead I would drop ScopedPyGIL and only declare PythonEvaluation thread-safe in Python free-threaded flavors (cp314t)

lib/src/Base/Algo/SimplicialCubature.cxx
regislebrun: Fixed: now checks `F.getImplementation()->isParallel()` before enabling TBB parallelism.

lib/src/Base/Func/openturns/SymbolicParserExprTk.hxx
regislebrun: Fixed: added explicit copy ctor and copy assignment to both `SymbolicParserExprTk` and `SymbolicParserMuParser` that allocate a fresh `std::make_shared<std::mutex>()` for each clone, so no two instances share the same mutex.

python/src/PythonEvaluation.cxx
regislebrun: Good point. `isParallel()` already returns `false` unconditionally, so TBB never dispatches work to `PythonEvaluation` and the `ScopedPyGIL` is a no-op in the standard path (the GIL is already held by the calling thread). It is kept as a safety net for direct C++ reentrant calls.

Dropping `ScopedPyGIL` and enabling `isParallel()` only in free-threaded builds (`cp314t`) would be a cleaner design but is a separate change. Happy to revisit in a follow-up.

jschueller: you can drop ScopedPyGIL, I wont be needed in free-threaded builds

adutfoy (changes requested):
See my remarks:

- detail the affine transformation 
- precise that the recurence coeffiicents are those of the polynomials orthonormal wrt the standard representative distribution
- in the table, prensent the general families and not the stanard ones

lib/src/Base/Algo/SimplicialCubature.cxx
regislebrun: Confirmed fixed: SimplicialCubature.cxx:239 now reads `const Bool parallel = (flatSize > 1) && F.getImplementation()->isParallel() && (TBBImplementation::GetThreadsNumber() > 1);`.

lib/src/Base/Func/openturns/SymbolicParserExprTk.hxx
regislebrun: Confirmed fixed: both `SymbolicParserExprTk` and `SymbolicParserMuParser` now have explicit copy constructors and copy assignment operators that allocate a fresh `std::make_shared<std::mutex>()` for `mutex_`, so copies no longer share the mutex.

python/src/PythonEvaluation.cxx
regislebrun: Done: `ScopedPyGIL` has been dropped entirely (was 6 call sites). `isParallel()` now returns `true` only under `Py_GIL_DISABLED` (free-threaded builds), `false` otherwise. No GIL acquisition in the evaluation path.

regislebrun:
@adutfoy, I addressed your three remarks:

- **Detail the affine transformation**: added a paragraph explaining that any member of a continuous parametric family maps onto its standard representative via the affine transformation `X = alpha * Z + beta`, and that the orthonormal polynomials wrt `X` follow from those wrt `Z` by the substitution `x -> (x - beta)/alpha`, with the `Z ~ N(0,1)`, `X = mu + sigma*Z` example.
- **Precise that the recurrence coefficients are those of the polynomials orthonormal wrt the standard representative distribution**: now explicit in both the class docstring and the theory page; also that the standard representative is given by `Distribution.getStandardRepresentative`.
- **Present the general families and not the standard ones in the table**: both tables now list the general parametric families (Normal `cN(mu,sigma)`, Uniform `cU(a,b)`, Gamma, Beta, Poisson, Binomial, Polya) instead of the standard representatives.

As a bonus, the ASCII grid tables were converted to the Sphinx `.. list-table::` directive, which is much easier to maintain.

Note that the two doc files (`StandardDistributionPolynomialFactory_doc.i` and `orthonormal_polynomials.rst`) were not part of the PR diff, so those edits now appear in the PR (commit `da230d89e`); if you prefer them in a dedicated doc PR, let me know and I will split them out.

adutfoy (approved):

lib/src/Base/Algo/IteratedQuadrature.cxx
```diff
@@ -155,11 +230,164 @@ class IteratedQuadraturePartialFunctionWrapper : public FunctionImplementation
...
+    std::copy(prefix_.begin(), prefix_.end(), fullSampleImplementation.data_begin() + i * fullDimension);
+    fullSample(i, prefixDimension) = sample(i, 0);
+  }
+  const Sample sampleA(lowerBounds_[depth_](fullSample));
+  const Sample sampleB(upperBounds_[depth_](fullSample));
+  // Evaluate the local integrals over the sample points, using one integrand
+  // per thread in order to keep the prefix-based reuse while being safe to
+  // call concurrently when the original integrand allows parallel evaluation
+  const IteratedQuadratureIntegratePolicy policy(*this, fullSample, sampleA, sampleB, sample, result);
+  TBBImplementation::ParallelForIf(sampleSize > 1 && TBBImplementation::GetThreadsNumber() > 1 && function_.getImplementation()->isParallel(), 0, sampleSize, policy);
```
jschueller: imho the condition can be dropped here (its implicit)

python/test/t_SimplicialCubature_std.py
```diff
@@ -4,24 +4,130 @@
...
+try:
+    ot.ResourceMap.SetAsUnsignedInteger("SimplicialCubature-EvaluationBlockSize", 2)
+    f = ot.SymbolicFunction(["x1", "x2", "x3"], ["exp(x1 + x2 + x3)"])
+    algo.setRule(1)
+    algo.setMaximumAbsoluteError(0.0)
+    algo.setMaximumRelativeError(1.0e-5)
+    algo.setMaximumCallsNumber(50000)
+    value = algo.integrate(f, mesh)[0]
+    ott.assert_almost_equal(value, (m.exp(1.0) - 1.0) ** 3 / 6, 1e-4, 1e-4)
+finally:
```
jschueller: you dont need try/finally blocks

lib/src/Base/Algo/SimplicialCubature.cxx
```diff
@@ -195,16 +245,40 @@ Point SimplicialCubature::integrate(const Function & F, const Mesh & mesh) const
...
+  // kept while the memory footprint stays bounded: the full mesh would need
+  // flatSize * nodeNumber points, which grows with the mesh and the rule.
+  const UnsignedInteger blockSize = ResourceMap::GetAsUnsignedInteger("SimplicialCubature-EvaluationBlockSize");
+  if (blockSize == 0)
+    throw InvalidArgumentException(HERE) << "SimplicialCubature-EvaluationBlockSize must be > 0";
+  for (UnsignedInteger blockBegin = 0; blockBegin < flatSize; blockBegin += blockSize)
+  {
+    const UnsignedInteger blockFlatSize = std::min(blockSize, flatSize - blockBegin);
+    Sample blockPoints(blockFlatSize * nodeNumber, dimension);
+    TBBImplementation::ParallelForIf(parallel, blockBegin, blockBegin + blockFlatSize,
```
jschueller, line 261: are we sure that these small parallel regions yield actual speedups ?
last time I benchmarked tbb regions I saw that it only yield speedups if each thread had enough work
so I had to increase the grain size almost in every case, and here I dont see any grain size being set in any of the new parallel regions in this branch

lib/src/Base/Algo/IteratedQuadrature.cxx
regislebrun: Done in 9b007bc30: the condition is now just function_.getImplementation()->isParallel(), matching the canonical ParallelForIf pattern used everywhere else.

python/test/t_SimplicialCubature_std.py
regislebrun: Done in 9b007bc30: both try/finally blocks removed, the ResourceMap overrides are now set and restored with plain statements as in the EvaluationBlockSize=0 guard above.

lib/src/Base/Algo/SimplicialCubature.cxx
regislebrun: Done in 9b007bc30: explicit grain size of 1024 added to the four new SimplicialCubature parallel regions, as in ProductUniVariateFunctionEvaluation, P1LagrangeEvaluation and ProductPolynomialEvaluation. Per-simplex work (rule-point transform + contraction) is light, so small refinement rounds now run in a single chunk while full 2048-simplex blocks still split. The IteratedQuadrature region is left at the default grain since each item is a full adaptive sub-integral. No local benchmark was possible here; CI must confirm, happy to tune the value with measurements if you have numbers.

jschueller (approved):

regislebrun:
Note: my 2026-09-18 comment answering @adutfoy's three affine-transformation remarks was posted on the wrong PR, and its doc edits (commit da230d89e) did not belong here either. The comment now lives on PR #3249, the edits are ported to branch affine_transformation (commit ef00473b7), and the stray hunks were removed from this PR (commit 0846fbe59, exact revert to da230d89e^). This PR is again quadrature-only. Sorry for the confusion.

regislebrun:
@jschueller Two thumbs up, no thumb down, go to master?

jschueller:
I cannot merge this one either because its not properly rebased

regislebrun:
@jschueller done

regislebrun:
@jschueller two thumbs up, one thumb down, ready to go?

Issue #3307, opened 2026-09-14
Gaussian Process Regression does not retain variable names


josephmure:
### What happened?

Previously, the metamodel produced by KrigingAlgorithm took its input (resp. output) description from the description of the input (resp. output) sample.

This is no longer the case, the metamodel does not take its input/output names from the training data, which is a bug IMO.

### How to reproduce the issue?

```Python
# %%
from numpy import sqrt
import openturns as ot
import openturns.viewer as otv
from openturns.usecases import branin_function

# %%
# We load the Branin-Hoo model from the usecases module and use the model (stored in `objectiveFunction`)
bm = branin_function.BraninModel()
model = bm.objectiveFunction

experiment = ot.LHSExperiment(
    ot.JointDistribution([ot.Uniform(0.0, 1.0), ot.Uniform(0.0, 1.0)]),
    28,
    False,
    True,
)
xdata = experiment.generate()
ydata = bm.objectiveFunction(xdata)


# This use case is defined in dimension 2 and we use a constant basis for the trend estimation :
dimension = bm.dim
basis = ot.ConstantBasisFactory(dimension).build()

# %%
# We choose a squared exponential covariance model in dimension 2 :
covarianceModel = ot.SquaredExponential([0.1] * dimension, [1.0])

# %%
# We have all the components to build a GP fitter  algorithm and run it :
fitter_algo = ot.GaussianProcessFitter(xdata, ydata, covarianceModel, basis)
fitter_algo.setOptimizeParameters(True)
fitter_algo.run()
fitter_result = fitter_algo.getResult()
gpr_algo = ot.GaussianProcessRegression(fitter_result)
gpr_algo.run()
gpr_result = gpr_algo.getResult()


# We get the metamodel (mean of the conditioned Gaussian Process) and the variance estimation.
# We draw the GP metamodel of the Branin function.
gprMetamodel = gpr_result.getMetaModel()


# %%
# Print the input descriptions
# They should be the same (X0, X1 in both cases)
print("metamodel input description = ", gprMetamodel.getInputDescription()) #x0, x1
print("input data description = ", xdata.getDescription()) # X0, X1
```

### Version

1.28dev

### Operating System

unknown

### Installation media

unknown

Issue #3227, opened 2026-06-09
TruncatedDistribution with 2 scalar lower ad upper bounds


adutfoy:
### What happened?

I try to truncate a scalar distributin giving both bounds.

### How to reproduce the issue?

```Python
import openturns as ot
# it does not work
dist_T = ot.TruncatedDistribution(Normal(1000, 50), 800, 1200)
# here it is ok
dist_T = ot.TruncatedDistribution(Normal(1000, 50), ot.Interval(800, 1200))
```
In the documentation, both declarations are possible!
### Version

1.27dev

### Operating System

Linux

### Installation media

from source

jschueller:
it works if you explicitely pass floats (else it uses the side/bound variant):
```
dist_T = ot.TruncatedDistribution(ot.Normal(1000, 50), 800.0, 1200.0)
```
this is kind of ambiguous, remove this ctor in favor of the interval ctor ?

Issue #2521, opened 2024-01-24
Normal::computeProbability does not account for infinite boundaries in dimension 1


josephmure:
### What happened?

The following line is bugged because it does not take into account the possibility that the upper or lower bound of an interval was declared infinite.
It is only triggered in dimension 1.

https://github.com/openturns/openturns/blob/61f0de3491571a933207c252b006a4ff23ea046b/lib/src/Uncertainty/Distribution/Normal.cxx#L446

A possible solution would be to remove the `setFiniteLower/UpperBound` methods and the constructor
```hxx
  Interval(const Point & lowerBound,
           const Point & upperBound,
           const BoolCollection & finiteLowerBound,
           const BoolCollection & finiteUpperBound);
```
and simply set the bounds to `MaxScalar` (e.g. infinity) when required.

### How to reproduce the issue?

```shell
import openturns as ot

dist = ot.Normal()
interval = ot.Interval([-1.0], [0.0], [False], [True])
print(dist.computeProbability(interval)) # expect 0.5, get 0.34134474606854337 instead (proba of [-1,0])
```

### Version

1.22

### Operating System

Linux

### Installation media

conda

jschueller:
its meant to use the numerical value of range when computing the intersection with the distribution range; I dont think we want to change that

josephmure:
This problem is not specific to Normal.cxx, all distributions should be reviewed.

JPelamatti:
Propostion from the technical committee:
Modify the computeProbabilityGeneral1D method so that it behaves in the following way
- Input is now an interval rather than 2 scalars
- if finitelowerbound = False && finiteupperbound = false -> return 1
- if only finitelowernbound = False -> use cdf in upperbound
- if only finiteupperbound = False -> use complementary cdf in lowerbound
- else -> as usual

Issue #2365, opened 2023-08-10
Document the order of the coefficients returned by KrigingResult.getTrendCoefficients


mdelozzo:
Since openturns 1.21, `KrigingResult.getTrendCoefficients` no longer returns a collection of `Point` but a unique `Point`.

First, a minor point: [in the documentation](https://openturns.github.io/openturns/latest/user_manual/response_surface/_generated/openturns.KrigingResult.html#openturns.KrigingResult.getTrendCoefficients),  I think that `Point` would be more relevant than _sequence of float_ (in 1.20, we can read _collection of `Point`_).

Then, a major point: surely there is a good explanation for this API change and so I'm not criticizing that. But I would like to share a feeling. It is now more difficult to read the coefficient associated with a given basis function. To illustrate that, I built a Kriging from some samples of $f(x)=(1+2x,3+4x)$ and I got `[1,3,2,4,1.60118e-15,-1.37244e-15]` while I would have expected `[1,2,1.60118e-15,3,4,1.60118e-15]`. In other words, I would have expected: first the coefficients of the first output, then those of the second output. 

I think that documenting the order of the coefficients would be a good thing. Not only in this case of 1 input and $p>1$ outputs but also in the general case of $d$ inputs and $p$ outputs.

sofianehaddad:
I will update the documentation
you might notice we grouped coefficients by blocks ( (1,3), (2,4), (1.60118e-15,-1.37244e-15)).

 In other words, $\phi_0$,..., $\phi_{m-1}$, `m` being the size of the basis, $\phi_j$ a p-vectorial function

mdelozzo:
Something like $(\phi_{1,1}\ldots\phi_{1,p}\ldots\phi_{m,1}\ldots\phi_{m,p})$ would help.

mbaudin47:
Indeed, the notations in this page are confusing, wrong and inconsistent with other help pages: it would be too long to point them all here.
Notice that "sequence of float" might be the right naming. Indeed, if you give to the class a list of float, the conversion to a Point is done at the Swig layer. So anything like a tuple of floats or even a 1d-numpy array will do the job. We cannot list them all everywhere, hence the name "sequence of floats".

mdelozzo:
[...]

PR #3310, opened 2026-09-15
Misc2

jschueller:
- Fix OOB access in DickeyFullerTest
- #2521
- #2365
- #3227
- #3307

## Summary by CodeRabbit

* **New Features**
  * Added Gaussian-process-based Box-Cox fitting through `buildWithGPF`.
  * Metamodels now retain input and output descriptions.
  * Added the `SlicedInverseRegression` API.
* **Bug Fixes**
  * Improved probability calculations for unbounded intervals and infinite distribution bounds.
  * Added validation for positive shifted samples and single-dimensional outputs.
* **API Changes**
  * Removed the scalar-bounds `TruncatedDistribution` constructor; use an `Interval`.
  * Deprecated `buildWithGLM` in favor of `buildWithGPF`.
* **Documentation**
  * Updated examples and API guidance for interval-based truncation and Gaussian-process workflows.


lib/src/Uncertainty/Algorithm/Transformation/ProcessTransformation/BoxCoxFactory.cxx
```diff
@@ -212,6 +213,85 @@ class BoxCoxGLMOptimization : public EvaluationImplementation
...
+  }
+
+  void computeSumLog()
+  {
+    const UnsignedInteger size = shiftedOutputSample_.getSize();
+    const UnsignedInteger dimension = shiftedOutputSample_.getDimension();
+    sumLog_ = 0.0;
+    for (UnsignedInteger k = 0; k < size; ++k)
+      for (UnsignedInteger d = 0; d < dimension; ++d)
+        sumLog_ += std::log(shiftedOutputSample_(k, d));
```
regislebrun, line 279: no need to check if shiftedOutputSample_(k, d)>0 here? -> ok, checked at the calling level

lib/src/Uncertainty/Algorithm/Transformation/ProcessTransformation/BoxCoxFactory.cxx
```diff
@@ -526,6 +617,82 @@ BoxCoxTransform BoxCoxFactory::buildWithGLM(const Sample &inputSample,
...
+  // Keep the shifted marginal samples
+  Sample shiftedSample(outputSample);
+  shiftedSample += shift;
+  for (UnsignedInteger i = 0; i < size; ++i)
+    if (!(shiftedSample(i, 0) > 0.0))
+      throw InvalidArgumentException(HERE) << "Error: shifted output sample must be strictly positive, got " << shiftedSample(i, 0) << " at index " << i;
+
+  // optimization process
+  BoxCoxGPFOptimization boxCoxOptimization(inputSample, shiftedSample, covarianceModel, basis);
+  Function objectiveFunction(boxCoxOptimization);
```
regislebrun: const

lib/src/Uncertainty/Algorithm/Transformation/ProcessTransformation/BoxCoxFactory.cxx
```diff
@@ -526,6 +617,82 @@ BoxCoxTransform BoxCoxFactory::buildWithGLM(const Sample &inputSample,
...
+  MemoizeFunction objectiveMemoizeFunction(objectiveFunction, Full());
+  objectiveMemoizeFunction.enableCache();
+  OptimizationProblem problem(objectiveMemoizeFunction);
+  problem.setMinimization(false);
+  OptimizationAlgorithm solver(solver_);
+  solver.setProblem(problem);
+  solver.setStartingPoint(Point(1, 1.0));
```
regislebrun: Is it reasonable to hard-code 1 as the starting value? No use-case where a clever starting point is known (eg you update a previous study with a few new data)?

lib/src/Uncertainty/Distribution/openturns/TruncatedDistribution.hxx
```diff
@@ -42,12 +42,6 @@ public:
   /** Default constructor */
   TruncatedDistribution();
 
-  /** Parameters constructor to use when the two bounds are finite */
```
regislebrun, line 45: q_q

lib/src/Uncertainty/Distribution/KernelMixture.cxx
```diff
@@ -380,13 +380,20 @@ Scalar KernelMixture::computeProbability(const Interval & interval) const
...
-  const Point upperBound(reducedInterval.getUpperBound());
+  Point lowerBound(reducedInterval.getLowerBound());
+  Point upperBound(reducedInterval.getUpperBound());
+  const Interval::BoolCollection reducedFiniteLower(reducedInterval.getFiniteLowerBound());
+  const Interval::BoolCollection reducedFiniteUpper(reducedInterval.getFiniteUpperBound());
+  for (UnsignedInteger j = 0; j < dimension; ++j)
+  {
+    if (!reducedFiniteLower[j]) lowerBound[j] = SpecFunc::LowestScalar;
+    if (!reducedFiniteUpper[j]) upperBound[j] = SpecFunc::MaxScalar;
+  }
```
regislebrun: The values of lower bound and upper bound are here to bound the numerical range when the mathematical range is infinite. If you replace these values by SpecFunc::LowestScalar and SpecFunc::MaxScalar you may break the numerical integration routines and the solvers. Here, the objective is to say that CDF(-inf)=0 and CDF(+inf)=1, do this explicitely.

lib/src/Uncertainty/Model/DistributionImplementation.cxx
```diff
@@ -1221,22 +1233,43 @@ Scalar DistributionImplementation::computeProbabilityGeneral1D(const Scalar a,
...
   if (reducedInterval.isEmpty()) return 0.0;
   if (reducedInterval == range_) return 1.0;
+  const Interval::BoolCollection reducedFiniteLower(reducedInterval.getFiniteLowerBound());
+  const Interval::BoolCollection reducedFiniteUpper(reducedInterval.getFiniteUpperBound());
   if (dimension_ == 1)
-    return computeProbabilityContinuous1D(reducedInterval.getLowerBound()[0], reducedInterval.getUpperBound()[0]);
+  {
+    const Scalar a = reducedFiniteLower[0] ? reducedInterval.getLowerBound()[0] : SpecFunc::LowestScalar;
+    const Scalar b = reducedFiniteUpper[0] ? reducedInterval.getUpperBound()[0] : SpecFunc::MaxScalar;
+    return computeProbabilityContinuous1D(a, b);
```
regislebrun: Same remark here. For a Gaussian-like distribution you replace a [mu-k*sigma, mu+k*sigma] range by something HUGE, and the adaptive integration algorithm will completely miss the mass.

regislebrun (commented):
I have serious doubts regarding the replacement of the bounds values by SpecFunc::LowestScalar and SpecFunc::MaxScalar in the probability computations. It will break the adaptive integration of the PDF, it was the actual purpose of lowerBound and upperBound to pilot integration routines in the case of infinite ranges. Otherwise, what is the purpose of making a distinction between exact and numerical range (see the methods of Interval). This point has to be discussed before this PR lands in master.

lib/src/Uncertainty/Algorithm/Transformation/ProcessTransformation/BoxCoxFactory.cxx
jschueller: ok, added an accessor

jschueller:
> I have serious doubts regarding the replacement of the bounds values by SpecFunc::LowestScalar and SpecFunc::MaxScalar in the probability computations. It will break the adaptive integration of the PDF, it was the actual purpose of lowerBound and upperBound to pilot integration routines in the case of infinite ranges. Otherwise, what is the purpose of making a distinction between exact and numerical range (see the methods of Interval). This point has to be discussed before this PR lands in master.

fixed 1d continuous and kernel mixture paths to handle it explicitly, integration only sees finite bounds

regislebrun (approved):
Thanks for the comments around interval bounds modifications

